PR #3, opened 2019-07-18
Remove old GNUmakefiles

mathomp4:
# NOTE NOTE NOTE NOTE NOTE

This one is an unknown. Does Harvard still require these `GNUmakefile` to exist? If so, this pull request should be closed and the branch deleted.


lizziel:
Do the GNU makefiles in the repo still work (or could work with a bit of work), or are you removing files that are beyond help at this point?

tclune:
If you desperately want them, they could be restored.  As you say they may need a bit of modification, but I can't imagine it would be very much.

Our strong preference is to delete them asap.   Please remind me GCHP's timeline for cmake?

lizziel:
We plan on moving to cmake at the same time that we incorporate the new git repos. We are meeting in a couple weeks to further discuss, but Seb and I both feel confident we will not issue a version with both GNU make and cmake, but will move straight to cmake only.

PR #22, opened 2019-08-08
Feature/rc template

weiyuan-jiang:
template and EASE grid fix


mathomp4:
I'll wait for @bena-nasa to be back Monday to look at this. @weiyuan-jiang Did you test these changes with the GCM to see that it doesn't affect that? If so, that's probably the first thing we need to test. If so, then approval here can trigger all the other pulls.

mathomp4:
I have marked this a blocker in Zenhub to the other two @weiyuan-jiang requests.

Issue #25, opened 2019-08-13
MAPL_Base depends on a file that isn't included in ESMF's installation


LiamBindle:
Hi everyone,

First off, thanks for all your work and for making MAPL open-source on GitHub! 

I've been working on moving GEOS-Chem's build to CMake and one of the things we'd like to do is make ESMF an external library. I'm running into problems in MAPL_Base's build because `MAPL_Exceptions.h:157` is trying to include a file that isn't included in ESMF's installation. Specifically, this is the line that's causing my issue

https://github.com/GEOS-ESM/MAPL/blob/898ce7b8ee9ce48909debc7eedc96e316aa60945/MAPL_Base/MAPL_Exceptions.h#L157

My problem is that `ESMF_ErrReturnCodes.inc` is in `ESMF/src/include` which therefore has to be included in MAPL's build's list of include directories. But this means that you can't build MAPL from a default ESMF install. It looks like there might be an easy fix though.

By the way, my ESMF version is v8.0.0-beta.

## Possible solution
It looks like the main purpose of `ESMF_ErrReturnCodes.inc` is to `#include "ESMC_ReturnCodes.h"`. `ESMF_ErrReturnCodes.inc` does define two extra preprocessor definitions, `ESMF_ERR_PASSTHRU` and `ESMCI_ERR_PASSTHRU`, but I don't see any references to these in MAPL or GMAO_Shared.

```diff
--- a/MAPL_Base/MAPL_Exceptions.h
+++ b/MAPL_Base/MAPL_Exceptions.h
@@ -154,7 +154,7 @@
 ! ESMF Error codes are defined here
 !                     
 
-#include "ESMF_ErrReturnCodes.inc"
+#include "ESMC_ReturnCodes.h"
```

This change did not break GCHP's build but I haven't validated it any further than that.

I've submitted PR #26 with this proposed fix.

Thanks in advance!

Liam

mathomp4:
@LiamBindle This is a good spot. I believe if we go back long ago, that `.inc` file was the only place to get ESMF error codes from ESMF. I mean *way* back in ESMF time. Also, Baselibs builds and installs ESMF in a wonky way because, again, I think it predates even the current ESMF make strategy. So there is a line in our build scripts that copies *all* of the `src/include` directory to our install path. 

Indeed, fixing this will make GEOS work better with `spack` if I can ever get that effort going. The spack build of ESMF I had to hack to install these extra inc files.

PR #26, opened 2019-08-13
Fixes 25. Changed how ESMF return codes are included in MAPL_Exceptions.h

LiamBindle:
Hi everyone, 

This is a PR for the possible fix that I discuss in #25. Let me know what you think.

Liam


mathomp4:
@tclune and @bena-nasa, I tested this in a build of `GEOSgcm` on the `dev/MAPL-2.0` branch and it seems zero-diff. I think it's a safe fix to take.

mathomp4:
Added (the newly created) @GEOS-ESM/mapl-team as a reviewer.

PR #33, opened 2019-08-29
Fixes #32 - Add option to pass in lm to ArrDescrInit

### LINKED ISSUE #32: ArrDescrInit does not init lm
The subroutine ArrDescrInit has not argument for the number of levels.

kgerheiser:
I create an ArrDescr in my NUOPC code and need to pass `lm` to it

### DIFF
--- a/MAPL_Base/MAPL_IO.F90
+++ b/MAPL_Base/MAPL_IO.F90
@@ -239,11 +239,12 @@ module MAPL_IOMod
 
   contains
 
-      subroutine ArrDescrInit(ArrDes,comm,im_world,jm_world,nx,ny,num_readers,num_writers,is,ie,js,je,rc)
+      subroutine ArrDescrInit(ArrDes,comm,im_world,jm_world,lm_world,nx,ny,num_readers,num_writers,is,ie,js,je,rc)
          type(ArrDescr), intent(INOUT) :: ArrDes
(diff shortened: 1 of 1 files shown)


MAPL_Base/MAPL_IO.F90
```diff
@@ -239,11 +239,12 @@ module MAPL_IOMod
 
   contains
 
-      subroutine ArrDescrInit(ArrDes,comm,im_world,jm_world,nx,ny,num_readers,num_writers,is,ie,js,je,rc)
+      subroutine ArrDescrInit(ArrDes,comm,im_world,jm_world,lm_world,nx,ny,num_readers,num_writers,is,ie,js,je,rc)
```
tclune: It would be better if lm_world could be an optional argument so that it does not break backward compatibility.  Unfortunately, rc is already optional, so doing this the "right way" (TM) would still force existing clients to add "rc=...".

kgerheiser:
There is only one other place in the code that uses this routine; restart_interp.F90 uses it in FVDyCoreCubed/. And I think having `lm` is the right decision because `lm_world` exists in the structure but isn't filled by this subroutine.

MAPL_Base/MAPL_IO.F90
tclune: ok

PR #35, opened 2019-09-04
add step size for average sun angle

weiyuan-jiang:


weiyuan-jiang:
I didn't see "Our high res model is already taking steps smaller than 300s". GEOSldas needs control of the step size. Now I pull request both on master and develop branch

PR #59, opened 2019-09-25
Fix i-server crash

bena-nasa:
I forgot to put the wait call in on the iclient when refactory this code. Causes a crash because data arrays being filled are not done being filled yet when used without the wait call. Also add a print so we know if the i-server is being started on dedicated resources.

### DIFF
--- a/MAPL_Base/MAPL_Cap.F90
+++ b/MAPL_Base/MAPL_Cap.F90
@@ -228,6 +228,12 @@ contains
            allocate(this%i_server, source = MpiServer(this%split_comm%get_subcommunicator(), s_name))
            call this%directory_service%publish(PortInfo(s_name,this%i_server), this%i_server)
            call this%directory_service%connect_to_client(s_name, this%i_server)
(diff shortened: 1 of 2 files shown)


weiyuan-jiang:
In the "if" statement, should this be (i) instead of (1)?

if (rank == 0 .and. this%cap_options%nodes_input_server(1) /=0 ) then
               write(*,'(A,I0,A)')"Starting pFIO input server on ",this%cap_options%nodes_input_server(i)," nodes"
            else if (rank==0 .and. this%cap_options%npes_input_server(1) /=0 ) then
               write(*,'(A,I0,A)')"Starting pFIO input server on ",this%cap_options%npes_input_server(i)," pes"
            end if

bena-nasa:
Yep, you you are right, also there is nothing to merge onto master as this code does not exist there.

bena-nasa:
Yep, you are right. It is changed. This code is only on develop, no need to get this in into master too

PR #75, opened 2019-10-24
speed up reading ascii tile files

weiyuan-jiang:
read ascii tile file in 2d by reading the line into the first dimension. The changes speed up the reading by about 100 times for a 200M tile file in openmpi.


weiyuan-jiang:
not necessary. changes were transferred to the dev/fix_tests branch

PR #112, opened 2019-11-26
Copy pointers for same grid in new IO layer

bena-nasa:
The new IO layer I have been working on has been always making a regridder between the input and output grids. Added logic to check if the grids are the same and just copy the pointers if they are indeed  on the same grid.


tclune:
I think there are reasons that this is not a good idea.     If/when we use asynchronous I/O,  you will need a separate copy to hold the "not-really-regridded" data anyway, so I convinced myself that the IdentityRegridder was a reasonably efficient option without too much complication.

Of course that was before we started using ESMF regridders.     Was there a measurable performance impact here?   I can imagine that ESMF was a wee bit expensive in computing the RouteHandle, but it should have been nearly as fast as a copy in subsequent operations.

PR #115, opened 2019-12-02
Fix MPI_Init call

mathomp4:
Test building on macOS found this issue. I'm assuming @bena-nasa wanted `MPI_Init` as the arguments aren't enough for `MPI_Init_Thread`

I'm not sure why Linux doesn't seem to care. 🤷‍♂


mathomp4:
Note I'm assuming 0-diff though I don't know how to test this.

PR #130, opened 2019-12-05
Feature/tclune/#119 use gfio shared in pfio

tclune:
Debatable as to which category of 0-diff this actually is.

Unfortunately, there as one use of containers in FV that relied upon the export from MAPL.   So in this weird instance, a PR on FV must precede approval of this PR.

Working on the other PR now, but need to log out.


mathomp4:
This should be based against `develop` not `master`, I think.

tclune:
Interesting.   I was able to change the base with the "Edit" button at the top.  In the past I've resorted to closing PR's and then resubmitting when I made this mistake.   Just in the off chance that others did not know this.

PR #131, opened 2019-12-05
Fixes #119.

### LINKED ISSUE #119: Leverage gFTL-shared in pFIO.
pFIO was developed before gFTL-shared was created.   But considerable simplification should now be possible.    

MAPL_Base should also be examined for adopting gFTL-shared as part of this change.

tclune:


mathomp4:
@tclune Is this zero-diff? I'm assuming it is, and I've enabled the "required labels" here in MAPL (though I'm using the same labels as elsewhere...perhaps they should be different).

PR #134, opened 2019-12-05
Changed intent(in) and intent(out) to intent(inout) for MAPL_VarReadN…

aoloso:
Changed intent(in) and intent(out) to intent(inout) for MAPL_VarReadNCpar routines. A number of the routines previously had intent(in) for array A that should at least be intent(out)

### DIFF
--- a/MAPL_Base/MAPL_IO.F90
+++ b/MAPL_Base/MAPL_IO.F90
@@ -5157,7 +5157,7 @@ module MAPL_IOMod
 
     type (MAPL_NCIO)          , intent(IN   ) :: ncioObj
     character(len=*)            , intent(IN   ) :: name
-    real(kind=ESMF_KIND_R8)     , intent(IN   ) :: A(:,:,:)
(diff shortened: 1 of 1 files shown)


aoloso:
This should have been meant for the "develop" branch in the meantime.

MAPL_Base/MAPL_IO.F90
```diff
@@ -5157,7 +5157,7 @@ subroutine MAPL_VarWriteNCpar_R8_3d(ncioObj, name, A, ARRDES, RC)
 
     type (MAPL_NCIO)          , intent(IN   ) :: ncioObj
     character(len=*)            , intent(IN   ) :: name
-    real(kind=ESMF_KIND_R8)     , intent(IN   ) :: A(:,:,:)
+    real(kind=ESMF_KIND_R8)     , intent(INOUT) :: A(:,:,:)
```
mathomp4: Query: Why would a *write* routine need INOUT on the array we are writing?

PR #135, opened 2019-12-05
Changed intent(in) and intent(out) to intent(inout) for MAPL_VarReadN…

aoloso:
…Cpar routines

### DIFF
--- a/MAPL_Base/MAPL_IO.F90
+++ b/MAPL_Base/MAPL_IO.F90
@@ -5157,7 +5157,7 @@ module MAPL_IOMod
 
     type (MAPL_NCIO)          , intent(IN   ) :: ncioObj
     character(len=*)            , intent(IN   ) :: name
-    real(kind=ESMF_KIND_R8)     , intent(IN   ) :: A(:,:,:)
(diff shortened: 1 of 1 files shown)


mathomp4:
Note: If there is a bug in `master` as #134 might suggest, I think we'd want to pull there first, then into `release` and then into `develop` so we keep the flow the right direction for future merges back the other way.

tclune:
We thought about putting this on master, but it is not impacting anyone.   Basically because we are using ```include 'mpif.h'``` instead of ```USE MPI``` the compiler is not able to detect the standard violation.  At least for contiguous arrays.    For noncontiguous arrays, where a copy in/out becomes necessary, the compiler is saying "Oh, it's INTENT(IN), so there's no reason to do the copy out."

But I'm happy to switch course and ask Hamid to add it to master first.

aoloso:
I have corrected the mistake of VarWrite routines by changing intent(inout) back to intent(in).

tclune:
@mathomp4 GitHub is claiming that Required labels have not been applied.  But ... they have.  ???

PR #172, opened 2019-12-18
Expand/simplify ExtData DEBUG_LEVEL prints and reformat for log reada…

lizziel:
…bility

With this update any DEBUG_LEVEL value greater than 0 will cause maximum
debug verbosity in ExtData. Previously values up to 20 would result in
different levels of verbosity, but the tiers were not documented anywhere.
Setting DEBUG_LEVEL to 1 in ExtData.rc to turn on all prints is a simpler
approach and generally results in finding bugs in input files faster.

Signed-off-by: Lizzie Lundgren <elundgren@seas.harvard.edu>


lizziel:
These updates originally went into GCHP earlier this year. I do not have release/2.0.0 working yet with https://github.com/geoschem/gchp_ctm/ so this has not been fully tested in the new version. All updates are prints so there should be zero diffs.

bena-nasa:
There appears to be a typo as it fails when I pulled this branch and tried to build. It is failing on line 3230. There is a trim() with no arguments. I can't quite see what it is you wanted to trim.

lizziel:
That is supposed to have fdata%get_file_name. With my older version of MAPL I used cfio%fname but cfio was replaced with fdata for that subroutine in 2.0, and then I botched the update for compatiblity. I'll update shortly.

PR #173, opened 2019-12-19
Add equation of time

dr0cloud:
Add optional equation of time correction to solar position used to calculate solar zenith angle.
Defaults to off. zero-diff if not enabled.


MAPL_Base/sun.H
```diff
@@ -131,30 +165,80 @@
...
             ZTH = max(ZTH, 0.0)
+
+            ! normalized downward solar flux at TOA
             SLR = ZTH*AA
 
             if(present(DIST)) DIST = AA
 
+            ! ---%---
+
+! pmn: Andrea said the following fixed cases are for single column tests,
```
tclune: Might want to put Andrea's full name.   Someday people may pull us out of retirement to ask us why something was done in such-and-such manner.

dr0cloud:
Tom, just sent you an email … Can we talk by phone … 240-505-1476 … Peter.

MAPL_Base/sun.H
tclune: Added a change in the log, and added Andrea's name as long as I was in there.

PR #206, opened 2020-01-21
Fixes #205. Adds mepo support

### LINKED ISSUE #205: Add mepo support to MAPL
MAPL should be able to use `mepo` like all fixtures.

mathomp4:
This commit adds support for `mepo`. At present, `mepo` is
"automatically" used only if `-DUSE_MEPO=ON` is passed in. For now we
default to `checkout_externals`.

Also, add some logic so that we don't try to clone `@cmake` *again* if
it already exists. That causes some ugly error output.

Finally, add some directories to `.gitignore` as well as add ESMA_env to
the `Externals.cfg`. This is *not* needed to build MAPL, but it is sort
of "expected" when building at NCCS/NAS. It's just some extra noise for
our Harvard friends though. Sorry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,7 @@
 *~
-/@cmake
+/@cmake/
+/@env/
+/BUILD/
+/build*/
+/install*/
+/.mepo/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ project (
   VERSION 2.0
   LANGUAGES Fortran CXX C)  # Note - CXX is required for ESMF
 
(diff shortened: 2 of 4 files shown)


mathomp4:
Note: I put a blocking label on this. The reason is that Harvard has said they don't use `checkout_externals` so, per discussion with @tclune, I might push a commit so that `mepo` is the default tool used instead of the optional one.

mathomp4:
Okay. I've moved to use `mepo` as the default "if nothing else is checked out" repo getter. I've removed the blocking flag as well.

PR #223, opened 2020-02-07
Added missing stretched-grid support to MAPL_ExtDataGridChangeLev()

LiamBindle:
Hi everyone,

I've been testing stretched-grid in GCHPctm and saw that some of emissions we're being regridded properly. The affected emissions were 3D emissions that *did not* have 72 vertical layers. These emissions were regridded onto a cubed-sphere that was missing the stretching transformation (i.e. the grid they were regridded onto was otherwise correct).

I tracked this down to the `MAPL_ExtDataGridChangeLev()` subroutine in `MAPL_ExtDataGridCompMod.F90`. It appears this subroutine is missing stretched-grid support. This PR adds stretched-grid support to `MAPL_ExtDataGridChangeLev()`.

Below are some plots of before and after this PR. Sorry they are a bit hard to see--I pulled them out of a benchmark file which auto-generates a bunch of plots. I'm just trying to show the difference between the LHS and RHS before and after this PR.

## Before
Below are some figures comparing a C48 simulation (Ref) to a C32 simulation with a stretch factor of 1.5 (Dev). These simulations were both ran with GCHPctm 13.0.0-alpha.0.

image

Note that if I undid the Dev simulation's stretch transformation before plotting, the emissions were in good agreement.

## After

Below are simulations ran with GCHPctm 13.0.0-alpha.0 + this PR.

image

Thanks,

Liam

### DIFF
--- a/MAPL_Base/MAPL_ExtDataGridCompMod.F90
+++ b/MAPL_Base/MAPL_ExtDataGridCompMod.F90
@@ -33,7 +33,7 @@
    use ESMF_CFIOUtilMod
    use MAPL_CFIOMod
    use MAPL_NewArthParserMod
-   use MAPL_ConstantsMod, only: MAPL_PI,MAPL_PI_R8
+   use MAPL_ConstantsMod, only: MAPL_PI,MAPL_PI_R8,MAPL_RADIANS_TO_DEGREES
    use MAPL_IOMod, only: MAPL_NCIOParseTimeUnits
    use MAPL_regridderSpecMod
    use, intrinsic :: iso_fortran_env, only: REAL64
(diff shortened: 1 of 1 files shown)


LiamBindle:
Sorry, I'm not familiar with the GMAO's PR naming convention. Should this be "Non 0-diff"?

atrayano:
Technically speaking, the answer should be “yes”. But since nobody at GMAO is using the stretched grid regularly, it really does not matter


On Feb 7, 2020, at 11:01 AM, Liam Bindle <notifications@github.com<mailto:notifications@github.com>> wrote:


Sorry, I'm not familiar with the GMAO's PR naming convention. Should this be "Non 0-diff"?

LiamBindle:
Sounds good. Let me know if there's anything you would like me to change.

PR #226, opened 2020-02-07
Updated CMakeLists to work with GEOS-ESM/ESMA_cmake#57

LiamBindle:
This PR accompanies https://github.com/GEOS-ESM/ESMA_cmake/pull/57

### DIFF
--- a/MAPL_Base/CMakeLists.txt
+++ b/MAPL_Base/CMakeLists.txt
@@ -22,7 +22,7 @@ set (srcs
   MAPL_GenericCplComp.F90  MAPL_Sort.F90
   MAPL_Generic.F90         MAPL_sun_uc.F90        MAPL_SatVapor.F90
   MAPL_HistoryCollection.F90
-  MAPL_HistoryGridComp.F90 MAPL_CapOptions.F90    MAPL_FlapCapOptions.F90 
+  MAPL_HistoryGridComp.F90 MAPL_CapOptions.F90    $<$<BOOL:${FLAP_FOUND}>:MAPL_FlapCapOptions.F90> 
   MAPL_Cap.F90         hash.c
   MAPL_CapGridComp.F90
   MAPL_GridType.F90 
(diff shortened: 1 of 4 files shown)


MAPL_Base/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set (srcs
   MAPL_GenericCplComp.F90  MAPL_Sort.F90
   MAPL_Generic.F90         MAPL_sun_uc.F90        MAPL_SatVapor.F90
   MAPL_HistoryCollection.F90
-  MAPL_HistoryGridComp.F90 MAPL_CapOptions.F90    MAPL_FlapCapOptions.F90 
+  MAPL_HistoryGridComp.F90 MAPL_CapOptions.F90    $<$<BOOL:${FLAP_FOUND}>:MAPL_FlapCapOptions.F90>
```
tclune: I'm weak on CMake generator expressions.  Does this construct build the source file only if FLAP is found?  If so, very nice, though might be good to put it on a separate line if we need to make any other changes.

LiamBindle: Yes that's correct. Personally, I prefer them to `if()` blocks because I find they simplify the logic of CMakeLists (e.g. you'd need a new variable to accomplish this otherwise).

Specifically, `$<BOOL:${FLAP_FOUND}>` evaluates `${FLAP_FOUND}` to 0 or 1 according to [CMake's if logic](https://cmake.org/cmake/help/latest/command/if.html). Then the ouside generator expression evaluates to "MAPL_FlapCapOptions.F90" iff the inside generator expression was 1.

I just added it to a new line. How does that look?

LiamBindle: `FLAP_FOUND` is the standard variable set by `find_package(FLAP ...)` (i.e. `find_package(XXX)` sets `XXX_FOUND` according to [here](https://cmake.org/cmake/help/v3.0/command/find_package.html))

mathomp4:
Huh. Was there a reason this was pulled into `release/2.0.0`? I'm pretty sure this means that branch is now unbuildable (on its HEAD) without GEOS-ESM/ESMA_cmake#57

tclune:
mea culpa  - we'll deal with this once we've settled the cmake PR.

PR #261, opened 2020-03-19
Fixes #260. The logic in timermodeset is just wrong.

### LINKED ISSUE #260: Funny logic in MAPL_ProfilerMod.F90, subroutine MAPL_SetTimerMode
Running the CTM/GCM sequentially uncovered funny logic in MAPL_SetTimerMode. The module has a global variable timerMode. This routine is supposed to set it which it does. However, the routine has an _ASSERT call before setting it and asserts that the mode is in the valid range. However the range is not big enough. We must have added the new timer mode but not adjusted this check accordingly.

bena-nasa:
Fixes #260 

This is a zero diff change. No effect on the model, just incorrect logic running the GCM/CTM sequentially through NUOPC uncovered.

### DIFF
--- a/MAPL_Base/MAPL_Profiler.F90
+++ b/MAPL_Base/MAPL_Profiler.F90
@@ -352,8 +352,8 @@
       character(len=ESMF_MAXSTR), parameter :: IAm="MAPL_ProfModeSet"
 
       ! Sanity check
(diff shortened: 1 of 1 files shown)


tclune:
Does this actually fix William's problem?     I agree the ASSERT after the assignment is better, but ...

(And I actually think the real fix should have been to keep the ASSERT before the assignment but to use the dummy argument rather than timerMode.   I.e. the assertion checks that the argument is in range.   Probably not worth another PR given that we're in the midst of switching to a new profiler utility.)

PR #294, opened 2020-04-10
Re-enable conservative regrid of vector quantities in ExtData

sdeastham:
## Description

Removes a logical check which halts the model if conservative regridding of vector quantities.

## Motivation and Context

Although it is not clear that conservative regridding of vector quantities is scientifically valid, it is not clear if (in the case of winds) vector regridding causes smaller or larger errors in terms of conservation of divergence. Meanwhile the use of conservative regridding does help to avoid errors in regridding for unusual grid definitions.

## Types of changes

## Checklist:


sdeastham:
Ack - made a mess of this. Will revise it and re-do as a single commit, not this giant stack of commits.

bena-nasa:
oops, didn't see that you had closed it.

PR #299, opened 2020-04-13
re-arrange timer

weiyuan-jiang:
## Description

Re-arrange and clean up time profiler 
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [x] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [ ] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #300, opened 2020-04-13
Revert "re-arrange timer"

tclune:
Reverts GEOS-ESM/MAPL#299

@weiyuan-jiang reports that there were problems with this merge.


weiyuan-jiang:
No need any more

Issue #301, opened 2020-04-13
Vertical flipping in ExtData fails for edge-aligned 3-D arrays


sdeastham:
When reading in data, ExtData gives grid-centered data indexing of 1:LM, but gives grid-edge data indexing of 0:LM. The ExtData vertical flipping subroutine however assumes a 1:LM indexing. This results in memory corruption if a grid-centered array needs to be flipped.

mathomp4:
Fixed by #302

PR #302, opened 2020-04-13
Fix vertical flipping for 0-indexed arrays

sdeastham:
## Description
When flipping vertical arrays, ExtData is implicitly assuming that the array indexing starts from 1. However, 3-D arrays which are defined at the grid edges (e.g. cloud mass flux) are indexed beginning at 0 (i.e. TOA or surface). This commit makes the routine robust by explicitly flipping the arrays based on lbound:ubound, rather than 1:size.

## Related Issue
Resolves https://github.com/GEOS-ESM/MAPL/issues/301.

## Motivation and Context
Fixes a bug when reading in meteorological data which is specified as positive=up.

## How Has This Been Tested?
Simulation in GCHP.

## Types of changes

## Checklist:


tclune:
@bena-nasa Please review asap.   I want this to go in 2.1 release.

PR #305, opened 2020-04-14
Fixes #412 - Integrate support for pFlogger

tclune:
Added support to use pFlogger for logging
  - Command line option: --logging_config=<file>
  - Each grid comp now has an associated logger with name given by the
    ancestry.  E.g.,   CAP.GCM.AGCM.PHYSICS
  - Added warning message:
      When not providing pFlogger a yaml file, a warning message is
      emitted (to last_resort handler most likely).

## Motivation and Context

Loggers are good and useful and the only thing that matters in life.

## How Has This Been Tested?

Simple run of GCM at coarse resolution

## Types of changes

## Checklist:


mathomp4:
I ran some tests with Intel and GCC. I ran a case with a `logging.yaml` and without one.

Both logging and no-logging throw FAIL messages on every PE:
```
pe=00001 FAIL at line=02807    MAPL_Generic.F90                         <status=1>
pe=00001 FAIL at line=00110    MAPL_CapGridComp.F90                     <status=1>
```

When you run without logging you also get a warning on every PE:
```
WARNING: MAPL: No configure file specified for logging layer.  Using defaults.
```

mathomp4:
Everything seems good with this fix. Both Intel and GCC 9 seem to run. I only did `Release` though I imagine @tclune tested with `Debug` as he found that MemUtils bug.

mathomp4:
Note. I also did a set of C48 runs of AMIP, Replay, and Replay-no-inc. They all ran without issue, which is nice. I also did gcm_regress in all three with successes. So if nothing else, we are good.

PR #306, opened 2020-04-14
Added % memory used, and skipped init for estimating run time remaining

wmputman:
I have updated the memory tracker to include the percent memory used in addition to committed, given that we (FMS) can and often do over commit memory without using a large percentage of the memory.

I also updated the estimated time remaining to ignore the model initialization.

This has been running regularly in my daily tests

## Types of changes

## Checklist:


tclune:
@wmputman I'm going to make this a merge into the develop branch as it is not a bug fix.   But the good news is that we are rolling out MAPL 2.1 shortly and this will go in.  So when Scott gets the updated MAPL it will be there for everyone.

Issue #293, opened 2020-04-10
Segfault with MAPL_Profiler and GNU


mathomp4:
Last night's GNUMAPLDEV test threw an error with MAPL_Profiler:
``` 
Using parallel NetCDF for file: fvcore_internal_checkpoint

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x2aaab5a7090f in ???
#1  0x2aaab5b65da8 in ???
#2  0x4e11ced in __mapl_multicolumn_MOD_get_header
        at /discover/nobackup/mathomp4/SystemTests/builds/AGCM_GNUMAPLDEV/CURRENT/GEOSgcm/src/Shared/@MAPL/MAPL_Profiler/MultiColumn.F90:134
#3  0x4e195d5 in __mapl_profilereporter_MOD_generate_report_profiler
        at /discover/nobackup/mathomp4/SystemTests/builds/AGCM_GNUMAPLDEV/CURRENT/GEOSgcm/src/Shared/@MAPL/MAPL_Profiler/ProfileReporter.F90:47
#4  0x4a0e629 in report_generic_profile
        at /discover/nobackup/mathomp4/SystemTests/builds/AGCM_GNUMAPLDEV/CURRENT/GEOSgcm/src/Shared/@MAPL/MAPL_Base/MAPL_Generic.F90:2232
#5  0x4a0e629 in __mapl_genericmod_MOD_mapl_genericfinalize
        at /discover/nobackup/mathomp4/SystemTests/builds/AGCM_GNUMAPLDEV/CURRENT/GEOSgcm/src/Shared/@MAPL/MAPL_Base/MAPL_Generic.F90:2185
```
The line in question is:
https://github.com/GEOS-ESM/MAPL/blob/ed08e9bd430a3805e41aa176110c5e2ac5a4e5d9/MAPL_Profiler/MultiColumn.F90#L134
I'm going to try and stare at the code, but I'm guessing @weiyuan-jiang or @tclune will have better luck 😄

mathomp4:
Annoying things.

One, it looks like the latest Arm Forge doesn't work on SLES 11, so I'll need to rebuild there.

Two, it looks like whatever is happening isn't part of the unit tests. Those all pass.

mathomp4:
Added in some prints, and the segfault moved. Prints I added:
```diff
[...]

PR #309, opened 2020-04-14
fix GNU

weiyuan-jiang:
works for both Intel and GNU compilers
## Types of changes

## Checklist:

### DIFF
--- a/MAPL_Profiler/TextColumn.F90
+++ b/MAPL_Profiler/TextColumn.F90
@@ -103,8 +103,8 @@ contains
    ! issue with allocatable arrays of deferred length strings.
    subroutine get_separator(this, separator, k)
       class(TextColumn), intent(in) :: this
-      character(*), intent(inout) :: separator(k)
       integer, intent(in) :: k
+      character(*), intent(inout) :: separator(k)
 
       integer :: w
       character(1) :: c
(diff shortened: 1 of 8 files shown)


weiyuan-jiang:
solve issue #293

MAPL_Profiler/TextColumn.F90
```diff
@@ -101,9 +101,10 @@ end subroutine set_separator
 
    ! Would be a function, but this is a workaround for gfortran 8.2
    ! issue with allocatable arrays of deferred length strings.
-   subroutine get_separator(this, separator)
+   subroutine get_separator(this, separator, k)
       class(TextColumn), intent(in) :: this
-      character(*), intent(inout) :: separator(:)
+      character(*), intent(inout) :: separator(k)
+      integer, intent(in) :: k
```
mathomp4, line 107: Ooh. You might want to flip these around. I know I've used a compiler that didn't like it in this order. Aka:
```fortran
real :: x(n)
integer :: n
```
failed but:
```fortran
integer :: n
real :: x(n)
```
succeeded.

But maybe that doesn't matter anymore?

weiyuan-jiang: If it is safer to flip, why not ? done !

PR #334, opened 2020-04-21
Hotfix/bmauer/fixes#327

### LINKED ISSUE #327: Error with replay and periodic update frequency in ExtData.
The CO2/Gocart developers have alerted me to an issue. One mode for ExtData is to update the import field at a specific frequency say 3 hours. This is used by the CO2_GridComp in GOCART with the CMS_EMIS: 1 option. This is handled by creating an alarm in ExtData. They were finding that in replay after rewinding the clock, entries being handled this way were not getting updated. For example if we start at 21z, it updates pointer with the 21z data, then runs the predictor segment. After this is done the clock is rewound to 21z and ExtData runs again to "reset" itself to the state before the predictor segment. Fields that are updated at a frequency like this (rather than every timestep, once a day thru the template mechanism) do not get "reset". 

After some searching there are 2 things going on. First when the alarm is created the ringtime was not set to the current time of the clock. Meaning the alarm does not ring until one ringinterval in. I.E. if I start the clock at 21z and the ring interval is 3 hours, and clock ticks in 15 minute increments, it does not ring until 0z. However, fixing that by setting the ringtime in the alarm when it is created brings up another problem. The alarm does now ring at 21z BUT after rewinding the clock after the predictor back to 21z, restoring the alarms, the alarm is ringing, but then it is still ringing the next time step at 21:15z  when it should not be, and then is not ringing 2 time steps from then at 21:30z. Note this alarm was created with sticky = false so it should not be persisting in ringing between steps.

The solution to this for now is to just give up on the alarm in ExtData as this seems like  flat out ESMF bug. I have coded a simple alarm like class in MAPL that simply remembers a start time, a time interval and when calling it's one method, you pass it a time, and if the current interval, relative to the start time is not divisible by the interval stored in the object you are not ringing. Else you are. This is all the functionality I need for ExtData.

bena-nasa:
Fixes #327 

This fixes a bug when specifying a time interval as the refresh template in ExtData
## Types of changes

## Checklist:

- [ X] I have tested this change with a run of GEOSgcm (if non-trivial)
- [ X] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [X ] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 ### Fixed
+- Fixed bug during replay when the refresh template in ExtData is a time interval
 ### Removed
 ### Added
 
--- a/MAPL_Base/CMakeLists.txt
+++ b/MAPL_Base/CMakeLists.txt
@@ -62,7 +62,7 @@ set (srcs
   MAPL_NUOPCWrapperMod.F90 MAPL_CubedSphereGridFactory.F90
   MAPL_ExtDataCollection.F90 MAPL_ExtDataCollectionManager.F90
   FileMetadataUtilities.F90 FileMetadataUtilitiesVector.F90
-  )
+  MAPL_SimpleAlarm.F90)
 
 esma_add_library(
   ${this} SRCS ${srcs}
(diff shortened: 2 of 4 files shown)


bena-nasa:
I don't think so. We don't support the ability to have fractional seconds in the start time or heartbeat so until that happens I don't see that we would need this.

Issue #336, opened 2020-04-22
Handling for mis-specification of latitude centers in archived GEOS collections


sdeastham:
The archived MERRA-2 and GEOS-FP data collections used to drive GEOS-Chem specify the first and last latitudes as being at (90 - dlat/2) degrees N/S, rather than at 90 N/S. This results in the grids being misinterpreted as "regional" by MAPL when read in through ExtData, and results in incorrect regridding at the poles. This issue is to request a small modification to the lat-lon metadata reading routines in MAPL so that these files can still be used safely. Although this could be resolved by file reprocessing, many GEOS-Chem users have already acquired large chunks of this dataset. While future files will *not* include this error, we would like to enable backwards compatibility.

PR #337, opened 2020-04-22
Targeted fix for incorrectly specified latitudes in archived data

sdeastham:
## Description
Some reprocessed GEOS-Chem input data specifies the first and last latitudes for pole-centered data as instead being at (90 - dlat/2) N/S. This results in MAPL treating the data as being regional, which in turns results in incorrect bilinear interpolation of (eg) wind data at the poles - giving non-physical results. This fix checks specifically for this incorrect latitude specification, correcting it to a conventional "pole-centered" specification if detected.

## Related Issue
This relates to a downstream issue when MAPL is used in GCHP,  It should have no effect outside of GCHP. The issues is described in https://github.com/GEOS-ESM/MAPL/issues/336.

## Motivation and Context
When reading in (eg) wind data from MERRA-2 or GEOS-FP data which was previously archived for use in GEOS-Chem, the incorrectly specified polar latitudes result in incorrect regridding.

## How Has This Been Tested?
GCHP was compiled with and without this fix. Implementing the fix resulted in a small change in polar ozone. It is labeled as zero diff on the basis that it should have no effect on GEOS or GEOSctm.

## Types of changes

## Checklist:


bena-nasa:
Seb,
I think I follow what is going on. Here you are taking the 2nd point and the 2nd to last point and subtracting or adding a d_lat to extrapolate the pole. I'm confused about this code:
`           if (is_valid) then
              ! Should the southernmost point actually be at the pole?
              extrap_lat = this%lat_centers(i_min+1) - d_lat
              if (extrap_lat <= ((d_lat/20.0)-90.0)) then
                 this%lat_centers(i_min) = -90.0
              end if
              ! Should the northernmost point actually be at the pole?
              extrap_lat = this%lat_centers(i_max-1) + d_lat
              if (extrap_lat >= (90.0-(d_lat/20.0))) then
                 this%lat_centers(i_max) =  90.0
              end if
           end if`

Why did you do the this:
 extrap_lat >= (90.0-(d_lat/20.0)) 
rather than just:
 extrap_lat >= 90.0
for example?
You are just trying to test if you have gone over +90 in this case, if so just set the top lat to 90? If in your case you have a dt between every point except the first/second and second-to-last/last where you have a dt/4 (this is what you files are like right?), I'm not see what the 90.0-(d_lat/20.0) does.

sdeastham:
Hi Ben,

Basically, I don’t quite trust floating point math. The overall idea is exactly as you said: the is_valid test beforehand made as sure as possible that the grid spacing everywhere (away from the poles) is regular, so the final test is to see if – when extrapolating using that same delta – the first and last points should be at the poles. In theory then the “extrap_lat” should be at exactly 90 – but I REALLY don’t trust floating point to get exactly anything right!  The delta_lat could well be very slightly under or very slightly over the true value. That could then mean that sometimes extrap_lat = 89.999999, which would fail the test “extrap_lat >= 90.0”. Because the spacing elsewhere is totally uniform, I’m thinking it’s reasonable to instead assume that if extrap_lat is within a fraction of a grid delta of the pole, it should probably be at the pole. Does that make sense?

Regards,

Seb

tclune:
My review from earlier this morning disappeared.  Going to do it again ...

MAPL_Base/MAPL_LatLonGridFactory.F90
```diff
@@ -775,6 +780,38 @@ subroutine initialize_from_file_metadata(this, file_metadata, unusable, rc)
...
+                    (size(this%lat_centers)-3)
+           is_valid = .True.
+           ! Check: is this a regular grid (i.e. constant spacing away from the poles)?
+           do i=(i_min+1),(i_max-2)
+              d_lat_temp = this%lat_centers(i+1) - this%lat_centers(i)
+              is_valid = (is_valid.and.(abs((d_lat_temp/d_lat)-1.0) < 1.0e-5))
+              if (.not. is_valid) then
+                 exit
+              end if
+           end do
```
tclune, line 799: This loop can be simplified.  Something like:
```f90
is_valid = all(abs((this%lat_centers(i0+2:i1-1)-this%lat_centers(i0+1:i1-2))/d_lat)-1.0) < tolerance)
```

MAPL_Base/MAPL_LatLonGridFactory.F90
```diff
@@ -775,6 +780,38 @@ subroutine initialize_from_file_metadata(this, file_metadata, unusable, rc)
...
+           if (is_valid) then
+              ! Should the southernmost point actually be at the pole?
+              extrap_lat = this%lat_centers(i_min+1) - d_lat
+              if (extrap_lat <= ((d_lat/20.0)-90.0)) then
+                 this%lat_centers(i_min) = -90.0
```
tclune, line 804: This formulation is confusing in multiple respects.  In particular "20.0"  superficially looks like an angle in this expression.   It's real role is as a tolerance.  so something like
```f90
tolerance = 0.05
if (extrap_lat <= -90.0 + tolerance*d_lat) then
```

MAPL_Base/MAPL_LatLonGridFactory.F90
```diff
@@ -775,6 +780,38 @@ subroutine initialize_from_file_metadata(this, file_metadata, unusable, rc)
...
+              end if
+              ! Should the northernmost point actually be at the pole?
+              extrap_lat = this%lat_centers(i_max-1) + d_lat
+              if (extrap_lat >= (90.0-(d_lat/20.0))) then
```
tclune, line 808: Ditto here.

lizziel:
Couldn’t a simpler algorithm be, after verifying constant dlat spacing between internal lat_centers, then if (90 – lat_center_max < dlat) use 90, and if (90 – abs(lat_center_min) < dlat) use -90? Our grids would have the difference between lat center extremes and the poles be dlat/2. This shouldn’t get close to floating point issues. Or is this trying to capture cases beyond our GEOS-Chem global met grids?

Lizzie

--
Lizzie Lundgren
Senior Scientific Programmer
GEOS-Chem Support Team
geos-chem-support@g.harvard.edu<mailto:geos-chem-support@g.harvard.edu>

Please direct all GEOS-Chem support emails to the entire GEOS-Chem Support Team<http://wiki.geos-chem.org/GEOS-Chem_Support_Team>. If you are reporting a bug, please create an issue on GitHub<https://github.com/geoschem/>.

sdeastham:
I was getting excited about this but then realized that I think this would clobber pole-edge grids, right? For our “bad” pole-centered grids, the last latitude is at (90 – dlat/4), but for legitimate pole-edged grids the last latitude is at (90 – dlat/2).

bena-nasa:
I was confusing myself earlier.
It seams like lizzie suggestion will overwrite the first and last point the PE grid, but what is currently in the PR will overwrite the first and last point in a PC grid, of course to the value it is supposed to be.

lizziel:
Is it possible to identify if the grid is PE or PC and only apply the correction if it is PC?

tclune:
@lizziel Basically you do something similar to what Seb did.   You look at the delta of the last center from the pole.   If it is "nearly" 0, then it is  PC.   If it is "nearly" dlat/2 then it is PE.    Until now, I'd have thought that a tolerance of dlat/4 would be robust, but if there really are cases where more games are played with the final spacing ...

bena-nasa:
something like this?
`
        tolerance = 0.05
        if (size(this%lat_centers) >= 4) then
           ! Assume lbound=1 and ubound=size for now
           i_min = 1 !lbound(this%lat_centers)
           i_max = size(this%lat_centers) !ubound(this%lat_centers)
           d_lat = (this%lat_centers(i_max-1) - this%lat_centers(i_min+1))/&
                    (size(this%lat_centers)-3)
           is_even = .True.
           ! Check: is this a regular grid (i.e. constant spacing away from the poles)?
           do i=(i_min+1),(i_max-2)
              d_lat_temp = this%lat_centers(i+1) - this%lat_centers(i)
              is_even = (is_even.and.(abs((d_lat_temp/d_lat)-1.0) < tolerance))
              if (.not. is_even) then
                 exit
              end if
           end do
           if (is_even) then
              ! Should the southernmost point actually be at the pole?
              extrap_lat = -90.0d0 + d_lat/4.0d0
              if ( (extrap_lat/this%lat_centers(i_min)-1.0d0) < tolerance ) then
                 this%lat_centers(i_min) = -90.0
              end if
              ! Should the northernmost point actually be at the pole?
              extrap_lat = 90.0d0 - d_lat/4.0d0
              if ( (extrap_lat/this%lat_centers(i_max)-1.0d0) < tolerance  ) then
                 this%lat_centers(i_max) =  90.0
              end if
           end if
        end if
`

bena-nasa:
something like this?
```
        tolerance = 0.05
        if (size(this%lat_centers) >= 4) then
           ! Assume lbound=1 and ubound=size for now
           i_min = 1 !lbound(this%lat_centers)
           i_max = size(this%lat_centers) !ubound(this%lat_centers)
           d_lat = (this%lat_centers(i_max-1) - this%lat_centers(i_min+1))/&
                    (size(this%lat_centers)-3)
           is_even = .True.
           ! Check: is this a regular grid (i.e. constant spacing away from the poles)?
           do i=(i_min+1),(i_max-2)
              d_lat_temp = this%lat_centers(i+1) - this%lat_centers(i)
              is_even = (is_even.and.(abs((d_lat_temp/d_lat)-1.0) < tolerance))
              if (.not. is_even) then
                 exit
              end if
           end do
           if (is_even) then
              ! Should the southernmost point actually be at the pole?
              extrap_lat = -90.0d0 + d_lat/4.0d0
              if ( (extrap_lat/this%lat_centers(i_min)-1.0d0) < tolerance ) then
                 this%lat_centers(i_min) = -90.0
              end if
              ! Should the northernmost point actually be at the pole?
              extrap_lat = 90.0d0 - d_lat/4.0d0
              if ( (extrap_lat/this%lat_centers(i_max)-1.0d0) < tolerance  ) then
                 this%lat_centers(i_max) =  90.0
              end if
           end if
        end if
```

sdeastham:
I suppose that will more specifically get at the GEOS-Chem use case, although I haven't tested it. With that in mind I'd make one more adjustment to avoid a potential div-by-zero (in case a regional grid is erroneously being tested, that regional grid may have a lat center at zero) - i.e.
```
        tolerance = 0.05
        if (size(this%lat_centers) >= 4) then
           ! Assume lbound=1 and ubound=size for now
           i_min = 1 !lbound(this%lat_centers)
           i_max = size(this%lat_centers) !ubound(this%lat_centers)
           d_lat = (this%lat_centers(i_max-1) - this%lat_centers(i_min+1))/&
                    (size(this%lat_centers)-3)
           is_even = .True.
           ! Check: is this a regular grid (i.e. constant spacing away from the poles)?
           do i=(i_min+1),(i_max-2)
              d_lat_temp = this%lat_centers(i+1) - this%lat_centers(i)
              is_even = (is_even.and.(abs((d_lat_temp/d_lat)-1.0) < tolerance))
              if (.not. is_even) then
                 exit
              end if
           end do
           if (is_even) then
              ! Should the southernmost point actually be at the pole?
              extrap_lat = -90.0d0 + d_lat/4.0d0
              if ( abs(extrap_lat - this%lat_centers(i_min)) < (tolerance*d_lat) ) then
                 this%lat_centers(i_min) = -90.0
              end if
              ! Should the northernmost point actually be at the pole?
              extrap_lat = 90.0d0 - d_lat/4.0d0
              if ( abs(extrap_lat - this%lat_centers(i_max)) < (tolerance*d_lat) ) then
                 this%lat_centers(i_max) =  90.0
              end if
           end if
        end if
```

lizziel:
Thinking more about this, if we want to target the GEOS-Chem grid only then we could also apply a symmetry condition test. Min and max latitudes should always be equivalent in GEOS-Chem met grids. If floating point comparison is an issue then applying a very small tolerance on the difference, such as 1e-20, would work. This would exclude all PE grids, unless I am missing something.

sdeastham:
I don't think a symmetry check would help much unfortunately - it doesn't distinguish between PE/PC grids and I'd expect the majority of grids to be symmetrical whether correct or not. With that in mind, here's my suggested final version:

```
         ! Check: is this a "mis-specified" pole-centered grid (sometimes used in GEOS-Chem)?
         tolerance = 0.05
         if (size(this%lat_centers) >= 4) then
            ! Assume lbound=1 and ubound=size for now
            i_min = 1 !lbound(this%lat_centers)
            i_max = size(this%lat_centers) !ubound(this%lat_centers)
            d_lat = (this%lat_centers(i_max-1) - this%lat_centers(i_min+1))/&
                     (size(this%lat_centers)-3)
            ! Check: is this a regular grid (i.e. constant spacing away from the poles)?
            is_even = all(abs(((this%lat_centers(i_min+2:i_max-1)-this%lat_centers(i_min+1:i_max-2))/d_lat)-1.0) < tolerance)
            ! If so, perform final checks on each pole
            if (is_even) then
               ! Should the southernmost point actually be at the pole?
               extrap_lat = -90.0d0 + d_lat/4.0d0
               if ( abs(extrap_lat - this%lat_centers(i_min)) < (tolerance*d_lat) ) then
                  this%lat_centers(i_min) = -90.0
               end if
               ! Should the northernmost point actually be at the pole?
               extrap_lat = 90.0d0 - d_lat/4.0d0
               if ( abs(extrap_lat - this%lat_centers(i_max)) < (tolerance*d_lat) ) then
                  this%lat_centers(i_max) =  90.0
               end if
            end if
         end if
```

If that looks good I can try to edit the PR accordingly.

bena-nasa:
Seb, Why don't you update your branch with the latest version you proposed? I can't see that it will effect the standard PC or PE grids we used and is zero diff in my test.

PR #353, opened 2020-05-01
release logger mpi resource before mpi_finalize

weiyuan-jiang:
Free pFlogger mpi resources before mpi_finalize. This should go with the new pflogger lib.
## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
    gcc-build-env:
       docker:
-         - image: gmao/geos-build-env-gcc-source:6.0.11
+         - image: gmao/geos-build-env-gcc-source:6.0.12
       environment:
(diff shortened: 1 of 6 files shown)


weiyuan-jiang:
This pull request requires new pflogger base lib

mathomp4:
I've added a blocker to this so I can update the CI on the branch as well.

mathomp4:
I wonder if we should remove the workaround now for MPI_comm_set_errhandler? If this "solves" all the issues we were seeing for a while?

tclune:
If it solves the known issues, then yes you should remove the bandaid.  That way we'll know if a similar issue is (re) introduced down the road.

mathomp4:
All seems good to me. @tclune can you look over the changes? If you approve, we can get a new release out to Scott, et al.

PR #361, opened 2020-05-07
Feature/wjiang/mapl shared

weiyuan-jiang:
Make a separate MAPL_shared directory that can be shared with GMAP_pFIO  and MAPL_profiler.  All the *.h and *.H files are moved to include directory.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Trivial change (affects only documentation or cleanup)

## Checklist:

- [ ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [ ] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #362, opened 2020-05-07
Feature/bmauer/execptio library

bena-nasa:
Note, I'm putting a contingent tag on this to elicit feedback before actually merging:

In order for us to add error checking to MAPL_Profiler and as part of longer term MAPL restructuring efforts Tom suggested moving the exception handling to an independent library.
Here is my crack at it. I've moved the .h's and all things related to exception handling to a new library.

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [X ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [ X] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)


CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ add_subdirectory (MAPL_cfio MAPL_cfio_r8)
 
 add_subdirectory (GMAO_pFIO)
 add_subdirectory (MAPL_Profiler)
+add_subdirectory (MAPL_ExceptionHandling)
```
tclune, line 46: Let's call the new subdirectory "Shared"  And in any event no "MAPL_" on the subdirectory names.

MAPL_Base/CMakeLists.txt
```diff
@@ -66,7 +64,7 @@ set (srcs
 
 esma_add_library(
   ${this} SRCS ${srcs}
-  DEPENDENCIES MAPL_Profiler GMAO_pFIO MAPL_cfio_r4 pflogger gftl-shared FLAP::FLAP MPI::MPI_Fortran)
+  DEPENDENCIES MAPL_Profiler MAPL_ExceptionHandling GMAO_pFIO MAPL_cfio_r4 pflogger gftl-shared FLAP::FLAP MPI::MPI_Fortran)
```
tclune, line 67: "Shared"

MAPL_Base/MAPL_CFIO.F90
```diff
@@ -9,7 +9,7 @@
 #include "unused_dummy.H"
 
 module MAPL_CFIOMod
-   use MAPL_ErrorHandlingMod
+   use MAPL_ExceptionHandling
```
tclune, line 12: Not worth changing,   but ErrorReporting is probably more accurate than ExceptionHandling.   But the latter sounds more professional.

MAPL_ExceptionHandling/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+  MAPL_ErrorHandling.F90
+  MAPL_Throw.F90
+  )
+esma_add_library (${this} SRCS ${srcs} DEPENDENCIES MPI::MPI_Fortran)
+#esma_add_library (${this} SRCS ${srcs} DEPENDENCIES gftl-shared gftl MPI::MPI_Fortran)
+target_include_directories (${this} PRIVATE ${MAPL_SOURCE_DIR}/include)
+
+file (COPY MAPL_ErrLogMain.h DESTINATION ${esma_include}/${this})
+file (COPY MAPL_ErrLog.h DESTINATION ${esma_include}/${this})
+file (COPY MAPL_Exceptions.h DESTINATION ${esma_include}/${this})
```
tclune, line 14: These should move into an include subdirectory under the main MAPL.

Then just add  cmake:   target_include_directories(${this}  ... )

Note that you'll need  a $<BUILD_INTERFACE...> and a $<INSTALL_INTERFACE ...> clause.

MAPL_ExceptionHandling/MAPL_ErrorHandling.F90
```diff
@@ -0,0 +1,174 @@
...
+   logical function MAPL_VRFYt(A,text,iam,line,rc)
+      integer,           intent(IN ) :: A
+      character*(*),     intent(IN ) :: iam,text
+      integer,           intent(IN ) :: line
+      integer, optional, intent(OUT) :: RC
+        MAPL_VRFYt =  MAPL_VRFY(A,iam,line,rc)
+        if(MAPL_VRFYt) print *, text
+   end function MAPL_VRFYT
+
+   subroutine MAPL_Abort_comm_world
```
tclune, line 168: Was there a specific driver to make this distinction?  Just curious.

MAPL_ExceptionHandling/CMakeLists.txt
tclune: You will also need to create a small CMakeLists.txt in the new include directory to install these files.

weiyuan-jiang:
I am closing my PR #361 . I think we need to change the directory's name to MAPL_Shared and put more files into it ( I will do it in my PR).  Otherwise there will be too many small directories in the future.

tclune:
NOT MAPL_Shared.    Just "Shared"

MAPL does not need to advertise MAPL at every turn.  We need it in module names to protect the namespace.

MAPL_ExceptionHandling/CMakeLists.txt
mathomp4: @tclune I'm not sure we need `$<INSTALL_INTERFACE>` here, or are you forward-thinking? This [StackOverflow post](https://stackoverflow.com/a/58729533/1876449) seems to say that `$<INSTALL_INTERFACE>` is only used during `install(EXPORT)`.

MAPL_ExceptionHandling/MAPL_ErrorHandling.F90
bena-nasa: @tclune 
What distinction, the MAP_Abort? I thought the point was this error handling library sits at the lowest level. MAPL_Abort lives  MAPL_Comms.F90 so I would not use MAPL_Abort. I guess I should have removed MAPL_Abort from MAPL_Comms and just moved it here rather than duplicate the code.

bena-nasa: I'll just move MAPL_Abort into the erorrhandling module, nothing else was using it but the include files.

PR #373, opened 2020-05-14
Feature/atrayano/split4d

atrayano:
## Description

This PR adds full support for handling 4d fields in MAPL. The main feature is the ability to split the 4d fields into series of 3d fields in History 
## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:


MAPL_Base/MAPL_Generic.F90
```diff
@@ -4997,7 +4997,7 @@ subroutine MAPL_GenericStateClockOff(STATE,NAME,RC)
     !EOPI
 
     character(len=ESMF_MAXSTR), parameter :: IAm = "MAPL_GenericStateClockOff"
-    integer :: STATUS, n
+    integer :: STATUS !, n
```
tclune, line 5000: If "n" is no longer needed, it really should be deleted.   Or do you expect another sweep through the code after this review.

MAPL_Base/MAPL_HistoryGridComp.F90
```diff
@@ -857,12 +866,12 @@ subroutine Initialize ( gc, import, dumexport, clock, rc )
...
-             call WRITE_PARALLEL('Can not have a vector pair in a conservative collection')
-             _ASSERT(.false.,'needs informative message')
-          end if
-       end if
+!@@       if (associated(list(n)%field_set%vector_list)) then
+!@@          if (list(n)%conservative /= 0) then
+!@@             call WRITE_PARALLEL('Can not have a vector pair in a conservative collection')
+!@@             _ASSERT(.false.,'needs informative message')
+!@@          end if
+!@@       end if
```
tclune, line 874: Again if this code should not be used, please delete it (and the preceding comment.

MAPL_Base/MAPL_HistoryGridComp.F90
```diff
@@ -1409,6 +1414,17 @@ subroutine Initialize ( gc, import, dumexport, clock, rc )
...
+    !--------------------------
+    call split4dFields(rc=status)
+    _VERIFY(status)
+
+    do n=1,nlist
+       m=list(n)%field_set%nfields
+       allocate(list(n)%r4(m), list(n)%r8(m), list(n)%r8_to_r4(m), stat=status)  
+       _VERIFY(STATUS)
+    end do
+
```
tclune, line 1427: Don't change now,  but we really need to start getting in the habit of using Arlindo's __RC__ now that we have collectively agreed on it as the general approach.

MAPL_Base/MAPL_HistoryGridComp.F90
atrayano: Agree. This block could be removed altogether...

MAPL_Base/MAPL_HistoryGridComp.F90
atrayano: Sure. Since this is new code, and I expected to find bugs, it was easier to set breakpoints with the "old" style

MAPL_Base/MAPL_Generic.F90
atrayano: "n" is no longer needed. But there was already commented section below the declaration which uses "n". My intent was to make it easier if anyone want to go back and uncomment that particular section to have a "ready" declaration staement

mathomp4:
Updated branch with new `develop` should allow this to merge. Sorry about that.

PR #375, opened 2020-05-19
Added ability to check a process of a specific rank is running

WilliamJamieson:
This is similar to `MAPL_Am_I_Root` except it supports any process rank.

## Description
Adds a new MAPL function, `MAPL_Am_I_Rank` which is functionally identical to `MAPL_Am_I_Root` except it adds the possibility of checking that a process is of any rank. Note that if no rank is provided it defaults to identical behavior to `MAPL_Am_I_Root`.

## Related Issue

None

## Motivation and Context

I needed to be able to filter debugging prints to processes of rank other than the root rank in several places. So I created a new function to test if one is on a process of rank other than root. This seems like it can be useful in the future.

## How Has This Been Tested?

Builds and runs with the desired behavior of indicating when a process of rank other than root is running. This is entirely new and so it should be zero diff.

## Types of changes

## Checklist:


mathomp4:
Looks like GCC 9 had an issue:
```
/__w/MAPL/MAPL/MAPL_Base/MAPL_Comms.F90:328:2:

  285 |   function MAPL_Am_I_Rank_Vm(VM, rank) result(R)
      |  2
......
  328 |   function MAPL_Am_I_Rank_Only(rank) result(R)
      |  1
Error: Ambiguous interfaces in generic interface 'mapl_am_i_rank' for ‘mapl_am_i_rank_only’ at (1) and ‘mapl_am_i_rank_vm’ at (2)
```

PR #386, opened 2020-05-26
Cap reorganization part 2

bena-nasa:
Part 2 of moving stuff out of the cap. Now the logging and profiler stuff has been moved to a support module. Now mkiau, gsi, and other programs can use these. I could not think of something better than ApplicationSupport for the name. Other things like ProfilerAndLoggerSupport seemed too long.

## Types of changes

## Checklist:


bena-nasa:
I make a a MAPL_Initialize and MAPL_Finalize and updated. Now both the logger and profiler get initialized when the cap is created. I did not see any reason why the global profiling singleton could not be initialized there and it ran without problem. It seems like it should be there anyway as that is the place to start it if you want the full application time.

PR #400, opened 2020-06-02
Use MPIEXEC_EXECUTABLE

mathomp4:
Just a little thing I saw when reading CMake docs. Per [CMake documentation](https://cmake.org/cmake/help/latest/module/FindMPI.html), we should use `MPIEXEC_EXECUTABLE`


mathomp4:
> Hmm. I suppose I should propagate this to pflogger and pfunit?

Maybe. It looks like back in early CMake 3 days it was `MPIEXEC` but then it moved to be `MPIEXEC_EXECUTABLE`

PR #408, opened 2020-06-08
Bugfix/tclune/#407 bug in python acg

tclune:
## Description
Quick bugfix for @gmao-esherman 

## How Has This Been Tested?
Tested against sample input.   ACG is not used by the model yet.

## Types of changes


tclune:
Was submitted in haste before a 1:00 meeting.  Should have been 'develop' and Matt caught it as well.

PR #426, opened 2020-06-15
create string templating module

bena-nasa:
As part of the effort to wean us out of use the "little" cfio layer, we still need to be able to untemplate grads style strings. Currently this lives in "little" cfio so just making a new module to do this. Also added an ESMF option that can be used for a future round of refactoring as most of the time we get an esmf time and then just construct the integer representation.

## Types of changes

## Checklist:


tclune:
```string_template``` is a bad name for a procedure.  Procedures should have verbs.   Variables should have nouns.

PR #431, opened 2020-06-17
Clean up CHANGELOG and update CMakeLists.txt

mathomp4:
* Fix incorrect dates in `CHANGELOG.md`
* Clean up some blank lines
* Add `@env` as a subdirectory so files like `g5_modules` are installed in `install/bin` and some rc files get installed in `install/etc` (that's all @env `CMakeLists.txt` does)
* Update the version to 2.2.0 in `CMakeLists.txt` preemptively so we don't forget.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 *~
-/@cmake/
-/@env/
+/ESMA_cmake/
+/ESMA_env/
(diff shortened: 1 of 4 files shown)


mathomp4:
Added blocking tag as @tclune found some issues

PR #434, opened 2020-06-18
Fixes #433 - wildcard short names

### LINKED ISSUE #433: Allow wildcard in short name in ACG for grid comp specs
In GOCART carbonaceous components, different sources for fields come from different entries in ExtData.   Since ExtData does not know the component name, the information must instead be encoded in the actual field.

```
foo*
```
Should be expanded as
```
foo<gc-name>
```
(cc'ing @gmao-esherman )

tclune:
Some gridded components are able to be instantiated multiple times but
need to distinguish source files for emissions in ExtData.   For this
purpose the imported field name is "mangled" to include the name of
the gridded component.

E.g., a local variable "Foo" in GridComp "anthro" would be declared as
"Fooanthro".

Note that the name internal to the gridcomp is still just "Foo".

Only the code generator has been tested.  Not used yet in main GEOS branches.

## Types of changes

## Checklist:


tclune:
I'm more confident of the former than the latter.   (Found a web page with an example.)   But I did manually test and got the right results.

PR #447, opened 2020-06-22
Hotfix/esherman/mapl acg python2

gmao-esherman:
The changes in MAPL_GridCompSpecs_ACG.py will allow MAPL_GridCompSpecs_ACG.py to be used by python 2.x which is what is used in g5_modules.


mathomp4:
So this is a hotfix? Will we need a release right away?

PR #450, opened 2020-06-22
Update Changelog and version for 2.2.2

mathomp4:
So. I just learned you don't enable superpowers on Github because it can merge `develop` into `main` for you automatically. Bad @mathomp4 for doing that. 

But if there was a day to do it, today's the day.

This update just pulls down my CI changes. If all works, this should only build the MAPL tests and not the GCM.


mathomp4:
Also, the "cancel" bits I added to this don't really work for label changes, only pushes. It's a start. I'll look for more comprehensive solutions soon.

Issue #451, opened 2020-06-22
Compiler errors in MAPL_FlapCapOptions.F90 if not using FLAP


lizziel:
The MAPL/base file MAPL_FlapCapOptions.F90 includes a dependency on FLAP regardless of whether USE_FLAP is defined. This causes compiler errors when building MAPL 2.2 within GCHP which does not use FLAP. I got around this issue by surrounding most of the MAPL_FlapCapOptions module in a USE_FLAP ifdef. I am submitting a separate PR with this fix to go with this issue.

tclune:
Thanks.  Our intent was for their to be an option to bypass FLAP.   I've not recently looked into the details, but look forward to your PR.   Please do it as a hotfix - and we'll get it into a patch for 2.2.

PR #452, opened 2020-06-22
Bypass the majority of MAPL_FlapCapOptions.F90 if not using FLAP

lizziel:
## Description

Nearly the entire MAPL_FlapCapOptions module in enclosed with an USE_FLAP ifdef to avoid compile errors when not using FLAP.

## Related Issue

See issue https://github.com/GEOS-ESM/MAPL/issues/451.

## Motivation and Context

When using MAPL 2.2 in GCHP, which does not use FLAP, there are compiler errors in MAPL_FlapCapOptions.F90 due to the FLAP dependency.

## How Has This Been Tested?

I am testing only in the context of building GCHP without FLAP.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+
+- USE_FLAP ifdef within MAPL_FlapCapOptions module
+
 ### Changed
 
 - Updated Github Actions to not build GCM if trivial PR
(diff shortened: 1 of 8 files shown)


mathomp4:
PR broke MAPL build. Removing trivial label as full model build should be required.

lizziel:
Thanks all. I was surprised to see it broke the build, and the errors were a bunch of MPI errors seemingly unrelated to my change. Could you explain what happened?

tclune:
I'd like to know as well.  Seems most likely that it was not your change per se but some other change that we're unaware is folded in.  (e.g., system issue)

mathomp4:
Tested at C48 with a run that passed in a command-line option.

PR #453, opened 2020-06-22
Merge main back into develop

mathomp4:
After the misadventures today, `main` is ready to pull into `develop`.


mathomp4:
Huh. I see four checks passed on the iOS app. Seems odd.

PR #474, opened 2020-06-26
Pull main into develop for timer hotfix

mathomp4:
As per usual for Git Flow, pull `main` into `develop` to absorb hotfix.


mathomp4:
Well nuts. Need to do this by hand again. 😦

PR #510, opened 2020-07-24
Fixes #509 - intrument ExtData with pFlogger.

### LINKED ISSUE #509: Instrument ExtData with pFlogger
A common failure mode in GEOS is for something to go wrong in processing files in ExtData.   These are usually not bugs in the code per-se and the generated stack trace is not so useful as it cannot show which file experienced problems.

pFlogger would allow us to introduce debug() calls that can be activated at run time to better analyze these situations.

tclune:
Mostly straightforward but some items are still not addressed.  In
particular there is a block with ```#ifdef DEBUG``` that uses
```ESMF_StatePrint()```, but that can only go to STDOUT.

## Types of changes

## Checklist:


bena-nasa:
If we are going to do this. We should add a lgr%info to print the new files when the brackets get updated as that is what we want 95% of the time when there is an error.

tclune:
OK - but I'm concerned that ```%info(...)``` will be too verbose.   That's a lot of output in a chem config.

tclune:
@bena-nasa  OK - I've now marked this one as "ready for review".       I don't expect to do more work on it until you've had a chance to find additional problems.

mathomp4 (approved):

PR #528, opened 2020-08-26
Fixes tripolar metadata ouput

### LINKED ISSUE #527: Missing metadata in ocean history collections.
Following metadata are missing in ocean history collections after update to MAPL2:
1. Geographical 2d latitudes and longitudes in tripolar collections.
2. Depth levels in 3d collections.

Current post processing scripts depend on these metadata. Since all our history output is on A-grid, A-grid coordinates are sufficient.  Before MAPL2, tripolar collections were conformant with CF conventions, it would be nice if they stay conformant.

bena-nasa:
Fixes #527 
This will add 2D lons and lats if you are outputting on the native tripolar grid. (assuming the lons/lats in the grid are correct, but that is another issue being worked on, they should be good for at least MOM5, MOM6 maybe not until we do a little more work). This only addresses the requests in the issue Yuri opened.

In addition it restores the functionality that you can specify the vertical metadata even if not selecting a subset of levels or vertically regridding as the old History/CFIO could do.

This is bit a of a kluge, but then the behaivor in the old CFIO layer was just as klugy. 
Basically when creating the file vertical information you pass the levels. It was assuming that if you passed levels, but no vertical variable for interpolation you wanted to select levels. Of course this breaks the cases they were using in the tripolar output of passing in a file that has the level metadata.
In the old CFIO layer the Kluge was, if the values were negative the level array, don't use that as an index to do level selection.
The new kluge is this. If you just pass levels assume at first we want to select levels, that is the existing behaivour. However after we parse the bundle and if we find that the number of levels in the bundle fields matches the input level variable, don't do try to do level subsetting with the passed in level variable because why would you be trying to select levels if the number of levels passed in matches the size of the variable, makes no sense. Since at this point the levels have been written to the metadata we can obliterate them and go back to the no special vertical handling behavior and output all the levels.

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_TripolarGridFactory.F90
+++ b/base/MAPL_TripolarGridFactory.F90
@@ -65,8 +65,6 @@ module MAPL_TripolarGridFactoryMod
 
       procedure :: equals
 
-      procedure :: read_grid_dimensions
-      procedure :: read_grid_coordinates
       procedure :: check_and_fill_consistency
       procedure :: generate_grid_name
       procedure :: to_string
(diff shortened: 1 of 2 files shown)


bena-nasa:
> @bena-nasa Not really part of this change, but ... as long as changes are being made in this file anyway ....
> 
> Could you update the netcdf calls to use the `nf90` interfaces and switch the `include netcdf.inc` to `use netcdf`?

I'm not sure what those routines that had the include netcdf.inc were even for. They were were there is the initial commit but I can find no where they are ever used so I removed them.

tclune:
@bena-nasa Even better!    I was assuming there was some reason that pFIO was insufficient.

PR #533, opened 2020-08-31
Update ServerManager.F90

weiyuan-jiang:
typo and bug fix


weiyuan-jiang:
This has been "fixed" in many branches of mine. I realize it may take some time to push my branches so I fixed it here.

weiyuan-jiang:
The build is not supposed to fail. @mathomp4 , is this caused by Discover?

mathomp4:
Ah. You are merging into `main` which does not have a CircleCI configuration to look for. The real question is: is there where you mean to merge into? If you meant `develop` then I think both @tclune and myself would say to _nuke this branch from orbit_ and restart fresh from `develop` make a new branch that you are sure is against `develop` and make a PR.

We are afraid sometimes of "changing base" on PRs...

mathomp4:
I added a blocking label until that is resolved...

weiyuan-jiang:
merge to wrong branch.  replaced with #534

PR #537, opened 2020-09-04
Update Regrid_Util.x to use MAPL/ESMF regridding machinery, other improvements as well.

bena-nasa:
This PR only affects the Regrid_Util.x code in MAPL which is a code I created for regridding History output a long time ago due to repeated user requests. I have modified this utility to use the grid/regridder managers and grid factories in MAPL. When this was first created the cubed-sphere grid factory was still in FV3 so if I wanted this to be self contained to MAPL so I could not use the factory. Now that the cubed-sphere factory is in MAPL I was able to clean this up.

I have modified the code to now use all our grid/regridder manager framework to create the grids, regridders, and applying the regridders. I have also added the options to pass in cubed-sphere stretch parameters if the output grid is a cubed-sphere. Finally I have instrumented it with the MAPL_Profiling layer.

## Types of changes

## Checklist:


atrayano (approved):
Looks OK

PR #539, opened 2020-09-08
Allow cap driver to retrieve GEOS state

bena-nasa:
This satisfies another request by the JEDI developers. Since they will ultimately be doing DA from one executable they want to be able to modify the GEOS state, i.e. maybe JEDI does a forecast, then an analysis, then backup, rerun this time doing something like IAU where you will have to perturb the model state. 

I've added a new call in Generic to retrieve the internal state of a named component, just like we already have for the import state of a component in a MAPL hierarchy.

Then I added two hooks in the CapGridComp so whatever is driving the Cap JEDI, or something else can reach in and pull the fields from the appropriate components it needs to act on.

Finally I cleaned up the memory checkpointing and simplified it into one routine each for the checkpoint and restore. There was no reason it needed to be broken up as the initial implementation was.

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_CapGridComp.F90
+++ b/base/MAPL_CapGridComp.F90
@@ -72,6 +72,8 @@ module MAPL_CapGridCompMod
      procedure :: record_state
      procedure :: refresh_state
      procedure :: destroy_state
+     procedure :: get_field_from_import
+     procedure :: get_field_from_internal
   end type MAPL_CapGridComp
 
   type :: MAPL_CapGridComp_Wrapper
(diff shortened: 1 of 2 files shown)


atrayano (approved):
Looks OK

tclune:
I'm wary of the concept of sharing "internal" state outside of the component.    I know we abuse this a fair bit in MAPL,  but in reality the most that I would expect an outside layer to be able to do with such is to make a copy and restore the value to some previously captured copy.   If additional operations are required they should be expressible in a generic fashion.  E.g., maybe it must be possible to create linear interpolations of two values of the internal state (i.e. weighted sums).    

The existing abuses in Generic can be forgiven as shortcuts from providing the items in an export state.

bena-nasa:
I'm sure we will have to revisit this many times and evolve. I don't like exposing the internal state. However, JEDI requires you can extract a "state" and pass a new "state" (i.e. something that is a result of the DA) back to be used in whatever model is ultimately be used at run time. With the "state" being the variables used in the DA. So we absolutely have to provide a mechanism to do this.  We already have a couple possible ways to accomplish this. What Kyle did for NUOPC where you could have imports bubbling up not passed to ExtData but put in a state exposable by the driver of Cap. Of course that requires explicitly creating them down at the component level and then doing something with them. In the existing systemDA system we do this all with files through the AGCM import to feed the result of GSI back to the model. 

The more powerful interfaces here will allow them to start playing on their end while we refine this and think about it more.

PR #540, opened 2020-09-09
Common tripolar format

bena-nasa:
This PR updates the tripolar grid factory to read the format that we have agreed upon to contain the coordinates of the MOM5/MOM6 grid. In addition the metadata for tripolar output was modified to match earlier tags so that the post processing does not choke in time_ave.x

## Types of changes

## Checklist:


atrayano (approved):
Looks OK. The root processor on each node reads the data. This definitely is OK, but we might consider reading on root, and broadcasting to node roots, as this is done in several places in MAPL

PR #542, opened 2020-09-10
Feature/cube corners

### LINKED ISSUE #541: Getting grid-box corners out of HISTORY
For seamless plotting of data on curvlinear grids with functions like matplotlib's [`pcolormesh(X, Y, C)`](https://matplotlib.org/3.3.1/api/_as_gen/matplotlib.pyplot.pcolormesh.html), the grid-box corner coordinates are needed. Would it be possible to get corner coordinates in diagnostics out of HISTORY? Functions like `pcolormesh()` expect `X` and `Y` to be size `N+1` if the shape of `C` is `(N, N)`, so corners with duplicates (between faces) would be preferred.

Following up on geoschem/GCHPctm#38 and this comment by @tclune 
> I have discussed with @atrayanov and @ben-nasa and they think this will be straightforward. Please open a ticket in MAPL and assign to @ben-nasa.
> 
> Basically we already have a per-grid factory with a method that specifies metadata. That routine would need a few more lines to add metadata for corners. And then the generic layer that sends data to the I/O server needs a few lines to extract the actual values of the coordinates from the ESMF Grid object. The biggest concern is the handling of the duplicate corners, and the MAPL ticket you would open is the best place to discuss that.

bena-nasa:
Fixes #541 
This adds the corners to the cubed-sphere output. I unfortunately had to modify all the grid factories since I needed to make a new routine to get the correct file bounds in the edge case and this is deferred from the abstract factory.

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_AbstractGridFactory.F90
+++ b/base/MAPL_AbstractGridFactory.F90
@@ -75,7 +75,7 @@ module MAPL_AbstractGridFactoryMod
       procedure(get_grid_vars), deferred :: get_grid_vars
       procedure(append_variable_metadata), deferred :: append_variable_metadata
       procedure(generate_file_bounds), deferred :: generate_file_bounds
-      procedure(generate_file_edge_bounds), deferred :: generate_file_edge_bounds
+      procedure(generate_file_corner_bounds), deferred :: generate_file_corner_bounds
       procedure(generate_file_reference2D), deferred :: generate_file_reference2D
       procedure(generate_file_reference3D), deferred :: generate_file_reference3D
    end type AbstractGridFactory
@@ -177,24 +177,24 @@ module MAPL_AbstractGridFactoryMod
          import AbstractGridFactory
          class (AbstractGridFactory), intent(inout) :: this
          type(ESMF_Grid), intent(inout)      :: grid
-         integer, allocatable, intent(inout) :: local_start(:)
-         integer, allocatable, intent(inout) :: global_start(:)
-         integer, allocatable, intent(inout) :: global_count(:)
+         integer, allocatable, intent(out) :: local_start(:)
+         integer, allocatable, intent(out) :: global_start(:)
+         integer, allocatable, intent(out) :: global_count(:)
          integer, optional, intent(out) :: rc
          
       end subroutine generate_file_bounds
 
-      subroutine generate_file_edge_bounds(this,grid,local_start,global_start,global_count,rc)
+      subroutine generate_file_corner_bounds(this,grid,local_start,global_start,global_count,rc)
          use esmf
          import AbstractGridFactory
          class (AbstractGridFactory), intent(inout) :: this
          type(ESMF_Grid), intent(inout)      :: grid
-         integer, allocatable, intent(inout) :: local_start(:)
-         integer, allocatable, intent(inout) :: global_start(:)
-         integer, allocatable, intent(inout) :: global_count(:)
+         integer, allocatable, intent(out) :: local_start(:)
+         integer, allocatable, intent(out) :: global_start(:)
+         integer, allocatable, intent(out) :: global_count(:)
          integer, optional, intent(out) :: rc
          
-      end subroutine generate_file_edge_bounds
+      end subroutine generate_file_corner_bounds
 
       function generate_file_reference2D(this,fpointer) result(ref)
          use pFIO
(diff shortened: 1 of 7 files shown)


base/MAPL_AbstractGridFactory.F90
```diff
@@ -183,6 +184,18 @@ subroutine generate_file_bounds(this,grid,local_start,global_start,global_count,
          
       end subroutine generate_file_bounds
 
+      subroutine generate_file_edge_bounds(this,grid,local_start,global_start,global_count,rc)
```
tclune, line 187: Do you mean edge here?  Or corner?

base/MAPL_AbstractGridFactory.F90
```diff
@@ -183,6 +184,18 @@ subroutine generate_file_bounds(this,grid,local_start,global_start,global_count,
...
+         use esmf
+         import AbstractGridFactory
+         class (AbstractGridFactory), intent(inout) :: this
+         type(ESMF_Grid), intent(inout)      :: grid
+         integer, allocatable, intent(inout) :: local_start(:)
+         integer, allocatable, intent(inout) :: global_start(:)
+         integer, allocatable, intent(inout) :: global_count(:)
```
tclune, line 194: I believe these should be `intent(out)`, not `inout`.    With allocatable, the usual concerns that often force us to conservatively use the broader INTENT are not a concern.

base/MAPL_newCFIO.F90
```diff
@@ -645,6 +646,37 @@ subroutine stage2DLatLon(this, fileName, oClients, rc)
...
+        _VERIFY(status)
+
+        call factory%generate_file_edge_bounds(this%output_grid,LocalStart,GlobalStart,GlobalCount,rc=status)
+        _VERIFY(status)
+        call ESMF_GridGetCoord(this%output_grid, localDE=0, coordDim=1, &
+        staggerloc=ESMF_STAGGERLOC_CORNER, &
+        farrayPtr=ptr2d, rc=status)
+        _VERIFY(STATUS)
+        if (.not.allocated(this%corner_lons)) allocate(this%corner_lons(size(ptr2d,1),size(ptr2d,2)))
+        this%corner_lons=ptr2d*MAPL_RADIANS_TO_DEGREES
```
tclune, line 667: I think you can do this more elegantly.  No need for the explicit allocate.   Just
```
this%corner_lons = ptr2d * MAPL_RADIANS_TO_DEGREES
```

base/MAPL_newCFIO.F90
```diff
@@ -645,6 +646,37 @@ subroutine stage2DLatLon(this, fileName, oClients, rc)
...
+        ref = ArrayReference(this%corner_lons)
+         call oClients%collective_stage_data(this%write_collection_id,trim(filename),'corner_lons', &
+              ref,start=localStart, global_start=GlobalStart, global_count=GlobalCount)
+        call ESMF_GridGetCoord(this%output_grid, localDE=0, coordDim=2, &
+        staggerloc=ESMF_STAGGERLOC_CORNER, &
+        farrayPtr=ptr2d, rc=status)
+        _VERIFY(STATUS)
+        if (.not.allocated(this%corner_lats)) allocate(this%corner_lats(size(ptr2d,1),size(ptr2d,2)))
+        this%corner_lats=ptr2d*MAPL_RADIANS_TO_DEGREES
```
tclune, line 676: As above - no need for the explicit allocate statement.

tclune (changes requested):
Mostly minor changes requested.  See the inline comments.   But edge vs corner is important to get right.

base/MAPL_newCFIO.F90
bena-nasa: Is '=' overloaded to do the allocation and match the shape/bounds of the right hand side if the left hand side is not allocated?

tclune: "Overloaded" is the wrong term here.  In F2003, _intrinsic_ assignment with allocatable LHS is required to allocate (or reallocated) the LHS unless it already matches the shape of the RHS.   Exactly suited to this case.

F2008 went a step further and included reallocating if the LHS is polymorphic and not of the same type as the rhs.

bena-nasa: Make the requested changes including an updating the mockgridfactory so hopefully the MAPL unit tests pass this time.

bena-nasa:
Made requested changes including update to mockgridfactory so hopefully units tests pass this time.

tclune (approved):

PR #544, opened 2020-09-16
Adds a mpeu StrTemplate Replacement

WilliamJamieson:
Adds a wrapper for `fill_grads_template` which allows MAPL to provide the same interface (name and arguments) as the `StrTemplate` from `GMAO_Shared/GMAO_mpeu` (see [`m_StrTemplate.F90`](https://github.com/GEOS-ESM/GMAO_Shared/blob/main/GMAO_mpeu/m_StrTemplate.F90)). 

## Description

Replicates all the functionality of the mpeu version except for true support of the `preserve` option.

## Motivation and Context

GOCART2G should be independent of the `GMAO_Shared`, but it currently needs the mpeu provided `StrTemplate`. Since GOCART2G relies on MAPL, it makes sense to provide this functionality directly from MAPL.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added routine to finalize the ioservers so that it can be called by another application using cap, like JEDI
 - Re-added CircleCI with FV3 standalone test
 - Add ability to run multiple forward time integrations within one execution for JEDI (#529)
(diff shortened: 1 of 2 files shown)


atrayano (approved):
Looks OK

tclune (changes requested):
Since preserve is not supported,  stat should be set to failure if preserve is passed in (and has value  .true.)

WilliamJamieson:
Added requested change.

tclune (approved):

PR #547, opened 2020-09-21
AT: Very first --alpha-- version of service-services. The code compil…

atrayano:
…es, but is not tested at all

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_Generic.F90
+++ b/base/MAPL_Generic.F90
@@ -164,6 +164,9 @@ module MAPL_GenericMod
   public MAPL_AddChild
   public MAPL_AddConnectivity
   public MAPL_TerminateImport
+  public MAPL_ServiceAddConnection
+  public MAPL_ServiceAddProvider
+  public MAPL_ServiceAddSubscriber
   
   ! MAPL_Util
   !public MAPL_GenericStateClockOn
@@ -380,6 +383,7 @@ end type  MAPL_InitialState
 type MAPL_Connectivity
    type (MAPL_VarConn), pointer :: CONNECT(:)       => null()
    type (MAPL_VarConn), pointer :: DONOTCONN(:)     => null()
+   type (MAPL_VarServiceConnectionPtr), pointer :: ServiceConnection(:) => null()
 end type MAPL_Connectivity
 
 type MAPL_LinkType
@@ -441,6 +445,10 @@ type  MAPL_MetaComp
    type (MAPL_Communicators)                :: mapl_comm
    type (TimeProfiler), public :: t_profiler
    character(:), allocatable :: full_name ! Period separated list of ancestor names
+
+   type(MAPL_VarServiceProviderPtr), pointer :: provider_list(:) => null()
+   type(MAPL_VarServiceSubscriberPtr), pointer :: subscriber_list(:) => null()
+
    class(Logger), pointer :: lgr
 
 !!$   integer :: comm
@@ -1693,6 +1701,23 @@ endif
       _VERIFY(STATUS)
    end if
 
+   ! Service services processing:
+   ! process any providers
+   if (associated(state%provider_list)) then
+      call MAPL_VarServiceProviderSet(state%provider_list, import, rc=status)
+      _VERIFY(STATUS)
+   end if
+
+   ! process any subscribers
+   if (associated(state%subscriber_list)) then
+      call MAPL_VarServiceSubscriberSet(state%subscriber_list, state%internal, rc=status)
+      _VERIFY(STATUS)
+   end if
+
+   ! process any service connections
+  call MAPL_ServiceProcessConnections(state, RC=status)
+  _VERIFY(STATUS)
+   
   call MAPL_GenericStateClockOff(STATE,"GenInitTot")
   call MAPL_GenericStateClockOff(STATE,"TOTAL")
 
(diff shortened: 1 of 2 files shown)


base/MAPL_Generic.F90
```diff
@@ -441,6 +445,10 @@ module MAPL_GenericMod
    type (MAPL_Communicators)                :: mapl_comm
    type (TimeProfiler), public :: t_profiler
    character(:), allocatable :: full_name ! Period separated list of ancestor names
+
+   type(MAPL_VarServiceProviderPtr), pointer :: provider_list(:) => null()
+   type(MAPL_VarServiceSubscriberPtr), pointer :: subscriber_list(:) => null()
+
```
tclune, line 451: I guess I am a little bit surprised to see this information in Meta.

tclune, line 451: This might be a good place to introduce Vector containers to support the growable list - depends on what I see about their use below.

base/MAPL_Generic.F90
```diff
@@ -4916,7 +4941,247 @@ subroutine MAPL_TerminateImportAll ( GC, ALL, RC )
   end subroutine MAPL_TerminateImportAll
 
 
+!new routines to handle ServiceServices
+  subroutine MAPL_ServiceAddConnection( GC, PROVIDER, SUBSCRIBER, SERVICE, RC)
```
tclune, line 4945: Lower case - please.

base/MAPL_Generic.F90
```diff
@@ -4916,7 +4941,247 @@ subroutine MAPL_TerminateImportAll ( GC, ALL, RC )
...
+    _VERIFY(STATUS)
+
+    provider => maplobj%provider_list
+    _RETURN(ESMF_SUCCESS)
+    
+  end subroutine MAPL_ServiceProviderGet
+
+  subroutine MAPL_ServiceSubscriberGet(GC, SUBSCRIBER, RC)
+    type(ESMF_GridComp),            intent(INOUT) :: GC ! Gridded component
+    type(MAPL_VarServiceSubscriberPtr), pointer :: subscriber(:)
```
tclune, line 5037: "subscriber" => "subscribers"  (plural)
The procedure name should also be plural.

base/MAPL_Generic.F90
```diff
@@ -4916,7 +4941,247 @@ subroutine MAPL_TerminateImportAll ( GC, ALL, RC )
...
+    do_subscriber = .false.
+    if (present(provider)) then
+       do_provider = .true.
+       cname = provider
+    end if
+    if (present(subscriber)) then
+       do_subscriber = .true.
+       cname = subscriber
+    end if
+    _ASSERT(IEOR(do_provider,do_subscriber)/=0, 'Only one of the arguments PROVIDER or SUBSCRIBER must be provided')
```
tclune, line 5138: IEOR takes integer arguments, but these are logicals.    Better would be
```
_ASSERT((do_provider .or. do_subscriber, ...
```

base/MAPL_Generic.F90
```diff
@@ -4916,7 +4941,247 @@ subroutine MAPL_TerminateImportAll ( GC, ALL, RC )
...
+    else ! must be provider
+       call MAPL_VarServiceProviderGet(cmeta%provider_list, service, bundle, rc=status)
+       _VERIFY(STATUS)
+    end if
+    ! return the saved bundle
+    
+    _RETURN(ESMF_SUCCESS)
+  end subroutine MAPL_ServiceGetBundle
+
+  recursive subroutine MAPL_FindChild(meta, name, result, rc)
```
tclune, line 5155: I am surprised the need for this procedure has not arisen before.

base/MAPL_VarSpecMod.F90
```diff
@@ -1973,4 +2020,248 @@ subroutine MAPL_ConnCheckReq(CONN, ImSpecPtr, ExSpecPtr, RC)
...
+    _VERIFY(STATUS)
+      
+    TMP(1:I) = PLIST
+    deallocate(PLIST)
+
+    allocate(TMP(I+1)%Ptr,stat=STATUS)
+    _VERIFY(STATUS)
+    
+    TMP(I+1)%Ptr%Service_name = SERVICE
+    TMP(I+1)%Ptr%Bundle_name = BUNDLE
```
tclune, line 2055: Definitely looks like Vector territory.

base/MAPL_VarSpecMod.F90
```diff
@@ -1973,4 +2020,248 @@ subroutine MAPL_ConnCheckReq(CONN, ImSpecPtr, ExSpecPtr, RC)
...
+    TMP(1:I) = SLIST
+    deallocate(SLIST)
+
+    allocate(TMP(I+1)%Ptr,stat=STATUS)
+    _VERIFY(STATUS)
+    
+    TMP(I+1)%Ptr%Service_name = SERVICE
+    TMP(I+1)%Ptr%VAR_LIST = VARS
+    TMP(I+1)%Ptr%BUNDLE = ESMF_FieldBundleCreate(NAME=SERVICE, RC=STATUS)
+    _VERIFY(STATUS)
```
tclune, line 2098: Vector

base/MAPL_VarSpecMod.F90
```diff
@@ -164,6 +173,44 @@ module MAPL_VarSpecMod
   type(MAPL_VarConnType), pointer :: ConnPtr => null()
 end type MAPL_VarConn
 
+type MAPL_VarServiceProviderType
+  private
+  character(len=ESMF_MAXSTR)               :: SERVICE_NAME
+  character(len=ESMF_MAXSTR)               :: BUNDLE_NAME
```
tclune, line 179: Prefer allocatable strings - contingent on whether something else forces these to be fixed length.

base/MAPL_VarSpecMod.F90
```diff
@@ -164,6 +173,44 @@ module MAPL_VarSpecMod
...
+
+type, public :: MAPL_VarServiceSubscriberPtr
+  private
+  type(MAPL_VarServiceSubscriberType), pointer :: Ptr => null()
+end type MAPL_VarServiceSubscriberPtr
+
+type, public :: MAPL_VarServiceConnectionPtr
+  private
+  type(MAPL_VarServiceConnectionType), pointer :: Ptr => null()
+end type MAPL_VarServiceConnectionPtr
```
tclune, line 212: Since you need the wrapper types anyway, Vector is probably a better paradigm.  (And the lowest risk FTL component.)    It will require that the basic types MAPL_VarServiceSubscriberType et al will need to be in their own module first.   But that's not so bad of a thing.

I'll also probably push that these adopt a more OO naming style.  (No MAPL prefix, and no "TYPE" suffix.

tclune (changes requested):

tclune (commented):
This is now very close.    I still see a few spots where request vs requester and provides vs provider are not quite right.    I'll clone and push my suggested changes back to here.

tclune (dismissed):

generic/MAPL_RequestedServiceItem.F90
```diff
@@ -0,0 +1,16 @@
+#include "MAPL_Generic.h"
```
tclune, line 1: Same file name comment here.

generic/MAPL_ServiceConnectionItem.F90
```diff
@@ -0,0 +1,16 @@
+#include "MAPL_Generic.h"
```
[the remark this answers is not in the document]
atrayano: Yes to all of the last 3 comments

generic/MAPL_VarSpecMod.F90
```diff
@@ -2267,4 +2281,210 @@ subroutine MAPL_ConnCheckReq(CONN, ImSpecPtr, ExSpecPtr, RC)
...
+    character (len=*)             , intent(IN   ) :: SERVICE
+    character (len=*)             , intent(IN   ) :: BUNDLE
+    integer,              optional, intent(  OUT) :: RC     !
+    
+    type (ProvidedServiceType)           :: item
+            
+    item%Service_name = SERVICE
+    item%Bundle_name = BUNDLE
+    
+    call plist%push_back(item)
```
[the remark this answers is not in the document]
atrayano: I'll make the changes shortly

atrayano:
Using the constructor was a nice idea, however, it created an issue. One of the fields of the ProvidedServiceList (in this case is BUNDLE) cannot not be initialized at the time of the call at line 2292.

generic/MAPL_ServiceServices.F90
```diff
@@ -0,0 +1,176 @@
...
+    
+    found = .false.
+    iter = provider_list%begin()
+    ! loop over provided services
+    do while (iter /= provider_list%end())
+       item => iter%get()
+       if(item%service_name == advertised_service) then
+          found = .true.
+          bundle = item%bundle
+          exit
```
tclune, line 81: Let's leave this for now, but wouldn't a MAP work better than a vector here?   you could then eliminate this procedure entirely:

```f90
bundle => list%at(advertised_service)
```

PR #553, opened 2020-09-25
Fixes from GEOSadas-5_27_0 Test Builds

mathomp4:
## Description

I am currently trying to build GEOSadas-5_27_0 inside Git. And as such I'm pulling in work that @bena-nasa did previously with an old test with MAPL 2.

This is a duplicate of code from the  `feature/bmuaer/adas_with_mapl2` branch. He figured out that the `ASSERT_` statement in `MAPL_ErrLogMain.h` can't have the message. My guess is at some point I added it there in an aggressive sed command or something. 

But I'm doing my test with MAPL `develop` (e.g. 2.3) as that will probably be the MAPL basis when we get to the final integration and figured let's just get this in.

Note that `MAPL_ErrLogMain.h` is really used only in a couple GSI_App files so this is trivially zero-diff for the GEOSgcm.

## Motivation and Context

Let's the GEOSadas build (well, one program).

## How Has This Been Tested?

Code builds with it, doesn't without.

## Types of changes

## Checklist:


mathomp4:
ETA: I also deleted `base/MAPL_base.mk`. That's an old GNU Make file and I don't think we care about it anymore.

mathomp4:
Adding blocking tag. More edits are needed in MAPL. Will consult with @bena-nasa

tclune (approved):

PR #554, opened 2020-09-28
Feature/restart by oserver refactory

weiyuan-jiang:
## Description

Refactoring the option WRITE_RESTART_BY_OSERVER
There is one minor but important change to get the rank of the writer in ServerThread.F90. When SimpleSocket is used, The writer rank should be 0 which is the same as the ArrayGather. That avoids data movement across the process. 

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_IO.F90
+++ b/base/MAPL_IO.F90
@@ -2907,8 +2907,9 @@ module MAPL_IOMod
                    lMemRef = LocalMemReference(pFIO_REAL32,[0])
                    call c_f_pointer(lMemRef%base_address, gvar_1d, shape=[0])
                 endif
-                
-                call ArrayGather(var_1d, gvar_1d, grid, mask=mask, rc=status)
+                if (DIMS == MAPL_DimsTileOnly .or. DIMS == MAPL_DimsTileTile) then 
+                   call ArrayGather(var_1d, gvar_1d, grid, mask=mask, rc=status)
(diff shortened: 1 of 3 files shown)


base/MAPL_IO.F90
```diff
@@ -2935,28 +2939,34 @@ subroutine MAPL_FieldWriteNCPar(formatter, name, FIELD, ARRDES, HomePE, oClients
              endif
 
              if (arrdes%write_restart_by_oserver) then
-                if(arrdes%writers_comm /= MPI_COMM_NULL) then
+                if(MAPL_AM_I_ROOT()) then
```
tclune, line 2942: These `MAPL_AM_I_ROOT()` references seem very unusual.  These are more typically for allocating buffers on the root process and/or log messages.

I'll approve, but would appreciate an explanation.

tclune (approved):

base/MAPL_IO.F90
weiyuan-jiang: By default, ArrayGather gathers everything to root. But I found it has optional argument "depe" ( destination PE , I guess). I am wondering to make changes to link the writing PE with the "pede"

weiyuan-jiang: The difficulty here is the client does not ( or should not)  know about oserver

PR #561, opened 2020-10-01
Fixes for ExtData Logger

mathomp4:
Runs with DEBUG logging on showed a few issues with logger lines. I'll let @tclune be the final arbiter on my changes.


tclune (approved):
Looks good.  Thanks!

PR #576, opened 2020-10-14
Fix coarse grid IO

### LINKED ISSUE #574: Coarse IO broken in MAPL 2.3.2
Running my ExtData tests with the latest MAPL version 2.3.2 resulted in a new failure of one particular case, the case that tests the ability to read/write a very course grid on a core count that results in a grid with no DE's on some processes.
I had some protections for this. For example, I made a MAPL_FieldGet that first checks if the grid has a DE on the PET and only then gets the pointer or have just explicitly check this.
However, on those processes with no DE ,the presumable null pointer still gets passed to another subroutine.
The test cases was failing at runtime on entry into the regridding routines which all look like some variation of this:
```
   subroutine regrid_scalar_2d_real32(this, q_in, q_out, rc)
      class (EsmfRegridder), intent(in) :: this
      real(kind=REAL32), intent(in) :: q_in(:,:)
      real(kind=REAL32), intent(out) :: q_out(:,:)
```
The input grid might have DE's on every process but the output might not or vice versa so everybody must make this call.
I confirmed that with MAPL 2.2.7, which was still using Intel 18.05 that the test worked and that moving to MAPL 2.3.2, which uses Intel 19.01 fails. I am presuming that intel must not be happy with the unassociated pointer being passed in now but was less strict in.

I found that if I do something like this 
```
         if (HasDE) then
            call ESMF_FieldGet(field,farrayPtr=ptr3d,rc=status)
            _VERIFY(status)
         else
            allocate(ptr3d(0,0,0))
         end if
```
so that the pointer is allocated to something it fixes the problem. The bad thing is that the code was already somewhat ugly and this just makes it worse.

The only other solution I can think of is extended the GetFieldArray.H macro to allocate a size zero pointer of the right rank if there is no DE

bena-nasa:
Fixes #574

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fix bug with using coarse grids in History and ExtData
+
 ### Added
 ### Changed
 ### Fixed
--- a/base/MAPL_newCFIO.F90
+++ b/base/MAPL_newCFIO.F90
@@ -411,8 +411,17 @@ module MAPL_newCFIOMod
         real, pointer :: ptr2d(:,:), outptr2d(:,:)
         real, allocatable, target :: ptr3d_inter(:,:,:)
         type(ESMF_Grid) :: gridIn,gridOut
+        logical :: hasDE_in, hasDE_out
 
         call ESMF_FieldBundleGet(this%output_bundle,itemName,field=outField,rc=status)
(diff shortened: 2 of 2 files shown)


tclune (approved):
The only other thought I have to improve this situation would be to introduce a helper function (well, one per kind/rank) that takes in a pointer and allocates it to 0 size if it is not allocated, otherwise does nothing.   It would at least eliminate the conditionals.

But that should not be part of this PR in any event.  It could be just another enhancement on develop

PR #584, opened 2020-10-20
Git Flow pull of main hotfix into develop

mathomp4:
Git Flow pull of main hotfix into develop


mathomp4:
I'll close this as #585 is the only one that'll work (yay weird conflicts!)

PR #594, opened 2020-11-09
Bugfix/atrayano/monthly fix seg alarm

atrayano:
## Description

This fixes a bug with new segment alarm for monthly mean collections

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Fix bug with segment alarm whem processing a monthly mean collectiom
 
 ### Added
 
@@ -16,7 +17,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ## [2.3.4] - 2020-10-20
-- Fix bug with segment alarm whem processing a monthly mean collectiom
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ## [2.3.4] - 2020-10-20
+- Fix bug with segment alarm whem processing a monthly mean collectiom
 
 ### Changed
 
--- a/base/MAPL_Generic.F90
+++ b/base/MAPL_Generic.F90
@@ -1352,7 +1352,7 @@ recursive subroutine MAPL_GenericInitialize ( GC, IMPORT, EXPORT, CLOCK, RC )
    _VERIFY(STATUS)
    if (isPresent) then
       nra = ESMF_ConfigGetLen( STATE%CF, RC = STATUS)
-      _ASSERT( NRA > 0,'needs informative message')
+      _ASSERT( NRA > 0,'Empty list is not allowed')
 
       allocate (ref_date(NRA), ref_time(NRA), ref_freq(NRA), stat=STATUS)
       _VERIFY(STATUS)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ## [2.3.4] - 2020-10-20
+- Fix bug with segment alarm whem processing a monthly mean collectiom
```
tclune, line 19: This line should go up in the "Unreleased "clause under `### Fixes` subclause.

base/MAPL_Generic.F90
```diff
@@ -1351,7 +1352,7 @@ recursive subroutine MAPL_GenericInitialize ( GC, IMPORT, EXPORT, CLOCK, RC )
    _VERIFY(STATUS)
    if (isPresent) then
       nra = ESMF_ConfigGetLen( STATE%CF, RC = STATUS)
-      _ASSERT( NRA > 0 .and. NRA < 10,'needs informative message')
+      _ASSERT( NRA > 0,'needs informative message')
```
tclune, line 1355: The "needs informative message" was generated during a brut force sweep that changed the macros.   Since you are changing the logic here, presumably you can come up with a better error message?

base/MAPL_Generic.F90
```diff
@@ -1379,7 +1380,8 @@ recursive subroutine MAPL_GenericInitialize ( GC, IMPORT, EXPORT, CLOCK, RC )
       _VERIFY(STATUS)
 
       DO  I = 1, NRA
-         AlarmName = "RecordAlarm" // CHAR(I+ICHAR('0')) 
+         write(alarmNum,'(I3.3)') I
```
tclune, line 1383: Is 3 digits enough?    We've generally been trying to aim for an order of magnitude beyond any use cases that we have now.  3 is probably enough, but I figured I'd make you think about it to be certain.

tclune (changes requested):
Just a few minor changes and a question.

base/MAPL_Generic.F90
atrayano: 3 digit goes way above all use cases currently cover (the "old" limit was a single digit, in GEOS we never exceeded it, only Lizzie currently ran into this trying to use this for 12 month worth of alarm). We have the ability to use frequency to specify these alarms (this is the typical usage) and these are still 1 record alarm. In addition, we have the ability to specify specific date/times at arbitrary points, and this is what Lizzie exceeded.

base/MAPL_Generic.F90
atrayano: It is hard to imagine what can cause this number to be non-negative - the counting is done by ESMF_ConfigGetLen (counts the number of entries for a specific keyword), so I am not sure what would be a "better" assert message. How about "This should never happen"

CHANGELOG.md
atrayano: Agree. I was not aware about this line

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Fix bug with segment alarm whem processing a monthly mean collectiom
```
tclune, line 9: Still not quite right.   There should be a section called `### Fixed` and this comment would go under that.   But I'll approve.  Matt will undoubtedly fix when there is a release.

tclune (approved):

CHANGELOG.md
mathomp4: Yeah. Since I'm making a release as soon as it's in, it'll get fixed soon!

atrayano: I also spotted a typo: the world "collectiom" should have been collection. @mathomp4 could you fix it when making the release

PR #601, opened 2020-11-10
fixes #600

### LINKED ISSUE #600: Subtle difference in MAPL_TerminateImportAllBut and MAPL_TerminateImportAll that causes different behaivor
When testing using ExtData to deliver fields to components on the ocean side of the model, I discovered a strange bug. Normally at the OGCM level we terminate all imports. Now if we want something to go to ExtData, I had to use a different interface for MAPL_TerminateImport which underneath is the MAPL_TerminateImportAllBut overload. This was added by myself, several years ago just for this purpose, to allow only select imports to move up the MAPL hierarchy but terminate the rest but should accomplish the same thing as MAPL_TerminateImportAll if the import name you are passing in doesn't exist.

However, changing this terminate import calling OGCM was causing a crash complaining of a connectivity error for FRACICE in OGGM. The issue turned out to be a subtle difference in the two terminate import calls. The AllBut overload was looping over each child, then over each importspec and if the name in the import spec does not matched any of the passed in names we call MAPL_DoNotConnect, passing the import name.

The other terminate import for all is MAPL_TerminateImportAll which called MAPL_DoNotConnectAnyImport which call MAPL_DoNotConnect passing the name of the import in each child as MAPL_AnyChildImport.

Atanas and I traced this through and found these cause different conditionals to be hit MAPL_WireComponent and when you do  MAPL_TerminateImportAllBut, it turns out the import does not get marked as connected causing crash because there is a cycle that causes earlier breaking of a loop. This was only affecting OGCM because right before the MAPL_TerminateImport we had a connectivity involving FRACICE. Also no one is actually using the AllBut version of terminate import. We think that connectivity was not there several years ago when I tested this call.

The solution is in MAPL_TerminateImportAllBut first check if the import is already connected and only if it is not make the do not connect call.

bena-nasa:
#fixes #600 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed integer overflow in memutils for big memory systems
+- Fixed logic to allow proper termination of all imports except those specified
 
 ### Removed
 
--- a/base/MAPL_Generic.F90
+++ b/base/MAPL_Generic.F90
@@ -4854,6 +4854,9 @@ end function MAPL_AddChildFromGC
     integer                               :: I,J
     logical                               :: SKIP
     character(len=ESMF_MAXSTR), allocatable :: SNAMES(:)
+    type (MAPL_Connectivity), pointer     :: conn
+    type (MAPL_VarConn), pointer          :: CONNECT(:)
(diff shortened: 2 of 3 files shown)


atrayano (approved):
Looks good!

atrayano (approved):
Looks OK

PR #608, opened 2020-11-12
Inject external grid

WilliamJamieson:
Completes system to bring an external grid into a MAPL based NUOPC component by providing a method to take a grid from the `Cap_GridComp` and set it into the `root_child`'s GridComp.

## Description

At the request of @rmontuoro, we have extended the external grid set API to automatically set the external grid into the `root_child` GridComp in addition to the `Cap_GridComp`.

## How Has This Been Tested?

Since this change involves code that will always be run, I performed a 3 day C12 run, which was zero diff with a run with no changes.

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_CapGridComp.F90
+++ b/base/MAPL_CapGridComp.F90
@@ -76,6 +76,7 @@ module MAPL_CapGridCompMod
      procedure :: get_field_from_import
      procedure :: get_field_from_internal
      procedure :: set_grid
+     procedure :: inject_external_grid
      procedure :: set_clock
   end type MAPL_CapGridComp
 
@@ -547,6 +548,12 @@ contains
 
     call MAPL_Get(MAPLOBJ, gcs = cap%gcs, gim = cap%child_imports, gex = cap%child_exports, rc = status)
     _VERIFY(status)
+
+
+    !  Inject grid to root child if grid has been set externally
+    !-----------------------------------------------------------
+
+    call cap%inject_external_grid(__RC__)
 
     ! Run as usual unless PRINTSPEC> 0 as set in CAP.rc. If set then
     ! model will not run completely and instead it will simply run MAPL_SetServices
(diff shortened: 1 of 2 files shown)


mathomp4:
Uh oh. The fact that you changed base is always concerning. Did you branch off of `develop` when you started this and just forgot to make the PR against `develop`?

WilliamJamieson:
Yes I originally branched off `develop` and just forgot to change that when I made the initial PR.

rmontuoro:
Do you envision a case where `set_grid()` and `inject_external_grid()` need to be called independently by the user? Otherwise, I'd suggest changing `inject_external_grid()` to be a private procedure.

WilliamJamieson:
@rmontuoro, I don't currently envision such a case. You make a good point, I just didn't make it private as none of the other procedures in the `MAPL_CapGridComp` class are private. I leave the final decision up to @tclune on this point.

tclune:
I generally don't make procedures private, simply because that interferes with testing.  Not that the testing at this level is very thorough.   I have no problem making it private, but as William said, its a drop in the bucket.

tclune (approved):

base/MAPL_CapGridComp.F90
```diff
@@ -1290,6 +1297,40 @@ subroutine set_grid(this, grid, unusable, rc)
...
+     type(ESMF_Grid)           :: cap_grid,            root_grid
+     logical                   :: cap_grid_is_present, root_grid_is_present
+     integer                   :: status
+
+     _UNUSED_DUMMY(unusable)
+
+     call ESMF_GridCompGet(this%gc, gridIsPresent=cap_grid_is_present, __RC__)
+
+     if (cap_grid_is_present) then
+        call ESMF_GridCompGet(this%gc, grid=cap_grid, __RC__)
```
rmontuoro, line 1315: The following logic is a bit obscure to me. If a grid is already set in the root grid comp, should it be validated? Is the assert() necessary? Should we introduce a "force" flag to override previously set grid?
In my opinion, limiting the scope of this method would increase its reliability as you are in control of the conditions under which it is called. However, I defer to your implementation's choice.

WilliamJamieson: The assert is there for the case that the root grid comp defines its own grid. In that case, it is safer for the model to fail gracefully, then it is for us to override the defined grid (the grid comp may make assumptions based on its defined grid). Thus it is my opinion that if one is trying to override the root grid comp's defined grid, then one has larger issues to address. 

As far as the validate calls are concerned, we are only calling this method once, so I placed them in there for error catching purposes.

tclune: The only config where it might be the case that the root has its own grid would be the CTM.   Might be worth checking.

In any event, I consider most of this logic to be "temporary", as I remain committed to evolving MAPL to the point where MAPL components can readily be used as NUOPC components.    But it will take some time to get there.

PR #618, opened 2020-11-20
Move MAPL_Cap and dependencies

bena-nasa:
This PR moves the MAPL_Cap, MAPL_CapGridComp, and the NUOPC driver to a new Cap directory, under a new gridcomp directory in anticipation of the new ExtData component that will not live in base. If CapGridComp stays where it is this is a circular dependency once ExtData is no longer in base.

## Types of changes

## Checklist:


tclune (changes requested):
Unfortunately for you, William's PR went through first.   You'll have to update.  May be easier just to start over.

PR #623, opened 2020-11-30
Move to CircleCI Medium Resource Class

mathomp4:
Due to circumstances beyond our control, we must move to a lower/free CircleCI Resource Class for a time.


mathomp4:
Note: using "skip changelog" because this is the definition of a PR that doesn't need a changelog entry!

PR #646, opened 2020-12-18
Make clobber default

### LINKED ISSUE #640: [FEATURE REQUEST] Allow clobbering of existing NetCDF files
Currently MAPL fails if an attempt is made to create a new NetCDF file when one already exists in that location. However, this is problematic, as it can result in frustrating loss of productivity. For example, we (the GCHP developers) frequently perform a series of tests in the same working directory; adding lines to the run script which clear out the working directory and storage directories would resolve this but would also raise the possibility of users accidentally destroying their output directory and all existing checkpoints. This behavior is also internally inconsistent, as MAPL will happily "clobber" (technically, write over the data within) 3-D gridded output through HISTORY.rc, but will fail if asked to clobber a restart file or 1-D non-gridded output.

A helpful solution would be to have a switch which explicitly permits such clobbering, perhaps in CAP.rc or another resource file. This would prevent unwanted behavior in GEOS (where presumably this clobbering is not desired) but would be very helpful to the GCHP community.

Tagging @tclune, @atrayano, and @lizziel with whom this was previously discussed.

bena-nasa:
Fixes #640
This is a very simple PR that changes the default behavior when creating NetCDF files with pfio from NOCLOBBER to CLOBBER (although I did add a new argument so if we need someday we can override).
This was requested by the GEOS-Chem developers. I can find no harm in doing this for GEOS. We must not be trying clobber files in the first place with our workflow as we would not be able to run. Hence changing to NOCLOBBER should have no effect.
Making this user configurable is difficult since the places the "create" is called are so low level they don't have access to any resource file so no ability for the user to change this for now.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added an `ExternalGridManager`, to allow MAPL to have knowledge of external grids (for NUOPC).
 - Added command line interface option `--isolate_nodes`. By default it is `.true.`
 - Add stretching factors to file if applicable in cubed-sphere output via History and uptick to cube version 2.91
+- Make the default clobber rather than no clobber in NetCDF formatter in PFIO
 
 ### Changed
 
--- a/pfio/NetCDF4_FileFormatter.F90
+++ b/pfio/NetCDF4_FileFormatter.F90
@@ -139,16 +139,23 @@ module pFIO_NetCDF4_FileFormatterMod
 
 contains
 
-   subroutine create(this, file, unusable, rc)
+   subroutine create(this, file, unusable, mode, rc)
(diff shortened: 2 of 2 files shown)


tclune (approved):
Approving.  But fear this one may bite later.

bena-nasa:
I don't know, on the one hand we must be pretty good about running everything in clean workspaces so I'm not seeing any harm currently. If we want to be safer, right now the netcdf formatter is used in some utility programs (mostly related to regridding) that once could just add the argument so they can't clobber existing output (harvard never uses these). History, via the o-server by fluke already does clobber in a way (overwrite data if file exists). The only other place the NetCDF formatter is used is writing restarts (non-oserver). And if you write the restarts by the server, they would behave like history, so you are clobbering.

tclune:
It is the unforeseen uses that will become dangerous.   But all things are relative.

PR #647, opened 2020-12-18
Preparatory work for new regrid option.

tclune:
To re-enable the ability for MAPL to provide custom regridding outside
of ESMF, the NewGridManager needed thep prototyping ability from the
older GridManager class.  But doing it a bit differently.  Have now
addeed a support() method to the AbstractRegridder hierarchy that allows
subclasses to determine whether they can handle a given combination
of grids and a method.

Also used the opportunity to do a bunch of house cleaning
 - moved modules into their own files
 - switched to newen file and module naming conventions
 - switched to the standard CPP macros in MAPL_Generic.h
Restored the prototyping capability from the old RegridderManager to
accomodate regridding capabilities not provided by ESMF.  A skeleton
of HorizontalFluxRegridder is implemented, but hos no relevant
functionality.

## Description
See above.  Note that the feature is not complete but the cleanup is worth merging into develop asap so that it is not lost.

## How Has This Been Tested?
Reran basic c12 experiment.  Changes are almost but not quite trivial so this should be sufficient.

## Types of changes
- [ X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I have tested this change with a run of GEOSgcm (if non-trivial)


bena-nasa (approved):

mathomp4 (approved):
Approving as @bena-nasa approved before and the only change was a fix to a unit test

PR #648, opened 2020-12-22
Wjiang/feature/check restart dims

weiyuan-jiang:
Check the dimension of variables in restart file. 
## Motivation and Context

If given a wrong restart file( higher resolution), the program would just read and move on. The checking makes sure the restart file is consistent with the original setup
## Types of changes

## Checklist:


weiyuan-jiang:
#655 has addressed this issue.

PR #655, opened 2021-01-05
Add simple checking that the restart file matches the grid for at lea…

### LINKED ISSUE #643: GEOS runs with wrong restarts!
This is something mentioned by @sdeastham in #639. Namely, GEOS doesn't die as fast as I'd expect if you have the wrong restarts. To wit, I made a C24 experiment, but used C48 restarts to run with. Now, GEOS at first will die because of the salt tile checker:
```
Error! Found 99389 tiles in openwater. Expect to find 69464 tiles.
Your restarts are probably for a different ocean.
```
But let's say we don't have that check. If we comment that out, the model will run a fair bit. Indeed, I think it only dies in the ***Run*** method of Catch!
```
...
 Real*4 Resource Parameter: SOLAR_CONSTANT:-1.000000
 Character Resource Parameter: SOLAR_CYCLE_FILE_NAME:ExtData/g5gcm/solar/NRLSSI2.v2017.txt
 Logical Resource Parameter: USE_NRLSSI2:T
 Logical Resource Parameter: PERSIST_SOLAR:T
 Integer*4 Resource Parameter: INTERPOLATE_ATMTAU:0
 Real*4 Resource Parameter: LANDICEEMISS:.999990
 Real*4 Resource Parameter: LANDICEALBEDO:.580000
 Real*4 Resource Parameter: LANDICEDEPTH:.070000
 Real*4 Resource Parameter: LANDICECOND:1.200000
 Real*4 Resource Parameter: LANDICETDEEP:230.000000
 Integer*4 Resource Parameter: INCL_LOUIS_EXTRA_DERIVS:1
 Character Resource Parameter: VISDF_FILE:visdf.dat
 Character Resource Parameter: NIRDF_FILE:nirdf.dat
 NOT using buffer I/O for file: visdf.dat
 Previous time forVISDFs is:
           1           4           7           0           0           0
           1           4          15           0           0           0
 NOT using buffer I/O for file: nirdf.dat
 Previous time forNIRDFs is:
           1           4           7           0           0           0
           1           4          15           0           0           0
 CATCH_INTERNAL_RST is NOT consistent with VEGDYN Data
 CATCH_INTERNAL_RST is NOT consistent with VEGDYN Data
 CATCH_INTERNAL_RST is NOT consistent with VEGDYN Data
CATCHRUN2                                     3629
LANDRun2                                      1642
SURFACERUN2_DOTYPE                            8588
SURFACERun2                                   6742
PHYSICS::Run                                  2575
AGCMRun                                       2382
GCMRun                                        1803
```
That actually amazes me.

The question is: is this something that MAPL should be checking? Or should this be something we do a priori to running `GEOSgcm.x`?

bena-nasa:
Fixes #643

This is a simple check that the restart matches the grid. The elegant way would be to create a grid from the file and check they match but the tiles, the fact that the restarts are not on the "new" cube format, vertical levels etc ... make that not practical. 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add stretching factors to file if applicable in cubed-sphere output via History and uptick to cube version 2.91
 - Ability to register protoype (non-ESMF) regridders in NewRegridderManager
 - Make the default clobber rather than no clobber in NetCDF formatter in PFIO
-
+- Add basic check that the restart files match the application grid
 
 ### Changed
 
--- a/base/MAPL_IO.F90
+++ b/base/MAPL_IO.F90
@@ -7336,7 +7336,9 @@ module MAPL_IOMod
     integer                            :: MAPL_DIMS
     integer, pointer                   :: MASK(:) => null()
     type(Netcdf4_Fileformatter)        :: formatter
+    type(FileMetaData)                 :: metadata
     character(len=:), allocatable      :: fname_by_face
+    logical :: grid_file_match
 
     call ESMF_FieldBundleGet(Bundle,FieldCount=nVars,rc=STATUS)
     _VERIFY(STATUS)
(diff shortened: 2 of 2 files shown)


atrayano (approved):
Looks OK

PR #667, opened 2021-01-15
Add missing declaration statement.

rmontuoro:
## Description

This PR adds a missing declaration statement for a logical variable used in `MAPL_GenericFinalize()`, fixing a bug which prevented MAPL from building when preprocessor macro `H5_HAVE_PARALLEL` is not defined (_e.g._ when building with serial NetCDF).

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [ ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [ ] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #668, opened 2021-01-19
Remove Docker Auth for GitHub Actions

mathomp4:
Per https://github.com/actions/virtual-environments/issues/1445#issuecomment-713861495 I think we can comment these out to at least let GitHub Actions work for outside users (see #667 )


mathomp4:
Closing in favor of #669

PR #669, opened 2021-01-19
Fixes to allow forked PRs to pass CI

mathomp4:
This is a testbed to fix the CI for forks...


mathomp4:
@tclune Well, this worked but I made a lot of commits. Should we take it as it is? Or should I nuke and redo in one nice commit?

tclune (approved):
We're too busy to clean up for this at the moment I think.  Let's just get this in.

PR #682, opened 2021-02-08
Feature/bmauer/generalized ij search

### LINKED ISSUE #680: Bug: Floating divide by zero error when reading input volcanic emissions in GOCART/SU with UFS
A `floating divide by zero` error occurs in `MAPL_BaseMod > MAPL_GetHorzIJIndex() > cube_xy()` while reading in volcanic emissions in `GOCART2G_GridComp/SU2G_GridComp/SU2G_GridCompMod.F90` when running UFS-GOCART built in debug mode on Orion. Relevant stack trace information are provided below:

```
forrtl: error (73): floating divide by zero
Image              PC                Routine            Line        Source             
ufs_model          000000000A5E9C4B  Unknown               Unknown  Unknown
libpthread-2.17.s  00002B512D4CB5D0  Unknown               Unknown  Unknown
ufs_model          00000000021385DB  mapl_basemodmapl_        3701  MAPL_Base.F90
ufs_model          0000000002134931  mapl_basemod_mp_m        3510  MAPL_Base.F90
ufs_model          0000000001DCAA5D  su2g_gridcompmod_         868  SU2G_GridCompMod.F90
ufs_model          0000000001DBC198  su2g_gridcompmod_         719  SU2G_GridCompMod.F90
```

The MAPL library (branch: `main`, revision `bc57d97c`) was built without debug flags and with `BUILD_WITH_FLAP=NO`.

bena-nasa:
Fixes #680 

This is an enhancement to the IJ search algorithm needed to bin point sources to a grid. This replaces the logic of the cubed-sphere with a generic bisection search testing if the point is in a spherical polygon. This fixes an issue when using this with the NOAA UFS which maybe different unit and rotation conventions. Also eliminates the save that was used, instead goes back to the old trick of stuffing the full corners in grid as attributes. This seemed like the least intrusive way to do this.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 =======
 - Fixes incorrect referrencing of R4 variables when reading R8 tile vars in base/MAPL_IO.F90
 - Use integer to represent logical internally in pFIO utilities to avoid non-standard transfer
+- Modified horizontal ij search algorithm in MAPL_Base.F90 to use general bisection search to fix capaibility issues with non-gmao created cubed sphere grids
 
 ### Removed
 
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -44,7 +44,7 @@ set (srcs
   MAPL_LocStreamFactoryMod.F90          MAPL_LocstreamRegridder.F90
   MAPL_ExternalGridFactory.F90
   ServerManager.F90 ApplicationSupport.F90
-  regex_module.F90 StringTemplate.F90
+  regex_module.F90 StringTemplate.F90 MAPL_SphericalGeometry.F90
   regex_F.c
   c_mapl_locstream_F.c  getrss.c  memuse.c
   )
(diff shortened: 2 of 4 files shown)


atrayano (changes requested):
@bena, could you review line 160 of base/MAPL_SphericalGeometry.F90. The usage of parentheses does not look right. If I need to guess, most likely you meant 
in_poly=(count(signs)==0) .or. (count(signs)==4)
instead of 
in_poly=(count(signs)==0) .or. (count(signs==4))

bena-nasa:
Fixed it

bena-nasa:
Making this contingent for now. When I ran both algorithms simultaneously and compared the assigned indices it never complained that they did not match. But now after stripping out the old algorithm I'm finding it is non-zero diff so I'm confused. I'll do some more checking but keep the PR as DNA to save retyping later.

tclune (approved):

tclune:
@atrayano - I think they all need to be positive if you traverse the polygon in a counter clockwise direction.   I oversimplified when I stated the theorem last week.

PR #683, opened 2021-02-08
Fixes #248. Update Python MAPL constants

### LINKED ISSUE #248: Python MAPL Constants are incorrect
Apparently MAPL's Python layer has its own constants file, [constants.py](../blob/master/MAPL_Base/Python/MAPL/constants.py), that is a bit out of date compared to [MAPL_Constants.F90](../blob/master/MAPL_Base/MAPL_Constants.F90).

As an example, in Fortran:
https://github.com/GEOS-ESM/MAPL/blob/4b99ddd45df0f5c3c3a25882087bc92aa2221087/MAPL_Base/MAPL_Constants.F90#L30
and in Python:
https://github.com/GEOS-ESM/MAPL/blob/4b99ddd45df0f5c3c3a25882087bc92aa2221087/MAPL_Base/Python/MAPL/constants.py#L6

First, the Python version should be updated, obviously. I suppose as a hotfix on `master` and then merged up to `develop` (right @tclune ?)

But the real question is, how can we make sure this never happens again? This might need some clever bit of engineering thought. Maybe when yaFyaml is integrated both Fortran and Python should read/parse a YAML file of constants? The trick then would be to get the Python and Fortran code to both know where the YAML file is. Maybe Arlindo (not on Github yet) should be consulted?

mathomp4:
## Description

This updates the Python MAPL constants to be in concert with `MAPL_Constants.F90`

## Related Issue

#248 

## Motivation and Context

Keeps the MAPL Python constants in concert with the MAPL Fortran constants as found in [`MAPL_Constants.F90`](https://github.com/GEOS-ESM/MAPL/blob/develop/shared/MAPL_Constants.F90).

## How Has This Been Tested?

I am not sure how to test this, but it seems like a good idea?

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,16 +8,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+
 - Add option fast_oclient that waits before using oserver. It would not wait after done message are sent 
+  
 ### Changed
 
 - For MultiGroupServer, the backend and frontend share each node
 - Moved `MAPL_HistoryGridComp.F90` and related files to `gridcomps/History`
+- Updated `Python/MAPL/constants.py` to have the same constants as
+  `MAPL_Constants.F90`
 
 ### Fixed
 
 - Fixes for allowing forked PRs to pass CI
-=======
 - Fixes incorrect referrencing of R4 variables when reading R8 tile vars in base/MAPL_IO.F90
 - Use integer to represent logical internally in pFIO utilities to avoid non-standard transfer
 
--- a/Python/MAPL/constants.py
+++ b/Python/MAPL/constants.py
@@ -3,31 +3,56 @@ Python version of MAPL Constants.
 """
 
 MAPL_PI     = 3.14159265358979323846
-MAPL_GRAV   = 9.80                   # m^2/s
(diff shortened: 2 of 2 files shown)


mathomp4:
Note. I'm adding @WilliamJamieson as an explicit reviewer as he can make sure things look "Python-right" (I essentially took the Fortran code and then "de-Fortran'd" it. So, there might be a lingering Fortran-ism in it. 

I've also added @amdasilva as a reviewer as in my brain he's the not-MAPL person to best approve this. *I* might think it's a good idea to keep the Python constants the same as the Fortran constants, but @amdasilva might have a different opinion which would be obviously correct!

mathomp4:
Also, I've labeled as both zero-diff and non-zero-diff. It's zero-diff to the GCM, etc. as, well, the GCM doesn't use this. But it is obviously non-zero-diff to anything that might use the Python constants.

bena-nasa (approved):

WilliamJamieson (approved):

mathomp4:
Arlindo seemed to approve, so I've removed the blocker from this.

bena-nasa (dismissed):

WilliamJamieson (dismissed):

bena-nasa (approved):

PR #684, opened 2021-02-09
Fixes #232. Add time and date validator

### LINKED ISSUE #232: Add sanity checks for times
This is inspired by an issue a user had when they had a reftime in history of `2100000`. Notice, that's a bit too big.

Thus, until we move (:crossed_fingers:) to using ISO times, it would be good to make sure that the first two digits are between 00 and 23, then 00 and 59, and 00 and 59, so that the requested time is a valid time.

mathomp4:
## Description

Adds a date and time checker for those places where MAPL expects dates and times to be a certain valid format. This is inspired by #232 where you could have a `ref_time` of `2100000` which would "work" unexpectedly.

## Related Issue

#232 

## Motivation and Context

Solves the issue where invalid times could be passed in.

## How Has This Been Tested?

Added some funit tests and I've tested to make sure things don't fall apart if, say, a bad `ref_time` is passed in.

## Types of changes

## Checklist:


shared/TimeUtils.F90
```diff
@@ -0,0 +1,99 @@
...
+      if (time < 0) then
+         is_valid = .false.
+      else if (time > 240000) then
+         is_valid = .false.
+      else if (hours > 24) then
+         is_valid = .false.
+      else if (minutes > 59) then
+         is_valid = .false.
+      else if (seconds > 59) then
+         is_valid = .false.
```
tclune, line 84: What about leap seconds?

tclune (dismissed):
Approving - but please do consider whether this should allow for leap seconds.

shared/TimeUtils.F90
mathomp4: I'm not sure *ESMF* would allow 23:59:60 as a valid time so we might be safe ignoring it.

tclune (approved):

PR #687, opened 2021-02-10
Feature/bmauer/locstream variance

bena-nasa:
Add option at request of GMAO user to compute tile variance when doing a T2G transform with locstream. 
They apparently need this for some sort of experimentation they want to do.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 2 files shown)


atrayano (commented):
@bena-nasa , the coding looks totally fine. I am not quite sure if summing the squares of (var-mean) is exactly "variance". At the very least it is the square of the variance. Also, in the traditional definition (i.e. all the weights are 1) the expression involves division by (n-1), while here we effectively divide by n

atrayano (approved):
Looks OK to me

PR #690, opened 2021-02-12
CFIORead/Write replacement

bena-nasa:
This implements very simple ESMF bundle IO routines to allow the eventual retirement of MAPL_CFIO.F90 and the MAPL_cfio library. This is a few simple bundle IO subroutines that can be used to replace the extensive use of MAPL_CFIORead/Write in GEOS. MAPL_CFIORead/Write is general used for non performance critical IO that respects and understands GEOS metadata conventions and extensively used through GEOS both for the model and other applications (GSI,mkiau, etc...). So while nothing terribly exciting this is a very crucial basic capability that many applications are relying on.

This relies entirely on the work done to use the IO server in ExtData and History and all IO is done through whatever server the application has instantiated. This does mean any application that wants to use these must start a pfio server but using this allows me to reuse everything already done with History and ExtData for file IO through the server.

In addition I added support for the "old" cubed sphere style with a new grid factory. The create a grid from metadata detection has now been modify accordingly for compatibility with the "old" format cubed sphere files. This means for example History and produce, and ExtData can now read old style cubed-sphere files.

I have converted Regrid_Util.x to use this layer instead of MAPL_CFIO for file IO. In addition I added a test (write a file, read it, make sure they match) to the tests directory under base.

The motivation for this is that MAPL_CFIORead still uses the old regridders which we have established cause a significant memory leak and the plan was always to try to retire MAPL_CFIO. See this issue: https://github.com/GEOS-ESM/GEOSgcm/issues/244

I have confirmed that if I use the new subroutine in place of MAPL_CFIORead if fixes the memory leak. In addition I tested  calls of MAPL_CFIORead elsewhere (i.e. in mkiau for example where it reads the increments) and make sure everything was still working like replay.

Once this is used, it may not be zero-diff. Since this is using the ESMF regridders, any replacement of MAPL_CFIORead with the new routine maybe not be zero diff since they are using different regridding layers so depends on the circumstances. So this is zero diff in that any existing call has not changed.

Finally I'm not wedded to the module/file/subroutines names, we can discuss if other names make more sense and if there are general design issues. I wanted to get the functionality there.

## Types of changes

## Checklist:


atrayano (approved):
@bena-nasa, the code looks OK to me. One minor comment, we decided that for Lizzie, we are changing the "anchor" to plural (i.e. "anchors"). I am going to approve this PR, but we should change it at some point

bena-nasa:
I do have one, very, very slight concern. Until this point, only History and ExtData use the pfio server. Now this opens the possibility that in an application like GEOS you now would have multiple bits of the application say History/Data and now any other odd read write using the server.
Since we never run the i-server asynchronously this is at present no concern for reads, but if we were to it could be. Also to my knowledge, MAPL_CFIOWrite (which we would be replacing in applications) is never called in an application that uses History like GEOSgcm.x, GEOSldas.x, and GEOSctm.x, they all use history, the only reason MAPL_CFIOWritre existed was for the odd applications like mkaiu.x and GISsa.x that need to write files and use the same metadata and file creation layer for compatibility. So there is no use case I can think of where these would presently cause some potential conflict.
Still, probably worth thinking about a little.

atrayano:
I cannot think at the moment of any use case either. Unless Ricardo has some old executable that both of us have not seen. 

Pert is using CFIO for reading and writing (the writing is done by History, but without MAPL_Cap). But the reading is not asynchronous

tclune (approved):
We should discuss whether the new file under base/tests could be converted to a suite of pFUnit unit tests.

tclune:
Regarding potential conflicts with side I/O:   The intent of the DirectoryService was to allow multiple clients to connect with one or more servers.     Currently the design is pretty much locked into 1 client per server.          

I think the DirectoryService persists though, so it is possible for other layers to add a server and a client.   In practice these would need to be synchronous cases as otherwise additional logic would be needed to carve off cores at the beginning of the run.    The long term ideal solution is that one server has a front end and not only listens for new requests from "known" clients, but also listens for new clients.        At one point I convinced myself this could work, but it was in the early design phases of pFIO.

PR #693, opened 2021-02-15
Remove need for HDF5 Fortran bindings

sdeastham:
## Description
HDF5 builds by default without Fortran bindings, and these do not seem to be used anywhere in MAPL. As such it should be sufficient to require only the HDF5 package, not also the HDF5 Fortran bindings. This will improve accessibility for non-Baselibs users.

## Motivation and Context
When building MAPL for GCHP, the `find_package(HDF5 REQUIRED Fortran)` line was causing issues, because some clusters which have prebuilt NetCDF libraries did not build the HDF5 Fortran bindings. Since the Fortran bindings do not seem to be necessary for MAPL, this requirement can be removed, enabling the users to proceed. 

## How Has This Been Tested?
Built GCHP 13.0 (including MAPL) after removing the ` Fortran` requirement.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Move to ESMA_cmake v3.3.6
+- Remove requirement for HDF5 Fortran bindings in MAPL
 
 ### Fixed
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if (NOT Baselibs_FOUND)
   add_definitions(-DNETCDF_NEED_NF_MPIIO)
   add_definitions(-DHAS_NETCDF3)
 
-  find_package(HDF5 REQUIRED Fortran)
+  find_package(HDF5 REQUIRED)
   if(HDF5_IS_PARALLEL)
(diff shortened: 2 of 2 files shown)


mathomp4:
@sdeastham I'm actually removing the trivial label and going full 0-diff. I'd like to see the CI work in full with GitHub Actions.

mathomp4:
@sdeastham Also: There is another MAPL CMake Hotfix coming in soon. So, we might need you to try your system with the merge of the two branches.

mathomp4:
Also, I'm going to ping @biljanaorescanin about this. The GEOSgcm doesn't need HDF5 Fortran interfaces, I think GEOSldas does. Indeed, I might want to look at adding a "build GEOSldas" CI test to MAPL as well. 

(Though, since we always have HDF5 with Fortran and HL, it will never fail for us.)

mathomp4:
Okay. I was able to build GEOSldas with out the `Fortran` here in MAPL. I've made a (draft) PR of adding it back to GEOSldas as seen here: https://github.com/GEOS-ESM/GEOSldas/pull/366

I'll wait to un-draft that until MAPL 2.6.1 is out.

mathomp4:
I've merged in #694 into this branch. And for some reason CI was canceled? Sigh.

mathomp4 (approved):

PR #695, opened 2021-02-16
Merge main into develop (hotfixes)

mathomp4:
This pulls in the hotfixes on `main` into `develop`


mathomp4:
Nuts. Need to do this by hand it looks like. Not sure why...

PR #697, opened 2021-02-16
add seconds to string template

bena-nasa:
Add %S (note capital!, lower case was already used) to allow seconds in the string template function used by History etc, note allows one to output to a unique file every timestep when the timestep is not a whole minute.

## Types of changes

## Checklist:


atrayano (approved):
Looks OK

PR #708, opened 2021-02-22
Fixes #538. Use MAPL Error Macros.

### LINKED ISSUE #538: Some files in MAPL define their own error macros
Just adding this as a reminder to clean this up as time permits.
I noticed that some files in MAPL are defining their own ASSERT, VERIFY, etc... macros. The grid factories in particular. These should be cleaned up to use the common error handling macros.

mathomp4:
## Description

Attempt to clean up the use of self-defined `_ASSERT` etc. In one case, `pfio/DownBit.F90` it looks like it doesn't use *any* macros at all. So, I just removed them. Maybe @weiyuan-jiang or @tclune can tell me if that is the "right" thing to do. Or should I re-introduce MAPL Macros and change how it's handling `rc`?

## Related Issue

#538 

## Motivation and Context

Keeps MAPL consistent.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Change one sided mpi_put to mpi_send and receive pair in the class MultiGroupServer
 - Change command line interface to --npes_backend_pernode to avoid confusion
+- Remove self-defined-in-file MAPL macros
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


mathomp4:
I just built and ran GEOS with this change and...nothing blew up? I guess if @weiyuan-jiang (for pfio) and @GEOS-ESM/mapl-team approves of the Range fix. it should be pretty simple.

@bena-nasa can you see any other places where this should be fixed? The other files all seemed to be Tests which I wasn't as concerned with.

tclune (approved):

PR #719, opened 2021-02-24
Remove duplicate and unnecessary use statements

rmontuoro:
## Description

This PR removes duplicate and unnecessary `use` statements.

## Motivation and Context

Code maintenance and cleanup
## How Has This Been Tested?

These changes were tested on Orion, building MAPL as standalone library and using it to build and run UFS-GOCART.
## Types of changes

## Checklist:

### DIFF
--- a/generic/MaplGeneric.F90
+++ b/generic/MaplGeneric.F90
@@ -2,7 +2,6 @@ module MaplGeneric
    use mapl_AbstractComponent
    use mapl_MaplComponent
    use mapl_ComponentSpecification
-   use mapl_VariableSpecification
    use mapl_StateSpecification
    use mapl_VarSpecVector
    use mapl_VarSpecMod
--- a/pfio/MultiCommServer.F90
+++ b/pfio/MultiCommServer.F90
@@ -17,7 +17,6 @@ module pFIO_MultiCommServerMod
    use pFIO_ServerThreadVectorMod
    use pFIO_AbstractSocketMod
    use pFIO_AbstractSocketVectorMod
-   use pFIO_AbstractDataReferenceMod
    use pFIO_AbstractServerMod
    use gFTL_StringInteger64Map
    use pFIO_AbstractMessageMod
--- a/pfio/MultiCommServer.F90
+++ b/pfio/MultiCommServer.F90
@@ -74,30 +74,32 @@ module pFIO_MultiCommServerMod
 
 contains
 
-   function new_MultiCommServer(server_comm, port_name, nwriter, rc) result(s)
+   function new_MultiCommServer(server_comm, port_name, nwriter_per_node, rc) result(s)
       type (MultiCommServer) :: s
       integer, intent(in) :: server_comm
       character(*), intent(in) :: port_name
-      integer, intent (in) :: nwriter
+      integer, intent (in) :: nwriter_per_node
       integer, optional, intent(out) :: rc
-      integer :: s_rank, s_size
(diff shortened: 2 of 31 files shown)


generic/MaplGeneric.F90
```diff
@@ -2,7 +2,6 @@ module MaplGeneric
    use mapl_AbstractComponent
    use mapl_MaplComponent
    use mapl_ComponentSpecification
-   use mapl_VariableSpecification
```
tclune, line 5: Is this change necessary?   The other layers of MAPL don't yet use this (new) data type, but they will eventually.

pfio/MultiCommServer.F90
```diff
@@ -17,7 +17,6 @@ module pFIO_MultiCommServerMod
    use pFIO_ServerThreadVectorMod
    use pFIO_AbstractSocketMod
    use pFIO_AbstractSocketVectorMod
-   use pFIO_AbstractDataReferenceMod
```
tclune, line 20: Just a question on this one:    are you just noticing that some of these USE's are unnecessary?    How do you detect this.

tclune (commented):
Will approve module response to the first question.

generic/MaplGeneric.F90
rmontuoro: @tclune This change attempts to fix the following compiler error:
```
[...]/GOCART/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90(894): error #6405: The same named entity from different modules and/or program units cannot be referenced.   [MAPL_VARSPEC]
    type(MAPL_VarSpec), pointer       :: InternalSpec(:)
```

Module `mapl_VariableSpecification` exposes derived type `MAPL_VarSpec`, which is also exposed by module `MAPL_VarSpecMod`. Based on last week's discussions, my understanding is that `MAPL_VarSpecMod` is the proper "owner" of `MAPL_VarSpec`.

mathomp4:
Actually, this looks to be against `main`. This should be a PR to `develop`

pfio/MultiCommServer.F90
rmontuoro: @tclune This `use` statement is duplicated. See the [identical statement](https://github.com/rmontuoro/MAPL/blob/beb095f4a14f008fbbef061a135b31a1dab12236/pfio/MultiCommServer.F90#L23) still in the source code (line 23).

rmontuoro:
> Actually, this looks to be against `main`. This should be a PR to `develop`

This PR fixes a compiler error (see comment above), that's why it was issued against `main`.

mathomp4:
> > Actually, this looks to be against `main`. This should be a PR to `develop`
> 
> This PR fixes a compiler error (see comment above), that's why it was issued against `main`.

Ah. GitHub decided not to refresh my page. Thanks web caching!

tclune:
This PR should also be for develop.  Sorry for the extra work.

rmontuoro:
Updated and issued against `develop`.

generic/MaplGeneric.F90
tclune: I thought the solution we discussed was a change in MAPL_Generic.F90.   `MAPL/generic` is the proper owner of this module and should be exporting this.

The error you refer to should not happen when starting a clean build.    If it does, lets handle this as a separate hotfix on main.

(These changes otherwise all need to be redone from a base off of develop.  Sorry)

PR #716, opened 2021-02-23
Fearure/wjiang/remove gfortran workaround

weiyuan-jiang:
This workaround which introduces memory leak is not necessary with newer gfortran compiler (9.3 and above). close #493 
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [ ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [x] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #725, opened 2021-02-25
Feature/bmauer/locstream pfaf grid

bena-nasa:
This PR allows one to "attach" to the pfafstetter grid under the land tiles. This is needed by the river routing model in development. The river routing model runs directly on the catchments but needs to import runoff from the catchment component. The solution was to create a fake 1D grid that indexes the catchment space. Then we "attach" this grid to the pfaffstetter grid. Then we have the information needed to regrid the tiles to the catchments. Unfortunately we can only specify 2 grids in the file, there is rather baked in assumptions that you will only attach to one of two named grids in the file and there can only be two named grids. Up til now the only two grids we ever had to attach to were the ocean and atmospheric. The least disruptive thing we decided on was to add a new option to force attachment to the 2nd grid (use_pfaf) which only makes sense when the land mask is applied. Note this is not different in that it only makes sense attach to the ocean grid if you are on ocean tiles. However, we need to know the global number of catchments. The global sizes are normally in the grid definition but as I said we really can't add a third. This was not in the file so Sarith and I have decided that when the river routing is implemented make new tile files that will put the pfafsetter number  as the 2nd number in the first line. Right now this line had, nt, nx, ny; the number of tiles, and the raster size used to generate it. Note that locstream never reads the raster size, it only cares about nt so this does not break any backwards compatibility with locstream.

In addition I had to make tweeks elsewhere for to make this "fake" 1d grid (I just use latlon grid factory).

For full functionality this is depend on PR https://github.com/GEOS-ESM/GMAO_Shared/pull/162  in GMAO_Shared, this will copy the 2nd number from the first line of the ascii to the binary. Note that this will not break any existing functionality and is functional with the routing work Sarith is doing already if you use an appropriate ascii tile file.

## Types of changes

## Checklist:


tclune (dismissed):

base/MAPL_LocStreamMod.F90
```diff
@@ -329,6 +338,30 @@ subroutine MAPL_LocStreamGet(LocStream, NT_LOCAL, TILETYPE, TILEKIND, &
        tilegrid = locstream%Ptr%TILEGRID
     end if
 
+    if (present(local_i)) then
+#ifdef __GFORTRAN__
```
mathomp4, line 342: Hmm. I wonder if we just shouldn't add the `#ifdef __GFORTRAN__` bits here? We are planning to remove them in #716

atrayano: @mathomp4  I agree, the #ifdef __GFORTRAN__ is not needed anymore. @bena-nasa could you simply remove this little section

base/MAPL_LocStreamMod.F90
```diff
@@ -666,6 +735,11 @@ subroutine MAPL_LocStreamCreateFromFile(LocStream, LAYOUT, FILENAME, NAME, MASK,
           call MAPL_CommsBcast(vm, DATA=STREAM%TILING(N)%IM, N=1, ROOT=0, RC=status)
           call MAPL_CommsBcast(vm, DATA=STREAM%TILING(N)%JM, N=1, ROOT=0, RC=status)
        enddo
+       if (use_pfaf_) then
+          STREAM%TILING(2)%IM = stream%pfafstetter_catchments
+          STREAM%TILING(2)%JM = 1
+          STREAM%TILING(2)%name = "ROUTE_GRID"
+       end if
```
atrayano, line 743: Can we call the new grid "Catchment grid", or "Pfafstetter grid" insead of "ROUTE GRID"

base/MAPL_LocStreamMod.F90
```diff
@@ -2821,6 +2897,7 @@ integer function GRIDINDEX(STREAM,GRID,RC)
         exit
      end if
   end do
+  if (trim(name)=="ROUTE_GRID") GridIndex=2
```
atrayano, line 2901: Same comment as above

base/MAPL_LocStreamMod.F90
bena-nasa: Removed all gfortran references.

atrayano (approved):
Looks good

PR #729, opened 2021-03-02
Prevent History from duplicating output fields whose names end with the component name

rmontuoro:
## Description

Fields included in History collections may accidentally be duplicated when their name end with their corresponding component name, if such a name includes non-alphabetical characters used to identify expressions in `MAPL_SetExpression()`. For example, this happens when running two instances of GOCART's carbonaceous aerosol (CA) component: one for black carbon, named `CA.bc`, and the other for organic carbon, named `CA.oc`. Exported fields from each component are differentiated by adding to the component's name to their original name (_e.g._ `CAphilicCA.bc`).

This PR introduces changes to explicitly take into account this case, preventing History from duplicating such fields and generating the error: `existing item CAphilicCA.bc not replacable`

## Motivation and Context

This PR addresses an issue potentially preventing History from writing exported fields whose names end with their component's name.

## How Has This Been Tested?

The changes were tested on Orion using a UFS-GOCART test case.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [ ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [x] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #730, opened 2021-03-02
Prevent History from duplicating output fields whose names end with their component name

rmontuoro:
## Description

Fields included in History collections may accidentally be duplicated when their name end with their corresponding component name, if such a name includes non-alphabetical characters used to identify expressions in `MAPL_SetExpression()`. For example, this happens when running two instances of GOCART's carbonaceous aerosol (CA) component: one for black carbon, named `CA.bc`, and the other for organic carbon, named `CA.oc`. Exported fields from each component are differentiated by adding to the component's name to their original name (_e.g._ `CAphilicCA.bc`).

This PR introduces changes to explicitly take into account this case, preventing History from duplicating such fields and generating the error: `existing item CAphilicCA.bc not replacable`

## Related Issue

This PR reintroduces changes submitted with PR #729 
## Motivation and Context

This PR addresses an issue potentially preventing History from writing exported fields whose names end with their component's name.

## How Has This Been Tested?

The changes were tested on Orion using a UFS-GOCART test case.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Have CMake automatically gitignore build and install dirs
+- Prevent History from mistakenly duplicating export fields
+  when field names end with their component's name
 
 ### Removed
 
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -4736,11 +4736,18 @@ ENDDO PARSER
   iRealFields = 0
   allocate(isBundle(nfield))
   do m=1,nfield
-    if (scan(trim(fields(1,m)),'()^*/+-.')/=0) then
+    VarName = fields(1,m)
+    i = len_trim(fields(1,m))
+    j = len_trim(fields(2,m))
+    idx = i-j
+    if (idx > 0) then
+       if (fields(1,m)(idx+1:i) == fields(2,m)(1:j)) VarName = fields(1,m)(1:idx)
+    end if
+    if (scan(trim(VarName),'()^*/+-.')/=0) then
        rewrite(m)= .TRUE.
        tmpfields(m)= trim(fields(1,m))
     else
-       if (index(fields(1,m),'%') == 0) then
+       if (index(VarName,'%') == 0) then
           iRealFields = iRealFields + 1
           rewrite(m)= .FALSE.
           isBundle(m) = .FALSE.


tclune:
Adding @bena-nasa as the reviewer.

tclune:
@rmontuoro   we can't see what the actual problem is here.   Could you propose some times where you, @bena-nasa and myself could discuss?

bena-nasa:
I think you have hit a corner case in History that we have not hit before.
If you add aliases to the fields that will get you past this. This issue has nothing to do with the component or instant name in the field, you just picked the field name that triggered this.  It is misinterpreting that as an expression because of the "." in the field name and causing a duplicate field to be added because fo the quirks of how the expressions are handled. So try adding the alias:
```
  inst_ca_3d.format:     'CFIO' ,
  inst_ca_3d.descr:      '3d,3-Hourly,Instantaneous,Model-Level,Analysis,Aerosol Concentrations',
  inst_ca_3d.template:   '%y4%m2%d2_%h2%n2z.nc4',
  inst_ca_3d.mode:       'instantaneous',
  inst_ca_3d.grid_label:  PC720x361-DC ,
  inst_ca_3d.splitField:  1,
  inst_ca_3d.frequency:   030000 ,
  inst_ca_3d.duration:    030000 ,
  inst_ca_3d.ref_time:    000000 ,
  inst_ca_3d.nbits:      10,
  inst_ca_3d.fields:   'CAphilicCA.bc' , 'CA.bc'   , 'CAphillicCA_bc' ,
                       'CAphobicCA.bc' , 'CA.bc'   , 'CAphobicCA_bc' ,
                       'CAphobicCA.oc' , 'CA.oc'   , 'CAPphobicCA_oc' ,
                       'CAphobicCA.oc' , 'CA.oc'   ,  'CAphobicCA_oc' ,
                      ::
```
This should get you past the crash, while I make a patch.

rmontuoro:
@bena-nasa - Adding aliases for the CA fields as you specified causes the following error, which is the same behavior I was observing early on:
```
FAIL at line=05086    MAPL_HistoryGridComp.F90                 <Field CAphillicCA_bc not found>
```

bena-nasa:
Ok, I have made a patch for what I assumed was the issue I was able to see which I assumed was the "." in the name. Now that I can run UFS/Gocart let me investigate this. I don't know why using the alias works for GEOS but would not work here.

mathomp4:
@bena-nasa @rmontuoro Can this PR be closed with #734 being in MAPL?

rmontuoro:
PR #734 superseded this PR and fixed the related issue.

PR #734, opened 2021-03-03
Bmauer/feature/#731 fixdotinname

### LINKED ISSUE #731: Handling for field names with "." needs to be fixed
The UFS/MAPL/GOCART work exposed a bug in History.
The issue is when you have "." in the field names and put them in a collection:
```
  inst_ca_3d.format:     'CFIO' ,
  inst_ca_3d.descr:      '3d,3-Hourly,Instantaneous,Model-Level,Analysis,Aerosol Concentrations',
  inst_ca_3d.template:   '%y4%m2%d2_%h2%n2z.nc4',
  inst_ca_3d.mode:       'instantaneous',
  inst_ca_3d.frequency:   030000 ,
  inst_ca_3d.fields:   'fooCA.bc' , 'CA.bc'    ,
                      ::
```
The problem is that History interprets this as an expression because of the "."
The causes History to unnecessarily add the field twice to the state. History tries to identify extra fields in the expression that are not already in the collection as it can pull these from the component specified in the expression. The "." in the name meant that "fooCA.bc" was not being scanned for in the other fields in the collection because of the presence of one of these special characters "()+-/*." and hence was not being added to the list of fields already in the collection. Then it thinks that is an expression because it was making this determination based on the presence of one of these special characters "()+-/*.". Technically this is an expression in that the name of a single field is a perfectly valid expression, it is just treated as an identity by the parser, if the expression is "fooCA.bc" and "fooCA.bc" is a field in the state delivered to the parser it fill the output field with "fooCA.bc". The bottom line, there is a routine that returns a list of fields that in this case is just "fooCA.bc" and adds it to a state. This should not have happened in the first place.

 Then history loops over the actual  list of fields in the collection and adds those to this same state. The name of the field added is actual the alias, which if none is provided is the same as the short name. Expressions must have aliases. 

At this point "fooCA.bc" is added to this state, which already as "fooCA.bc" and is an error!

So the routine that parses the expression should be smarter. It should first do a check, is "fooCA.bc" in the component named in entry for that field. If that is the case, we are done, it is not an expression and it should be added to the list of fields already in the collection. I will make the appropriate PR to fix this.

bena-nasa:
Fixes #731 
This fixes the issue that Raffaele covered, that having "." in the variable name confused History. The issue is **NOT** the parser. The issue was the logic I have to determine if the variables in the expression are already in the collection and that not being properly detected leading to duplicate fields being placed in a state.
I modified the code so when scanning the fields it first checks if the field string actually is a field in the export state, rather than just assuming if you find one of these characters "+-/*()." it is an expression.
I was able to output fields with "." in the name with this fix.

## Types of changes

## Checklist:

### DIFF
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -4750,6 +4750,7 @@ ENDDO PARSER
           isBundle(m) = .FALSE.
           tmpfields(m)= trim(fields(1,m))
        else 
+          isBundle(m) = .false.
           rewrite(m)= .TRUE.
           tmpfields(m)= trim(fields(1,m))
        end if
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -4731,26 +4731,34 @@ ENDDO PARSER
 ! rather than the actual output field variables (i.e., fields(1,:)).
 ! Also do check that there are no illegal operations
 !-------------------------------------------------------------------
+  allocate ( exptmp (1), stat=status )
+  _VERIFY(STATUS)
+  exptmp(1) = ExpState
   ! check which fields are actual exports or expressions
   nPExtraFields = 0
   iRealFields = 0
   allocate(isBundle(nfield))
   do m=1,nfield
-    if (scan(trim(fields(1,m)),'()^*/+-.')/=0) then
-       rewrite(m)= .TRUE.
-       tmpfields(m)= trim(fields(1,m))
-    else
-       if (index(fields(1,m),'%') == 0) then
+
+    call MAPL_ExportStateGet(exptmp,fields(2,m),state,rc=status)
+    _VERIFY(STATUS)
+    if (index(fields(1,m),'%') == 0) then
+       call ESMF_StateGet(state,fields(1,m),field,rc=status)
+       if (status==_SUCCESS) then
           iRealFields = iRealFields + 1
           rewrite(m)= .FALSE.
           isBundle(m) = .FALSE.
           tmpfields(m)= trim(fields(1,m))
-       else
-          isBundle(m)=.true.
-          rewrite(m)= .FALSE.
+       else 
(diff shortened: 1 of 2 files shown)


atrayano (dismissed):
Looks OK

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -4731,26 +4731,34 @@ subroutine MAPL_SetExpression(nfield,fields,tmpfields,rewrite,nPExtraFields, &
...
+       if (status==_SUCCESS) then
           iRealFields = iRealFields + 1
           rewrite(m)= .FALSE.
           isBundle(m) = .FALSE.
           tmpfields(m)= trim(fields(1,m))
-       else
-          isBundle(m)=.true.
-          rewrite(m)= .FALSE.
+       else 
+          rewrite(m)= .TRUE.
```
tclune, line 4753: This conditional branch no longer provides a value for isBundle(m)

mathomp4:
Does this PR supersede #730?

atrayano:
@mathomp4 I think it does supersede #730. But to be 100% sure, please check with @bena-nasa

mathomp4:
> @mathomp4 I think it does supersede #730. But to be 100% sure, please check with @bena-nasa

And I guess we should ask @rmontuoro to try out this branch as well. If it works for him, then #730 can be closed in favor of this.

gridcomps/History/MAPL_HistoryGridComp.F90
bena-nasa: I fixed that.

bena-nasa:
> Does this PR supersede #730?

Yes

rmontuoro:
The current `develop` branch crashes UFS-GOCART (see issue #737). However, applying the changes from this PR to my MAPL fork ([bugfix/rmontuoro/ufs](https://github.com/rmontuoro/MAPL/tree/bugfix/rmontuoro%2Fufs) - cccff393) fixes the CA field duplication issue. Thank you, @bena-nasa!

tclune (approved):

PR #736, opened 2021-03-03
Fix bug in HorzIJ routine

bena-nasa:
Fix issue with HorzIJ routine. It had been generalized to work for the UFS/GOCART/NOAA work. However, I forgot to check the centers that are retrieved from the grid is those are degrees as well and convert those to radians too if needed.

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_Base.F90
+++ b/base/MAPL_Base.F90
@@ -2679,6 +2679,10 @@ and so on.
 
          call ESMF_GridGet(grid,coordSys=coordSys,rc=status)
          _VERIFY(status)
+         if (coordSys==ESMF_COORDSYS_SPH_DEG) then
+            gridCornerLons=gridCornerLons*MAPL_DEGREES_TO_RADIANS
+            gridCornerLats=gridCornerLats*MAPL_DEGREES_TO_RADIANS
+         end if
          allocate(lons1d(size(gridCornerLons,1)*size(gridCornerLons,2)),stat=status)
          _VERIFY(status)
          allocate(lats1d(size(gridCornerLons,1)*size(gridCornerLons,2)),stat=status)
@@ -2691,10 +2695,6 @@ and so on.
                lats1d(idx)=gridCornerLats(i,j)
             enddo
          enddo
-         if (coordSys==ESMF_COORDSYS_SPH_DEG) then
-            lons1d=lons1d*MAPL_DEGREES_TO_RADIANS
-            lats1d=lats1d*MAPL_DEGREES_TO_RADIANS
-         end if 
          call ESMF_AttributeSet(grid, name='GridCornerLons:', &
                itemCount = idx, valueList=lons1d, rc=status)
          _VERIFY(STATUS)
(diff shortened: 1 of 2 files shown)


rmontuoro:
While I could not test this branch since MAPL `develop` crashes UFS-GOCART (see issue #737), applying these changes to my MAPL fork ([bugfix/rmontuoro/ufs](https://github.com/rmontuoro/MAPL/tree/bugfix/rmontuoro%2Fufs) - cccff393) resulted in `SO2EMVN` still showing a pattern resembling the one in GOCART issue [#28](https://github.com/GEOS-ESM/GOCART/issues/28) during the first few output time steps. This pattern blends into a more realistic one as GOCART steps forward.

tclune:
Hmm.  It is hard to see how there could be any transitory aspects of this layer.

bena-nasa:
I found the bug, I was correcting the units on the arrays that get stuffed into the grid if those were not already attached but not the output arrays with the coordinates in the MAPL_GridGetCorners subroutine. On subsequent calls when we pull the corners from the grid, they are correct. This explains why the first time was wrong.

base/MAPL_Base.F90
```diff
@@ -2679,6 +2679,10 @@ subroutine MAPL_GridGetCorners(grid,gridCornerLons, gridCornerLats, RC)
 
          call ESMF_GridGet(grid,coordSys=coordSys,rc=status)
          _VERIFY(status)
+         if (coordSys==ESMF_COORDSYS_SPH_DEG) then
+            gridCornerLons=gridCornerLons*MAPL_DEGREES_TO_RADIANS
+            gridCornerLats=gridCornerLats*MAPL_DEGREES_TO_RADIANS
+         end if
```
tclune, line 2685: Still needs an else clause for CARTESIAN case

base/MAPL_Base.F90
```diff
@@ -3310,7 +3310,8 @@ subroutine MAPL_GetHorzIJIndex(npts,II,JJ,lon,lat,lonR8,latR8,Grid,IMGlob,JMGlob
      real(kind=REAL64), parameter :: PI_R8     = 3.14159265358979323846_REAL64
```
tclune, line 3310: Why can't we use MAPL_PI here?

base/MAPL_Base.F90
```diff
@@ -3347,14 +3348,27 @@ subroutine MAPL_GetHorzIJIndex(npts,II,JJ,lon,lat,lonR8,latR8,Grid,IMGlob,JMGlob
...
+        allocate(center_lons(im,jm),center_lats(im,jm))
+
+        if (coordSys==ESMF_COORDSYS_SPH_DEG) then
+           center_lons=lons*MAPL_DEGREES_TO_RADIANS
+           center_lats=lats*MAPL_DEGREES_TO_RADIANS
+        else if (coordSys==ESMF_COORDSYS_SPH_RAD) then
+           center_lons=lons
+           center_lats=lats
+        else if (coordSys==ESMF_COORDSYS_CART) then
+           _RETURN(_FAILURE)
```
tclune, line 3364: Should be
```
_FAIL('Unsupported coordinate system:  ESMF_COORDSYS_CART')
```

tclune (changes requested):
A few minor changes requested.

tclune (approved):

Issue #743, opened 2021-03-08
Ability to run MAPL with two phases


WilliamJamieson:
@rmontuoro has requested the ability to run both run phases of GOCART independently in the UFS. This is so that the UFS can better emulate how GOCART is run within GEOS. This will require changes to MAPL so that the run method for the MAPL_CapGridComp supports the optional ability to run multiple phases.

Currently, the UFS does not support running multiple phases; however, there is a push on the NOAA side to get the UFS ready to run GOCART in multiple phases. Thus we need to ready MAPL (and GOCART) to support this feature.

weiyuan-jiang:
It seems to me, in GEOS_PhysicsGridCompMod, it already runs two phases of GEOS_ChemGridComp, which runs two phases of GOCART.  What is the eventual order to run the phases of GOCART?

WilliamJamieson:
GOCART in the UFS is run independently of the GEOS_PhysicsGridComp and the GEOS_ChemGridComp (neither of these is used at all).

tclune:
NOAA is not using Physics, nor using Chemistry.       So they cannot drive it indirectly the way it currently happens in GEOS.

atrayano:
@weiyuan-jiang MAPL_CapGridComp should simply run whatever phase the caller wants. The order is determined by the caller, not by MAPL_Cap. Since you asked, the GEOS Physics calls phase 1 of Chem between calls to Surf, and Turb (also phase 1), and then calls phase 2 of Surf, Turb, Chem. But again, this is how GEOS calls Chem. NOAA will decide how they will call it. We just need to make it possible.

PR #749, opened 2021-03-11
support for multi run phases in root grid comp

weiyuan-jiang:
Add support to run multi phases of root grid comp. Address #743 
## Types of changes

## Checklist:

### DIFF
--- a/gridcomps/Cap/MAPL_CapGridComp.F90
+++ b/gridcomps/Cap/MAPL_CapGridComp.F90
@@ -58,7 +58,7 @@ module MAPL_CapGridCompMod
      type(ESMF_Time),  allocatable :: AlarmRingTime(:)
      logical,          allocatable :: ringingState(:)
      logical :: compute_throughput
-     integer :: total_phases
+     integer :: n_run_phases
    contains
      procedure :: set_services
      procedure :: initialize
@@ -93,14 +93,14 @@ module MAPL_CapGridCompMod
 contains
 
   
-   subroutine MAPL_CapGridCompCreate(cap, root_set_services, cap_rc, name, final_file, unusable, total_phases, rc)
+   subroutine MAPL_CapGridCompCreate(cap, root_set_services, cap_rc, name, final_file, unusable, n_run_phases, rc)
       use mapl_StubComponent
     type(MAPL_CapGridComp), intent(out), target :: cap
     procedure() :: root_set_services
     character(*), intent(in) :: cap_rc, name
     character(len=*), optional, intent(in) :: final_file
     class(KeywordEnforcer), optional, intent(in) :: unusable
-    integer, optional, intent(in)  :: total_phases
+    integer, optional, intent(in)  :: n_run_phases
     integer, optional, intent(out) :: rc
 
     type(MAPL_CapGridComp_Wrapper) :: cap_wrapper
@@ -116,8 +116,8 @@ contains
     if (present(final_file)) then
        allocate(cap%final_file, source=final_file)
     end if
-    cap%total_phases = 1
-    if (present(total_phases)) cap%total_phases = total_phases
+    cap%n_run_phases = 1
+    if (present(n_run_phases)) cap%n_run_phases = n_run_phases
 
     cap%config = ESMF_ConfigCreate(rc=status)
     _VERIFY(status)
@@ -196,7 +196,7 @@ contains
     logical                      :: cap_clock_is_present
 
 
-    type (MAPL_MetaComp), pointer :: maplobj
+    type (MAPL_MetaComp), pointer :: maplobj, root_obj
     procedure(), pointer :: root_set_services
     type(MAPL_CapGridComp), pointer :: cap
     class(BaseProfiler), pointer :: t_p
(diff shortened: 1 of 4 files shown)


gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -1103,146 +1128,193 @@ subroutine run_MAPL_GridComp(gc, rc)
...
-
-    ! Call Record for intermediate checkpoint (if desired)
-    ! ------------------------------------------------------
+      if (this%compute_throughput) then
+         if (.not.this%started_loop_timer) then
+            this%loop_start_timer = MPI_WTime(status)
+            this%started_loop_timer=.true.
+         end if
+         start_timer = MPI_Wtime(status)
+      end if
```
tclune, line 1170: This should move into the first_phase() procedure.

gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -1103,146 +1128,193 @@ subroutine run_MAPL_GridComp(gc, rc)
...
+
+       call last_phase(rc=status)
+       _VERIFY(STATUS)
+
+       ! Estimate throughput times
+       ! ---------------------------
+       if (this%compute_throughput) then
+          call print_throughput(rc=status)
+          _VERIFY(STATUS)
+       end if
```
tclune, line 1208: All of this should be in last phase.

tclune (changes requested):
Just a couple of straightforward changes for quality as explained in the individual comments.

gridcomps/Cap/MAPL_CapGridComp.F90
weiyuan-jiang: changed

WilliamJamieson:
I would like to bring @rmontuoro into this conversation.

gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -1103,146 +1128,192 @@ subroutine run_MAPL_GridComp(gc, rc)
...
-      integer           :: HRS_R, MIN_R, SEC_R
-      real(kind=REAL64) :: START_RUN_TIMER,END_RUN_TIMER,START_TIMER,END_TIMER
-      real(kind=REAL64) :: TIME_REMAINING
-      real(kind=REAL64) ::  LOOP_THROUGHPUT=0.0_REAL64
-      real(kind=REAL64) ::  INST_THROUGHPUT=0.0_REAL64
-      real(kind=REAL64) ::   RUN_THROUGHPUT=0.0_REAL64
-      real              :: mem_total, mem_commit, mem_committed_percent
-      real              :: mem_used, mem_used_percent
+    integer           :: HRS_R, MIN_R, SEC_R
+    real(kind=REAL64), save :: START_RUN_TIMER,END_RUN_TIMER,START_TIMER,END_TIMER
```
tclune, line 1141: Sorry - one more change.     We don't want SAVE here - just to dangerous.  Create a small derived type for these items and then add a component of that type into the CapGridComp type.

atrayano: I agree with Tom, 'save' could have implications (especially if we are going to run multi-threaded)

weiyuan-jiang: Good point.  Changed

gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -1103,146 +1134,192 @@ subroutine run_MAPL_GridComp(gc, rc)
...
-      real(kind=REAL64) ::  INST_THROUGHPUT=0.0_REAL64
-      real(kind=REAL64) ::   RUN_THROUGHPUT=0.0_REAL64
-      real              :: mem_total, mem_commit, mem_committed_percent
-      real              :: mem_used, mem_used_percent
+    integer           :: HRS_R, MIN_R, SEC_R
+    real(kind=REAL64) :: END_RUN_TIMER, END_TIMER
+    real(kind=REAL64) :: TIME_REMAINING
+    real(kind=REAL64) ::  LOOP_THROUGHPUT=0.0_REAL64
+    real(kind=REAL64) ::  INST_THROUGHPUT=0.0_REAL64
+    real(kind=REAL64) ::   RUN_THROUGHPUT=0.0_REAL64
```
tclune, line 1151: These variables also have the SAVE attribute.  (Just more subtle because of the implied initialization.)  Note that on the 2nd trip through, their value will _not_ be 0.

tclune: They should also be relocated into the last_phase() procedure.

gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -38,6 +38,12 @@ module MAPL_CapGridCompMod
 
   public :: MAPL_CapGridComp, MAPL_CapGridCompCreate, MAPL_CapGridComp_Wrapper
 
+  type :: Throughput_Timer
+     real(kind=real64) :: loop_start_timer
+     real(kind=REAL64) :: start_run_timer
+     real(kind=REAL64) :: start_timer
+  end type
+
```
tclune, line 46: Since I'm asking for additional changes anyway, I'm going to request a style change here:

Use CamelCase for the type name: `ThroughputTimers`   (note the rare plural for a type)
I was also going to suggest new names for the components inside, but I did not really come up with anything I liked.    I suspect at some point we should just use  ordinary timers from profiler.  (Won't help with the names though.)

gridcomps/Cap/MAPL_CapGridComp.F90
weiyuan-jiang: Actually print_throughput() is a better place

gridcomps/Cap/MAPL_CapGridComp.F90
weiyuan-jiang: changed

tclune (approved):

PR #751, opened 2021-03-12
Gitflow

tclune:
## Types of changes

## Checklist:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,6 +6,10 @@
 # The MAPL Team owns all the files
 * @GEOS-ESM/mapl-team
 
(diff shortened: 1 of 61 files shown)


atrayano (approved):
I agree that this file does not belong here. If we need to keep it, a good clean up would be desirable - some values are hardcoded (e.g 140, 340, etc). I would suggest we move it to GEOS_Shared (again, I am not sure if we need it at all)

mathomp4 (approved):
Indeed, the whole SatVapor code should be moved somewhere other than Base. it's Physics!

mathomp4 (approved):
Re-approve

Issue #673, opened 2021-01-20
Variable splitting in multiple History collections


atrayano:
This is a bug, that can be reproduced when we try to split the same variable in more that 1 collection. The issue is that result of the splitting is a new field, and currently we put it in the same export state as the original field. This works as long as the same variable is not split again in another collection. The second time we do this, ESMF will return an error that the export state already contains the split field

mathomp4:
@atrayano I believe this is fixed, yes?

atrayano:
@mathomp4 this is definitely fixed but I am not 100% if it got merged into develop. I will check and get back to you.

atrayano:
@mathomp4 I just checked. It is not merged to develop. I have two branches that I passed to Elliot for testing (and 1 of them addresses #673 and 1 for #674). But Elliot needed one more thing from me to fully test (overload for MAPL_ConfigSetAttribute for array of integers and reals). This is all done, but Tom suggested a freeze MAPL until we deliver a stable version to NOAA, so I am putting my PR's on hold until then

atrayano:
This issue was addressed with PR#758

PR #758, opened 2021-03-17
Fixes #756. The bug was related to putting the split fields in the sa…

### LINKED ISSUE #756: Bug with splitField and identical fields in different collections
User @rmontuoro reported an issue when testing the UFS that if he tried to write two distinctly named collections with identical fields (one is time averaged, other is not) like so:
```
  inst_3d_aer.format:     'CFIO' ,
  inst_3d_aer.descr:      '3d,3-Hourly,Instantaneous,Model-Level,Analysis,Aerosol Concentrations',
  inst_3d_aer.template:   '%y4%m2%d2_%h2%n2z.nc4',
  inst_3d_aer.mode:       'instantaneous',
  inst_3d_aer.grid_label:  PC720x361-DC ,
  inst_3d_aer.splitField:  1,
  inst_3d_aer.frequency:   010000 ,
  inst_3d_aer.duration:    010000 ,
  inst_3d_aer.ref_time:    000000 ,
  inst_3d_aer.nbits:      10,
  inst_3d_aer.fields:   'DU'           , 'DU'   ,
                        'SS'           , 'SS'   ,
                      ::

  tavg_3d_aer.format:     'CFIO' ,
  tavg_3d_aer.template:   '%y4%m2%d2_%h2%n2z.nc4',
  tavg_3d_aer.mode:       'time-averaged',
  tavg_3d_aer.grid_label:  PC720x361-DC ,
  tavg_3d_aer.splitField:  1,
  tavg_3d_aer.frequency:   020000 ,
  tavg_3d_aer.duration:    020000 ,
  tavg_3d_aer.ref_time:    000000 ,
  tavg_3d_aer.nbits:      10,
  tavg_3d_aer.fields:   'DU'           , 'DU'   ,
                        'SS'           , 'SS'   ,
                      ::
```
This looks like an issue with the split option. I was able to reproduce this in a hack of my ExtDataDriver and see this:
```
pe=00000 FAIL at line=03051    MAPL_Base.F90                            <status=13>
pe=00000 FAIL at line=02867    MAPL_HistoryGridComp.F90                 <status=13>
pe=00000 FAIL at line=01445    MAPL_HistoryGridComp.F90                 <status=13>
pe=00000 FAIL at line=01839    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=00379    ExtDataDriverGridComp.F90                <status=1>
pe=00000 FAIL at line=00535    ExtDataDriverGridComp.F90                <status=1>
pe=00000 FAIL at line=00128    ExtDataDriverMod.F90                     <status=1>
pe=00000 FAIL at line=00024    ExtDataDriver.F90                        <status=1>
```
From what I can tell, it looks like the split option adds the split fields back to the export state of the component when first initializing history. It looks like if you have two collections with the split and the same fields it fails with a trying to add duplicate field error. So the logic here probably needs to check if we have already added the split field for the original in that component.

atrayano:
…me export state as the original field. When more than 1 collection is splitting the same field, this resulted in duplicate items in the export state

## Description
we now check if the split field is already in the export state. If "yes" we do not put it again

## Related Issue

#756, #673

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -2775,6 +2775,7 @@ ENDDO PARSER
       character(ESMF_MAXSTR) :: fldName, stateName
       logical :: split
       integer :: k, i
+      logical :: hasField
       
       ! Restrictions:
       ! 1) we do not split 4d vectors
@@ -2863,8 +2864,17 @@ ENDDO PARSER
                   call appendArray(newExpState,idx=list(n)%expState(k),rc=status)
                   _VERIFY(status)
 
-                  call MAPL_StateAdd(expState, field=splitFields(i), rc=status)
+                  ! ALT: this is ONLY a very simple test to make sure that this is not a duplicate
+                  ! this issue might be revisited to assure that possible duplicates have
+                  ! identical content. Otherwise the split fields should be put in its own container
+                  ! perhaps per collection, but this 
+                  hasField = .false. ! initialize just in case
+                  call checkIfStateHasField(expState, fieldName=fldName, hasField=hasField, rc=status)
                   _VERIFY(status)
+                  if (.not. hasField) then
+                     call MAPL_StateAdd(expState, field=splitFields(i), rc=status)
+                     _VERIFY(status)
+                  end if
 
                   item%itemType = ItemTypeScalar
                   item%xname = trim(fldName)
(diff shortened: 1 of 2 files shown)


gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -5170,5 +5180,41 @@ subroutine RecordRestart( gc, import, export, clock, rc )
     _RETURN(ESMF_SUCCESS)
   end subroutine RecordRestart
 
+  subroutine  checkIfStateHasField(state, fieldName, hasField, rc)
```
tclune, line 5183: No change requested,  but in the future consider that a LOGICAL function would have expressed the functionality more cleanly.

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -5170,5 +5180,41 @@ subroutine RecordRestart( gc, import, export, clock, rc )
...
+       if(itemTypeList(I)/=ESMF_STATEITEM_FIELD) cycle
+       if(itemNameList(I)==fieldName) then
+          hasField = .true.
+          exit
+       end if
+    end do
+    deallocate(itemNameList, stat=status)
+    _VERIFY(STATUS)
+    deallocate(itemTypeList, stat=status)
+    _VERIFY(status)
```
tclune, line 5214: Unnecessary.  Unsaved ALLOCATABLE's are automatically deallocated on exit.

Again - no change requested.   Just for the future.

tclune (approved):

rmontuoro:
This PR fixes the issue occurring with UFS-GOCART.

PR #768, opened 2021-03-22
Fixes #511. Use INT64 for SYSTEM_CLOCK

### LINKED ISSUE #511: negative throughput numbers reported
Several users have reported negative values for throughput from MAPL_Cap.

The timers are using `system_clock()` with default integers instead of 64 bit which greatly increases the odds that the clock will "roll over" during a run.      Should be a simple fix.

mathomp4:
## Description

Uses `INT64` for `system_clock()` calls in MAPL. This should prevent negative throughputs.

I also changed all the other places I saw `system_clock()` that could be in `libMAPL.a`. Note there are some programs and tests that are still `INT32` but I figure they don't quite need it.

## Related Issue

Closes #511 

## Motivation and Context

Can prevent rare negative throughputs, and other timers.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed error handling in profiler/BaseProfiler.F90
 - Fix memory leak when using fast_oserver in write_restart_by_oserver
 - Bumped cube version to 2.91 in global metadata
-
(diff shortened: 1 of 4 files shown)


mathomp4:
Note: I need to actually build and run this, so keeping this as draft at the moment.

mathomp4:
Tested with a run of GEOS with GNU Debug. Runs and is zero-diff.

tclune (approved):

PR #772, opened 2021-03-24
Updates to fix NAG build

mathomp4:
## Description

A set of CMake updates to let NAG Fortran build on my laptop. 

The main change is adding `FLAP::FLAP` to a couple`CMakeLists.txt` files where apparently NAG needs them. Not sure why GNU and Intel didn't.

Also, I turned off the NAG build of `ExtDataDriver.x` because it doesn't seem to work at the moment (ICE). 

Finally, cleaned up a lot of the "CMake + NAG + OpenMP" doesn't work comments. It seemed like it was saying CMake and Fortran never worked with OpenMP, but it's just CMake and NAG Fortran.

## Motivation and Context

Let's us use NAG to compile MAPL.

## How Has This Been Tested?

It builds and runs the MAPL tests. I'll test it out with a full GEOSgcm run soon. 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix memory leak when using fast_oserver in write_restart_by_oserver
 - Bumped cube version to 2.91 in global metadata
 - Change calls to `system_clock()` to be `INT64` (#511)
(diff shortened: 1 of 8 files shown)


mathomp4:
Tested with GEOSgcm. Seems to be zero-diff!

tclune (approved):

PR #803, opened 2021-04-12
Fixes #396 duplicate KeywordEnforcer module

### LINKED ISSUE #396: Keyword enforcer duplicated in two places
Making a note of this so I don't forget.

Both MAPL_Shared and GMAO_pFIO have an identical keyword enforcer. 
GMAO_pFIO has a file KeywordEnforcer.F90 with a module pFIO_KeywordEnforcerMod

MAPL_Shared has a file MAPL_KeywordEnforcer.F90 with module MAPL_KeywordEnforcerMod

Other than the names the code is identical.

This should be consolidated under shared. 
If we are following the MAPL conventions document the file should be 
KeywordEnforcer.F90
the module should be MAPL_KeywordEnforcer

tclune:
## Motivation and Context

Eliminated duplicate file.   Lowers general entropy.

## Types of changes

## Checklist:

I have not tested a run of the full model, but MAPL self tests pass.  (Low risk of model failure at this point.)


pfio/ClientManager.F90
```diff
@@ -5,7 +5,7 @@ module pFIO_ClientManagerMod
 
    use MAPL_ExceptionHandling
    use MAPL_SortMod
-   use pFIO_KeywordEnforcerMod
+   use mapl_KeywordEnforcerMod
```
mathomp4, line 8: Random comment: Should this be `MAPL_KeywordEnforcerMod`? We seem to use the `MAPL` in most use statements.

tclune: yes - how did I miss it?   Will never know.  Unfortunately I compiled before I deleted the old module, but I thought after I had removed it from CMake.

tclune: I'm adopting a policy of making the namespace prefix be lower case.   This generally improves legibility IMO, as the prefix contains the least important information.   This is esp. true when scanning a long list where the prefix dominates the quick scan.   OTOH, I did not feel like pushing a change through all the files to enforce this.  Just doing it as I go.

mathomp4 (approved):

Issue #790, opened 2021-04-06
Feature Request: TimestampStart


lizziel:
MAPL History names time-averaged files with mid-point datetime by default, and has the capability to use end datetime instead (https://github.com/GEOS-ESM/MAPL/issues/602). This feature request is for implementation of using start datetime in time-averaged History filenames. Like `TimestampEnd`, `TimestampStart` would be an optional setting.

This request is for use in GCHP. We are going to recommend usage of the monthly average for collections. Monthly output files are timestamped with start datetime in the filenames. Having a `TimestampStart` option for _all_ time-averaged output, not just monthly, would allow GCHP output files to have consistent format. An added benefit is they would then match the format of GEOS-Chem Classic.

atrayano:
Yes, it is related. I'll fix both

atrayano:
Actually the code is still doing what it supposed to do, but it is plain messy. I'll fix  it anyway, but the issue was a "red" herring

lizziel:
Do you mean TimestampStart was already an option in History?

atrayano:
@lizziel I made a mistake and my comment went to the wrong issue. But at the moment I am working on a branch that should address both issues (#790) the TimeStampStart and (#784) the incorrect time increment for monthly. I'll send you email once it is ready

lizziel:
Thanks @atrayano for implementing this so quickly. I manually added the feature to GCHP (uses MAPL v2.6.3) and it all looks good.

PR #804, opened 2021-04-13
Fixes #784, #790, #802. Fixed several issues regarding timestamping o…

### LINKED ISSUE #784: Incorrect attribute 'time_increment' in History monthly collection files
I ran a 1-year GCHP simulation using MAPL 2.6.3 with the `monthly` collection option turned on in `HISTORY.rc`. The time variable has attribute `time_increment` which I would expect to be the duration of the month per file. Instead it is 7440000 for each month. This corresponds to the `duration` setting otherwise ignored when `monthly` is enabled.

For example, the time information for June is this:
```
        float time(time) ;
                time:begin_date = 20180601 ;
                time:begin_time = 0 ;
                time:long_name = "time" ;
                time:time_increment = 7440000 ;
                time:units = "minutes since 2018-06-01 00:00:00" ;
```
 This is a problem since no information other than `time_increment` is present to infer the file is a monthly average. The way it is now someone may think it is the average over 31 days starting on June 1 which is incorrect.

### LINKED ISSUE #802: Option 'timestampEnd' incorrectly prevents time-averaging in History
If a time-averaged collection specifies 'timestampEnd' then the variable "average" is not set properly, and as a result, no averaging is performed.

atrayano:
…ffsets and incorrect time_increment attribute for monthly averaged output. Added new option "timestampStart".

## Related Issue

Closes #802, #790 and #784

## Motivation and Context

This fixes several issues regarding time stamping of averaged files, as well as incorrect time increment metadata for a monthly averaged collections
## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_TimeMethods.F90
+++ b/base/MAPL_TimeMethods.F90
@@ -63,7 +63,7 @@ contains
   end subroutine get
        
   subroutine setFrequency(this,frequency,rc)
-     class(TimeData) :: this
+     class(TimeData), intent(inout) :: this
      integer, intent(in) :: frequency
      integer, optional, intent(out) :: rc
 
--- a/base/MAPL_TimeMethods.F90
+++ b/base/MAPL_TimeMethods.F90
@@ -26,6 +26,7 @@ module MAPL_TimeDataMod
      procedure :: compute_time_vector
      procedure :: get_start_time
      procedure :: get
+     procedure :: setFrequency
   end type timeData
 
   interface timeData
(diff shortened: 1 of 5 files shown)


base/MAPL_TimeMethods.F90
```diff
@@ -61,6 +62,16 @@ subroutine get(this,clock,rc)
      _RETURN(_SUCCESS)
   end subroutine get
        
+  subroutine setFrequency(this,frequency,rc)
+     class(TimeData) :: this
```
tclune, line 66: missing `INTENT` attribute.

tclune (changes requested):
One minor fix requested.

atrayano:
Thanks for catching! Done. Added the missing INTENT

tclune (approved):

Issue #787, opened 2021-04-05
Create a stub pFlogger library and an option to use it


tclune:
NOAA wishes to avoid a dependency on pFlogger.    The idea is to instead have a subdirectory of MAPL that builds a stub version of pFlogger and then also have a CMake switch that controls whether to add that subdirectory.

The following types must be defined: `Logger`, `LoggerManager`

`Logger` needs these methods:
- debug()
- info()
- warn()
- error()
- critical()
The hard part with these is that they have up to 9 optional arguments of class(*).

`LoggerManager` needs just `get_logger()`.  

Initialization uses additional entities from pFlogger, but for starters those can be handled with a #ifdef.

On the CMake side, a new OPTION is needed:  `USE_PFLOGGER` which defaults to ON.   Then the cmake in MAPL/base should add a compile definition for ApplicatioSupport.F90  `-DUSE_PFLOGGER` which will be used to mask out the initialization logic.

I'm sure I'm forgetting things, so just update the ticket as issues come up.

tclune:
Adding @rmontuoro for situational awareness.

PR #805, opened 2021-04-14
add pflogger stub to skip pflogger library

weiyuan-jiang:
add pflogger stub to skip pflogger library.  If cmake command line is with -DBUILD_WITH_PFLOGGER=OFF, a stub is created to replace the pflogger
## Description

add pflogger_stub subdirectory
## Related Issue

This PR resolves issue #787 

## Types of changes

## Checklist:


weiyuan-jiang:
I added DNA label because two more changes are needed : 1) component.yaml for new ESMA_cmake and 2) CHANGELOG.md

weiyuan-jiang:
CHANGELOG.md updated. Now it needs components.yaml.

mathomp4:
> CHANGELOG.md updated. Now it needs components.yaml.

I've released ESMA_cmake v3.3.9. and pushed an update. I'm still a bit concerned that the GitHub Actions failed...

mathomp4:
Oy. Okay. The GEOSgcm build will fail until https://github.com/GEOS-ESM/GEOSgcm/pull/281 is pulled into GEOSgcm because it needs ESMA_cmake v3.3.9

mathomp4:
@tclune @weiyuan-jiang Now I'm a bit interested in this. Why would this fail? Okay, without ESMA_cmake v3.3.9, then the GEOSgcm build would not see the `BUILD_WITH_PFLOGGER` as `ON` and so default to `OFF`. But doesn't this mean that the stub would take over?

tclune (dismissed):

mathomp4:
```
CMake Error at @cmake/@ecbuild/cmake/ecbuild_add_library.cmake:268 (add_library):
  add_library cannot create target "pflogger" because an imported target with
  the same name already exists.
Call Stack (most recent call first):
  @cmake/@ecbuild/cmake/ecbuild_add_library.cmake:576 (ecbuild_add_library_impl)
  @cmake/esma_add_library.cmake:93 (ecbuild_add_library)
  src/Shared/@MAPL/pflogger_stub/CMakeLists.txt:12 (esma_add_library)
```
@tclune is this some Baselibs thing? pFlogger is "magically" in the CMake chain because Baselibs 6.1.0 doesn't have the latest version of some library which is implicitly bringing in the pflogger target?

tclune:
Ugh.  Good point.   This is the advantage of using the opposite sentiment.   I think there might be a way to check to see if an option exists at all, and maybe we should add that to the other pflogger PR's.   I.e.,  if the option exists _and_ it is OFF then use the stub.

tclune:
But yes - the stub should have taken over, which should not have failed presuming the stub has all the necessary interfaces.   I suspect @weiyuan-jiang and myself may have missed a few.  E.g., there is a function that returns the severity level of a logger so that you can skip entire blocks of logging.    (That should always return something very high - CRITICAL in the stub.)

tclune:
This happens when `find_package()` is finding the baselibs package despite the intended guards.  So either the option was not set early enough, or it is not being used in advance of that find_package.

weiyuan-jiang:
@tclune I actually added function isEnableFor() as always .false.  The issue here is not the stub, it is probably about cmake. I guess we need to move the option BUILD_WITH_PFLOGGER way up

weiyuan-jiang:
Is that  possible that cmake is too smart so it detects pflogger automatically in the baselib?

tclune:
No that does not happen.   It can be too smart about finding it in the wrong directory,  but it will only "find" it if an explicit `find_package()` is used.

mathomp4:
Okay. Scott has merged. Let me make a trivial edit to force new builds.

mathomp4:
Okay. Trivial commit in and it looks like the GCM CMake is happier thanks to @sdrabenh. 

The next issue is that some of the "act like pflogger" bits disappeared. I'm looking at how:
```cmake
add_library (${this} STATIC ${SRCS})
add_library(PFLOGGER::pflogger ALIAS pflogger)

set (dest "PFLOGGER-${PFLOGGER_VERSION_MAJOR}.${PFLOGGER_VERSION_MINOR}")

target_include_directories(${this} PUBLIC
  $<BUILD_INTERFACE:${PFLOGGER_BINARY_DIR}/include>
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
  $<INSTALL_INTERFACE:${dest}/include>
  )

target_link_libraries(${this} PUBLIC GFTL_SHARED::gftl-shared GFTL::gftl)

install (TARGETS pflogger EXPORT PFLOGGER DESTINATION ${dest}/lib)
install (DIRECTORY  ${PROJECT_BINARY_DIR}/include/ DESTINATION ${dest}/include)
install (FILES  ${PROJECT_SOURCE_DIR}/src/error_handling_macros.fh DESTINATION ${dest}/include)
```
became:
```cmake
if (ESMA_USE_GFE_NAMESPACE)
   esma_add_library (${this} SRCS ${srcs} DEPENDENCIES GFTL_SHARED::gftl-shared TYPE SHARED)
else ()
   esma_add_library (${this} SRCS ${srcs} DEPENDENCIES gftl-shared TYPE SHARED)
endif ()

target_include_directories (${this} PUBLIC $<BUILD_INTERFACE:${MAPL_SOURCE_DIR}/include>)
```

I think we might need some combination?

And should the pflogger stub be a `TYPE SHARED` library?

weiyuan-jiang:
I think in the baselib, pflogger should be type SHARED, or at least we should have both STATIC and SHARED

mathomp4:
I removed the Contingent flag as this seems to be good...though for some stupid reason that triggered new CI? Ugh. Time to stare at the CI yaml...

mathomp4:
Passed. I'm re-requesting the @tclune review. I think we are close to new MAPL release!

tclune (approved):

PR #810, opened 2021-04-16
Fixes #809. Added explicit initialization after allocation

### LINKED ISSUE #809: Possible uninitialized members of the derived type Segment_T in MAPL_Shmem
Since shmid was statically initialized to -1 in the declaration of the type Segment_T, no further initialization was done after allocation. But this code generated strange shmem related error on the Rome nodes at NAS. Fixing them helped. Might be a compiler bug

atrayano:
## Description

Initialize the members of Segment_T after allocation

## Related Issue

#809

## Types of changes

## Checklist:

- [x I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

### DIFF
--- a/shared/MAPL_ShmemMod.F90
+++ b/shared/MAPL_ShmemMod.F90
@@ -55,7 +55,7 @@
 
     type Segment_T
        integer (c_int) :: shmid=-1
-       type    (c_ptr) :: addr
+       type    (c_ptr) :: addr=C_NULL_PTR
     end type Segment_T
 
     type(Segment_T), pointer :: Segs(:) => NULL()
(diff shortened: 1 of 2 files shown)


shared/MAPL_ShmemMod.F90
```diff
@@ -198,6 +198,8 @@ subroutine MAPL_InitializeShmem(rc)
 
       allocate(Segs(CHUNK),stat=STATUS)
       _ASSERT(STATUS==0,'needs informative message')
+      Segs(:)%shmid = -1
+      Segs(:)%addr=C_NULL_PTR
```
tclune, line 202: Not worth the change at this point, but this could be more clearly implemented as:
```
segs(:) = SegmentT(-1_C_INT, C_NULL_PTR)
```
(esp not worth it because of how ugly the "-1" must be declared.)

tclune (approved):
I'm still fairly confident that this is merely masking some other problem.  Many codes would break if the automatic initialization were not happening without these additional changes.

But I'm going to double check the rules for POINTER to be safe.

PR #744, opened 2021-03-08
add --with_io_profiler option

weiyuan-jiang:
Add an option whether or not turn on io_profiler
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [ ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [ ] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [ ] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #811, opened 2021-04-19
clean up comms

weiyuan-jiang:
1) add function to free communicators that is spit by SimpleCommSplitter
2) remove some unnecessary mpi_comm_dup calls. Some of those call are actually bugs
  
## Types of changes

## Checklist:

### DIFF
--- a/base/ApplicationSupport.F90
+++ b/base/ApplicationSupport.F90
@@ -31,8 +31,7 @@ module MAPL_ApplicationSupport
          logging_configuration_file=''
       end if
       if (present(comm)) then
-         call MPI_comm_dup(comm,comm_world,status)
-         _VERIFY(status)
+         comm_world = comm
       else
          comm_world=MPI_COMM_WORLD
       end if
@@ -56,8 +55,7 @@ module MAPL_ApplicationSupport
       _UNUSED_DUMMY(unusable)
       
       if (present(comm)) then
-         call MPI_comm_dup(comm,comm_world,status)
-         _VERIFY(status)
+         comm_world = comm
       else
          comm_world=MPI_COMM_WORLD
       end if
@@ -100,8 +98,7 @@ module MAPL_ApplicationSupport
          logging_configuration_file=''
       end if
       if (present(comm)) then
-         call MPI_Comm_dup(comm,comm_world,status)
-         _VERIFY(status)
+         comm_world = comm
       else
          comm_world=MPI_COMM_WORLD
       end if
@@ -153,10 +150,9 @@ module MAPL_ApplicationSupport
 
       _UNUSED_DUMMY(unusable)
       if (present(comm)) then
-         call MPI_Comm_dup(comm,world_comm,status)
-         _VERIFY(status)
+         world_comm = comm
       else
-         world_comm=MPI_COMM_WORLD
+         world_comm = MPI_COMM_WORLD
       end if
       t_p => get_global_time_profiler()
       t_p = TimeProfiler('All', comm_world = world_comm)
@@ -185,8 +181,7 @@ module MAPL_ApplicationSupport
 
       _UNUSED_DUMMY(unusable)
       if (present(comm)) then
-         call MPI_comm_dup(comm,world_comm,ierror)
-         _VERIFY(ierror)
+         world_comm = comm
       else
          world_comm=MPI_COMM_WORLD
       end if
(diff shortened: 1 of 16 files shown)


base/ApplicationSupport.F90
```diff
@@ -31,8 +31,7 @@ subroutine MAPL_Initialize(unusable,comm,logging_config,rc)
          logging_configuration_file=''
       end if
       if (present(comm)) then
-         call MPI_comm_dup(comm,comm_world,status)
-         _VERIFY(status)
+         comm_world = comm
```
tclune, line 34: How do we know it is safe not to dup the communicator here?

base/ApplicationSupport.F90
```diff
@@ -56,8 +55,7 @@ subroutine MAPL_Finalize(unusable,comm,rc)
       _UNUSED_DUMMY(unusable)
       
       if (present(comm)) then
-         call MPI_comm_dup(comm,comm_world,status)
-         _VERIFY(status)
+         comm_world = comm
```
tclune, line 58: Why does Finalize need to dup a communicator at all?

tclune (commented):
It looks like you decided against using a FINAL method?

base/ApplicationSupport.F90
weiyuan-jiang: It should be safe if  line 36 comm_world=MPI_COMM_WORLD is safe

base/ApplicationSupport.F90
weiyuan-jiang: I believe it is a bug to duplicate here

weiyuan-jiang:
Yes. There is a SimpleCommSplitter in MAPL_Cap.  If we keep the FINAL in SimpleCommSplitter, , then we should add FINAL in MAPL_Cap. In the FINAL of MAPL_Cap, it has to call MPI_Finalize , which is not explicit enough for me.

base/ApplicationSupport.F90
tclune: I disagree with the reasoning.   In the MPI_COMM_WORLD case, GEOS is not being embedded inside of someone else's application.   (Or if it is, they've already not done anything to insulate themselves.)      Some of GEOS comms in MAPL use TAGs that could potentially conflict with TAGs of the external application.   Hard to imagine any problems in practice, but that is the main reason for DUP in the first place.     Usually more of a concern when going back and forth between two layers.

I suppose we could accept and just  reverse when an issue is found later.

tclune:
I'm not understanding.   FINAL would only be needed for the type that encapsulates the comm.   The other containing types would not need  a FINAL method.

base/ApplicationSupport.F90
weiyuan-jiang: Even we don't duplicate here, it is still a bug because the original comm was duplicated. Remove those mpi_comm_dups may be the best bet.

weiyuan-jiang:
When is the FINAL of SimpleCommSplitter called? I believe it is called when MAPL_Cap is out of scope or right before it is destroyed. But at that time, MPI_Finalize is already called and the free_sub_comm() is either buggy or meaningless.

tclune:
Ideally it should go out of scope before MPI_Finalize is called.  After all, MPI_Finalize() does not need it.  But I agree that we don't want a major refactoring just to accomplish this.

mathomp4:
@weiyuan-jiang I pushed the "Update branch" button, so please pull if you are developing.

mathomp4:
@weiyuan-jiang Also: did you test this on Orion (I guess along with #744 ) ?

weiyuan-jiang:
Not yet.

tclune (approved):

weiyuan-jiang:
@tclune I think it is quite awkward here : https://github.com/GEOS-ESM/MAPL/blob/0ad9d599a972aeb81e2628045ebbd7200780dda4/shared/SimpleCommSplitter.F90#L328-L331  .

tclune:
I cannot follow the precise rationale for the check, so I think more explanation for your concern is warranted.

mathomp4:
I did a run with GEOSgcm and this seems zero-diff.

tclune:
@weiyuan-jiang Please elaborate on your concerns about the code here:  

> @tclune I think it is quite awkward here :
> 
> https://github.com/GEOS-ESM/MAPL/blob/0ad9d599a972aeb81e2628045ebbd7200780dda4/shared/SimpleCommSplitter.F90#L328-L331
> 
> .

weiyuan-jiang:
@tclune , if  a splitter is copied after slpit() is called, then both splitters can call free_sub_comm(). But there is only one comm to be freed.  It should be the users' responsibility to free only once just like allocated pointer.  So I print a "WARNING" message.

tclune (approved):

PR #818, opened 2021-04-29
Use ESMF_Finalize in Cap

### LINKED ISSUE #817: Use ESMF_Finalize instead of MPI_Finalize
Per discussions with @rsdunlapiv and @theurich, it was determined that to use ESMF built-in profiling with GEOS, one needs to use `ESMF_Finalize` and not `MPI_Finalize` to end GEOS. This is a change in Cap.

mathomp4:
## Description

This PR moves to use `ESMF_Finalize` instead of `MPI_Finalize`. This comes from the want of the ESMF folks (@rsdunlapiv, @theurich) to use ESMF Profiling. Turns out, that doesn't happen if you don't have 

## Related Issue

Closes #817 

## Motivation and Context

Allows the use of the ESMF profiler.

## How Has This Been Tested?

I've tested using `ESMF_Finalize` and the model doesn't seem to crash GEOSgcm. 

That said, we should ask our friends at Harvard (@lizziel, et al) and UFS (@rmontuoro or @weiyuan-jiang as a proxy) to try this out in their codes. That is, if they call the Cap? I do not know if they do. If they don't, then this is moot for them.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 ### Added
 ### Changed
+
+- Use `ESMF_Finalize` instead of `MPI_Finalize` in Cap
+
 ### Fixed
 
 ## [2.6.5] - 2021-04-28
--- a/gridcomps/Cap/MAPL_Cap.F90
+++ b/gridcomps/Cap/MAPL_Cap.F90
@@ -412,13 +412,13 @@ contains
       class (KeywordEnforcer), optional, intent(in) :: unusable
       integer, optional, intent(out) :: rc
 
-      integer :: ierror
+      integer :: status
       _UNUSED_DUMMY(unusable)
 
       if (.not. this%mpi_already_initialized) then
          call MAPL_Finalize(comm=this%comm_world)
-         call MPI_Finalize(ierror)
-         _VERIFY(ierror)
+         call ESMF_Finalize(rc=status)
+         _VERIFY(status)
       end if
 
       _RETURN(_SUCCESS)


mathomp4:
I've added a blocking tag just until we know if this is happy with how UFS and Harvard run (@tclune might be able to tell us immediately).

atrayano (dismissed):
Looks good!

lizziel:
GCHP calls `cap%run` which includes the call to `mpi_finalize` so I think theanswer is yes. Will this update be be backwards compatible with ESMF 8.0?

tclune (changes requested):
I thought we were going to add a run-time switch (if possible) just to be safe.   If @atrayano is correct, it might cause problems, and it would be better not to have to patch again.

weiyuan-jiang:
I am trying to run ufs system in debugging mode. The error shifted. But I can try this branch without debugging flag first


 88: forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable DTDT_M when it is not allocated
 88:
 88: Image              PC                Routine            Line        Source
 88: ufs_model          000000000A6A6BCF  Unknown               Unknown  Unknown
 88: ufs_model          0000000002917902  fv_dynamics_mod_m         762  fv_dynamics.F90
 88: ufs_model          00000000026B8A06  atmosphere_mod_mp        1746  atmosphere.F90
 88: ufs_model          00000000025E3EEA  atmosphere_mod_mp         501  atmosphere.F90
 88: ufs_model          000000000240BD5D  atmos_model_mod_m         490  atmos_model.F90
 88: ufs_model          00000000023D5609  module_fcst_grid_         380  module_fcst_grid_comp.F90

weiyuan-jiang:
Still print out the error message in the end

tclune (approved):

mathomp4:
> GCHP calls `cap%run` which includes the call to `mpi_finalize` so I think the answer is yes. Will this update be backwards compatible with ESMF 8.0?

@lizziel That should be safe as `ESMF_Finalize()` has always been in ESMF

mathomp4:
Well, it looks like this should be safe? @weiyuan-jiang says it ran on Orion and from what @lizziel said, I think GCHP should be safe. Shall we merge, @tclune ?

mathomp4:
Or should I build in a run-time flag?

lizziel:
Would it be advantageous for GCHP to use ESMF profiling as well? Do you have an example of what it looks like?

mathomp4:
> Would it be advantageous for GCHP to use ESMF profiling as well? Do you have an example of what it looks like?

@lizziel I don't have much experience with it as, well, without this change, it doesn't work! But you can see some examples here:
https://earthsystemmodeling.org/docs/nightly/develop/ESMF_refdoc/node6.html#SECTION060130000000000000000

The text output is roughly like the new MAPL profiler output in ways. It can also output in binary so that [Cupid](https://cupid.readthedocs.io/en/latest/index.html) can show traces that look a bit like Ninja traces you can view in Chrome. I think?

atrayano (approved):
Looks OK

tclune (approved):

PR #828, opened 2021-05-06
Update CircleCI to use xlarge

mathomp4:
This updates the CI to use the newly re-enabled xlarge resource class on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,8 +11,8 @@ executors:
       OMPI_ALLOW_RUN_AS_ROOT: 1
       OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
       OMPI_MCA_btl_vader_single_copy_mechanism: none
-    #XLARGE# resource_class: xlarge
-    resource_class: medium
+    resource_class: xlarge
+    #MEDIUM# resource_class: medium
 
 workflows:
   version: 2.1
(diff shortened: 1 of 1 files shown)


mathomp4:
Let's get some timings.

With `medium` resource:

- build-and-test-MAPL: 11m 40s
- build-GEOSgcm: 45m 41s
- make-FV3-experiment: 51s
- run-FV3-standalone:  2m 6s

With `xlarge` resource:

- build-and-test-MAPL: 
- build-GEOSgcm: 
- make-FV3-experiment: 
- run-FV3-standalone:

tclune (approved):

tclune (approved):

tclune:
BTW, this looks like it goes _from_ XLARGE _to_ MEDIUM.  Not the other way around.

mathomp4:
> BTW, this looks like it goes _from_ XLARGE _to_ MEDIUM. Not the other way around.

Yeah. I know. The `#MEDIUM#` looks impressive, but I like the "labeled" comment so if we need to revert, that makes it easy.

mathomp4:
@tclune  So, it looks like we might want to focus our attention on the GNU build's tall poles. I think `MAPL_HistoryGridComp.F90` is still the big gun in the GNU Debug build. All that extra parallelism and the MAPL build only goes down by 1 minute (compared to GEOSgcm being 3x faster).

PR #834, opened 2021-05-10
Feature/bmauer/#833 getresource

### LINKED ISSUE #833: New overload for MAPL_GetResource that takes a plain ESMF_Config?
Bill was noting that some things that get read from the main AGCM.rc get echeod in the output as they use MAPL_GetResource which is useful to know what parameters you ran with. However, the surface components now read GEOS_SurfaceGridComp.rc which uses plain ESMF_ConfigGetAttribute calls so these don't get recorded to the output. This is apparently causing problems as there was some confusion what parameters were used. Bill was asking if we could do something about this.

Looking at MAPL_GetResource the easiest option would be to split it up. Right now it takes a MAPL object and pulls the config from there. So this only works with that config. We could split it up into two routines. A MAPL_GetResourceFromConfig that takes the config and does what MAPL_GetResource currently does with the config and MAPL_GetResourceFromMeta object that takes a MAPL object, pulls the config and call MAPL_GetResourceFromConfig.

This leads to a confusion as now we could have multiple configs echo their information so we don't know which file one came from as now we just echo the key and value. This is not known since the config does not store the file name. One thought we had was adding an optional argument for the filename that defaults to "ROOT_CF" or something like that.

bena-nasa:
Fixes #833 
This adds an overload to MAPL_GetResource to pass a config. This was requested so that for example when reading the GEOS_SurfaceGridComp.rc file the value read would be echoed the same way as when the configf from the mapl object is read.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added new CI test using Intel oneAPI
 - Add function to free communicators that is split by SimpleCommSplitter
 - Add with_io_profiler option
+- New interface to MAPL_GetResource to pass config rather than  MAPL object
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


atrayano (approved):
Looks OK

PR #831, opened 2021-05-07
Update CircleCI to use both gfortran and ifort

mathomp4:
This PR enables both gfortran and ifort builds of MAPL and GEOSgcm

Closes #830

PR #838, opened 2021-05-11
Use large resource at CircleCI

mathomp4:
As seen in #831, the `large` resource class at CircleCI is good enough for our testing. Move to use it.


atrayano (approved):
Looks OK

PR #840, opened 2021-05-12
Update components and CI in MAPL

mathomp4:
Let's fix up the CI and sync the components in MAPL to be the same as GEOSgcm before we release.


atrayano (approved):
Looks OK

PR #846, opened 2021-05-17
change FlapCapOptions to FlapCLI

weiyuan-jiang:
change FlapCapOptions to FlapCLI which is not a sub class of MAPL_CapOptions any more.
## Description

After this replacement, in GEOSgcm.F90, it would become something like 

command_line = MAPL_FlapCLI(....)
call command_line%parse_command_line_argumens(cap_options)
cap = MAPL_Cap(..,cap_options,..)
call cap%run()

## Types of changes

## Checklist:


weiyuan-jiang:
@tclune , we can start discussion from here

atrayano (changes requested):
The changes look OK to me, but you need to add details in CHANGELOG.md file

mathomp4:
So why is this needed? UFS?

weiyuan-jiang:
Yes, we are trying to solve ufs's problem. Also, I think it is more reasonable.  Mapl_capOptions can be constructed from flap or the orther way. They are not supposed to be parents-child relationship.   

Maybe we can have MAPL_CapOptions( flapCLI ) constructor

mathomp4:
Note if this goes in, it will need to have an *immediate* release and we'll want to work with @sdrabenh because we'd have to change:

* `@FVdycoreCubed_GridComp/StandAlone_DynAdvCore.F90`
* `@FVdycoreCubed_GridComp/StandAlone_FV3_Dycore.F90`
* `@FVdycoreCubed_GridComp/StandAlone_AdvCore.F90`
* `@GEOSgcm_App/GEOSgcm.F90`

If not, all my nightly tests that test off MAPL `develop` would fail is my guess.

tclune (dismissed):

weiyuan-jiang:
I removed DNA label and it is ready for review @tclune @mathomp4 . Other repositories I can think of are here
1)  https://github.com/GEOS-ESM/GEOSgcm_App/tree/feature/wjiang/flap_cli
2) https://github.com/GEOS-ESM/FVdycoreCubed_GridComp/tree/feature/wjiang/flap_cli

I will make PR there and refer back to this PR

base/MAPL_FlapCLI.F90
```diff
@@ -1,35 +1,31 @@
...
    contains
       procedure, nopass :: add_command_line_options
-      procedure :: parse_command_line_arguments
-      procedure :: set_esmf_logging_mode
-   end type MAPL_FlapCapOptions
+   end type MAPL_FlapCLI
 
-   interface MAPL_FlapCapOptions
-      module procedure new_FlapCapOptions
+   interface MAPL_FlapCLI
```
tclune, line 21: Is it possible to keep the original constructor names (and thus backward compatibility)?

tclune (changes requested):
I think it should be possible to keep CapOptions as a data struct with public components.  (Ideally with no methods - I have to look yet to see about the esmf setter.)   The files which had been subclasses will now just contain constructors for CapOptions objects.    File names can change, but the constructors should retain the name that they had when they were subclasses to avoid breaking backward compatibility.

weiyuan-jiang:
I can try to preserve the name, but eventually, the interface  has to change. For example, in GEOSgcm.F90,    **type (MAPL_FlapCapOptions) :: cap_options** must be changed to    **type (MAPL_CapOptions) :: cap_options**

   character(len=*), parameter :: Iam="GEOS5_Main"
   type (MAPL_Cap) :: cap
   **type (MAPL_FlapCapOptions) :: cap_options**
   integer :: status

   cap_options = MAPL_FlapCapOptions(description = 'GEOS AGCM', &
                                     authors     = 'GMAO')
   cap = MAPL_Cap('GCM', ROOT_SetServices, cap_options = cap_options)

tclune:
Well, arguably that declaration should have been
```f90
class(MAPL_CapOptions), allocatable :: cap_options
```
and there was just a quality-of-implementation bug for not having done so.   But even so we'd still be wanting to change `CLASS` to `TYPE` and skipping the `ALLOCATABLE`.  But they ought to continue to work.

We can't quite get to complete backwards compatible, but let's try to get close.   

A variant for that (which I'm not yet advocating) would be to use rename on the USE statement:
```f90
USE MAPL ::   MAPL_FlapCapOptions => MAPL_CapOptions
```
But then there are conflicts between the renamed type and the un-renamed constructor.   So lets not risk that problem.

base/MAPL_CapOptions.F90
```diff
@@ -80,5 +88,97 @@ function new_CapOptions(unusable, cap_rc_file, egress_file, ensemble_subdir_pref
 
    end function
 
+#ifdef USE_FLAP
```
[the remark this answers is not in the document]
weiyuan-jiang: I made a change and kept CapOptions as a data struct by moving set_esmf_looging_mode into the constructor. That makes sense because only Flap knows about flag_name.  I did have another implementation that moves  "constructors" to FlapCLI,  the GEOSgcm main would look like this
cli = MAPL_FlapCLI(.....)
cap_options = cli%get_mapl_cap_options(...)
cap = MAPL_Cap(...)

But I think this is not as good as the one now we have

base/MAPL_FlapCLI.F90
weiyuan-jiang: It is probably not worthy to keep the interface but make the codes messy

weiyuan-jiang:
@tclune , What else do you think should be improved for this PR?

tclune (dismissed):

tclune:
I've now aproved.  Waiting on @atrayano for a second review.

atrayano (dismissed):
Code looks OK to me. There is minor typo in the comments (line 160 in MAPL_CapOptions.F90, ouput => output

weiyuan-jiang:
@mathomp4 , it seems ready to merge. But may I leave this to you? It affects at least these two PRs,

    https://github.com/GEOS-ESM/GEOSgcm_App/tree/feature/wjiang/flap_cli
    https://github.com/GEOS-ESM/FVdycoreCubed_GridComp/tree/feature/wjiang/flap_cli

mathomp4:
@weiyuan-jiang I'll have to work with @sdrabenh on this as we'll have to make all the releases in one day (otherwise my nightly tests will blow up in some way)

mathomp4:
Also, @bena-nasa might need to look at all the ADAS programs (if they call Cap like `GEOSgcm.F90`) and make similar changes.

rmontuoro:
@weiyuan-jiang, @tclune - This PR fixes #862.

tclune (approved):

mathomp4:
@atrayano Can you also re-approve this? I updated the changelog and cmakelists to get it ready for a release, and GitHub dismissed the review

atrayano (commented):
Looks OK

Issue #854, opened 2021-05-19
ESMF_Initialize() and ESMF_Finalize() mismatch


weiyuan-jiang:
ESMF_Initialize() is called under the communicator model_comm, but the ESMF_Finalize() is called under world_comm. When GEOSgcm main branch is run with the command

$RUN_CMD 48 $GEOSBIN/GEOSgcm.x --npes_model 24 --oserver_type multigroup --nodes_output_server 1 --npes_backend_pernode 4 --logging_config 'logging.yaml'

the error messages are printed at the end.

ESMF_LogWrite(): ESMF_Log not open -- cannot ESMF_LogWrite(). Log message =
Finalizing ESMF
...
...
ESMF_Finalize: Error writing into the default log
ESMF_Finalize: Error flushing log file: Operation not yet supported
ESMF_Finalize: Error flushing log file: Operation not yet supported
ESMF_LogWrite(): ESMF_Log not open -- cannot ESMF_LogWrite(). Log message =
Finalizing ESMF
ESMF_LogWrite(): ESMF_Log not open -- cannot ESMF_LogWrite(). Log message =
Internal error: Bad condition - - Invalid GlobalVM
ESMF_Finalize: Error finalizing VM: Internal error: Bad condition
pe=00029 FAIL at line=00425 MAPL_Cap.F90 <status=18>
pe=00029 FAIL at line=00131 MAPL_Cap.F90 <status=18>
pe=00029 FAIL at line=00032 GEOSgcm.F90 <status=18>
ESMF_Finalize: Error flushing log file: Operation not yet supported
ESMF_LogWrite(): ESMF_Log not open -- cannot ESMF_LogWrite(). Log message =
Internal error: Bad condition - - Invalid GlobalVM
ESMF_Finalize: Error finalizing VM: Internal error: Bad condition

tclune:
ok - do you have a fix?   They certainly should both be the same.  Eventually both will become world_comm, but for now they should both be world_comm.

weiyuan-jiang:
I haven't found a fix even I tried many ways yesterday. I would try again

weiyuan-jiang:
initialization does not have the condition this%mpi_already_initialized, why finalize need it?  @tclune @mathomp4 

     if (.not. this%mpi_already_initialized) then
         call MAPL_Finalize(comm=this%comm_world)
         call ESMF_Finalize(rc=status)
          _VERIFY(status)
      end if

PR #857, opened 2021-05-19
match esmf_initialize() and esmf_finalize()

weiyuan-jiang:
This PR solve issue #854.  The argument for ESMF_Finalize() is necessary to not finalize mpi at that point
## Types of changes

## Checklist:

### DIFF
--- a/gridcomps/Cap/MAPL_Cap.F90
+++ b/gridcomps/Cap/MAPL_Cap.F90
@@ -32,7 +32,6 @@ module MAPL_CapMod
       class (MAPL_CapOptions), allocatable :: cap_options
       ! misc
       logical :: mpi_already_initialized = .false.
-
       type(MAPL_CapGridComp), public :: cap_gc
       type(ServerManager) :: cap_server
       type(SimpleCommSplitter), public :: splitter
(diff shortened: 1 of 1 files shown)


gridcomps/Cap/MAPL_Cap.F90
```diff
@@ -419,10 +420,9 @@ subroutine finalize_mpi(this, unusable, rc)
       integer :: status
       _UNUSED_DUMMY(unusable)
 
+      call MAPL_Finalize(comm=this%comm_world)
```
tclune, line 423: This call is illegal if MPI is already finalized - the comm may no longer exist.  So either there is a problem here or the conditional on the next line is unnecessary.

weiyuan-jiang: I don't think MPI_finalize should have been called at this point.  If MPI_Initialize is called by Cap's parents, then it is its parents' responsibility to call MPI_finalize() after our Cap's  finalize_mpi() call.

tclune: I'm now realizing that I had misinterpreted the conditional that follows this line.

tclune (approved):

PR #860, opened 2021-05-20
Merge develop into main for MAPL 2.6.8 Release

mathomp4:
I've added a blocking label to this to do some testing on discover to make sure all is well. If so, then I'll release the label.


mathomp4:
Zero diff on discover with GCC and Intel!

bena-nasa (approved):

Issue #861, opened 2021-05-20
MAPL 2.6.7 crashes upon Cap initialization


rmontuoro:
When built with Intel Fortran Compiler 18.0.5.274 and Intel MPI 2018.0.4, MAPL crashes at runtime upon initialization of MAPL Cap with the following error messages:

_standard error_:
```
FAIL at line=00135    MAPL_CapGridComp.F90                     <status=19>
FAIL at line=00311    MAPL_Cap.F90                             <status=19>
```
`PET000.ESMF_LogFile:`
```
ESMF_Config.F90:2840 ESMF_ConfigLoadFile_1proc_       Unable to open file  - error opening text file:
ESMF_Config.F90:2779 ESMF_ConfigLoadFile       Unable to open file  - unable to load file:
```

The crash is caused by the `cap_options` data structure not being properly populated when calling `MAPL_Cap()`.

@weiyuan-jiang created a reproducer to investigate this issue.

rmontuoro:
Fixed in MAPL 2.7.0

PR #862, opened 2021-05-20
Fix runtime failure when building with Intel 18.0.5.274 and Intel MPI 2018.0.4

rmontuoro:
## Description

When building with Intel Fortran Compiler 18.0.5.274 and Intel MPI 2018.0.4, MAPL crashes at runtime upon Cap initialization, as described in further detail in issue #861.

This PR declares as Fortran derived type (`type`) a few variables previously declared as `class` to prevent the observed runtime errors.

## How Has This Been Tested?

This PR was tested within the UFS model on NOAA R&D Hera platform.
## Types of changes

## Checklist:

### DIFF
--- a/gridcomps/Cap/MAPL_Cap.F90
+++ b/gridcomps/Cap/MAPL_Cap.F90
@@ -29,7 +29,7 @@ module MAPL_CapMod
       integer :: rank
       integer :: npes_member
  
-      class (MAPL_CapOptions), allocatable :: cap_options
+      type (MAPL_CapOptions), allocatable :: cap_options
       ! misc
       logical :: mpi_already_initialized = .false.
       type(MAPL_CapGridComp), public :: cap_gc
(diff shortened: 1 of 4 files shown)


weiyuan-jiang:
The PR #846 will address this issue.  But I don't mind taking a small step here

weiyuan-jiang:
After taking another look, I think this is OO fix we want.  MAPL_Cap here takes a MAPL_CapOptions type, but in other applications, MAPL_FlapCapoptions is passed in

gridcomps/Cap/MAPL_Cap.F90
```diff
@@ -29,7 +29,7 @@ module MAPL_CapMod
       integer :: rank
       integer :: npes_member
  
-      class (MAPL_CapOptions), allocatable :: cap_options
+      type (MAPL_CapOptions), allocatable :: cap_options
```
tclune, line 32: I suspect the ALLOCATABLE is no longer needed.

mathomp4:
Per @rmontuoro This should be taken care of with MAPL 2.7.0, correct?

rmontuoro:
MAPL 2.7.0 addressed this issue.

PR #864, opened 2021-05-25
Merge main into develop

mathomp4:
I keep seeing weird "develop and main" are not in sync errors. So let's try merging `main` into `develop` after that last release.


mathomp4:
How weird. There must be a bug in how GitHub Actions works with `main` → `develop` PRs.

Issue #870, opened 2021-06-01
Query with MAPL FLAP True/False Options


mathomp4:
This is maybe for @weiyuan-jiang or @tclune.

Namely, I'm looking at adding a CLI option (see #869) and I noticed these options:

```fortran
      call options%add(switch='--use_comm_world', &
           help='# The model by default is using MPI_COMM_WORLD : .true. or .false.', &
           required=.false., &
           act='store',      &
           def='.true.',     &
           error=status)
      _VERIFY(status)

...
      call options%add(switch='--isolate_nodes', &
           help='Padding extra processes in the last nodes with idle', &
           required=.false., &
           def='.true.', &
           act='store', &
           error=status)
      _VERIFY(status)

      call options%add(switch='--fast_oclient', &
           help='Copying data before isend. Client would wait until it is re-used', &
           required=.false., &
           def='.false.', &
           act='store', &
           error=status)
      _VERIFY(status)

     call options%add(switch='--one_node_output', &
           help='Specify if each output server has only one nodes', &
           required=.false., &
           def='.false.', &
           act='store', &
           error=status)

      call options%add(switch='--with_io_profiler', &
           help='Turning on io_profler', &
           required=.false., &
           def='.false.', &
           act='store', &
           error=status)
      _VERIFY(status)
```
My question is: Do these take values? Or are they just "true/false" switches? If they are just "true if set" switches, then they should probably be:
```fortran
act='store_true'
```
a la argparse, so that if set, they are `.true.`, else `.false.`

weiyuan-jiang:
They take true or false values.  By default, they are false.

mathomp4:
@weiyuan-jiang Hmm. Is there a reason they are like that instead of just making them a "switch"? A la, if the switch is there, it's true, else false? Like a `--verbose` or `--help` option?

weiyuan-jiang:
I am not sure. But I will try

PR #872, opened 2021-06-02
set logical values through Flap without specifying true or false

weiyuan-jiang:
Now it can get the logical values without specifying true or false.
For example, --with_io_profiler  would give a true value. To avoid confusion, we change the command_line from --use_comm_world to --use_sub_comm and --isolate_nodes to --no_isolated_nodes .

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Regrid_Util.x now checks if file exsts and captures the units and long_name of the input for the output file
 ### Changed
-
+- set logical values in flap commmand line without true or false values
 - Set required CMake Version to 3.17
 - Update `components.yaml`
   - ESMA_cmake v3.4.2 (to match GEOSgcm)
--- a/base/MAPL_CapOptions.F90
+++ b/base/MAPL_CapOptions.F90
@@ -98,7 +98,7 @@ contains
       integer, optional, intent(out) :: rc
       integer :: status
       character(80) :: buffer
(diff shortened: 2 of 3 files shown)


weiyuan-jiang:
address issue #870

mathomp4:
@weiyuan-jiang You tested all this, correct? 

And would we need to inform anyone about this change? (I don't know if people outside of GEOS are using these options or if they mainly exist for testing/development work)

tclune (commented):
GCHP should be fine - they only use CapOptions struct which is unchanged here.

I'm more concerned about backward compatibility.  @mathomp4 do the gcm run scripts already use the options being modified?

mathomp4:
@tclune Nope. By default the only options the GCM would use are:

* `--npes_model`
* `--nodes_output_server`
* `--logging_config`

Anything beyond that is usually "Ask @weiyuan-jiang" for me because you are often in Fancy MAPL Land™

weiyuan-jiang:
I changed two interfaces ( --use_comm_world to --use_sub_comm,  --isolate_nodes to --no_isolated_nodes) because it does not make sense if default is true and "--use_comm_world" becomes false.  The same applies to "--no_isolated_nodes".  However, as far as I can tell, nobody is using these two parameters at this moment.  So it is probably safe

tclune (commented):
My remaining concern is that it is a bit awkward to have options that have negation in their name:  `--no_isolated_nodes`.

Of course at the same time, it is highly desirable to have default values be `.false.`.

Maybe something like `--compress_nodes` would work better?

weiyuan-jiang:
@tclune That seems better.

tclune (approved):

Issue #866, opened 2021-05-26
URGENT: UFS-GOCART crashes upon finalize when built with MAPL v2.7.0


rmontuoro:
The error below occurs upon finalizing GOCART within the UFS weather model built with MAPL v2.7.0:
```
*** Error in `/scratch1/NCEPDEV/nems/Raffaele.Montuoro/dev/nasa/dev/latest/ufs-weather-model/build/ufs_model': corrupted size vs. prev_size: 0x000000001bdafd60 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7f474)[0x2aee295c6474]
/lib64/libc.so.6(+0x8156b)[0x2aee295c856b]
/lib64/libc.so.6(__cxa_finalize+0x9a)[0x2aee295810aa]
/scratch1/NCEPDEV/nems/Raffaele.Montuoro/dev/nasa/dev/latest/libs/mapl/2.7.0/lib64/libMAPL.history.so(+0x3a2393)[0x2aee1bae3393]
```

`gdb` backtrace returns the following:
```
(gdb) bt
#0  0x00002b6afbb513d7 in raise () from /lib64/libc.so.6
#1  0x00002b6afbb52ac8 in abort () from /lib64/libc.so.6
#2  0x0000000007a3e18b in for__signal_handler ()
#3  <signal handler called>
#4  0x00002b6afbb513d7 in raise () from /lib64/libc.so.6
#5  0x00002b6afbb52ac8 in abort () from /lib64/libc.so.6
#6  0x00002b6afbb93f67 in __libc_message () from /lib64/libc.so.6
#7  0x00002b6afbb9c329 in _int_free () from /lib64/libc.so.6
#8  0x00002b6afbb550aa in __cxa_finalize () from /lib64/libc.so.6
#9  0x00002b6aec4c2e43 in __do_global_dtors_aux () from /scratch1/NCEPDEV/nems/Raffaele.Montuoro/dev/nasa/dev/latest/libs/mapl/2.7.0-debug/lib64/libMAPL.history.so
#10 0x00007ffcef6d4eb0 in ?? ()
#11 0x00002b6aeb4ff08a in _dl_fini () from /lib64/ld-linux-x86-64.so.2
Backtrace stopped: frame did not save the PC
```

which shows an issue occurring in the C runtime library while destroying a global C++ object in `libMAPL.history.so` via destructor ` __do_global_dtors_aux ()`

mathomp4:
[...]

PR #875, opened 2021-06-03
Build as shared/static according to BUILD_SHARED_LIBS

rmontuoro:
This PR allows MAPL to be built as a set of shared (default) or static libraries based on the value of BUILD_SHARED_LIBS
## Description

The included changes remove hardcoded settings aimed at building MAPL as shared libraries and allow users to choose between shared and static builds based on the CMake `BUILD_SHARED_LIBS` flag.

The previous build behavior is retained by setting `BUILD_SHARED_LIBS` to `ON` by default.
## Related Issue

This PR addresses issue #866.
## Motivation and Context

These changes allow building MAPL as static, which is required for the UFS

## How Has This Been Tested?

This PR was tested on NOAA RDHPCS platforms Hera and Orion by means of the UFS FV3-GOCART application.
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 ### Added
 - Regrid_Util.x now checks if file exsts and captures the units and long_name of the input for the output file
+- Build libraries either as shared (default) or static based on BUILD_SHARED_LIBS flag
 ### Changed
 
 - Set required CMake Version to 3.17
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,9 @@ project (
   VERSION 2.7.0
   LANGUAGES Fortran CXX C)  # Note - CXX is required for ESMF
 
+# build as shared libraries by default
+option (BUILD_SHARED_LIBS "Build shared libraries" ON)
+
 # mepo can now clone subrepos in three styles
 set (ESMA_CMAKE_DIRS
   ESMA_cmake
--- a/MAPL/CMakeLists.txt
+++ b/MAPL/CMakeLists.txt
@@ -6,7 +6,6 @@ esma_add_library (${this}
   DEPENDENCIES MAPL.base MAPL.pfio MAPL_cfio_r4 MAPL.gridcomps
                esmf NetCDF::NetCDF_Fortran MPI::MPI_Fortran
                $<$<BOOL:${BUILD_WITH_FLAP}>:FLAP::FLAP>
-  TYPE SHARED
   )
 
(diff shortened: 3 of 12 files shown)


weiyuan-jiang (commented):
It is nice to have options. Your approach will work if MAPL is built independently.  But if it is built with other applications ( like GEOSgcm and GEOSldas), it is too late to set the options here and it will affect all components if it is set at the very beginning.  To be specific to MAPL, we can

option (BUILD_SHARED_MAPL "Build shared MAPL  libraries" ON)
if (BUILD_SHARED_MAPL)
   set (MAPL_TYPE SHARED)
else ()
   set (MAPL_TYPE STATIC)
endif()

esma_add_library(
  ......
  TYPE ${MAPL_TYPE})

rmontuoro:
@weiyuan-jiang - Good point. Either approach would work. Note that this is a priority for the UFS.

weiyuan-jiang:
please check my PR #876 @rmontuoro

rmontuoro:
PR #876 can replace this PR.

mathomp4:
I'm throwing a blocker on this until I can really test it. My fear is that it might have broken one of our workarounds:
```cmake
if (APPLE AND CMAKE_Fortran_COMPILER_ID MATCHES Intel AND CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 20.2.1)
   set (LIBRARY_TYPE STATIC)
   ecbuild_warn (
      "Found Intel oneAPI on macOS.\n"
      "MAPL developers have found an issue with Intel oneAPI on macOS\n"
      "where GEOSgcm.x would not work. Debugging found the issue was\n"
      "that command_argument_count() would return -1 which should *NEVER*\n"
      "happen per Fortran Standard and then this broke FLAP.\n"
      "A workaround was found that if the ${this} library was compiled\n"
      "as TYPE STATIC, the model would work. So we are setting ${this} as\n"
      "a TYPE STATIC library. Note: This might interfere with coupled model.")
else ()
   set (LIBRARY_TYPE SHARED)
endif ()
```
As that says, for Intel oneAPI on macOS, our only workaround for a bug was to *force* MAPL_cfio as `TYPE STATIC`.  My guess is this would now build as `SHARED` no matter what.

Maybe we can get away with "If Intel oneAPI on macOS, STATIC everywhere" but that would need to be tested.

mathomp4:
Also changing label to just "0 diff". I can never say a change in how a library is linked and archived is a trivial PR!

rmontuoro:
@mathomp4 - I agree with closing this PR and merging PR #876 instead. @weiyuan-jiang's approach seems safer for your environment.

mathomp4:
Closing as #876 has been merged into `develop`

PR #888, opened 2021-06-21
Fixes for some stray mod files

### LINKED ISSUE #848: Bad path for .mod files in ./Tests
I think there is a problem with the Fortran module path in the MAPL Tests directory.  @amdasilva  spotted some stray .mod files and from his quick screen share it looks like they were in the default directory that ecbuild uses unless we override per-target as we do elsewhere.

mathomp4:
## Description

This PR fixes a few places where `.mod` files were not being put in `build/include` but in the "source" tree in `build`.

## Related Issue

Closes #848 

## Motivation and Context

Keeps MAPL clean 😄 

## How Has This Been Tested?

It builds and all MAPL tests passed. I'll do a GEOSgcm test soon with it (though if it builds, it should be fine).

## Types of changes

## Checklist:


atrayano (approved):
Looks very reasonable. I am not 100% happy with hard-coded names (shared/tests/CMakeLists.txt), but if we do not forget about it, it should be fine

mathomp4:
Sigh. Dang it. I missed some! Give me a bit to clean those up!

atrayano (approved):

mathomp4:
Okay. Turns out `make tests` also makes `.mod` files. Those I missed. And, unfortunately for @atrayano and all involved, it's more hard-coded paths. But, well, no other good way 🤷🏼‍♂️

mathomp4:
Oops. I think I accidentally removed @atrayano 's re-review. GitHub confuses me sometimes! Grr...

mathomp4:
Blocking again because I want to make this cleaner

atrayano:
@mathomp4 Let me know if you need another re-review

mathomp4:
> @mathomp4 Let me know if you need another re-review

I will. I discussed with @tclune and he knew of a CMake trick that should make this at least a bit more palatable. I'll let you look at the change when I get it in and tested 😄

mathomp4:
Well, it's not perfect, but it's better. At least the "hard-coded" paths are now in one place per file. I did try experimenting with using `set(CMAKE_Fortran_MODULE_DIRECTORY foo)` but I got something very wrong. So for now I figure, let's get this in, and then we can fix things if needed.

atrayano (approved):

mathomp4:
It is zero-diff for the GEOSgcm as expected as well.

atrayano (approved):

tclune (approved):

PR #890, opened 2021-06-21
Add stale bot yaml

mathomp4:
## Description

Recently, @tclune got a [Stale Bot](https://probot.github.io/apps/stale/) message from GCHP and we thought, huh, that's a good idea.

So, we have ~stolen~borrowed some ideas from @msulprizio. Hopefully she doesn't mind that our `stale.yml` looks suspiciously like her [`stale.yml`](https://github.com/geoschem/GCHP/blob/main/.github/stale.yml). But, well, we know her version works! 😄 

Note that we are using `long term` as our "do not stale" label (I kept in `security` because I started from the Stale Bot template...and it seems like something we might need? If not, I can remove it.)

## Motivation and Context

Helps keep stale issues and PRs from gumming up the repos (we hope).

## How Has This Been Tested?

Uh. Well, I guess it hasn't been. I'm a bit wary to turn this on and see what happens to the repo. At the moment we have a *LOOOOOOT* of issues that are older than 60 days.  So, as a "just in case" I have added a `before stale` label to all current open issues and PRs. That way we can go find them if the bot decides to close them en masse.

## Types of changes

## Checklist:

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,19 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - long term
+  - security
+# Label to use when marking an issue as stale
+staleLabel: stale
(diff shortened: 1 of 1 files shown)


atrayano (approved):
Looks good

Issue #899, opened 2021-06-24
Wrong coordinate units provided to the UFS aerosol component


rmontuoro:
Both MAPL and GOCART assume that grid coordinates are in radians. However, this may not be the case if the grid is provided externally, as it happens for the UFS aerosol component.

MAPL should check the grid's coordinate system flag and convert the coordinates to radians if needed.

tclune:
@rmontuoro Can you give a more specific scenario?    We are generally aware of this issue, but since we defer regridding to ESMF, that handles most cases where the coordinate system matters.   If an individual component wants to directly access lat/lon  that would be the place that a check should be made.    And I'm sure that GEOS components just assume the radians convention at this point and are thus fragile in that regard.

I'm more than happy to have someone add the necessary hooks if there is a MAPL wrapper layer in use that is pulling this information out of an ESMF grid.   Just want to be certain that this is an actual issue first.

Mentioning @atrayano in case he can think of other cases where we might need this.

tclune:
Ah - I see there is already a PR.   I'll check that - so maybe nothing more is needed here.

PR #900, opened 2021-06-24
Ensure grid coordinates are always provided in radians

rmontuoro:
based on the grid's coordinate system flag.

## Description

This PR adds changes ensuring that grid coordinates are always provided in units of radians regardless of the ESMF Grid object used.

Grid coordinates are assumed to be in radians throughout MAPL and GOCART. This assumption may no longer hold if an ESMF Grid object is provided externally to MAPL (as done in the UFS). This PR implements automatic unit conversion to radians for grid coordinates based on the coordinate system used to create the ESMF Grid object.

Note that `ESMFL_GridCoordGet()` will return with a generic `ESMF_FAILURE` error code if the coordinate system flag is neither `ESMF_COORDSYS_SPH_DEG`, nor `ESMF_COORDSYS_SPH_RAD`.

## Related Issue

Fixes issue #899 
## How Has This Been Tested?

This change was tested on the NOAA RDHPCS platform Hera within the UFS weather model.

## Types of changes

## Checklist:

### DIFF
--- a/base/ESMFL_Mod.F90
+++ b/base/ESMFL_Mod.F90
@@ -379,7 +379,7 @@ subroutine ESMFL_GridCoordGet(GRID, coord, name, Location, Units, rc)
   else if (crdSys == ESMF_COORDSYS_SPH_RAD) then
      conv2rad = 1._ESMF_KIND_R8
   else
-     _RETURN(ESMF_FAILURE)
+     _FAIL('Unsupported coordinate system:  ESMF_COORDSYS_CART')
   end if
 
   if (tk == ESMF_TYPEKIND_R4) then
--- a/base/ESMFL_Mod.F90
+++ b/base/ESMFL_Mod.F90
@@ -317,11 +317,13 @@ subroutine ESMFL_GridCoordGet(GRID, coord, name, Location, Units, rc)
 
 !	local variables
   integer                   :: rank
+  type(ESMF_CoordSys_Flag)  :: crdSys
   type(ESMF_TypeKind_Flag)  :: tk
   integer                   :: counts(ESMF_MAXDIM)
   integer                   :: crdOrder
   real(ESMF_KIND_R4), pointer :: r4d2(:,:)
   real(ESMF_KIND_R8), pointer :: r8d2(:,:)
+  real(ESMF_KIND_R8)        :: conv2rad
   integer                   :: STATUS
   integer                   :: i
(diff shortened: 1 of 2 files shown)


atrayano (commented):
This PR is based on a assumption that coordSys is properly set during grid creation

atrayano (approved):
It looks OK since MAPL always sets coordSys

mathomp4:
Blocking this until I can test with GEOS.

base/ESMFL_Mod.F90
```diff
@@ -372,6 +374,14 @@ subroutine ESMFL_GridCoordGet(GRID, coord, name, Location, Units, rc)
      _RETURN(ESMF_SUCCESS)
   end if
 
+  if (crdSys == ESMF_COORDSYS_SPH_DEG) then
+     conv2rad = MAPL_PI_R8 / 180._ESMF_KIND_R8
+  else if (crdSys == ESMF_COORDSYS_SPH_RAD) then
+     conv2rad = 1._ESMF_KIND_R8
+  else
+     _RETURN(ESMF_FAILURE)
```
mathomp4, line 382:
```suggestion
     _FAIL('Unsupported coordinate system:  ESMF_COORDSYS_CART')
```

mathomp4: Just to make the failure more explicit.

base/ESMFL_Mod.F90
```diff
@@ -301,7 +301,7 @@ end subroutine ESMFL_StateGetField
 
 !BOP
 
-! !IROUTINE: ESMFL_GridCoordGet - retrieves the coordinates of a particular axis 
+! !IROUTINE: ESMFL_GridCoordGet - retrieves the coordinates of a particular axis in radians
```
mathomp4, line 304: @rmontuoro We updated this doc string as well. We couldn't do it through the GitHub "make a suggestion" feature since it wasn't within three lines of a change...I guess?

mathomp4:
This does seem zero-diff for GEOS. Though I only tried a C24 L91 test.

tclune (approved):

mathomp4:
NOTE: Labeled as 0-diff as my testing shows it is...but hard to guarantee that.

tclune:
Multiplying by 1. is safe.   Anything that actually had the wrong units would be a big change - you would have seen it.

PR #901, opened 2021-06-24
Update Changelog Enforcer, add error message

mathomp4:
This is an update to the changelog enforcer that now apparently allows a more exciting error message. This is to test that!


mathomp4:
Okay. The error message is found in the [Details](https://github.com/GEOS-ESM/MAPL/pull/901/checks?check_run_id=2905779437) link under the checks. I wish it would be in the box on this page, but better than nothing 🤷🏼‍♂️

mathomp4:
And now in a bit of irony, I label this change as trivial so that it skips the Changelog Enforcer 😆

tclune (approved):

Issue #887, opened 2021-06-17
MAPL 2.7.0 and MAPL 2.7.1 non-zero diff with OPS emissions


bena-nasa:
When MAPL was upticked to 2.7.1 it was found that GEOS went non-zero diff when using the OPS emissions. On investigation I did find very subtle differences (in the order of the 5 or 6th decimal place in the mantissa) in the emissions in ExtData between MAPL 2.7.0 and 2.7.1 when using the OPS emissions. After much work I traced it down to the fact that two different input files used in the OPS emissions the QFED real time emissions and the "HTAP" emission, both on 3600x1800 DE/PE lat-lon grids have ever so slightly different lon/lat values.
In one the lons are
```-179.95, -179.85, -179.75 .... 179.75, 179.85, 179.95```
In the other the lons are
```-179.949996948242, -179.850006103516, -179.75 .... 179.649993896484, 179.75, 179.850006103516, 179.949996948242```

likewise for the lats.
Now when both files are opened by the Lat-Lon grid factory, it determines that both are 3600x1800 PE/DE regular grids based on the currently logic, but uses the lons and lats from the file.

Now the upshot is, whichever file gets opened first wins and the 3600x1800 PE/DE grid has the lats/lons from that file.

Well of course the the logic change I made to ExtData in MAPL 2.7.1 causes the order of which file gets opened first with the result that the 3600x1800 PE/DE grid is now slightly different which means the route handle for the regriding is slightly different, which means slightly different emissions and non-zero diff.

So on the one hand there is no reason that this could not in principle happen again, a user changes the RC file, new things are being provided by ExtData.

So what are the options
1. Do nothing
2. Modify the equals overload in the factory so it would have caught and made two different grids
3. If it detects that the grid one of our "standard" grids PC/PE/DC/DE ignore the values in the file and have the factory compute them.

The GEOSgcm gatekeeper was not in favor of 1 because what would prevent this from popping up again potentially and flagging a non-zero diff we have to track down.

atrayano:
I am not in favor of 1) either. But if we do 2) chances are we still will get non-zero diffs because one way or the other, in the past we were using coordinates from a single grid (whichever the first file happen to be). 3) is a good option, and with some luck it might be zero diff against older tags

bena-nasa:
Either way, this is a non-zero diff once going forward, the question is what can we do to ensure than something like this will not happen again.

atrayano:
[...]

PR #906, opened 2021-06-29
Feature/bmauer/#887 consistent fgrid

bena-nasa:
Fixes issue #887 
Note this is a non-zero diff change.
This now forces consistent behavior when using the coordinates from a file. We hit a non-zero diff as described in the issue #887. The problem was that we had two files with every so slightly different coordinates that were identified as a standard geos grid, but use the file coordinates. If the grids are defined as standard we say they are regular and PC/PE/DC/PE and in this case the equals operator in gridfactory was not checking the coordinates.

This forces consistent behavior. Now if the grid is identified as a standard grid (i.e. is evenly space and PC/DC/PE/DE up to a numerical tolerance) we use the computed coordinates from the routines in the factory. This way each grid has consistent coordinates. This can be overwritten to use the coordinates from the file. In this case we say the grid is not regular which forces a different chain in the "equals" operator in the grid factory and if the two grids are not regular will compare the coordinates themselves.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix format for writing out large number
 - Fixed CMAKE_Fortran_MODULE_DIRECTORY for some directories
+- Update handling of file coordinates when creating grids from file. Now if identified as a standard grid compute coordinates. Option to allow this to be overrided and use file coordinates. Fixed issue if two files are identified as a standard grid but has very slightly different coordinates causing one or the other to be used depending on which file is used first.
 
 ## [2.7.3] - 2021-06-24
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix format for writing out large number
 - Fixed CMAKE_Fortran_MODULE_DIRECTORY for some directories
+- Update handling of file coordinates when creating grids from file. Now if identified as a standard grid compute coordinates. Option to allow this to be overrided and use file coordinates. Fixed issue if two files are identified as a standard grid but has very slightly different coordinates causing one or the other to be used depending on which file is used first.
```
tclune, line 20: Probably should specify how to use that option here in the changelog rather than making interested users chase it down.   Low priority so ignore if I don't find other issues.

tclune (approved):

PR #908, opened 2021-06-29
Fixes #907 ESMF warnings in log.

### LINKED ISSUE #907: ESMF throws warnings when running tests
If one activates ESMF logging, the regular tests throw many ESMF errors.

tclune:
Despite the documentation, one does need to pass an importState,
exportState, and clock to the ESMF GridComp run methods.  Previously,
ESMF_TestCase tests would present numerous warnigs when logging was
turned on.  Logging was not on, which is why this was not noticed for
so long.

## Types of changes

## Checklist:

### DIFF
--- a/pfunit/ESMF_TestCase.F90
+++ b/pfunit/ESMF_TestCase.F90
@@ -44,14 +44,32 @@ contains
 
       logical :: discard
       type (ESMF_GridComp), target :: gc
+      type (ESMF_State) :: state
+      type (ESMF_Clock) :: clock
+      type (ESMF_TimeInterval) :: step
+      type (ESMF_Time) :: time0, time1
       integer :: rc
       integer :: pet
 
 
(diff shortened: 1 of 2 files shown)


atrayano (approved):
@tclune the PR looks fine. If we want to be super strict, all the ESMF_GridComp methods have an extra argument: userRc. The subtle difference is that userRc will be set to non-zero value if there is an error in the user code. RC will be set to non-zero value if there is an error in the ESMF code. In MAPL we do use userRc and rc. I decided to approve the PR, but  you might want to make changes

tclune:
I agree about capturing the user-rc and several semi-related issues of capturing rc's further down in that file.   Those were not in the way, and the run method is so trivial here that I doubt there is added value.   If I have energy this evening maybe I'll do a second pull to clean this up.

PR #910, opened 2021-06-30
Fix bug in latlon factory for use_file_coords logic

bena-nasa:
Thanks to Matt's nightly testing I found a corner case I had not protected against yesterday with my PR that was to address this issue #887 
If you weren't using the file coordinates by force it tested independently the lons and the lats if they were regular and one of our standard PC/PE/DC/PE and if they indepently matched they were computed.

But I did not consider the case where it only detected one of the two dimensions as regular. In that case it was computing one but using the file for the other. And that a side effect was that the file coordinate was not being converted to degrees.

So I have modified the logic to say both dimensions must be regular to use the computed  coordinates. If one is not the grid is not regular and we use the the file coordinates, and I make sure they are converted to degrees.

## Types of changes

## Checklist:


atrayano (approved):
Looks good

PR #921, opened 2021-07-12
Prepare CHANGELOG and version for MAPL 2.8.0 release

mathomp4:
After discussing with @tclune, the change to requiring the latest GFE libraries (i.e, CMake change to `GFTL::gftl`), means the next MAPL should be 2.8.0 

And since the @bena-nasa grid fix has to go to @sdrabenh at some point, this means we need a release to give him. The MAPL 2.8.0 sort of also implies the non-zero-diff ness, though that's not why we change the minor number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 ### Added
+### Changed
+### Fixed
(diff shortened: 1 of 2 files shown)


tclune (approved):

mathomp4:
Note: the odd spacing changes in `CHANGELOG.md` are because I saw some tab characters in there. And spaces > tabs.

PR #925, opened 2021-07-15
Remove testPresets from CMakePresets.json file

mathomp4:
So it turns out I can't quite figure out how to get the MAPL tests to work through presets. So for now, I'm just going to remove the `testPresets` bits of the `CMakePresets.json` file. Better to not make people think functionality exists if it doesn't work!


mathomp4:
Okay. Let's see if the CI is happy this time...

bena-nasa (approved):

PR #927, opened 2021-07-22
Add stretching parameters to restart files

### LINKED ISSUE #915: Feature request: stretched grid file attributes and run-time error checking
The MAPL stretched grid capability is now available for use by the general public in GCHP and we are starting to receive feedback. This week a user asked a question about the model not crashing when using a restart on the wrong grid. The grid size was correct but the restart was not generated using the same grid stretching as the simulation. The question was would this output be scientifically valid. The answer was no, there would be a silent bug.

To avoid this problem there could be (1) stretched grid information (target lat, lon, stretch factor) stored in the checkpoint file, and (2) grid validation of the restart file upon simulation start beyond just size. Having this information in the checkpoint files for easy referencing would help users stay organized and avoid mistakes. Having this information in History files would also be beneficial for the same reasons.

Pinging @LiamBindle, @bena-nasa, @atrayano, @tclune, @sdeastham for discussion.

bena-nasa:
Fixes #915 
This brings the capability requested in the mentioned issue. This adds the stretching parameters to restart files and then when reading restart files checks that the grid on the file matches the application (including the stretching parameters).
This did require some effort as the restarts technically use what we were calling the "old" cubed-sphere so I ended up getting rid of that and combining the capability into the plain cubed-sphere grid factory. It also required modifying the equals operator in the grid factories so that we could introduce a routine to just compare the physical grid parameter rather than the full factory.

## Types of changes

## Checklist:


base/MAPL_CubedSphereGridFactory.F90
```diff
@@ -723,10 +771,65 @@ elemental subroutine set_with_default_bounds(to, from, default)
       
    end subroutine set_with_default_bounds
    
+   logical function test_decomp_equals(this, a)
```
tclune, line 774: I would prefer a name that does not start with `test_` which is the semaphore use for unit tests.  How about just `decoms_equal` or `equal_decomps`?

base/MAPL_LatLonGridFactory.F90
```diff
@@ -1387,60 +1389,96 @@ subroutine initialize_from_esmf_distGrid(this, dist_grid, lon_array, lat_array,
 
    end subroutine initialize_from_esmf_distGrid
 
+   logical function test_decomp_equals(this,a)
```
tclune, line 1392: Note you can still declare the result variable to be `result(equal)` to make the procedure read a bit better.  I only mention since you are changing the proc name anyway.

base/MAPL_CubedSphereGridFactory.F90
bena-nasa: NP, I will change that

base/MAPL_LatLonGridFactory.F90
```diff
@@ -1387,60 +1389,96 @@ subroutine initialize_from_esmf_distGrid(this, dist_grid, lon_array, lat_array,
...
+         if (.not. test_decomp_equals) return
 
-         equals = (a%is_regular .eqv. b%is_regular)
-         if (.not. equals) return
+      end select
+
+   end function test_decomp_equals
+
+
+   logical function test_physical_params_equals(this, a)
```
tclune, line 1416: Again - avoid `test_`.    Maybe `physical_params_are_equal()`

base/MAPL_AbstractGridFactory.F90
```diff
@@ -89,6 +91,18 @@ logical function equals(a, b)
          class (AbstractGridFactory), intent(in) :: b
       end function equals
 
+      logical function test_decomp_equals(this,a)
```
tclune, line 94: I suggest the name `decomps_are_equal`  with a result variable called `equal`

base/MAPL_CubedSphereGridFactory.F90
tclune: Better suggestion is `decomps_are_equal` as per above.

base/MAPL_LatLonGridFactory.F90
tclune: Or variant `are_physical_params_equal()`

tclune (changes requested):
Requesting changes to procedure names.

base/MAPL_LatLonGridFactory.F90
bena-nasa: changed to physical_params_are_equal

tclune (approved):

mathomp4:
Well, the GCM built but the MAPL tests were not happy.

CHANGELOG.md
```diff
@@ -24,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unit tests can now use the `_RC` macro for checking results from
   calls to ESMF.  The file must first CPP define either `I_AM_PFUNIT`
   or `I_AM_FUNIT` (serial) and then `#include "MAPL_ErrLog.h"`.
+- Add stretch parameters to restarts and check the file grid compared to MAPL grid
```
mathomp4, line 30: @bena-nasa I think this changelog change should be in the Unreleased section.

tclune:
Yes - the unit tests involving factories need updating to correspond to these changes.  It should be straightforward ...

mathomp4 (approved):
Well, GEOS runs which is about all I can see. Approving!

tclune (approved):

PR #937, opened 2021-07-29
Only check cubed-sphere restarts for consistency with component grid

bena-nasa:
Well, most unfortunately it appears are restarts are so unlike our history output the only safe one I can check by making a factory to see that it matches the grid comp grid when reading the file is the cubed-sphere. When the LDAS was run with develop that now has the checking of the file grid against the component grid it failed. The LDAS apparently made a lat-lon grid and then when the initialization tried to make a grid from the restart file it failed. It went to the lat-lon factory but the particular restart had no lat/lon variables so the factory threw an error.

So we already discovered that this was not safe with the tripolar grid but apparently not even for a lat-lon grid as if you hit a restart like this it would fail.

So until we refactor MAPL_IO and modify the restart format we can really only do this for the cubed-sphere.
I have modified the check accordingly. This is too bad but at least the organization requesting this feature only uses the cubed-sphere.

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_Generic.F90
+++ b/base/MAPL_Generic.F90
@@ -5905,6 +5905,8 @@ end function MAPL_AddChildFromDSO
              call ESMF_AttributeGet(MPL%GRID%ESMFGRID,'GridType',value=grid_type,rc=status)
              _VERIFY(status)
           end if
+          !note this only works for geos cubed-sphere restarts currently because of 
+          !possible insufficent metadata in the other restarts to support the other grid factories
           if (trim(grid_type) == 'Cubed-Sphere') then
              app_factory => get_factory(MPL%GRID%ESMFGRID)
              allocate(file_factory,source=grid_manager%make_factory(trim(filename)))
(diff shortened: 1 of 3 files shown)


mathomp4:
I made a push to this PR as I'd want to release MAPL 2.8.2 with this fix. So I updated the Changelog and Cmakelists to fix it.

mathomp4:
Also, can confirm that this branch lets the LDAS global assim case run.

tclune (dismissed):

base/MAPL_Generic.F90
```diff
@@ -5905,7 +5905,7 @@ subroutine MAPL_ESMFStateReadFromFile(STATE,CLOCK,FILENAME,MPL,HDR,RC)
              call ESMF_AttributeGet(MPL%GRID%ESMFGRID,'GridType',value=grid_type,rc=status)
              _VERIFY(status)
           end if
-          if (trim(grid_type) /= 'Tripolar' .and. trim(grid_type) /= 'llc' .and. trim(grid_type) /= 'External') then
+          if (trim(grid_type) == 'Cubed-Sphere') then
```
tclune, line 5908: I think a comment here is probably important.    Just so the next person does not come along and say "Hey we use a factory everywhere this should work for lat-lon too.".    Just some sort of cautionary note.

tclune (approved):

base/MAPL_Generic.F90
bena-nasa: I added a comment

Issue #926, opened 2021-07-22
Refactor Constants into separate modules


Gvilla1000-nasa:
All constants are contained in a single file making physical/mathematical/internal distinction hard to observe. These need to be refactored into separate modules for easier identification.

Gvilla1000-nasa:
#42 This seems to be a related topic. Should this be addressed in this effort as well?

tclune:
My vote would be that we would address #42 during this process.  But ... new values should be hidden by `#ifdef` or some similar mechanism until we do a  "major" release of MAPL due to the backward compatibility issues.

Gvilla1000-nasa:
Question: I identified the following parameter: `ESMFL_UnitsRadians = 99` from `base/ESMFL_Mod.F90` as a potential parameter that can be moved into the new module. It is used in 4 different modules. Is this variable okay to be moved or because it is already identified as an `ESMFL` it should be left untouched?

bena-nasa:
I think it should be left untouched, this seems to be another software constant rather than a mathematical or physical constant

tclune:
Anyone care to comment on how much of ESMFL could be eliminated at this point?

bena-nasa:
It is hard to say for sure, I'm sure there are some routines that are either duplicates or not used but we would have to careful go through this and check the DAS. However, there are still some of these that are for sure used in the DAS and have no ESMF or other implementation in MAPL.  For example, off the top of my head, there are these bundle2state and state2bundle routines that do what the name suggests for example that I know the DAS uses.

Gvilla1000-nasa:
Ok. I think some software/internal constants could be added into MAPL_Constants. For example, `UNDEFINED_INTEGER`, `GRID_NAME_DEFAULT`, etc. I also identified `MAPL_UseStarrQsat` and other `use*Qsat` parameters that could be moved to the constants module but it seems these are only parameters that were used in a module that is not longer compiled. Should these `MAPL_Use` be removed to prevent confusion or will the `tstqsat.F90` be enabled in the future?

tclune:
If there are modules that are not being compiled, we (collectively) should decide if the module should be eliminated entirely.   (And if it is a test, as the name suggests it should at the very least be relocated to another direcotry.)

For now I think just maintain a list of questionable cases that we can review in a focused meeting.  

[...]

Issue #42, opened 2019-09-13
Update Physical Constants to CODATA 2018


mathomp4:
This is a tracking issue because the current MAPL constants are incorrect per CODATA. For example, the Avogadro constant is now *defined* exact. In MAPL:
```fortran
real, parameter, public :: MAPL_AVOGAD = 6.023E26               ! 1/kmol
```
Per CODATA:
```
Avogadro constant     6.022 140 76 e23         (exact)                  mol^-1
```

There are others that are now exactly defined because they are defined by exact constants. An example is the Stefan-Boltzmann constant. In MAPL:
```fortran
real, parameter, public :: MAPL_STFBOL = 5.6734E-8              ! W/(m^2 K^4)
```
Per CODATA:
```
Stefan-Boltzmann constant        5.670 374 419... e-8     (exact)                  W m^-2 K^-4
```
MAPL_RUNIV is also now exact as well and maybe others?

mathomp4:
Here's a list of changes that need to be made I think:
```fortran
real, parameter, public :: MAPL_STFBOL = 5.670374419E-8              ! W/(m^2 K^4)
real, parameter, public :: MAPL_RUNIV  = 8314.462618                 ! J/(Kmole K)
real, parameter, public :: MAPL_AVOGAD = 6.02214076E26               ! 1/kmol
```

PR #946, opened 2021-08-05
MAPL Constants Refactoring

Gvilla1000-nasa:
## Description
This PR is a bit extensive but overall it performs the following things:
1. Creates new directory shared/Constants/ that includes three modules housing internal, math, and physical constants.
2. As a consequence of (1), `shared/Enumerators.F90` and shared/MAPL_Constants.F90 are now removed.
3. Moved several `MAPL_` parameters found around MAPL into the internal constant modules. ~~One parameter in particular, `MAPL_UNDEF` is renamed to `disabled` to make sure the changes introduced here are still backward compatible since lots of places uses MAPL_UNDEF from BaseMod.~~
4. Added new physical variables and consolidated the use of some constants like PI to be from MAPL_Constants rather than being defined in each module.
5. Removed several unused variables

## Related Issue

#926, #42 

## Motivation and Context
Refactoring effort.

## How Has This Been Tested?
Ran a c12 for one day in release and debug mode.

## Types of changes

## Checklist:


Gvilla1000-nasa:
Query: In file `MAPL_cfio/ESMF_CFIOUtilMod.F90` the following parameter can be found:
` PARAMETER (IGREG=15+31*(10+12*1582))` and
`PARAMETER (IGREG=2299161)`

These are the first day of the gregorian calendar and the julian day count of the initial gregorian calendar respectively. Thoughts on adding these to either InternalConstants.F90 or MathConstants.F90? Their usage is limited but I've seen some components using these.

Gvilla1000-nasa:
Query: There are still some `MAPL_` variables lying around in Base.F90, Comms.F90,  ExtDataGridCompMod.F90 among others (e.g. `MAPL_StateItemOrderList`, MAPL_ExtDataVectorItem, MAPL_Distribute, MAPL_SunAutumnalEquinox). These are mostly kept within their own modules but because of their `MAPL_` prefix, these could be moved into `InternalConstantsMod.F90`. Thoughts?

Gvilla1000-nasa:
Query: Functions `deg2km` and `km2deg` inside `base/MAPL_NominalOrbitsMod.F90` aren't used. Furthermore, new constants are available inside PhysicalConstants.F90 that easily replace these functions. Are these safe to remove?

tclune:
If they are not used anywhere, then deleting fine.   (Probably do it as a separate PR though.)  If someone wants them later, we can recreate (using the MAPL constants) and give them better names.

tclune:
> Query: In file `MAPL_cfio/ESMF_CFIOUtilMod.F90` the following parameter can be found:
> ` PARAMETER (IGREG=15+31*(10+12*1582))` and
> `PARAMETER (IGREG=2299161)`
> 
> These are the first day of the gregorian calendar and the julian day count of the initial gregorian calendar respectively. Thoughts on adding these to either InternalConstants.F90 or MathConstants.F90? Their usage is limited but I've seen some components using these.

Can you summarize where these are used (outside of MAPL_cfio)?    The model itself should not be depending on such things I would hope.

Python/MAPL/constants.py
```diff
@@ -50,9 +41,127 @@
...
+MAPL_RUNIV   = 8314.47                # J/(Kmole K)
+
+MAPL_H2OMW   = 18.015                 # kg/Kmole
+MAPL_O3MW    = 47.9982                # kg/Kmole
+MAPL_ALHL    = 2.4665E6               # J/kg @15C
+MAPL_ALHF    = 3.3370E5               # J/kg
+MAPL_ALHS    = MAPL_ALHL+MAPL_ALHF    # J/kg
+
+
+!Internal constants
```
mathomp4, line 60:
```suggestion
#Internal constants
```
The comment character in Python is a hash. Fun with f2py has taught me this a lot!

Gvilla1000-nasa:
> > Query: In file `MAPL_cfio/ESMF_CFIOUtilMod.F90` the following parameter can be found:
> > ` PARAMETER (IGREG=15+31*(10+12*1582))` and
> > `PARAMETER (IGREG=2299161)`
> > These are the first day of the gregorian calendar and the julian day count of the initial gregorian calendar respectively. Thoughts on adding these to either InternalConstants.F90 or MathConstants.F90? Their usage is limited but I've seen some components using these.
> 
> Can you summarize where these are used (outside of MAPL_cfio)? The model itself should not be depending on such things I would hope.

Sure.
I found usage of iGreg (defined every time) in:
`Components/@GEOSgcm_GridComp/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM_GEOS5PlugMod/@mom/src/mom5/ocean_param/gotm-4.0/util/time.F90`

`Shared/MAPL/MAPL_cfio/ESMF_CFIOUtilMod.F90`

`Shared/MAPL/base/MAPL_NominalOrbitsMod.F90`

`Shared/GMAO_Shared/GMAO_etc/zeit_jday.f`

and

`Shared/GMAO_Shared/GMAO_gfio/timesubs.f`

base/Base/Base_implementation.F90
```diff
@@ -10,11 +10,11 @@
   ! !USES:
   !
   use ESMF
-  use MAPL_ConstantsMod, only: MAPL_PI, MAPL_PI_R8,MAPL_DEGREES_TO_RADIANS
+  !use MAPL_Constants, only: MAPL_PI, MAPL_PI_R8,MAPL_DEGREES_TO_RADIANS
```
mathomp4, line 13: I suppose this comment can be deleted?

base/ESMFL_Mod.F90
```diff
@@ -1262,9 +1261,8 @@ subroutine ESMFL_RegridStore (srcFLD, SRCgrid2D, dstFLD, DSTgrid2D, &
    !                           name="SRC 2D grid", rc=status)
    !_VERIFY(status)
    ! instead use the following...
-   pi = 4.0 * atan(1.0)
-   deltaX = 2.0*pi/gccpd(1)
-   deltaY = pi/(gccpd(2)-1) 
+   deltaX = 2.0*MAPL_PI/gccpd(1)
```
mathomp4, line 1264: @bena-nasa @atrayano  This seems like it would be non-zero-diff. Do you know who cares about π in ESMFL?

base/MAPL_TilingRegridder.F90
```diff
@@ -8,8 +8,9 @@ module MAPL_TilingRegridderMod
    use MAPL_RegridderSpec
    use MAPL_RegridMethods
    use MAPL_DirPathMod
-   use MAPL_BaseMod, only: MAPL_UNDEF, MAPL_TileNameLength
+   use MAPL_BaseMod, only: MAPL_UNDEF
```
mathomp4, line 11: Hmm. Does this need to be here if the code does `use MAPL_Constants`?

Gvilla1000-nasa: Yes. Because right now `MAPL_UNDEF` is being used from `MAPL_BaseMod` and not MAPL_Constants. This is because certain GridComps use this same line and I was getting conflicts in MAPL_UNDEF usage.

Gvilla1000-nasa: You'll notice that inside MAPL_Constants, MAPL_UNDEF is renamed to Disabled (a new commit will rename it to DisabledVar since Disabled is too general of a name and it is in use elsewhere. This is the cause of the gnu CI build failure).

shared/Constants/Constants.F90
```diff
@@ -0,0 +1,21 @@
+module MAPL_Constants
+
+use, intrinsic :: iso_fortran_env, only: REAL64, REAL32
+use MAPL_InternalConstantsMod, disabled => MAPL_UNDEF !For temporary backward compatibility
```
mathomp4, line 4: @tclune is `disabled` a new Fortran thing? Not sure I've seen it before...

shared/Constants/PhysicalConstants.F90
```diff
@@ -0,0 +1,80 @@
...
+
+!=============================================================================
+!BOP
+
+! !MODULE: -- A container module for MAPL physical constants
+
+! !PUBLIC VARIABLES:
+
+   ! Universal Constants
+#if defined(CODATA_CONSTANTS)
```
mathomp4, line 15: oooh! 😄 

@tclune Should we add a CMake `option()` for enabling this? Default to `OFF` of course?

shared/Constants/Constants.F90
mathomp4: Ahh. @Gvilla1000-nasa explained this...

base/MAPL_TilingRegridder.F90
mathomp4: Ahh. I thought it was a new fancy Fortran thing where you can `disable` a constant 😄

shared/Constants/Constants.F90
Gvilla1000-nasa: it is not. This is just a proxy rename to avoid conflicts between MAPL_UNDEF from BaseMod and MAPL_UNDEF from MAPL_Constant. I had to do it this way because several grid components external to MAPL still use MAPL_UNDEF from BaseMod and not doing this caused conflicts.

base/Base/Base_implementation.F90
tclune: > > > Query: In file `MAPL_cfio/ESMF_CFIOUtilMod.F90` the following parameter can be found:
> > > ` PARAMETER (IGREG=15+31*(10+12*1582))` and
> > > `PARAMETER (IGREG=2299161)`
> > > These are the first day of the gregorian calendar and the julian day count of the initial gregorian calendar respectively. Thoughts on adding these to either InternalConstants.F90 or MathConstants.F90? Their usage is limited but I've seen some components using these.
> > 
> > 
> > Can you summarize where these are used (outside of MAPL_cfio)? The model itself should not be depending on such things I would hope.
> 
> Sure.
> I found usage of iGreg (defined every time) in:
> `Components/@GEOSgcm_GridComp/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM_GEOS5PlugMod/@mom/src/mom5/ocean_param/gotm-4.0/util/time.F90`
> 
> `Shared/MAPL/MAPL_cfio/ESMF_CFIOUtilMod.F90`
> 
> `Shared/MAPL/base/MAPL_NominalOrbitsMod.F90`
> 
> `Shared/GMAO_Shared/GMAO_etc/zeit_jday.f`
> 
> and
> 
> `Shared/GMAO_Shared/GMAO_gfio/timesubs.f`

The mom reference is odd, as mom does not use MAPL.   But the others are enough to mean that we can not get rid of them.  How does CFIO export both of these, as they have the same name?

tclune:
> Query: There are still some `MAPL_` variables lying around in Base.F90, Comms.F90, ExtDataGridCompMod.F90 among others (e.g. `MAPL_StateItemOrderList`, MAPL_ExtDataVectorItem, MAPL_Distribute, MAPL_SunAutumnalEquinox). These are mostly kept within their own modules but because of their `MAPL_` prefix, these could be moved into `InternalConstantsMod.F90`. Thoughts?

Unless these are used in multiple files, I think it is better to leave them where they are.  The consolidation is to make it easier to find things, but if they are truly internal like that, we've not helped by moving them.

base/ESMFL_Mod.F90
tclune: Agreed, but ... if we are lucky, and they are zero diff, then we should make the switch.

shared/Constants/PhysicalConstants.F90
tclune: Why is the conditional needed?    Seems like should just always provide these.   Other modules can use `ONLY` to exclude if they have their own.

shared/Constants/Constants.F90
tclune: I don't think this solution helps us much.  If we can't get MAPL_UNDEF from MAPL_Constants, it defeats the purpose of putting it there.   Not saying I have a better solution though.

Python/MAPL/constants.py
Gvilla1000-nasa: Done. Problems of copy paste lol. Thanks for catching this.

base/Base/Base_implementation.F90
Gvilla1000-nasa: Deleted

Gvilla1000-nasa: @tclune They are declared in different subroutines so they don't conflict with one another. What I was thinking was of adding two new variables to Constants `IGREG` or fully spelt out `INITIAL_GREGORIAN` and `JIGREG` or fully spelt out `JULIAN_INITIAL_GREGORIAN` if you think it is worth it so that other repos like GMAO_Shared could simply use our definitions from MAPL_Constant rather than them having to define it every single time. But if its not worth it as usage is not common, then this can be left as is.

base/ESMFL_Mod.F90
Gvilla1000-nasa: Based on this comment, seems this code was not tested in the run I performed since I got zero diff. Is there a suggestion as to what I can do to test if this change truly is zero diff?

shared/Constants/Constants.F90
Gvilla1000-nasa: Based on the meetings we had, couldn't this change be done when MAPL moves to the next major revision at which point users will have to change their code to use MAPL_Constants rather than BaseMod? This was my motivation for doing it this way. Paving the road for that point in time.

mathomp4:
Tested with NAG and this branch seemed to build. Huzzah!

base/Base/Base_implementation.F90
tclune: Ok - those IGREG are not used anywhere then - they are internal to the subroutines.   The other layers you pointed out must define their own values.

`MAPL_cfio` is  deprecated (mostly replaced by pfio, but not quite all the strings have been cut yet), so no point in fixing it.

shared/Constants/Constants.F90
tclune: I'd rather just not export it in Constants at all than under a different name.  

@mathomp4 are the external uses of BaseMod all in DAS?  I thought we did a sweep a while back to make those all MAPL, but maybe this was just the GCM or just the targets in cmake?

Not really liking any of our options here.

shared/Constants/PhysicalConstants.F90
mathomp4: This is related to #42 which is to update the constants to CODATA 2018 values. So, you'd change the Stefan-Boltzmann constant, for example. This is an instant non-zero-diff of course so the thought is we can at least have the new constants here, make them non-default for now, then make the switch and eventually remove the old ones.

tclune: Ah - I did not recognize the name, so thought it was some obscure GEOS option.  (Well, I suppose it now will be.)   

Don't put the cmake option in the ESMA-cmake layer though.  Keep it all in MAPL.

mathomp4: Yeah. I'll work with @Gvilla1000-nasa on this.

Gvilla1000-nasa: latest push addresses this.

shared/Constants/Constants.F90
Gvilla1000-nasa: I did a quick search for uses of `MAPL_UNDEF` from BaseMod and found the following results:
```
Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/GMIchem_GridComp/GMI_GridComp/GmiChemistry/photolysis/CloudJ/CLD_SUB_MOD.F90
24:      use MAPL_BaseMod, only: MAPL_UNDEF

Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/GMIchem_GridComp/GMI_GridComp/GmiChemistry/photolysis/CloudJ/FJX_SUB_MOD.F90
29:      use MAPL_BaseMod,      only: MAPL_UNDEF

Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/GMIchem_GridComp/GMI_GridComp/GmiChemistry/photolysis/fast_JX65/fastJX65_mod.F90
7:      use MAPL_BaseMod, only: MAPL_UNDEF

Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/cloudnew.F90
23:   use MAPL_BaseMod,      only: MAPL_UNDEF

Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/cldmacro.F90
23:   use MAPL_BaseMod,      only: MAPL_UNDEF

Shared/@GMAO_Shared/GEOS_Util/post/ec_eta2fv.F
408:      use MAPL_BaseMod, only: MAPL_UNDEF

Shared/@GMAO_Shared/GEOS_Util/post/ec_prs2eta.F
1761:      use MAPL_BaseMod, only: MAPL_UNDEF

Shared/@GMAO_Shared/GEOS_Util/post/gg2fv.F
544:      use MAPL_BaseMod, only: MAPL_UNDEF
990:      use MAPL_BaseMod, only: MAPL_UNDEF

Shared/@GMAO_Shared/GEOS_Util/post/gg2eta.F
236:      use MAPL_BaseMod, only: MAPL_UNDEF

Shared/@GMAO_Shared/GEOS_Util/post/mpi_util.F
716:      use MAPL_BaseMod, only: MAPL_UNDEF
841:      use MAPL_BaseMod, only: MAPL_UNDEF

Shared/@GMAO_Shared/GEOS_Util/post/era5_prs2eta.F
1829:      use MAPL_BaseMod, only: MAPL_UNDEF
```

However, curiously enough, when I removed the `MAPL_UNDEF` declaration from Base.F90 and made it so that the MAPL code used `MAPL_UNDEF` from `MAPL_Constants` rather than `MAPL_BaseMod`, the compilation didn't fail at any of them. It did fail for  `src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/@FVdycoreCubed_GridComp/AppGridCreate.F90` because it couldn't find `pi` from `MAPL_ConstantsMod` which I find odd because why did changing MAPL_UNDEF impact pi?

tclune: I'm almost tempted to update GMAO_shared to use MAPL_Constants now, but there is a risk that the GCM will need an update there for other reasons and then need the newer MAPL and ...

It might be worth double checking things with the FV "pi mystery".     Maybe some module did not correctly recompile for reasons that I cannot see.    Or maybe Base is also exporting Pi for reasons that are not immediately obvious?

Make clean would be the safest thing to try.   Be sure to read any error messages carefully they might not quite indicate what you think ...

Gvilla1000-nasa: My apologies for the delay. So this was a doozy to fix but the fix turned out to be very simple but perplexing to me. The FV "pi mystery" was fixed by changing `use MAPL_ConstantsMod` to `use MAPL_ConstantsMod, only: MAPL_PI_R8` inside `base/MAPL_Mod.F90`. Not sure why this fix was needed only when `MAPL_UNDEF` was removed from `MAPL_BaseMod` though. After making this fix, no other compilation errors were reported.  I re-compiled GEOSgcm with intel and gnu in both debug and release (making clean everything each time) and no issues were seen. Running a quick intel debug run still showed zero diff to previous runs. Should I commit the removal of `MAPL_UNDEF` from `BaseMod`? I'm concerned about what sort of effects it may cause at runtime or compilation of other components that were not tested since there are still a lot of `MAPL_UNDEF` from `BaseMod` uses in GridComp and GMAO_Shared.

tclune: My only strict requirement for the current PR is that it is provably 0 diff for GCM and, as much as we can test, the DAS.  (The latter is probably just source inspection for now unfortunately.)

But ... I want to go ahead and start a MAPL3 branch (name TBD) where we will start putting our desired changes that break backward compatibility.   (@mathomp4 - you might want to consider this in terms of the CODATA constants.)

Gvilla1000-nasa: Ok. Then as things stand now, it is 0 diff (without removing `MAPL_UNDEF` from `BaseMod`). The question I have is should I revert the addition of `MAPL_UNDEF` to `MAPL_Constants` since I pretty much hacked it to be disabled through the use of the `disabledvar => MAPL_UNDEF` rename? The change proposed in the last comment can be left to the new branch since a much better fix would be for the FV to use the `mapl.constants` library.

tclune: Yes - revert.  Save that for the 3.0 branch.

base/ESMFL_Mod.F90
mathomp4: As I said at the meeting, I think we are safe. A grep of the full GEOSadas shows that no one actually seems to call `ESMFL_RegridStore`:
```
GMAO_Shared/MAPL_Base/ESMFL_Mod.F90
6: ESMFL_RegridStore
55:  public ESMFL_RegridStore ! only used for regridding using ESMF_FieldRegrid
1062:! !IROUTINE: ESMFL_RegridStore
1066:   subroutine ESMFL_RegridStore (srcFLD, SRCgrid2D, dstFLD, DSTgrid2D, &
1398:   end subroutine ESMFL_RegridStore
```

Now, maybe some odd one-off repo (a la AeroApps or GAAS or something) does, but my guess is they are far enough behind MAPL develop we can just blame any non-zero-diff on them moving up in time.

base/MAPL_NominalOrbitsMod.F90
```diff
@@ -625,7 +621,7 @@ SUBROUTINE find_sweeppoints(iii, latwayp, longwayp, l, r, wrapon, latlonshift)
                CALL linspace(r,l,iii,myvec)
        end if
        do i=1, size(myvec)
-          myvec_deg(i) =  ckm2deg * myvec(i)
+          myvec_deg(i) =  MAPL_KM2DEG * myvec(i)
```
tclune, line 624: As long as we are giving a new name here,  I would like to avoid the "cute" use of "2" to represent English "to".   How about  `MAPL_DEG_PER_KM`?

tclune: @Gvilla1000-nasa if you could do a sweep of such things.   (But only where we are already changing the name to have the MAPL prefix anyway.)

tclune: Hmm.   Looking at other examples, I guess "PER" is not common.    So a backup would be  `MAPL_KM_TO_DEG` though I still think "PER" is clearer.

tclune (changes requested):
One small item flagged.

base/MAPL_NominalOrbitsMod.F90
Gvilla1000-nasa: Sounds good. I'll address in next push. I also noticed that I didn't remove `MAPL_UNDEF` from the Python Constants so I'll also take care of that.

tclune (approved):

PR #969, opened 2021-08-16
specified npes for pfio_MAPL_demo for default interface

weiyuan-jiang:
pfio_MAPL_demo.x would not run without specifying --npes_model. This PR fixes the issue.
## Types of changes

## Checklist:


weiyuan-jiang:
@JulesKouatchou , This fixes the problem that mpiservber would not run without --npes_model

atrayano (approved):
Looks OK. I would have preferred the call to MPI_Comm_size to be inside the following if block, and also not to mention MPI_COMM_WORLD in several places. But this is only a demo

Issue #961, opened 2021-08-13
Multi-day GEOS Runs are non-zero-diff with latest MAPL develop


mathomp4:
So, this is interesting. Last night, my tests showed that SCM with MAPL `develop` (1726321964680cdfdc43eed418f502937076b649) is now non-zero-diff with MAPL v2.8.0. Odd. I did some more tests and found something interesting. The model itself is non-zero-diff as well, but you have to run more than one day. Indeed, it seems you have to run a while. I set up some 10-day runs (I was going to do 1 week but my makeoneday script has a 10day option) and:
```
$ cmphistory stock-gcm-2021Aug13-1week-c24/ mapldev-gcm-2021Aug13-1week-c24/
Comparing geosgcm_buda.20000414_2230z...
Success!
Comparing geosgcm_buda.20000415_0130z...
Success!
Comparing geosgcm_buda.20000415_0430z...
Success!
Comparing geosgcm_buda.20000415_0730z...
Success!
...
Comparing geosgcm_buda.20000419_2230z...
Success!
Comparing geosgcm_buda.20000420_0130z...
Failure!
Comparing geosgcm_buda.20000420_0430z...
Failure!
...
```
or with our friend prog:
```
Comparing geosgcm_prog.20000415_0000z...
Success!
Comparing geosgcm_prog.20000415_0600z...
Success!
Comparing geosgcm_prog.20000415_1200z...
Success!
Comparing geosgcm_prog.20000415_1800z...
Success!
...
Comparing geosgcm_prog.20000419_1800z...
Success!
Comparing geosgcm_prog.20000420_0000z...
Success!
Comparing geosgcm_prog.20000420_0600z...
Failure!
Comparing geosgcm_prog.20000420_1200z...
Failure!
Comparing geosgcm_prog.20000420_1800z...
Failure!
...
```

Now, to check, I also did some experiments with the previous MAPL `develop` (6bf423d3ff4c9dee2a3e3e2cb908ccf5b82b01e5) and it is zero-diff. 

So something in https://github.com/GEOS-ESM/MAPL/compare/6bf423d3ff4c9dee2a3e3e2cb908ccf5b82b01e5...1726321964680cdfdc43eed418f502937076b649 seems to cause the model to be non-zero-diff after ... 5, 6 days?? And the only change in there that should affect GEOS is Constants refactor.

[...]

PR #973, opened 2021-08-17
Fix to multi day run

Gvilla1000-nasa:
## Description
This PR fixes the Multi-Day GEOS non-zero diff bug. Bug appeared because `base/MAPL_sun_uc.F90` is the only module in MAPL that uses `MAPL_DEGREES_TO_RADIANS` in single precision. Rather than creating a `MAPL_DEGREES_TO_RADIANS_R4`, I went with changing the double precision instances to `MAPL_DEGREES_TO_RADIANS_R8` as there is precedence for other double precision variables. However, I didn't create a `MAPL_RADIANS_TO_DEGREES_R8` because compilation issues would have appeared since this module `Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/GEOS_MoistGridComp.F90` would still use the old name.

If there are any other recommendations, let me know.

## Related Issue

#961 

## Motivation and Context
Fixes bug that appeared from Constants refactoring

## How Has This Been Tested?
I ran the GEOS-5 model past 5 days and 5hours which was the threshold identified where non-zero diffs started appearing and it was finally zero-diff.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add option to flip native level output in History relative to input
 - Added `MAPL_AllocNodeArray_6DR8` and `MAPL_DeAllocNodeArray_6DR8` to Shmem
 - Refactors Constants into its own library and consolidated mathematical/physical constants used throughout code to use those from library
+- Added single precision Degrees to Radian Conversion
 
 ### Changed
 - Simplified implementation of MAPL_FieldCopyAttributes
--- a/base/Base/Base_implementation.F90
+++ b/base/Base/Base_implementation.F90
@@ -2206,10 +2206,10 @@ contains
 
     !  Compute the coordinates (the corner/center is for backward compatibility)
     !  -------------------------------------------------------------------------
-    deltaX      = MAPL_DEGREES_TO_RADIANS * DelLon_
-    deltaY      = MAPL_DEGREES_TO_RADIANS * DelLat_
-    minCoord(1) = MAPL_DEGREES_TO_RADIANS * BegLon_ - deltaX/2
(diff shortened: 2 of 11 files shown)


tclune (approved):
The next question is whether you should follow up with a PR onto the release/MAPL-v3 branch which eliminates the `_R8`.   Lot of trouble for minimal benefit.

mathomp4:
Oooh. Very nice spot, @Gvilla1000-nasa!

And, yeah, I wonder if perhaps we should instead make a `MAPL_DEGREES_TO_RADIANS_REAL32` as it's the *one* place that needs it? Rather than rename the R8 ones?

mathomp4:
Also, for consistency, should we have something similar for `MAPL_RADIANS_TO_DEGREES`?

mathomp4:
> Also, for consistency, should we have something similar for `MAPL_RADIANS_TO_DEGREES`?

Ahh. I see what you mean about Moist! Never mind. Me dumb :)

Gvilla1000-nasa:
@mathomp4 that was my initial thought to be honest...but went with the way of consistency. 2 file change vs 11 right now. But if that is the preferred method (which would also eliminate the need of doing a new PR to the MAPL-3 branch), I can go ahead and do it.

mathomp4:
> @mathomp4 that was my initial thought to be honest...but went with the way of consistency. 2 file change vs 11 right now. But if that is the preferred method (which would also eliminate the need of doing a new PR to the MAPL-3 branch), I can go ahead and do it.
> 
> EDIT: I am referring to creating a `MAPL_DEGREES_TO_RADIANS_REAL32` and `MAPL_RADIANS_TO_DEGREES_REAL32`

Yeah. I guess it's an @tclune decision (passing the buck... 😄 ). Hmm...

mathomp4:
I just did a test with this branch and turning on Replay and ran for 10 days and it too is zero-diff. Huzzah!

PR #1006, opened 2021-08-30
Fixes #545. Adds default build type to MAPL

### LINKED ISSUE #545: MAPL CMake build broken with GNU compilers with no build type specified
This was a fun one found by @WilliamJamieson. To wit, if you grab MAPL and do:
```
cmake .. -DBASEDIR=$BASEDIR/Darwin
```
you can't build with GNU. Why? Well, we have bits of MAPL that depend on `-ffree-line-length-none`.

The reason? Well the "default" build type for CMake is `RelWithDebInfo`...and we don't supply flags for `RelWithDebInfo` in our ESMA_cmake. 

I never saw this because I use aliases that always do `-DCMAKE_BUILD_TYPE=Release` or `-DCMAKE_BUILD_TYPE=Debug`.

We don't see this in GEOSgcm (etc.) because we have:
```cmake
# Set the default build type to release
if (NOT CMAKE_BUILD_TYPE)
  message (STATUS "Setting build type to 'Release' as none was specified.")
  set (CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
  # Set the possible values of build type for cmake-gui
  set_property (CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
    "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
endif ()
```
in our top-level `CMakeLists.txt` file.

I suppose this question is now for @tclune: What should we do? Do we want to have a chunk like this in our MAPL `CMakeLists.txt`? Or should I extend all our ESMA_cmake bits to include `RelWithDebInfo`? Both?

mathomp4:
## Description

This adds a default build type to MAPL. As GEOS really only supports `Release`, `Debug` and `Aggressive`, those are what we list here.

## Related Issue

Closes #545 

## Motivation and Context

This let's someone do:
```
cmake ..
```
in a build directory for MAPL and it'll work on GNU. It worked with Intel only because of differences in compiler and line lengths and all.

## How Has This Been Tested?

Did a test of MAPL standalone with GNU and it worked.

Built GEOSgcm on discover with this as well and it worked.

I asked @weiyuan-jiang to test with UFS and it built there.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Added default `CMAKE_BUILD_TYPE` for MAPL standalone. Defaults to `Release` build if not set on command line
+
 ## [2.8.4] - 2021-08-27
 
 ### Added
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,15 @@ project (
   VERSION 2.8.4
   LANGUAGES Fortran CXX C)  # Note - CXX is required for ESMF
 
+# Set the default build type to release
+if (NOT CMAKE_BUILD_TYPE)
+  message (STATUS "Setting build type to 'Release' as none was specified.")
+  set (CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
+  # Set the possible values of build type for cmake-gui
+  set_property (CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+    "Debug" "Release" "Aggressive")
+endif ()
+
 # mepo can now clone subrepos in three styles
 set (ESMA_CMAKE_DIRS
   ESMA_cmake


mathomp4:
I suppose I should ping @LiamBindle and @lizziel. Do you usually build MAPL with weird `CMAKE_BUILD_TYPE`?

LiamBindle:
We don't. We use the standard Release, RelWithDebInfo, and Debug. 

I haven't seen Aggresive before. How is it different from Release?

tclune (approved):

mathomp4:
> We don't. We use the standard Release, RelWithDebInfo, and Debug.
> 
> I haven't seen Aggresive before. How is it different from Release?

Hmm. You use RelWithDebInfo? Interesting. I don't think this PR should break that...

As for Aggressive, it's something I came up with to handle "more aggressive" flags. Ones that are "speed for speed's sake". It breaks layout regression (4x24 doesn't have the same answer as 6x36) but it's faster!

lizziel:
Looks like it's not in our `ESMA_cmake` since our version is old (3.0.6). "Aggressive" in the latest 3.5.2 so we'll get it when we eventually update libraries in GCHP.

LiamBindle:
I see, thanks!

Issue #486, opened 2020-07-07
MAPL Crashes when asked to write an empty checkpoint: What to do?


mathomp4:
This is a fun one. In working on the new NCAR GWD code, it was found that we needed to write out the gwd import state (for the first time in a long time) to preserve start-stop regression. @atrayano helped me add that in an elegant way.

I then started looking closer and realized that the *NCEP* GWD code that @wmputman sometimes runs has more bits in the GWD Import State (which would be captured). BUT! it also adds a GWD **Internal** State which only exists in its case.

So, I did an experiment, what if I'm *not* running NCEP GWD, but ask MAPL to write out the gwd_internal_checkpoint. (With New NCAR GWD we are saved in that no matter what, the import restart has a field). It's a checkpoint where no AddInternalSpec was called. Answer: Bit of a crash! See here a traceback from a GCC 10 Debug run (I had that build around).

```
pe=00000 FAIL at line=05227    MAPL_Generic.F90                         <status=57>
pe=00000 FAIL at line=02144    MAPL_Generic.F90                         <status=57>
pe=00000 FAIL at line=01840    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=02072    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=01840    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=02072    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=01840    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=02072    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=01840    MAPL_Generic.F90                         <needs informative message>
pe=00000 FAIL at line=00794    MAPL_CapGridComp.F90                     <status=1>
```

NOTE 1: This is against MAPL 2.1.5 not develop.
NOTE 2: My guess is this might be what would happen if we also, say, asked for a checkpoint of a state where all the fields were RestartSkip. 

I suppose my question is how to best handle this. Should MAPL support writing out 0-length/variable restarts? Maybe...? I mean, it would allow flexibility that if you turn on the NCEP GWD, say, you have a chance at preserving the state. But then in a non-NCEP GWD case, you'd have to deal with "writing" and "reading" a zero-variable restart...which seems weird (and might not be possible with netCDF).

[...]

PR #1010, opened 2021-09-02
Bugfix/atrayano/#486 no read or write empty

atrayano:
Added protection to not read nor write restarts for empty states. Also printing an appropriate warning in that case.

## Related Issue

 #486 

## Motivation and Context
 Fixes issue #486. 

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_Generic.F90
+++ b/base/MAPL_Generic.F90
@@ -5371,6 +5371,20 @@ end function MAPL_AddChildFromDSO
     !real(kind=ESMF_KIND_R8),save ::  total_time = 0.0d0
     !logical                               :: amIRoot
     !type (ESMF_VM)                        :: vm
+    logical :: empty
+    class(Logger), pointer :: lgr
+
+! Check if state is empty. If "yes", simply return
+    empty = MAPL_IsStateEmpty(state, __RC__)
+    if (empty) then
+       if (MAPL_Am_I_Root()) then
+          call MAPL_GetLogger(mpl, lgr, __RC__)
+          call lgr%warning('Checkpoint '//trim(filename) //&
+               ' requested, but state is empty. Ignored...')
+       end if
+
+       _RETURN(ESMF_SUCCESS)
+    end if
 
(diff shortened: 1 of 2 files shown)


base/MAPL_Generic.F90
```diff
@@ -5371,6 +5371,20 @@ subroutine MAPL_ESMFStateWriteToFile(STATE,CLOCK,FILENAME,FILETYPE,MPL,HDR, oCli
...
+    empty = MAPL_IsStateEmpty(state, __RC__)
+    if (empty) then
+       if (MAPL_Am_I_Root()) then
+          call MAPL_GetLogger(mpl, lgr, __RC__)
+          call lgr%warning('Checkpoint '//trim(filename) //&
+               ' requested, but state is empty. Ignored...')
+       end if
+
+       _RETURN(ESMF_SUCCESS)
+    end if
```
tclune, line 5387: You should introduce a procedure `handle_empty()` that takes one argument `empty`.    This eliminates the duplication. Were it not for the multiple error possibilities, I'd say:
```f90
if (MAPL_isStateEmpty(state)) call handle_empty()
```
instead you'll need to split it into two statements:
```f90
empty = ...
if (empty) then
    call handle_empty(__RC__)
    _RETURN(ESMF_SUCCESS)
end if
```
A case where lack of exceptions makes this much more verbose than it should be.  Sih.

atrayano: A procedure cannot save us - the point is to return from the caller (and log a warning)

atrayano: We simply do not want to execute the rest of this subroutine. I cannot see how you do it by calling a procedure. Only way to save typing would be a macro, but I saw little value in that

tclune: or maybe
```f90
empty = MAPL_isStateEmpty(state, __RC__)
if (empty) __RETURN(status)
```

tclune (approved):

PR #1021, opened 2021-09-09
Apply shave bit to tile space for histroy output

weiyuan-jiang:
When a history collection in tile space is binary, the "nbits" is not used. The PR adds that function. It will be zero-diff for test but will be non-zero diff for binary history output.
## Types of changes

## Checklist:

### DIFF
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -3629,31 +3629,9 @@ ENDDO PARSER
                        list(n)%descr, intstate%output_grids,rc )
                   INTSTATE%LCTL(n) = .false.
                endif
-              
-               nbits: block
-                 integer :: fieldRank
-                 type(ESMF_Field) :: field
-                 real, pointer :: ptr1d(:), ptr2d(:,:), ptr3d(:,:,:), ptr4d(:,:,:,:)
-                 if (list(n)%nbits <24) then
-                   do m=1,list(n)%field_set%nfields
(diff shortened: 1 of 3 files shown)


bena-nasa:
Could you make this a subroutine call in the gridcomp rather than just a new block of code in the run routine? Maybe just a subroutine that takes a state and does the bit shaving, that seems cleaner.

weiyuan-jiang:
good suggestion. The block has been moved to a subroutine @bena-nasa

bena-nasa (approved):

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -38,7 +38,7 @@ module MAPL_HistoryGridCompMod
   use mapl_RegridMethods
   use MAPL_newCFIOitemVectorMod
   use MAPL_newCFIOitemMod
-  use pFIO_ClientManagerMod, only: o_Clients
+  use pFIO
```
mathomp4, line 41: Just a question: do we need to use all of pFIO here? Or just one routine a la:
```fortran
use pFIO, only: shavebits
````

I'm not sure what @tclune thinks of this. I don't see anything in the [Style Guide](https://github.com/GEOS-ESM/MAPL/wiki/MAPL-Coding-standards) for how to handle use/use-only stuff...

mathomp4: Oh I'm an idiot. it's in the file. Time to end the day! :)

weiyuan-jiang: I am making changes to use smallest set of routines from pfio

mathomp4 (approved):
Approved

Issue #942, opened 2021-08-02
Feature request: configurable lev positive attribute in checkpoint files 


lizziel:
Currently MAPL reads and writes all restart and checkpoint files assuming vertical dimension direction `down`. This is indicated by the `positive` attribute included for the `lev` dimension in the file. For restart read this behavior is true regardless of whether the `positive` attribute is included in the file.

This feature request is to expand the vertical direction capability of checkpoint files, both upon read and write, to increase flexibility of how the data is stored externally. This would be useful to GCHP which uses a vertical direction `up` convention internally in GEOS-Chem state arrays and therefore has users more familiar with `up` than `down` files. 

To be clear, the internal state arrays would remain `positive:down` within MAPL. It would be the responsibility of grid comps, e.g. in `Chem_GridCompMod.F90` in GCHP, to make appropriate internal data flips to fill the internal state.

I envision the update would include adding an optional setting in the main model config file to set the checkpoint lev positive attribute, e.g. in `GCHP.rc` for GCHP which is the equivalent to `AGCM.rc` in GEOS. Ideally this would be done per gridcomp rather than one setting for all. For example, users would be able to write `positive:up` restart files for GCHP but write a `positive:down` checkpoint for FV3 (if they also implement an internal state for that grid comp - currently we don't have one). Exclusion of the setting for a grid comp could be interpreted as `positive:down` so that no changes to GEOS files are necessary to retain current behavior.

The file content would then look something like this in GCHP:

> GCHPchem_INTERNAL_CHECKPOINT_FILE:  gcchem_internal_checkpoint
> GCHPchem_INTERNAL_CHECKPOINT_TYPE:  pnc4
> GCHPchem_INTERNAL_CHECKPOINT_POSITIVE:  up

The config file setting is for the writing of checkpoint files. The behavior for MAPL restart read would be different in that it would base behavior on the metadata in the input file. MAPL would vertically flip the input arrays if `lev:positive` in the file is `up`, and do nothing if it is `down`. If the file does not contain the attribute then it should fail since `positive` should always be there for COARDS-compliance if lev is not pressure (lev can never be pressure for restarts, correct?).

This request is an off-shoot of https://github.com/GEOS-ESM/MAPL/issues/914. I am pinging a couple other GCHP developers so they are in the loop on this request. @sdeastham, @LiamBindle

bena-nasa:
@lizziel I've implemented this feature as requested in the issue and it is in our develop branch so should be in the next MAPL release issued.

PR #1029, opened 2021-09-13
Feature/bmauer/#942 checkpoint positive handling

bena-nasa:
Implement features requested in issue #942 

## Types of changes

## Checklist:


base/MAPL_IO.F90
```diff
@@ -7370,8 +7372,17 @@ subroutine MAPL_BundleReadNCPar(Bundle, arrdes, filename, rc)
        _VERIFY(status)
        grid_file_match=compare_grid_file(metadata,grid,rc=status)
        _VERIFY(status)
+       flip = check_flip(metadata,rc=status)
```
tclune, line 7375: No need to change now,  but please do use these types of opportunities to introduce the `_RC` macro.

base/MAPL_IO.F90
```diff
@@ -9291,4 +9331,148 @@ function get_fname_by_face(fname, face) result(name)
...
+                  select type(vpos)
+                  type is (character(*))
+                     positive => vpos
+                  class default
+                     _ASSERT(.false.,'units must be string')
+                  end select
+               else
+                  positive => null()
+               end if
+               if (positive == "up") then
```
tclune, line 9369: This line will fail if positive is `null()` from the `else` block just above.

base/MAPL_IO.F90
```diff
@@ -9291,4 +9331,148 @@ function get_fname_by_face(fname, face) result(name)
...
+      real(KIND=ESMF_KIND_R4), pointer :: ptr_r4(:,:,:)
+      real(KIND=ESMF_KIND_R8), pointer :: ptr_r8(:,:,:)
+      real(KIND=ESMF_KIND_R4), allocatable :: alloc_r4(:,:,:)
+      real(KIND=ESMF_KIND_R8), allocatable :: alloc_r8(:,:,:)
+      type(ESMF_TypeKind_Flag) :: tk
+      integer :: vloc,i,lb,ub,ii
+     
+      call ESMF_FieldGet(field,rank=rank,typeKind=tk,rc=status)
+      _VERIFY(status)
+      if (rank==3) then
```
tclune, line 9394: Might this be clearer with just an early return if the rank is not 3?

base/MAPL_IO.F90
```diff
@@ -9291,4 +9331,148 @@ function get_fname_by_face(fname, face) result(name)
...
+               call ESMF_FieldGet(field,farrayPtr=ptr_r4,rc=status)
+               _VERIFY(status)
+               allocate(alloc_r4,source=ptr_r4)
+               lb = lbound(ptr_r4,dim=3)
+               ub = ubound(ptr_r4,dim=3)
+               ii=0
+               do i=lb,ub
+                  ptr_r4(:,:,i)=alloc_r4(:,:,ub-ii)
+                  ii=ii+1
+               enddo
```
tclune, line 9408: Someday we'll have templates and won't have to duplicate code for r4/r8 each time we turn around.  (Just venting.)

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added option to flip vertical orientation of checkpoint files
```
tclune, line 14: You might want a bit more documentation here so that users that want to use this feature know how to use it.

tclune (approved):

Issue #1020, opened 2021-09-08
Refactor ExtData modules from Base


Gvilla1000-nasa:
Now that ExtDataGridCompMod will be moved into its own subdirectory inside gridcomps, it is time to refactor other ExtData modules into it.

PR #1031, opened 2021-09-14
Refactors remaining ExtData Modules

Gvilla1000-nasa:
## Description

Refactors remaining ExtData modules from base. Due to dependencies from other modules, this PR also addresses newcfio modules by renaming them and moving them to a different directory. Additionally, ExtDataCollection has been moved with what was known as newcfio and renamed to removed the Ext prefix.

Only things to note is if the selected names are appropriate and the placement of Regrid_Util.F90. It can't go in base anymore due to its dependency on what was known as newcfio.

~~The first commit of this PR is the most basic change in which all the required modules that needed to be moved due to the ExtData refactoring were moved to the same directory.~~

~~The second commit cleans and improves this refactoring by moving some modules to what I considered more appropriate places. That being said, this PR is submitted as a draft as I want the team's opinion on where to move some of these and any other possible changes deemed necessary.~~

~~Things to note and pay attention to:~~
~~1. I moved `base/ESMF_CFIOPtrVectorMod.F90` to `MAPL_cfio/` because all other files in this directory are ESMF_CFIO related. I tried renaming this directory to `ESMF_cfio` but external components like GMAO_hermes explicitly use the the library dependency of `MAPL_cfio_r8` rather than `MAPL`. Due to this, `MAPL_CFIO.F90` and `CFIOCollection.F90` were moved from `base` to a new directory `MAPL_ESMFcfio` which I would have rather called `MAPL_cfio`.~~

~~2. `MAPL_newCFIO.F90` and `MAPL_newCFIOitem.F90` are inside `gridcomps/ExtData` and not `MAPL_ESMFcfio` or other cfio directory because these files depend on some of the ExtData modules and some ExtData modules depend on some of the subroutines inside these newCFIO modules. Moving them to another directory would have caused circular dependency.~~

~~3. Created a new subdirectory `gridcomps/FieldBundle` to hold the FieldBundle Read/Write and the `MAPL_SimpleBundleMod.F90` Suggestions about where to place these modules are welcome.~~

## Related Issue

#1020 

## Motivation and Context

Refactoring of MAPL

## How Has This Been Tested?

Tested with compiling Intel/GNU in release/debug and running a simple GEOSgcm run.

## Types of changes

## Checklist:


Gvilla1000-nasa:
@tclune @atrayano @bena-nasa @mathomp4 I have created this PR as a draft due to some liberties I took during the refactoring in moving files and creating subdirectories. I can open it for review if preferred but I wanted to make sure it didn't merged before the PR got discussed.

bena-nasa:
MAPL_newCFIO.F90 and MAPL_newCFIO should not be in the ExtData directory you created. Those are also used by History. (How is this building? History, now depends on ExtData) They also do not belong in the MAPL_cfio directory, this is something I'm actively working to depreciate and eliminate
The historical context is that MAPL_CFIO.F90 was a set of higher level routines that make use what is in MAPL_cfio. to perform IO for ESMF objects like bundles, etc...
ExtData and History used this. A few years ago we took the first steps to eliminating MAPL_cfio and MAPL_CFIO.F90 by having ExtData and History go through  PFIO server that was implemented in PFIO. Of course that need some intermediate layer between the pfio server and ExtData/History which out of a complete lack of originality I called MAPL_newCFIO.
So the bottom line is MAPL_newCFIO.F90 belongs in neither History or ExtData or MAPL_cfio directories.

bena-nasa:
I would not mess with MAPL_cfio, this is a legacy library that I'm working to eliminate. It's not worth effort unless there is a good reason

tclune (commented):
I completely defer to @bena-nasa on these changes.    He is most aware of how these elements are currently planned for evolution/deprecation.

bena-nasa:
I'm not sure where anything related to the newCFIO layer, including the bundleread/write module should go. What the point of this layer is, is to be a sandwich layer for writing between ESMF fields/bundles which are data containers and the pfio layer that is the connection to the filesystem (and pfio does not and should know about ESMF, it is simply a generic io layer for fortran arrays).
Maybe just MAPL_ESMFio? @tclune @atrayano any thoughts about this?
I see you mentioned MAPL_SimpleBundle.F90, that is another legacy module that should not be grouped with what I have just referenced above.

Gvilla1000-nasa:
@bena-nasa. I've been working with:
 1. Reverting all CFIO changes back to base.
 2. Moving newCFIO modules away from ExtData.

So far it looks like I've been able to achieve this. Going to start looking at the FieldBundle and SimpleBundle changes I made. Quick question though, should the FieldBundleRead/Write be placed with the newcfio layer as well?

bena-nasa:
> @bena-nasa. I've been working with:
> 
> 1. Reverting all CFIO changes back to base.
> 2. Moving newCFIO modules away from ExtData.
> 
> So far it looks like I've been able to achieve this. Going to start looking at the FieldBundle and SimpleBundle changes I made. Quick question though, should the FieldBundleRead/Write be placed with the newcfio layer as well?

If you are referring to FieldBundleRead.F90 and FieldBundleWrite.F90, yes they should go with MAPL_newCFIO.F90

Gvilla1000-nasa:
Latest round of commits in summary does the following:
1. All ExtData modules are now in gridcomps/ExtData. However, two new subdirectories are now present inside ExtData (Collection and IOBundle). This was done because ExtData/IOBundle depends on newcfio and newcfio depends on ExtData/Collection. This was necessary so that newcfio modules could be moved away from ExtData and not be kept in base.
2. Moved newcfio and FieldBundleRead and FieldBundleWrite modules into new MAPL_newcfio directory

Gvilla1000-nasa:
image

This is a quick image showing the current dependency tree and the new/proposed one discussed offline. Would this be acceptable? The proposed dependency tree is only possible if the ExtData Collection gets moved to newcfio.

tclune:
Ah - thanks for making it easy to see!    Looks fine to me.    

Do we have a better name than "newcfio"?

bena-nasa:
Yes, that looks good. NewCFIO does need a better name. If anyone has any ideas, the idea is the same as the old cfio, server as a glue layer between ESMF containers and the library (in this case pfio) that knows how to talk to the file system and understands the SDF format we are using (netcdf in this case).

Gvilla1000-nasa:
> Ah - thanks for making it easy to see! Looks fine to me.
> 
> Do we have a better name than "newcfio"?

Based on @bena-nasa explanation and previous comments, it could be MAPL_ESMFcfio, or GenericCFIO. Now that ExtDataCollection is part of this, perhaps DataCFIO instead of newcfio and ExtDataCollection becomes simply DataCollection? Another variation of this last alternative, since all of this is for CF compliant data, would be CFDataio instead of newcfio and CFDataCollection instead of ExtDataCollection. These are my two cents.

tclune:
No strong opinion on my part, except I'd prefer to avoid the `MAPL_` prefix.    That would reverse the path I'm on to eliminate so many redundant `MAPL_` prefixes throughout the code.

bena-nasa:
We should not have CF in the name, our cubed-sphere files are not CF compliant. But then CFDataCollection becomes what? I'm almost wondering if the renamed files should just stay in base as the implementation in what is MAPL_newCFIO is closing tied to our grid factories for example.

Gvilla1000-nasa:
> We should not have CF in the name, our cubed-sphere files are not CF compliant. But then CFDataCollection becomes what? I'm almost wondering if the renamed files should just stay in base as the implementation in what is MAPL_newCFIO is closely tied to our grid factories for example.

ExtDataCollection could just become DataCollection or does that present any issues?

bena-nasa:
> > We should not have CF in the name, our cubed-sphere files are not CF compliant. But then CFDataCollection becomes what? I'm almost wondering if the renamed files should just stay in base as the implementation in what is MAPL_newCFIO is closely tied to our grid factories for example.
> 
> ExtDataCollection could just become DataCollection or does that present any issues?

I don't see DataCollection being an issue. I'm just not sure what to rename MAPL_newCFIO.F90, GriddedIO.F90? That's really what is for to write or read ESMF Gridded data where the grids are one of the grids we support via our grid factories

tclune:
But, isn't it true that the intent is to be CF compliant?       I'm not terribly upset by it only being an ideal rather than the reality.  E.g., if a single file CS grid format were to be published, we would immediately update that for future GEOS runs.   The backward compatibility option would not be compliant.

Gvilla1000-nasa:
The latest push does away with newcfio and ExtDataCollection. It renames them to griddedio and DataCollection respectively. That being said, if the preferred name is to cfdataio instead of griddedio and CFDataCollection instead of DataCollection, let me know and I can quickly make the change.

tclune (approved):

CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Moved newcfio modules from base into new griddedio directory
+    - Renamed newCFIO modules and routines to GriddedIO
+- Refactored ExtData modules. Because of the dependencies, the following changes were also done:
+    - Moved Collection ExtData modules into griddedio directory. Removed the Ext prefix for collection modules and subroutines and types
+    - Moved BundleRead and BundleWrite modules from base to MAPL_newcfio
```
Gvilla1000-nasa, line 23:
```suggestion
    - Moved BundleRead and BundleWrite modules from base to griddedio
```

Gvilla1000-nasa:
Sorry about that @tclune. Noticed a typo and was experimenting with commiting through here via suggestions and it seems to have dismissed your approval.

tclune (approved):
NP.  Just wish GitHub sent a notification that another review is needed in such situations.

PR #1036, opened 2021-09-16
Closes #1034. Cleans up redundant CircleCI code

### LINKED ISSUE #1034: CircleCI Config has a lot of Duplicated Code
The MAPL CircleCI config is a bit...much. A lot of the code is the same for GEOSgcm, GEOSldas, GEOSadas, etc. I think CircleCI has better ways to do this with reusable commands, etc.:

https://circleci.com/docs/2.0/reusing-config/#authoring-reusable-commands

Take a look at this, Matt.

mathomp4:
## Description

There was a lot of duplicated code in the CircleCI config. This tries to clean that up. It...sort of succeeds? I wonder what @tclune's feelings are...

## Related Issue

Closes #1034 

## Motivation and Context

There were large blocks of nigh-identical code in the CircleCI config. This makes commands that are then re-used.

## How Has This Been Tested?

Well, if CI passes, it worked? 🤷🏼‍♂️ 

## Types of changes

## Checklist:


mathomp4:
> looks like much less code. What is not to like?

Well, I *really* wish YAML had the ability to include files. Then I could hide the "ugly" bits in `commands.yml` and then do a `include commands.yml` but...sadly not. (Though it looks like [GitLab figured out a way](https://docs.gitlab.com/ee/ci/yaml/includes.html).)

tclune:
Hmm.   The GitLab way is not standard, but probably forgiveable in the context.

GMAO workflows will probably have $(value) references which will also not have standard interpretations.

Issue #641, opened 2020-12-15
Time for MAPL_IO refactoring?


bena-nasa:
Multiple feature requests have been made for the restart IO layer in MAPL.
https://github.com/GEOS-ESM/MAPL/issues/639
https://github.com/GEOS-ESM/MAPL/issues/635

Putting this out here for comment about the general state of MAPL restart IO. Those requests referenced above made me think it is time we at least have a discussion about refactoring of MAPL_IO.

MAPL_IO has reached a state that it is in need to of refactoring to make such requests easier to handle and just for future maintainability if we want to leverage redundant capabilities used by ExtData and History. It is also probably one of the longest or longest source file in MAPL.

The problem: it must currently support both netcdf and binary input/output for:
1D/2D/3D/4D R4/R8 variables needed by MAPL components on both regular gridded distributions (2 distributed dimensions), on the tile distribution (1 distributed dimension, arbitrarily distributed, must be reassembled in original order), and not distributed at all (1D variables like AK/BK). The undistributed dimensions represent edges, centers, or multiple arbitrary undistributed dimensions and generally exist in combination with the distributed dimensions (AK/BK is the exception).
Also supports some IO parallelism which does provide some benefit at higher resolutions for regular gridded variables.

Due to MERRA2 and older OPS version still having produced binary restarts, support for the this is regrettably probably a requirement for GEOS for the near future.

At the very least, it has probably reached a point where the code could be refactored into separate binary and netcdf layers so the netcdf can evolve, and the binary can just exist without getting in the way of future developments until it can be retired.

Other considerations.
Conceptually there should be some overlap between History/ExtData/restart IO in terms of the code base. Currently there is pretty much none. For example, it would seem logical that gridded output, whether for History or restarts really is no different and could certainly share a fair amount of code, for example, specification of metadata.

mathomp4:
Reexamined on 2021-Apr-12 with @bena-nasa and @tclune

bena-nasa:
Just adding a comment as a reminder of this. I added a new feature that reminded me that we really should do this. At the very least split apart MAPL_IO into separate binary and NetCDF modules so we can have some hope of cleaning this up the NetCDF handling someday.

PR #1055, opened 2021-09-28
Refactors MAPL_IO to separate Binary and NetCDF subroutines.

Gvilla1000-nasa:
## Description

This PR separates MAPL_IO.F90 into a binary (BinIO.F90) and NetCDF (NCIO.F90) modules. A shared file containing common subroutines has been created called FileIOShared.F90.

I tried refactoring this into a new directory but the MAPL_IO modules depend on base and sun_uc and gridded factory depend on MAPL_IO so there needs to be some refactoring effort done there first before moving this outside of base. 

## Related Issue

#641
## Motivation and Context

By separating MAPL_IO into binary and NetCDF modules it allows to continue development of the NetCDF routines without needing to touch the Binary routines.

## How Has This Been Tested?

Compiled with intel/gnu in release and debug mode. MAPL by itself was also compiled with NAG.
Single day run for gnu/intel were performed

## Types of changes

## Checklist:


tclune (dismissed):
I'm fine with the overall approach, but am relatively unfamiliar with this layer.   I defer to @bena-nasa to spot any further gotchas.

bena-nasa:
I'm looking at this now. Before merging we should test do additional tests for safety
1.) Make sure replay is still zero-diff as that uses the memory checkpointing
2.) Test with binary restarts. I see Matt has some MERRA2 binary restarts that can be used.

Gvilla1000-nasa:
Let me provide some background for the change in the CMakeLists.txt. When I was splitting MAPL_IO into NCIO and BinIO, I was getting a compilation error in cub2latlon specifically with MAPL_Sort (part of MAPL.Shared) while everything else was compiling fine. In the end, that addition. wasn't strictly needed since the compilation issue was due to a missing `use MAPL_SortMod` in one of the new files I added. Thats why in the cleanup I removed that line from the CMakeLists.txt. Should I remove it again?

Gvilla1000-nasa:
Thanks Ben for the suggestions. I ran the second test using binary restarts and it was 0diff. As discussed offline, I'm currently unable to run the 1st test due to permission issues.

bena-nasa:
I'm seeing more things that should be in the FileIOSharedMod
Arrdescr was moved there, but NCIO.F90 still has
```
  public ArrDescrSet
  public ArrDescrInit
```
Those should probably be in shared as well.

bena-nasa (changes requested):
See previous comment I made

bena-nasa:
The replay test was zero diff

bena-nasa (approved):

tclune (approved):

PR #1060, opened 2021-10-04
Fixes #1059 - cmake install dir

### LINKED ISSUE #1059: Create mapl-import.cmake file
ecbuild puts this blob in the mapl-config.cmake file:
```cmake
### include the <project>-import.cmake file if there is one
if(EXISTS ${MAPL_CMAKE_DIR}/mapl-import.cmake)
  set(MAPL_IMPORT_FILE "${MAPL_CMAKE_DIR}/mapl-import.cmake")
  include(${MAPL_IMPORT_FILE})
endif()
```
So we should create this file and put the extra paths that specify where MAPL cmake macros are installed for projects that need them.  E.g., UFS.   @rmontuoro @mathomp4 

Ref:  https://github.com/GEOS-ESM/GOCART/pull/80

tclune:
- Changed location of installed MAPL cmake macros to facilitate UFS.
  The new location is more conventional.  Also added ecbuild logic to
  add new location to module path.  I.e., now it's done in a better
  way, and maybe even the "right"(TM) way.

## Description

Changed CMake logic for installing cmake macros for upstream projects.

## Motivation and Context

Simplifies use of MAPL by _external_ projects.  No impact on GEOS.

## Types of changes

## Checklist:

Only affects upstream projects, so cannot be tested inside GEOS.  Have tested with a toy cmake project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Refactored MAPL_IO by separating it into a Binary (BinIO.F90) and NetCDF (NCIO.F90) modules. Shared subroutines and
   types have been moved to FileIOShared.F90. MAPL_IO becomes a package module to hold these aforementioned three modules.
 
+- Changed location of installed MAPL cmake macros to facilitate UFS.   The new location is more conventional.  Also added ecbuild logic to add
+  new location to module path.  I.e., now it's done in a better way, and maybe even the "right"(TM) way.
+
+
 ### Fixed
 
 - Fixed #338. Added a workaround for a gfortran bug that handles end-of-file incorrectly (returns IOSTAT=5001).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,7 +142,7 @@ install (
 
 install (
    DIRECTORY cmake
-   DESTINATION ${INSTALL_DATA_DIR}
+      DESTINATION ${INSTALL_CMAKE_DIR}
    )
 
 # https://www.scivision.dev/cmake-auto-gitignore-build-dir/


bena-nasa (approved):

mathomp4:
Blocking until it can be tested

mathomp4:
Doesn't seem to do anything to the model (as expected). So good on our end.

mathomp4 (approved):

rmontuoro:
Unfortunately this change does not solve our original issue:
```
CMake Error at GOCART/CMakeLists.txt:69 (include):
  include could not find requested file:

    mapl_acg


CMake Error at GOCART/ESMF/Aerosol_GridComp/CMakeLists.txt:9 (mapl_acg):
  Unknown CMake command "mapl_acg".
```

mathomp4:
Blocking again per @rmontuoro until @tclune can look at this

tclune:
We may have to iterate on this.   I created a toy external project that pretty much just did a `find_package(MAPL)` and it was able to use those macros.

tclune:
```cmake
cmake_minimum_required (VERSION 3.17)
cmake_policy (SET CMP0053 NEW)
cmake_policy (SET CMP0054 NEW)

project (
  FOO
  VERSION 0.0.1
  LANGUAGES Fortran CXX C)  # Note - CXX is required for ESMF

find_package(MAPL REQUIRED)

include(mapl_create_stub_component)
mapl_create_stub_component (xyz foofoo)

```

tclune:
I think it may be possible that you just need to remove the `include` line from the GOCART cmakelists.  But I'll check.  I would have thought that the module path extension done within the `find_package()` would persist, but it maybe makes sense that it does not.    I'll try my toy project ...  no.  It works either way.  the `include` is not needed, but it should also work.    

I'll try building GOCART standalone in a while - very busy with meetings today though.

rmontuoro:
Thanks @tclune - No hurry. I'll look into it as well.

tclune:
I was able to build the develop branch of GOCART with the following command on my system:
```shell
cmake .. -DCMAKE_PREFIX_PATH=$HOME/swdev/GEOS-ESM/MAPL/build/install -DCMAKE_MODULE_PATH="$HOME/swdev/GEOS-ESM/MAPL/ESMA_cmake;$HOME/swdev/GEOS-ESM/MAPL/ESMA_cmake/ecbuild/cmake" -DUFS_GOCART=YES
```
caveat - this got past the `mapl_acg` error, but since I did not include FLAP and such there were other errors.   I'm fairly confident those are unrelated.

rmontuoro:
@tclune - In my tests I found out that `mapl-import.cmake` does not get installed. I've submitted PR #1101 with an alternative set of changes that work with UFS. These changes also restore the original install location for the MAPL macros for consistency with other ecbuild-based installations.

PR #1082, opened 2021-10-14
Fixes #1064. Fixed error related to NewSeg logic

### LINKED ISSUE #1064: Incorrect file timestamp with 24-hour duration and 1-hour frequency time-average
Follow up from today's GCHP-MAPL call.

Collections with mode time-average, frequency 1-hour, and duration 24-hour aren't timestamped correctly (with the default Reference of 000000). If you enable `timestampStart` the files are stamped the previous day at 23:00. I *believe* that without `timestampStart` they are stamped at 23:30 of the previous day.

atrayano:
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
-- Added a proper handling for new segment logic in History. This addressed isues #1064 and #1067
+
+## [2.8.9] - 2021-10-15
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


tclune (approved):

mathomp4:
> We're using 2.6.3. Is this an easy patch on 2.6?
> 
> (FYI we're planning to upgrade to the latest MAPL in GCHP 13.4 or 13.5 -- soonish, but far enough away that it would be ideal if we didn't have the timestamp bug unitl then)

@LiamBindle Well, the next in the 2.6 would be 2.6.10, but I *think* it should be possible. Let me see...

LiamBindle:
Thanks @mathomp4 - testing it now

Issue #782, opened 2021-03-30
Migration to ESMF_Info


mathomp4:
As I'm doing some test builds with ESMF 8.1.0, from their release notes:

> The ESMF_Info class was introduced as a replacement for ESMF_Attribute. ESMF_Info is based on a modern C++ JSON implementation to provide efficient key-value pair storage. Most of the legacy ESMF_Attribute API is preserved for backward compatibility.

and from [the documentation](https://earthsystemmodeling.org/docs/release/ESMF_8_1_0/ESMF_refdoc/node6.html#SECTION06020000000000000000):

> The ESMF_Info class is a replacement for the ESMF_Attribute class and is the preferred way of managing metadata attributes in ESMF moving forward. It is recommended that users migrate existing ESMF_Attribute calls to the new ESMF_Info API. The ESMF_Info class provides the backend for ESMF_Attribute since ESMF version 8.1. The ESMF_Attribute docs are located in appendix 55. In practice, users should experience no friction when migrating client code. Please email ESMF support in the case of a migration issue.

I suppose the fact that the docs for ESMF_Attribute are now in an Appendix is a sign to start looking at this.

Though, as @atrayano, @bena-nasa and @tclune know, ESMF_Attribute is nigh-omnipresent in GEOS. So, this might take a while...

tclune:
My limited understanding is that this was more of an internal design change and that the existing interfaces are backward compatible.  No?

mathomp4:
@tclune Yes, they are. And I imagine if we request ESMF to keep it in the code they will, I just wanted to bring up the fact that they seem to be pushing ESMF_Info.

Indeed, I think one of the early reasons we couldn't use ESMF 8.1.0 snapshots was that the migration sort of "broke" ESMF_Attribute (something to do with 32 or 64 bits? Can't remember).

Annoyingly, the ESMF_Info class seems to be *more* verbose than ESMF_Attribute:

> 39.1.1 Setting an Attribute
> With ESMF_Attribute:
> ```fortran
> call ESMF_AttributeSet(array, "aKey", 15, rc=rc)
> ```
> With ESMF_Info:
> ```fortran
> call ESMF_InfoGetFromHost(array, info, rc=rc)
> call ESMF_InfoSet(info, "aKey", 15, rc=rc)
> ```
>

tclune:
We can always write our own wrappers if the 1-2 combo is frequent enough.

mathomp4:
[...]

PR #1114, opened 2021-10-20
ESMF_Info refactoring

Gvilla1000-nasa:
## Description
Because ESMF_AttributeGet and ESMF_AttributeSet will be deprecated by ESMF at some point in the near future, those calls needed to be changed to ESMF_InfoGet and ESMF_InfoSet respectively.

This PR is divided into several commits in which each individual one focuses on grids, state, bundle, and fields individually.

This PR is directed towards the MAPL-v3 branch because by changing some field ESMF_AttributeGet/Set calls to ESMF_InfoGet/Set required changes in some external gridcomps to address the FRIENDLYTO variables. The following files required changes:
`@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/GEOSdatmodyn_GridComp/GEOS_DatmoDynGridComp.F90`
`@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/GEOSCHEMchem_GridComp/@geos-chem/Interfaces/GCHP/Chem_GridCompMod.F90`
`@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/GEOS_MoistGridComp.F90`
`@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSturbulence_GridComp/GEOS_TurbulenceGridComp.F90`
`@GEOSgcm_GridComp/GEOSogcm_GridComp/GEOS_OgcmGridComp.F90`

## Related Issue

#782 

## Motivation and Context

To prevent issues when ESMF gets updated and deprecates ESMF_AttributeGet/Set calls

## How Has This Been Tested?

Each commit in which major changes were done (grid, state, bundle, field) a one day run was performed and compared to a control case. Each time it was zero diff.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+
 - Changed all ESMF_AttributeGet and ESMF_AttributeSet to ESMF_InfoGet and ESMF_InfoSet respectively as old calls will be deprecated soon.
 
 ### Removed


tclune (dismissed):

tclune:
Returning both errors is easy just uses "status1" and "status2" , but as you say, you lose the ability to prevent evaluating the conditional if the function errors.   So no choice but verbose.

tclune:
Of course, probably the only way for that function to return an error is for the `infoh` object to be already in an invalid state.   So one could argue that it has already been checked by this point.   Indeed, functional programming would insist that the constructor always generates a valid object, even if it just reports empty on all calls.

PR #1122, opened 2021-10-21
refactor regrid_util and add option to regrid multiple files at once

bena-nasa:
This is a refactoring of Regrid_Util.x
I've had requests to add the ability to regrid multiple files in one execution (this saves some time as you don't have to regenerate the route handle for each file). While I was at this I did a refactor to make it easy to extend and do some other things I want to do like convert to flap for example.
This should have no bearing on the model, I did have to update a few things for the bundlewrite, but only Regrid_Util.x uses this right now.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+### Added
+
+### Changed
+
+### Removed
+
+## [2.10.0] - 2021-10-22
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reduced runtime impact of communication barriers in Cap gridded component
 
 ### Added
+- Added ability toregrid multiple files in one execution of Regrid_Util.x
```
tclune, line 15:
```suggestion
- Added ability to regrid multiple files in one execution of Regrid_Util.x
```

Issue #1144, opened 2021-10-27
[Feature request] Run in reverse


TerribleNews:
In order to support adjoint integration in GCHP, MAPL should be able to run in reverse. My first attempt at this was to use the settings with a negative timestep and end time before the start time but this has many side effects. Instead, I have found a better solution to be to run the clocks forward in a loop without any child grid component run calls, then to reverse the clocks and run the regular time loop with a modified step function. There are also issues with the alarms running in reverse which need to be addressed. I am mainly submitting this issue because the merge request template says there should be an issue before a merge request.

mathomp4:
Labeling long term to keep open

PR #1166, opened 2021-11-01
Changes to allow GCHP to run in reverse mode for adjoint integration

TerribleNews:
## Description

This change adds a time loop to CapGridComp run the clock forward before the time loop that actually executes the grid components. There are also some small changes to handle issues with alarms not ringing properly in reverse.

## Related Issue

https://github.com/GEOS-ESM/MAPL/issues/1144

## Motivation and Context

This feature is to support to adjoint development in GCHP, which requires running the model in reverse. There may be other advantages to a generic capability to run MAPL in reverse as well.

This will be my first merge request to the team, so I am submitting mostly to start the conversation and get guidance / feedback.

## How Has This Been Tested?

Unfortunately, because GCHP uses a fork of MAPL it is non-trivial to test code changes with the latest version. All these changes are patterned after changes that have been tested in GCHP using the verison of MAPL that is there, which is based on 2.6.3. Is there a way for me to test these exact changes?

## Types of changes

## Checklist:


atrayano:
@TerribleNews Thanks for submitting the PR. While many of your changes are very well protected and won't impact GEOS at all, some would require additional logic. My main concern are the changes in History, and in particular the usage of the existing variable "list(n)%backwards". GEOS also has an adjoint model, and this variable might be used there. My preference would be to leave the existing "backwards: logic intact, and add a new variable, for example "reverse_time". I have many more comments regarding style, for example, the usage of debugging prints, etc., but I will articulate them in separate comments

TerribleNews:
Thanks for the feedback @atrayano. I am not thrilled with the changes I made to the history alarms in general; they were mostly trial and error, and possibly not all of them are necessary / useful. For example, when there's a new alarm declared, if it's backwards, it's declared as sticky, but I couldn't get my alarms to ring without `sticky=.false.`, but I couldn't never figure out exactly why, or why the original code was sticky. I have a somewhat limited testing platform. I used the backwards attribute because the comment said it's currently unused, but if GEOS is using it, then I don't want to mess that up. I think we have a few options here:

1. I revert the `backwards` logic and add in a new `reverse_time` attribute to replace my changes
2. I take the history changes out of this PR and we can come up with a plan to better test that part of the code so that I'm not just making changes to it willy-nilly
3. Something else I haven't thought of

mathomp4:
Huh. I didn't know the stale bot could mark *PRs* as stale. I think I need to look at how it's configured!

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3336,7 +3460,45 @@ subroutine Run ( gc, import, export, clock, rc )
...
+      write(6,*) "Checking history time"
+      call ESMF_ClockGet ( clock,  currTime=CurrTime ,rc=STATUS ) ; _VERIFY(STATUS)
+
+      call ESMF_TimeGet  ( CurrTime, timeString=TimeString, rc=status ) ; _VERIFY(STATUS)
+
+      read(timestring( 1: 4),'(i4.4)') year
+      read(timestring( 6: 7),'(i2.2)') month
+      read(timestring( 9:10),'(i2.2)') day
+      read(timestring(12:13),'(i2.2)') hour
+      read(timestring(15:16),'(i2.2)') minute
```
tclune, line 3474: Aren't there better ways to get these directly from ESMF_Time?

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3417,6 +3579,10 @@ subroutine Run ( gc, import, export, clock, rc )
       if (list(n)%disabled .or. ESMF_AlarmIsRinging(list(n)%end_alarm) ) then
          list(n)%disabled = .true.
          Writing(n) = .false.
+         if (MAPL_am_I_Root()) THEN
+            WRITE(*, 1999) n
+         ENDIF
+1999     FORMAT('Not Writing alarm ', i3, ' because end_alarm is ringing')
```
tclune, line 3585: Prefer to avoid FORMAT statements:
```suggestion
         if (MAPL_am_I_Root()) THEN
            WRITE(*, '("Not Writing arlarm ", i3, " because end alarm is ringing.") n
         ENDIF
```

tclune (changes requested):
Suggested one change and would like clarification from @bena-nasa on how to retrieve min/hr/... from an ESMF_Time.

I also note that the changes to CapGridComp are likely fragile under planned MAPL3 changes, and thus may require re-implementation down the road.

tclune:
Did a quick attempt to resolve conflicts.  May easily have done damage.

tclune:
Closing here.  Replaced by #1795 .

PR #1007, opened 2021-08-30
Attempt to refactor use of FLAP.

tclune:
- Migrated CapOptions into ./gridcomps/cap
 - Reduced use of `#ifdef`

## Description

- Relocated 2 files from `./base` to `./gridcomps/Cap`
- Improved separation of CapOptions and FlapCLI layers.
- 
## Motivation and Context
Cleanup.

## How Has This Been Tested?
Just MAPL tests.  Further testing with GCM is warranted.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Trivial change (affects only documentation or cleanup)

## Checklist:

- [ ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [x] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

Issue #996, opened 2021-08-26
MAPL_CapOptions in base: why?


mathomp4:
I'm not the MAPL guru @tclune and the rest @GEOS-ESM/mapl-team are but as I was looking around today, I saw that `MAPL_CapOptions.F90` is in `base` and not in `gridcomps/Cap`. Is there a reason for that I don't know about?

tclune:
These should be in gridcomps/Cap.    Further, the Flap extension is still not quite implemented in the cleanest way.  This can be done without `#ifdef's`

PR #1170, opened 2021-11-02
Refactors FLAP and CapOptions

Gvilla1000-nasa:
## Description
This PR should take the place of #1007 
* Takes @tclune FLAP refactoring attempt and CapOptions refactoring and adds extra stuff to make sure it works.
*  Changes include: 
     1) CapOptions refactoring from base to gridcomps/Cap
     2) FLAP refactoring and simplification
     3) Ensures backward compatibility

## Related Issue

#996 

## Motivation and Context

Refactors CapOptions and simplifies FLAP usage

## How Has This Been Tested?

Compilation tested with GNU and Intel in both release and debug. Also tested with NAG. One day sample case was also run for GEOS and it was zero diff.

## Types of changes

## Checklist:


Gvilla1000-nasa:
@tclune or anyone. If you want to look at changes done on top of #1007, please look at following commit:
https://github.com/GEOS-ESM/MAPL/pull/1170/commits/d911bc625c6dee4dd08e2b93e7563413c9f0c781

tclune (changes requested):
Does this induce a change on GEOSgcm.F90 which uses CapOptions?

Gvilla1000-nasa:
> Does this induce a change on GEOSgcm.F90 which uses CapOptions?

Honestly unsure. Still not too familiar with all aspects of the code. I'll take a look.

tclune (approved):

Tests/pfio_MAPL_demo.F90
```diff
@@ -74,9 +74,10 @@ program main
 !BOC
 
       ! Read and parse the command line, and set parameters
-      cli = MAPL_FlapCLI(description = 'GEOS AGCM', &
-                         authors     = 'GMAO')
-      cap_options = MAPL_CapOptions(cli)
+      cap_options = MAPL_FlapCLI( &
```
mathomp4, line 77: Ooooooh. This might mean this is only for MAPL3. Otherwise we have to change things like `GEOSgcm.F90` (see https://github.com/GEOS-ESM/GEOSgcm_App/blob/develop/GEOSgcm.F90)

Is there a way to do this and not to change the interface?

Gvilla1000-nasa: this is backward compatible. Notice the difference between GEOSgcm:
```   
cli = MAPL_FlapCLI(description = 'GEOS AGCM', &
                                     authors     = 'GMAO')
   cap_options = MAPL_CapOptions(cli)
```

and pfio that only uses cap_options = MAPL_CapOptions(cli)

mathomp4:
Putting a blocking label on this until @Gvilla1000-nasa can test with GEOSgcm.x (etc.)

Tests/pfio_MAPL_demo.F90
tclune: I asked @Gvilla1000-nasa about this privately.   He claims this version works with GEOSgcm.F90.

mathomp4: Ohh. Cool. Well, if it does, then I guess no blocking label needed. I'm all for changes in MAPL that don't need anything outside MAPL.

@bena-nasa Query: Do all bits of GEOS that use FLAP (including any in the ADAS if there are any?) use this sort of initialization of FLAP?

Gvilla1000-nasa: to elaborate a bit more. MAPL_FlapCLI is an interface to two different functions. One whose output is of MAPL_FlapCLI type and another whose output is of MAPL_CapOptions type. The new version is the latter but to prevent errors during compilation time because of the module procedures having the same number of argument, I was forced to add a dummy argument to the new version.

tclune: It's all a bit jury rigged at the moment to preserve backward compatibility.   This is probably a good time to create a MAPL 3 branch for GEOSgcm_App and change MAPL to only have the desired interfaces.

mathomp4 (approved):

PR #1189, opened 2021-11-08
Fixes #1186 - cmake version logic.

### LINKED ISSUE #1186: Broken version check for gFTL
Recent changes to support projects that build gFTL within their source tree are masking heightened version requirements for gFTL in MAPL.    If the package has already been found, a separate check must then be made for the correct version.    The fix will be a kludge.   external projects should change the way this is handled.

tclune:
## Description

Fixes #1186 bug in cmake version check.

## Motivation and Context

Older baselibs won't work, but no useful error message is given.  Fails deep inside of a gFTL template otherwise.

## How Has This Been Tested?

Verified that useful error is emitted for older gFTL in baselibs.

## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fragile CMake logic for checking minimum version requirements for gFTL.
```
mathomp4, line 12: Could you add this to a new:
```
## [2.12.1] - 2021-11-08
...
```
section above `[2.12.0]` and below `[Unreleased]`?

tclune: will do

mathomp4 (approved):
Approving!

PR #1202, opened 2021-11-18
Fixes #1199 - HorizontalFluxRegridder numerical bug.

### LINKED ISSUE #1199: Bug in HorizontalFlux regridder
The regridder assumes that the fluxes have already been multiplied by edge length, but they have not.

tclune:
## Description

Fixing numerical bug in HorizontalFluxRegridder

## Motivation and Context

bugfix

## How Has This Been Tested?

compiles on Intel
is not exercised by GEOS

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Changed default decomposition algorithm in Base/base.
+  . More optimal for CS case
+  . Hopefully aligns with common choices for native decomp to reduce need for nontrivial regridding.
 - Changed usage of MAPL_IO subroutines in CubedSphere and LatLon Grid Factories to open command with newunit clause
 - Updated `components.yaml`
   - ESMA_env v3.6.0 (Use MPT 2.25 at NAS on TOSS)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Corrected bug in HorizontalFluxRegridder.  Fluxes need to be
+  multiplied by edge length for correct treatment.
+
 ### Added
 
 - Added ability to generate monthly checkpoints (fixes issue #1065)
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -832,7 +832,7 @@ contains
 
     type (ESMF_VM) :: vm
     integer :: pet_count
-
+    integer :: bias
     character(len=*), parameter :: Iam= __FILE__ // '::MAPL_MakeDecomposition()'
     integer :: status
 
@@ -842,11 +842,18 @@ contains
     _VERIFY(status)
     call ESMF_VMGet(vm, petCount=pet_count, rc=status)
     _VERIFY(status)
-    if (present(reduceFactor)) pet_count=pet_count/reduceFactor
+    if (present(reduceFactor)) then
+       pet_count=pet_count/reduceFactor
+       ! Assume CS
+       bias = 1
+    else
+       ! Assume Lat-Lon
+       bias =2
+    end if
 
     ! count down from sqrt(n)
     ! Note: inal iteration (nx=1) is guaranteed to succeed.
-    do nx = floor(sqrt(real(2*pet_count))), 1, -1
+    do nx = nint(sqrt(real(bias*pet_count))), 1, -1
        if (mod(pet_count, nx) == 0) then ! found a decomposition
           ny = pet_count / nx
           exit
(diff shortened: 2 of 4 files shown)


bena-nasa (dismissed):

tclune:
Here are the results of regridding "pure" values of 2 for x and y fluxes:

output  varies from 1.9996 to 2.041.  Before it was identically 4.

VARY_in_flux_out 2004
VARX_in_flux_out 2004

bena-nasa (dismissed):

tclune:
@sdeastham Careful interpreting the above results.  Because the input fluxes were constant, the regrid essentially returns
```
y = (d_1 * x_1 + d_2 *x_2) / (d_12) =  (d_1 + d_2)*x/(d_12).   
```
I.e., it just measures how accurate the formulae for the arc lengths, as mathematically `d_12 = d_1 + d_2`

tclune:
Errors appear mostly near the poles.  I think that is because "longitude" is very sensitive to position in that vicinity and increases error in the trig expressions.

sdeastham:
Thanks @tclune ! So, to clarify, what's being plotted is the error in the new calculation? Something I'm curious about - is the calculation of length identical to the method used in FV3? My thinking is that, if so, then this error shouldn't actually matter (because the same error was present in FV3 when calculating the fluxes).

tclune:
No the output fields from a test signal are being output.  The variation from "2" is the error (entirely due to finite precisions FP, not to regridding).

@bena-nasa has set up a new test that has spatially varying input which will be a better indicator.   But harder to interpret. :-)   Hopefully I can exercise that next week.

sdeastham:
Ha sorry, that's what I meant - so if `x` is the plotted value, the relative error is `x/2 - 1`. Do you know though if this error would actually affect the output? If I understand you and the code correctly, the calculation proceeds as follows (see caveat underneath regarding distance calculation):

 - At C180, FV3 calculates mass fluxes across two faces. To figure out the flux per unit face length two adjacent cells `F1` and `F2`, it uses lengths `d1` and `d2`.
 - When we reuse that data at C180, that's fine - we have a completely consistent understanding of the arc lengths, whether or not those lengths are strictly correct. FV3 calculates the total flux as `F1 * d1` and `F2 * d2`.
 - When we reuse the data at a coarser resolution (say, C90), the regridder says that the total flux per unit length is `F12 = ((F1 * d1) + (F2 * d2))/(d12)`. FV3 doesn't know anything about the finer grid so it calculates the total flux across the face as `F12 * d12`, so the error `e` is `e = F12 * d12 - ((F1 * d1) + (F2 * d2)) = ((F1 * d1) + (F2 * d2)) * d12/d12 - ((F1 * d1) + (F2 * d2)) = 0`. So even if `d12` is not quite the same as `d1` and `d2`... I think?
 
 The above wouldn't be true if the regridder calculated the coarse flux as `F12 = ((F1 * d1) + (F2 * d2))/(d1 + d2)`, but as long as it uses `d12` in the denominator, I *think* this error doesn't affect the result? The total mass entering the C90 grid cell will be the same as the net flux which would have entered the set of 4 cells at C180, and - as long as the distance calculation exactly matches that used internally by FV3 - that should result in the same amount of mass loss or accumulation in the cell. I think there might still be an error associated with differing calculations of the cell area/footprint, though. Is that correct or am I missing an important detail?

tclune:
Yes.  I could try to make it use "d1 + d2".  Would not even be a lot of work.   It just makes the error more subtle though, I think.

sdeastham:
Wouldn't using `d1 + d2` be worse? If my logic above is correct, using `d12` directly in the regrid calculation as we already do should mean that errors due to precision problems don't actually cause a problem, even though it's mathematically incorrect?

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Changed default decomposition algorithm in Base/base.
+  . More optimal for CS case
+  . Hopefully aligns with common choices for native decomp to reduce need for nontrivial regridding.
```
mathomp4, line 23: The dots don't render as a list it seems

```suggestion
  - More optimal for CS case
  - Hopefully aligns with common choices for native decomp to reduce need for nontrivial regridding.
```

mathomp4:
Do the cool pictures change with this update? I like seeing them. 😄

mathomp4:
Pinging @tclune. Should this stay open?

tclune:
Nah - we know where to find it if @sdeastham returns to the issue.

tclune:
@sdeastham Is this the bug you mentioned in the AIST meeting today?

mathomp4:
I've restored the `bugfix/tclune/#199-hflux-fix-2nd-try` branch. It is pretty out-of-date with `develop` and threw some conflicts when I tried to merge `develop` in. @tclune might need to do this because I'm not sure what to keep and what to not...

That said, if the conflicts can be fixed up. I'm glad to test out to see if it affects GEOSgcm in anyway.

tclune:
Not sure how/why the Base_base stuff got mixed into this branch.   Presumably we want all the other stuff.   I'll try a hand merge.

tclune:
OK- made an attempt.   I had to do a new branch "3rd-try" because git would not let me push to origin after I rebased.  Going to try to edit this PR, but may have to do a new one.

PR #1206, opened 2021-11-22
Fixes #1205. Make AddChildFromDSO OS agnostic

### LINKED ISSUE #1205: Make AddChildFromDSO OS Agnostic
Per https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/489, the DSO AddChild calls in GEOSgcm_GridComp assume that the DSO file extension is `.so`. On macOS this is `.dylib`. So, `MAPL_AddChildFromDSO` should be more agnostic and strip off the extension and add the "right one" for the OS as detected by CMake.

In the future, we should in fact *enforce* that no suffix is used with the AddChild calls, but for now, just make it ourselves.

mathomp4:
## Description

This PR adds some string codes that allow us to construct a correct DSO filename that is not dependent on the OS.

## Related Issue

Closes #1205. 

Also see: https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/489

## Motivation and Context

GEOS currently has DSO calls that are looking for `.so` files. But on a Mac, shared object libraries are `.dylib` files. This code makes it 'agnostic' by just constructing the right name (assuming the basename of the library is correct).

## How Has This Been Tested?

I tested this in two ways. First in `GEOS_OceanGridComp.F90`, I changed the `libMOM6_GEOSPlug.so` in:
```fortran
call MAPL_GetResource ( MAPL, sharedObj,  Label="MOM6_GEOSPLUG:", DEFAULT="libMOM6_GEOSPlug.so", __RC__ )
```
to be `libMOM6_GEOSplug.jojo`. This caused a crash. With this PR, it works.

I also tried using just `libMOM6_GEOSplug` with no extension (which might be enforced in MAPL3) and it also ran.

## Types of changes

## Checklist:

### DIFF
--- a/generic/MAPL_Generic.F90
+++ b/generic/MAPL_Generic.F90
@@ -4637,7 +4637,7 @@ end subroutine MAPL_DateStampGet
 
   integer                                     :: I
   type(MAPL_MetaComp), pointer                :: CHILD_META, tmp_meta
-  class(AbstractFrameworkComponent), pointer :: tmp_framework
+  class(AbstractFrameworkComponent), pointer  :: tmp_framework
   character(len=ESMF_MAXSTR), allocatable     :: TMPNL(:)
   character(len=ESMF_MAXSTR)                  :: FNAME, PNAME
   type(ESMF_GridComp)                         :: pGC
@@ -4651,6 +4651,7 @@ end subroutine MAPL_DateStampGet
   type(StubComponent) :: stub_component
   type(ESMF_VM) :: vm
   integer :: comm
+  integer :: userRC
 
   lgr => logging%get_logger('MAPL.GENERIC')
 
@@ -4700,21 +4701,21 @@ end subroutine MAPL_DateStampGet
 
      gridcomp => child_meta%gridcomp
      if (present(configfile)) then
-        gridcomp = ESMF_GridCompCreate   (     &
-             NAME   = trim(FNAME),                 &
-             CONFIGFILE = configfile,             &
-             grid = grid,                         &
-             petList = petList,                   &
-             contextFlag = contextFlag,           &
+        gridcomp = ESMF_GridCompCreate   ( &
+             NAME   = trim(FNAME),         &
+             CONFIGFILE = configfile,      &
+             grid = grid,                  &
+             petList = petList,            &
+             contextFlag = contextFlag,    &
              RC=STATUS )
         _VERIFY(STATUS)
      else
-        gridcomp = ESMF_GridCompCreate   (     &
-             NAME   = trim(FNAME),                 &
-             CONFIG = META%CF,                    &
-             grid = grid,                         &
-             petList = petList,                   &
-             contextFlag = contextFlag,           &
+        gridcomp = ESMF_GridCompCreate   ( &
+             NAME   = trim(FNAME),         &
+             CONFIG = META%CF,             &
+             grid = grid,                  &
+             petList = petList,            &
+             contextFlag = contextFlag,    &
              RC=STATUS )
         _VERIFY(STATUS)
      end if
@@ -4723,16 +4724,16 @@ end subroutine MAPL_DateStampGet
 ! ----------------------------------
 
      child_import_state => META%get_child_import_state(i)
-     child_import_state = ESMF_StateCreate (                         &
+     child_import_state = ESMF_StateCreate (             &
           NAME = trim(META%GCNameList(I)) // '_Imports', &
-          stateIntent = ESMF_STATEINTENT_IMPORT, &
+          stateIntent = ESMF_STATEINTENT_IMPORT,         &
           RC=STATUS )
      _VERIFY(STATUS)
 
      child_export_state => META%get_child_export_state(i)
-     child_export_state = ESMF_StateCreate (                         &
+     child_export_state = ESMF_StateCreate (             &
           NAME = trim(META%GCNameList(I)) // '_Exports', &
-          stateIntent = ESMF_STATEINTENT_EXPORT, &
+          stateIntent = ESMF_STATEINTENT_EXPORT,         &
        RC=STATUS )
   _VERIFY(STATUS)
 
@@ -4765,7 +4766,8 @@ end subroutine MAPL_DateStampGet
   call CHILD_META%t_profiler%start(__RC__)
   call CHILD_META%t_profiler%start('SetService',__RC__)
   gridcomp => META%GET_CHILD_GRIDCOMP(I)
-  call ESMF_GridCompSetServices ( gridcomp, SS, RC=status )
+  call ESMF_GridCompSetServices ( gridcomp, SS, userRC=userRC, RC=status )
+  _ASSERT(userRC==ESMF_SUCCESS .and. STATUS==ESMF_SUCCESS,'issue with MAPL_AddChildFromMeta')
   call CHILD_META%t_profiler%stop('SetService',__RC__)
   call CHILD_META%t_profiler%stop(__RC__)
   call t_p%stop(trim(NAME),__RC__)
@@ -4821,14 +4823,15 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
   integer, optional  , intent(  OUT) :: rc
   !EOP
 
-  character(len=ESMF_MAXSTR)                  :: IAm
-  integer                                     :: STATUS
+  character(len=ESMF_MAXSTR) :: IAm
+  integer                    :: STATUS
+  integer                    :: userRC
 
   type(MAPL_MetaComp), pointer                :: META
 
   integer                                     :: I
   type(MAPL_MetaComp), pointer                :: CHILD_META, tmp_meta
-  class(AbstractFrameworkComponent), pointer :: tmp_framework
+  class(AbstractFrameworkComponent), pointer  :: tmp_framework
   character(len=ESMF_MAXSTR), allocatable     :: TMPNL(:)
   character(len=ESMF_MAXSTR)                  :: FNAME, PNAME
   type(ESMF_GridComp)                         :: pGC
@@ -4899,21 +4902,21 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
 
         gridcomp => child_meta%gridcomp
         if (present(configfile)) then
-           gridcomp = ESMF_GridCompCreate   (     &
-                NAME   = trim(FNAME),                 &
-                CONFIGFILE = configfile,             &
-                grid = grid,                         &
-                petList = petList,                   &
-                contextFlag = contextFlag,           &
+           gridcomp = ESMF_GridCompCreate   ( &
+                NAME   = trim(FNAME),         &
+                CONFIGFILE = configfile,      &
+                grid = grid,                  &
+                petList = petList,            &
+                contextFlag = contextFlag,    &
                 RC=STATUS )
            _VERIFY(STATUS)
         else
-           gridcomp = ESMF_GridCompCreate   (     &
-                NAME   = trim(FNAME),                 &
-                CONFIG = META%CF,                    &
-                grid = grid,                         &
-                petList = petList,                   &
-                contextFlag = contextFlag,           &
+           gridcomp = ESMF_GridCompCreate   ( &
+                NAME   = trim(FNAME),         &
+                CONFIG = META%CF,             &
+                grid = grid,                  &
+                petList = petList,            &
+                contextFlag = contextFlag,    &
                 RC=STATUS )
            _VERIFY(STATUS)
         end if
@@ -4922,16 +4925,16 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
 ! --   --------------------------------
 
         child_import_state => META%get_child_import_state(i)
-        child_import_state = ESMF_StateCreate (                         &
+        child_import_state = ESMF_StateCreate (             &
              NAME = trim(META%GCNameList(I)) // '_Imports', &
-             stateIntent = ESMF_STATEINTENT_IMPORT, &
+             stateIntent = ESMF_STATEINTENT_IMPORT,         &
              RC=STATUS )
         _VERIFY(STATUS)
 
         child_export_state => META%get_child_export_state(i)
-        child_export_state = ESMF_StateCreate (                         &
+        child_export_state = ESMF_StateCreate (             &
              NAME = trim(META%GCNameList(I)) // '_Exports', &
-             stateIntent = ESMF_STATEINTENT_EXPORT, &
+             stateIntent = ESMF_STATEINTENT_EXPORT,         &
           RC=STATUS )
         _VERIFY(STATUS)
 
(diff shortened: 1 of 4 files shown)


mathomp4:
I'm going to keep this in draft until @sanAkel (or I) can test on a Mac laptop and until @tclune can critique the Fortran. 😄

mathomp4:
Also to @tclune or anyone who looks at this, the EditorConfig file did a number on MAPL_Generic. It found a *LOT* of trailing whitespace and decided to clean it up. You might want to "ignore whitespace" in the comparison window

mathomp4:
Also: I just know at some point there will be a missing `trim()` here. I found one, I wonder when the next will be found!

generic/MAPL_Generic.F90
```diff
@@ -4955,16 +4960,25 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
   call ESMF_VMGet(vm, mpiCommunicator=comm, __RC__)
   CHILD_META%t_profiler = TimeProfiler(trim(NAME), comm_world=comm)
 
+  ! Construct a valid shared object library name
+
+  ! 1. Get the basename of the file without the extension
```
tclune, line 4965:
```suggestion
```
Redundant comment.

generic/MAPL_Generic.F90
```diff
@@ -4955,16 +4960,25 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
...
+  shared_object_library_basename = get_file_basename(SharedObj)
+  ! 2. Get the system dso suffix
```
tclune, line 4967:
```suggestion
```
Redundant

generic/MAPL_Generic.F90
```diff
@@ -4955,16 +4960,25 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
...
+  system_dso_suffix = get_system_dso_suffix()
+  ! 3. Make a new shared object name
```
tclune, line 4969:
```suggestion
```
Redundant

generic/MAPL_Generic.F90
```diff
@@ -4955,16 +4960,25 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
...
+  shared_object_library_with_system_dso_suffix = trim(shared_object_library_basename)//trim(system_dso_suffix)
```
tclune, line 4970:
```suggestion
  shared_object_library_with_system_dso_suffix = shared_object_library_basename // system_dso_suffix)
```
Trim not needed - we are carefully using deferred length strings

generic/MAPL_Generic.F90
```diff
@@ -4955,16 +4960,25 @@ recursive integer function MAPL_AddChildFromDSO(NAME, userRoutine, grid, ParentG
...
+
   t_p => get_global_time_profiler()
   call t_p%start(trim(NAME),__RC__)
   call CHILD_META%t_profiler%start(__RC__)
   call CHILD_META%t_profiler%start('SetService',__RC__)
   gridcomp => META%GET_CHILD_GRIDCOMP(I)
-  call ESMF_GridCompSetServices ( gridcomp, userRoutine, sharedObj=sharedObj,__RC__)
+  call ESMF_GridCompSetServices ( gridcomp, userRoutine, sharedObj=shared_object_library_with_system_dso_suffix,__RC__)
```
tclune, line 4977: Don't we need a `userRC` here?

shared/StringUtils.F90
```diff
@@ -0,0 +1,121 @@
...
+
+   public :: get_file_extension
+   public :: get_file_basename
+   public :: get_system_dso_suffix
+   public :: validate_dso_suffix
+
+contains
+
+   pure integer function find_extension_index(filename) result(dot_index)
+
```
tclune, line 14: Reduce vertical white space.

generic/MAPL_Generic.F90
mathomp4: @tclune No, it was necessary. I don't know why, but I needed them. It's the only way the no-suffix test worked.

Maybe that means I need trims in my StringUtils?

shared/StringUtils.F90
```diff
@@ -0,0 +1,121 @@
...
+      character(len=*), intent(in) :: suffix
+
+      character(len=6), dimension(3), parameter :: allowed_dso = [character(len=6) :: '.so','.dylib','.dll']
+      
+      integer :: i
+      logical :: valid_dso_suffix
+
+      valid_dso_suffix = .false.
+
+      validation_loop: do i = 1, size(allowed_dso)
```
tclune, line 110: Loop label is a bit overkill for such a short procedure.  Indeed, sort of the reason one wants to have short procedures, so that the beginning of the loop is easy to find.

generic/MAPL_Generic.F90
mathomp4: Oh. I suppose so. I'll have to learn how to do that.

generic/MAPL_Generic.F90
mathomp4: Ahh. I have a thought. In the Ocean Grid Comp the `SharedObj` variable being passed in is:
```fortran
character(len=ESMF_MAXSTR)         :: sharedObj
```

Maybe that's the reason?

sanAkel:
works for me! 😄  Thanks!

shared/StringUtils.F90
mathomp4: Removed

generic/MAPL_Generic.F90
tclune: Yes - StringUtils must be careful not to _generate_ padded strings.   But it should also trim any input strings.

generic/MAPL_Generic.F90
weiyuan-jiang: It should be fine with this
call ESMF_GridCompSetServices ( gridcomp, userRoutine, sharedObj=shared_object_library_with_system_dso_suffix, userRC=userRC, __RC__)
_VERIFY(userRC)

weiyuan-jiang: oops, __RC__) should still be there without being formatted here

weiyuan-jiang: _ _ RC _ _

generic/MAPL_Generic.F90
mathomp4: Okay. I think I have this now. Doing some testing.

generic/MAPL_Generic.F90
mathomp4: Fixed.

generic/MAPL_Generic.F90
mathomp4: I've added trims to stringutils and removed these trims.

mathomp4:
Okay. Changes are pushed. I added some `userRC` bits and now the traceback on failure looks like:
```
pe=00295 FAIL at line=04979    MAPL_Generic.F90                         <DSO library was not found>
pe=00295 FAIL at line=00124    GEOS_OceanGridComp.F90                   <status=1>
pe=00295 FAIL at line=04770    MAPL_Generic.F90                         <issue with MAPL_AddChildFromMeta>
pe=00295 FAIL at line=04806    MAPL_Generic.F90                         <status=1>
pe=00294 FAIL at line=00124    GEOS_OceanGridComp.F90                   <status=1>
OGCMSetServices                                236
GCMSetServices                                 239
pe=00295 FAIL at line=04770    MAPL_Generic.F90                         <issue with MAPL_AddChildFromMeta>
pe=00295 FAIL at line=04806    MAPL_Generic.F90                         <status=1>
pe=00295 FAIL at line=04770    MAPL_Generic.F90                         <issue with MAPL_AddChildFromMeta>
pe=00295 FAIL at line=00550    MAPL_CapGridComp.F90                     <status=1>
pe=00295 FAIL at line=00935    MAPL_CapGridComp.F90                     <status=1>
pe=00295 FAIL at line=00260    MAPL_Cap.F90                             <status=1>
pe=00295 FAIL at line=00218    MAPL_Cap.F90                             <status=1>
pe=00295 FAIL at line=00152    MAPL_Cap.F90                             <status=1>
pe=00295 FAIL at line=00129    MAPL_Cap.F90                             <status=1>
pe=00295 FAIL at line=00034    GEOSgcm.F90                              <status=1>
```
I don't know enough to really figure out what the right `ASSERT_` messages should be

mathomp4:
Also: probably we should have some sort of macro for this sort of check:
```fortran
  _ASSERT(userRC==ESMF_SUCCESS .and. STATUS==ESMF_SUCCESS,'needs informative message')
```

All I did here was copy-paste other instances in MAPL_Generic. Perhaps @weiyuan-jiang or @Gvilla1000-nasa could generalize/refactor it, but I didn't want to do a full refactor of generic for this PR.

mathomp4:
> I think the error message is not accurate. Many thing can go wrong, not just 'DSO library was not found'

True. I guess I coded the message for the one way I know it did fail. I'll change the message to whatever others would like.

weiyuan-jiang:
> > I think the error message is not accurate. Many thing can go wrong, not just 'DSO library was not found'
> 
> True. I guess I coded the message for the one way I know it did fail. I'll change the message to whatever others would like.

I still think we should check RC and userRC separately. They tell us different information.

tclune:
If the file is not found or fails to open, I would prefer that ESMF diagnose that, as that is the layer that opens the DSO.  My current understanding is that ESMF is not detecting this problem, and we should thus file a report.    Indeed, I'm a bit surprised that userRC is returning non-zero in this case now that I think about it.

But to @weiyuan-jiang a better message might be "Problem with using the specified DSO."  Meaning that it could be missing or corrupted, ...

weiyuan-jiang:
> 

@tclune , ESMF detects the error and return code "506".  You may have missed Santha's message too.

tclune:
Oh - sorry I've been in meetings.   Can someone remind me here we failed to trap it then?   Inside AddChild?

weiyuan-jiang:
The first error message is right after ESMF_GridCompSetservices in addchild call and the code is correct.  The profiler error is  due to the early return and not initialized properly.  @tclune 

pe=00105 FAIL at line=04947 MAPL_Generic.F90 <status=506>
pe=00105 FAIL at line=00122 GEOS_OceanGridComp.F90 <status=506>
pe=00105 FAIL at line=00145 BaseProfiler.F90

weiyuan-jiang:
> 

We didn't fail to trap the error.

tclune:
Then why did the error go away when Santha used the right name for the DSO?

sanAkel:
> Then why did the error go away when Santha used the right name for the DSO?

I guess "it" (whatever _that_ is) knows to be _nice_! 😮

weiyuan-jiang:
> 

The error is supposed to go away with the right DSO name.   I am confused!

tclune:
Yes - the error should have gone away with the right DSO name, but the _error_ that went away was about _timers_.   So it was the wrong error for the mistake.

sanAkel:
Not sure if it helps or not, please check what does the `gcm log file` (i.e., STDOUT) echo for:

`Character Resource Parameter: MOM6_GEOSPLUG:`

Is something like following:

1. `Character Resource Parameter: MOM6_GEOSPLUG:libMOM6_GEOSPlug.so`
2. `Character Resource Parameter: MOM6_GEOSPLUG:libMOM6_GEOSPlug.dylib`
3. ??

In my testing, when I was tripping on this [issue](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/489), with this code: 

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/f2272cbb6aa7b9e163ac3c4e41a910e94a90885b/GEOSogcm_GridComp/GEOS_OceanGridComp/GEOS_OceanGridComp.F90#L123 

it threw the error that I [mentioned in the issue](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/f2272cbb6aa7b9e163ac3c4e41a910e94a90885b/GEOSogcm_GridComp/GEOS_OceanGridComp/GEOS_OceanGridComp.F90#L123). 

To get past this problem, following @tclune's suggestion, I added following line to my `AGCM.rc`
>MOM6_GEOSPLUG:libMOM6_GEOSPlug.dylib

Of course, you guys are _fixing_ the _problem_!

I would like **report** that with @mathomp4's [MAPL branch](https://github.com/GEOS-ESM/MAPL/pull/1206), I did not look at my gcm log carefully when I was talking to him earlier today (11/22), but with:
**AGCM.rc** having `MOM6_GEOSPLUG:libMOM6_GEOSPlug.dylib`
I got in the log file: `Character Resource Parameter: MOM6_GEOSPLUG:libMOM6_GEOSPlug.so` which makes no sense, I tested twice! @mathomp4 would you please check your log file to make sure if you have something like following line  


>call MAPL_GetResource ( MAPL, sharedObj,  Label="MOM6_GEOSPLUG:", DEFAULT="libMOM6_GEOSPlug-xxxxxxxxxxxxxxxx", __RC__ )

Indeed it echoes that ⤴️ in the log file.

I hope this very _verbose_ comments helps!

weiyuan-jiang:
@sanAkel @tclune The most important thing here is what the first error message is.  Here is mine:

pe=00105 FAIL at line=04947 MAPL_Generic.F90 <status=506>
pe=00105 FAIL at line=00122 GEOS_OceanGridComp.F90 <status=506>
pe=00105 FAIL at line=00145 BaseProfiler.F90

Santha agreed that was what he saw too. The timer error is on the third line.  It is caused by the early return of  mapl_addchild call and the timer is not properly initialized.

mathomp4:
> I would like **report** that with @mathomp4's [MAPL branch](https://github.com/GEOS-ESM/MAPL/pull/1206), I did not look at my gcm log carefully when I was talking to him earlier today (11/22), but with: **AGCM.rc** having `MOM6_GEOSPLUG:libMOM6_GEOSPlug.dylib` I got in the log file: `Character Resource Parameter: MOM6_GEOSPLUG:libMOM6_GEOSPlug.so` which makes no sense, I tested twice! @mathomp4 would you please check your log file to make sure if you have something like following line
> 
> > call MAPL_GetResource ( MAPL, sharedObj,  Label="MOM6_GEOSPLUG:", DEFAULT="libMOM6_GEOSPlug-xxxxxxxxxxxxxxxx", **RC** )
> 
> Indeed it echoes that ⤴️ in the log file.
> 

@sanAkel The print is an interesting thing. I'm going to look at it. If nothing else, I can have the assert message print out what it tried to find.

mathomp4:
Okay. Have some updates. With these we get a bit more verbose. Maybe too much? I don't know.

So, let's work with `libMOM6_GEOSPlug.so`. If that is what is passed in and you are on Linux, it's all good. If you pass in `libMOM6_GEOSPlug.dylib` or `libMOM6_GEOSPlug.dll` you get:
```
 Character Resource Parameter: MOM6_GEOSPLUG:libMOM6_GEOSPlug.dylib
        GENERIC: INFO: MAPL_AddChildFromDSO: Requested shared library libMOM6_GEOSPlug.dylib for component MOM6 has extension .dylib but system uses .so
        GENERIC: INFO: MAPL_AddChildFromDSO: Will use shared library libMOM6_GEOSPlug.so for component MOM6
```
and if you pass in `libMOM6_GEOSPlug`, you get:
```
 Character Resource Parameter: MOM6_GEOSPLUG:libMOM6_GEOSPlug
        GENERIC: INFO: MAPL_AddChildFromDSO: Requested shared library libMOM6_GEOSPlug for component MOM6 has no extension but system uses .so
        GENERIC: INFO: MAPL_AddChildFromDSO: Will use shared library libMOM6_GEOSPlug.so for component MOM6
```
(In the future if we start requiring no suffix, then maybe this becomes a `lgr%debug` message.)

Now if you pass in `libMOM6_GEOSPlug.badsuffix`, you get:
```
pe=00134 FAIL at line=04983    MAPL_Generic.F90                         <Provided [libMOM6_GEOSPlug.badsuffix] for component [MOM6]. Only .so, .dylib, and .dll are allowed DSO suffixes>
```

And finally, if you pass in a library like `libMOM6_GEOSPlugXXXXX.so`:
```
pe=00212 FAIL at line=05000    MAPL_Generic.F90                         <DSO library load of [libMOM6_GEOSPlugXXXXX.so] from [MOM6] failed>
```

mathomp4:
Tested together with #1207 . Zero-diff for GEOS

atrayano (approved):
Looks OK

Gvilla1000-nasa (approved):

Gvilla1000-nasa:
cmake changes are OK

PR #1208, opened 2021-11-29
Changes usage of _ASSERT for SUCCESS error checking and _ASSERT with and conditional

Gvilla1000-nasa:
## Description
Adds new macro for `_ASSERT` when it is used specifically for checks against SUCCESS. The new macro is called `_ASSERT_SUCCESS` and compares two inputs against a success return code (0). 

Additionally, this PR adds a macro for when `_ASSERT ` is used with conditional `.and.` through the use of `_ASSERTS` (which can be renamed to another thing if preferred). 

My original intention with this PR was to create a macro for conditional `.and.` and conditional `.or.` but particularly with `.or.` there are multiple instances with varying number of arguments. I tried experimenting with having optional arguments but was unsuccessful. Because of this, the implementation of the `_ASSERTS` one is limited in usage to only two arguments so perhaps it might be better to just leave the `_ASSERT_SUCCESS` in this PR and remove the other macro?

## Motivation and Context

Usage of a specific _ASSERT case can be converted to a macro.

## How Has This Been Tested?

Tested in intel by changing the error code from good to bad to ensure macro worked as expected. When bad, error message was provided as expected. When good, result was zero diff.

## Types of changes

## Checklist:

### DIFF
--- a/Tests/ExtDataDriverGridComp.F90
+++ b/Tests/ExtDataDriverGridComp.F90
@@ -545,7 +545,7 @@ contains
     integer :: userRc
 
     call ESMF_GridCompRun(this%gc, userRC=userRC,rc=status)
-    _ASSERT(userRC==ESMF_SUCCESS .and. STATUS==ESMF_SUCCESS,'run failed')
+    _ASSERT_SUCCESS(userRC,STATUS,'run failed')
     _RETURN(ESMF_SUCCESS)
 
   end subroutine run
--- a/Tests/ExtDataDriverGridComp.F90
+++ b/Tests/ExtDataDriverGridComp.F90
@@ -544,8 +544,8 @@ contains
     integer :: status
     integer :: userRc
 
-    call ESMF_GridCompRun(this%gc, userRC=userRC,rc=status)
-    _ASSERT_SUCCESS(userRC,STATUS,'run failed')
+    call ESMF_GridCompRun(this%gc, userRC=userRC,__RC__)
+    _VERIFY(userRC)
     _RETURN(ESMF_SUCCESS)
 
   end subroutine run
--- a/base/MAPL_CubedSphereGridFactory.F90
+++ b/base/MAPL_CubedSphereGridFactory.F90
@@ -637,7 +637,8 @@ contains
       if ( (this%target_lon /= MAPL_UNDEFINED_REAL) .and. &
            (this%target_lat /= MAPL_UNDEFINED_REAL) .and. &
            (this%stretch_factor /= MAPL_UNDEFINED_REAL) ) then
-         _ASSERT( (this%target_lat >= -90.0) .and. (this%target_lat <= 90), 'latitude out of range')
+         _ASSERT(this%target_lat >= -90.0, 'Latitude should be greater than -90.0 degrees')
+         _ASSERT(this%target_lat <= 90, 'Latitude should be less than 90.0 degrees')
          this%stretched_cube = .true.
          this%target_lon=this%target_lon*pi/180.d0
          this%target_lat=this%target_lat*pi/180.d0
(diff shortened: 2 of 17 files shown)


tclune (commented):
- For `_ASSERTS` wouldn't it be clearer to implement it as calling `_ASSERT` with an `.and.`?
- Am thinking `_MASSERT` or `_M_ASSERT` might be better name.
- `_ASSERT_SUCCESS`  seems like could just be `_ASSERTS` - is the intent just to make it a bit shorter on the line? (Not even sure it does that.)

Tests/ExtDataDriverGridComp.F90
```diff
@@ -545,7 +545,7 @@ subroutine run(this, rc)
     integer :: userRc
 
     call ESMF_GridCompRun(this%gc, userRC=userRC,rc=status)
-    _ASSERT(userRC==ESMF_SUCCESS .and. STATUS==ESMF_SUCCESS,'run failed')
+    _ASSERT_SUCCESS(userRC,STATUS,'run failed')
```
tclune, line 548:
```suggestion
  call ESMF_GridCompRun(this%gc, userRC=userRC,__RC__)
 _VERIFY(userRC)
```

base/MAPL_CubedSphereGridFactory.F90
```diff
@@ -637,7 +637,7 @@ subroutine check_and_fill_consistency(this, unusable, rc)
       if ( (this%target_lon /= MAPL_UNDEFINED_REAL) .and. &
            (this%target_lat /= MAPL_UNDEFINED_REAL) .and. &
            (this%stretch_factor /= MAPL_UNDEFINED_REAL) ) then
-         _ASSERT( (this%target_lat >= -90.0) .and. (this%target_lat <= 90), 'latitude out of range')
+         _ASSERTS(this%target_lat >= -90.0, this%target_lat <= 90, 'latitude out of range')
```
tclune, line 640:
```suggestion
         _M_ASSERT(this%target_lat >= -90.0, this%target_lat <= 90.0, 'latitude out of range')
```
```suggestion
         _ASSERT2(this%target_lat >= -90.0, this%target_lat <= 90.0, 'latitude out of range')
```

base/MAPL_Profiler.F90
```diff
@@ -348,7 +348,7 @@ subroutine MAPL_TimerModeSet(MODE, RC)
 
       ! Sanity check
       timerMode = mode
-      _ASSERT(timerMode >= MAPL_TimerModeOld .and. timerMode <= MAPL_TimerModeMinMax,'needs informative message')
+      _ASSERTS(timerMode >= MAPL_TimerModeOld, timerMode <= MAPL_TimerModeMinMax,'needs informative message')
```
tclune, line 351:
```suggestion
      _ASSERTS(timerMode >= MAPL_TimerModeOld, timerMode <= MAPL_TimerModeMinMax,'inconsistent value for timerMode')
```

Gvilla1000-nasa:
I believe the latest push addresses all the suggestions.

mathomp4:
> I believe the latest push addresses all the suggestions.

@Gvilla1000-nasa I think we'd also discussed making the non-userRC `.and.` asserts into two separate asserts? Though I suppose that could be a separate PR.

tclune (approved):

Gvilla1000-nasa:
> > I believe the latest push addresses all the suggestions.
> 
> @Gvilla1000-nasa I think we'd also discussed making the non-userRC `.and.` asserts into two separate asserts? Though I suppose that could be a separate PR.

Working on it right now. Found a missing .and. with error checking as well

base/MAPL_sun_uc.F90
```diff
@@ -2994,7 +2994,8 @@ subroutine MAPL_SunGetDaylightDurationMax(ORBIT,LATS,DAYL,currTime,CLOCK,RC)
          ! obliquity fixed in this case
          OBQ = ORBIT%OB * (MAPL_PI/180.)
       endif
-      _ASSERT(0. <= OBQ .and. OBQ < MAPL_PI, 'Strange obliquity!')
+      _ASSERT(OBQ >= 0, 'Obliquity less than 0 detected!')
```
tclune, line 2997: Not going to ask for changes here, but we really need to develop a consistent policy about how to word such messages.  And then _gradually_ propagate.

tclune (approved):

PR #1217, opened 2021-12-06
Feature/wjiang/optimized writing node

weiyuan-jiang:
The way how to choose a processor to write data is changed. Previously, a writing processor is chosen as long as it is idle. That may lead to a situation that some nodes are crowded but some others are idle.  Now, an idle processor is chosen from a node with the most idle processors. This is the first step to optimize the o-server. The next step would be using memory size as a criteria to choose writers.
## Types of changes

## Checklist:

### DIFF
--- a/pfio/MultiGroupServer.F90
+++ b/pfio/MultiGroupServer.F90
@@ -419,338 +419,352 @@ contains
    subroutine start_back(this, rc)
      class (MultiGroupServer),target, intent(inout) :: this
      integer, optional, intent(out) :: rc
+     integer, parameter :: stag = 6782
      integer :: ierr
 
      integer :: MPI_STAT(MPI_STATUS_SIZE)
-     integer :: i, j, idle, no_job, idle_worker
      integer :: collection_id 
-     integer :: msg_size, back_local_rank
-     
-     type (MessageVectorIterator) :: iter
-     class (AbstractMessage), pointer :: msg
-     class (ServerThread),pointer :: thread_ptr
-     integer, parameter :: stag = 6782
-     integer, allocatable :: buffer_fmd(:)
-
-     integer, pointer :: g_1d(:), l_1d(:), g_2d(:,:), l_2d(:,:), g_3d(:,:,:), l_3d(:,:,:)
-     integer, pointer :: g_4d(:,:,:,:), l_4d(:,:,:,:), g_5d(:,:,:,:,:), l_5d(:,:,:,:,:)
-     integer :: msize_word, d_rank, request_id
-     integer :: s0, e0, s1, e1, s2, e2, s3, e3, s4, e4, s5, e5
-     type (StringAttributeMap) :: vars_map
-     type (StringAttributeMapIterator) :: var_iter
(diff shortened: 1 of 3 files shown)


pfio/MultiGroupServer.F90
```diff
@@ -447,65 +447,103 @@ subroutine start_back(this, rc)
...
-                  endif
-               enddo
-
-               ! if all workers are busy, wait for one
-               if (idle_worker == 0) then 
-                  call MPI_recv( idle_worker, 1, MPI_INTEGER, &
-                      MPI_ANY_SOURCE, stag, this%back_comm, &
-                      MPI_STAT, ierr)
-               endif
+
```
tclune, line 483: This subroutine is getting long.   I encourage you to try to move some of this logic into well-named helper procedures to keep the logic in reasonably sized chunks.

tclune (changes requested):
Complexity is just getting a bit high.  Time to refactor while it is fresh in your head.

bena-nasa:
@weiyuan-jiang  Is there any danger here:
https://github.com/GEOS-ESM/MAPL/blob/feature/wjiang/optimized_writing_node/pfio/MultiGroupServer.F90#L486
since technically the "first" node has one less writer. I presume this works but it's getting late in the day. I'm trying to do something similar in my ESMF server prototype and was looking at your code when I saw this.

pfio/MultiGroupServer.F90
```diff
@@ -418,301 +419,352 @@ end subroutine receive_output_data
...
+                flag = .false.
+                call MPI_Iprobe( local_rank, stag, this%back_comm, flag, MPI_STAT, ierr)             
+                if (flag) then
+                  call MPI_recv(idle_worker, 1, MPI_INTEGER, &
+                                local_rank, stag, this%back_comm, &
+                                MPI_STAT, ierr)
+                  _ASSERT(local_rank == idle_worker, "local_rank and idle_worker should match")
+                  node_rank = this%node_ranks(local_rank)
+                  num_idlePEs(node_rank) = num_idlePEs(node_rank) + 1
+                  nth_writer = mod(local_rank, nwriter_per_node)
```
bena-nasa, line 486: See previous comment I made. Should have made it here.

pfio/MultiGroupServer.F90
weiyuan-jiang: I have re-organized the codes. And it seems more clear now

pfio/MultiGroupServer.F90
weiyuan-jiang: I re-organized the codes before I saw your comment. Can you comment again and let me what is the problem?

weiyuan-jiang: I think the fact that the first node has one processor less has been carefully taken care of all the time.

tclune (approved):
Definitely an improvement.   But there are still are some longish procedures.   

However, we want to get this released for benchmarking, so approving ...

tclune (approved):

PR #1237, opened 2021-12-16
Feature/wjiang/get short name work around

weiyuan-jiang:
The intel compiler does not deallocate the allocated string in the condition block when optimized compiling flag is used. That leads to crash because the memory is used up. 
## Types of changes

## Checklist:

### DIFF
--- a/generic/VarConn.F90
+++ b/generic/VarConn.F90
@@ -147,6 +147,7 @@ contains
       integer                               :: I
       class(Logger), pointer :: lgr
       type(VarConnType), pointer :: conn
+      character(len=ESMF_MAXSTR)            :: from_name, to_name
 
       associate (conn_v => this%conn_v)
         checkUnused = .true.
@@ -157,9 +158,11 @@ contains
            if (.not. conn%USED) then
               checkUnused = .false.
               lgr => logging%get_logger('MAPL.GENERIC')
+              from_name = conn%FROM%get_short_name()
+              to_name   = conn%TO%get_short_name()
               call lgr%error( &
                    'SRC_NAME: <%a~>  DST_NAME: <%a~> is not satisfied', &
-                   trim(conn%FROM%get_short_name()),trim(conn%TO%get_short_name()))
+                   trim(from_name),trim(to_name))
            end if
         end do
       end associate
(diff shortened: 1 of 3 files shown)


weiyuan-jiang:
Hopefully, this patch [ ](https://github.com/GEOS-ESM/MAPL/tree/feature/wjiang/v2.6.3_patch) can solve your issue temporarily before you upgradeto the latest MAPL.

weiyuan-jiang:
Should this be "exit" ? @tclune 
https://github.com/GEOS-ESM/MAPL/blob/eb3eebb18c47299e3aa87879ac49feeb7a835229/generic/MAPL_Generic.F90#L7087

weiyuan-jiang:
Should we add if (i == j) cycle between the lines? @tclune 
https://github.com/GEOS-ESM/MAPL/blob/eb3eebb18c47299e3aa87879ac49feeb7a835229/generic/MAPL_Generic.F90#L7098-L7099

generic/VarConn.F90
```diff
@@ -147,6 +147,7 @@ logical function checkUnused(this)
       integer                               :: I
       class(Logger), pointer :: lgr
       type(VarConnType), pointer :: conn
+      character(len=ESMF_MAXSTR)            :: from_name, to_name
```
tclune, line 150: Of course part of the functionality of pFlogger is to _avoid_ needing to declare additional variables like this.  Sigh.

tclune (dismissed):

tclune:
> Should this be "exit" ? @tclune
> 
> https://github.com/GEOS-ESM/MAPL/blob/eb3eebb18c47299e3aa87879ac49feeb7a835229/generic/MAPL_Generic.F90#L7087

Really the question should go to @atrayano    I'm wary of changing the logic in such a fundamental way.  Why do you think it should be `exit`?

tclune:
> Should we add if (i == j) cycle between the lines? @tclune
> 
> https://github.com/GEOS-ESM/MAPL/blob/eb3eebb18c47299e3aa87879ac49feeb7a835229/generic/MAPL_Generic.F90#L7098-L7099

Again really a question for @atrayano

mathomp4 (approved):
Passes with GCM.

mathomp4:
This passed the GCM. Once the CI is happy, I'll merge and make a dev to main pr for release.

PR #1245, opened 2021-12-18
change some associate block to fix GNU crash

weiyuan-jiang:
gfortran crashes at blocks that  associate an allocated string.  Those blocks are changed to fix the crash. There are some minor cleanups in this PR too.
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- gfortran can not assoicate an allocated string. Such blocks are changed 
+
 ### Added
 
 ### Changed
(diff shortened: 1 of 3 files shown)


atrayano (changes requested):
The variable 'extension' has not been initialized yet when used in the __ASSERT at line 4852

atrayano (approved):

mathomp4:
Note: I'm going to tag @tclune here. I think he is more a fan of snake case (`parent_gc`) than camel case (`parentGC`). So he might be doing a bit of a posteriori work on this...

weiyuan-jiang:
Here parentGC is just a little bit more consistent among several related subroutines.  I agree the snake style makes more sense in Fortran in general.

tclune:
Already done, but we often have to preserve existing oddities with argument names for backwards compatibility.  Probably not even worth risking for MAPL 3 unless someone is very confident ...

Issue #1244, opened 2021-12-17
MPICH4: [WARNING] yaksa: 3 leaked handle pool objects


mathomp4:
Recently I decided to try out MPICH with GEOS. This was after the issue the GCHP folks saw with someone running GCHP with MPICH3.

So, I grabbed MPICH 4.0b1 and with the help of @hzhou on the [MPICH mailing list](https://lists.mpich.org/pipermail/discuss/2021-December/006307.html) (see also https://github.com/pmodels/mpich/pull/5728), I got a working MPICH! I could run HelloWorld on Discover, so I built Baselibs and no errors were seen.

I then found that MPICH would crash if you (accidentally) set:
```
  ROMIO_FSTYPE_FORCE="gpfs:"
```
and ran GEOS. (I've asked [the MPICH Discuss list](https://lists.mpich.org/pipermail/discuss/2021-December/006312.html) about that.)

Okay. Worked around that and GEOS runs! Huzzah! And is zero-diff to Open MPI! Double huzzah!

But, at the very end of the run (after all the timers print) I see:
```
[WARNING] yaksa: 3 leaked handle pool objects
```

Obviously, I'll make sure to ask on the MPICH list about this, but since the ROMIO issue might presage a "You built MPICH wrong", I'm waiting to hear back on that before asking there.

Still, I thought I should add an issue here since "leaked handle pool objects" at the end of a run sounds like a possible @tclune or @weiyuan-jiang sort of "we might have not cleaned up some MPI" sort of thing...

raffenet:
It might indicate you didn't call `MPI_Type_free` on all user-defined datatypes before `MPI_Finalize`.

mathomp4:
> It might indicate you didn't call `MPI_Type_free` on all user-defined datatypes before `MPI_Finalize`.

@raffenet Is there a way to make yaksa (or whatever) more verbose? Like can it give hints where we might find the un-free'd types? I have an idea though...

raffenet:
I would try adding `--enable-g` to your MPICH configuration. That turns on MPI object handle debugging, and should be able to tell you if there are unfreed objects at finalize time.

mathomp4:
> I would try adding `--enable-g` to your MPICH configuration. That turns on MPI object handle debugging, and should be able to tell you if there are unfreed objects at finalize time.

Ooh. Rebuild time! Okay. I've put it on my to-do list. Now to find out if I need to rebuild all our dependent libraries, or if just our code is enough... :)

[...]

PR #1261, opened 2022-01-04
initailize and finalize profiler in MAPL

weiyuan-jiang:
Initailize and finalize  global profiler in MAPL initialize and finalize
## Types of changes

## Checklist:

### DIFF
--- a/base/ApplicationSupport.F90
+++ b/base/ApplicationSupport.F90
@@ -39,8 +39,8 @@ module MAPL_ApplicationSupport
       call initialize_pflogger(comm=comm_world,logging_config=logging_configuration_file,rc=status)
       _VERIFY(status)
 #endif
-      call start_global_profiler(comm=comm_world,rc=status)
-      _VERIFY(status)
+      call initialize_global_time_profiler(comm=comm_world)
+      call start_global_time_profiler()
       _RETURN(_SUCCESS)
 
    end subroutine MAPL_Initialize
@@ -59,9 +59,10 @@ module MAPL_ApplicationSupport
       else
          comm_world=MPI_COMM_WORLD
       end if
-      call stop_global_profiler()
+      call stop_global_time_profiler()
       call report_global_profiler(comm=comm_world)
       call finalize_pflogger()
+      call finalize_global_time_profiler()
 
    end subroutine MAPL_Finalize
 
@@ -140,32 +141,6 @@ module MAPL_ApplicationSupport
 
    end subroutine initialize_pflogger
 #endif
-
-   subroutine start_global_profiler(unusable,comm,rc)
-      class (KeywordEnforcer), optional, intent(in) :: unusable
-      integer, optional, intent(in) :: comm
-      integer, optional, intent(out) :: rc
-      class (BaseProfiler), pointer :: t_p
-      integer :: world_comm
-
-      _UNUSED_DUMMY(unusable)
-      if (present(comm)) then
-         world_comm = comm
-      else
-         world_comm = MPI_COMM_WORLD
-      end if
-      t_p => get_global_time_profiler()
-      t_p = TimeProfiler('All', comm_world = world_comm)
-      call t_p%start()
-      _RETURN(_SUCCESS)
-   end subroutine start_global_profiler
-
-   subroutine stop_global_profiler()
-      class (BaseProfiler), pointer :: t_p
-
-      t_p => get_global_time_profiler()
-      call t_p%stop('All')
-   end subroutine stop_global_profiler
 
    subroutine report_global_profiler(unusable,comm,rc)
       class (KeywordEnforcer), optional, intent(in) :: unusable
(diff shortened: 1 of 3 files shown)


weiyuan-jiang:
@mathomp4 Would you please test  this PR for me that the issue #1244 can be addressed by this PR?

base/ApplicationSupport.F90
```diff
@@ -39,8 +39,8 @@ subroutine MAPL_Initialize(unusable,comm,logging_config,rc)
       call initialize_pflogger(comm=comm_world,logging_config=logging_configuration_file,rc=status)
       _VERIFY(status)
 #endif
-      call start_global_profiler(comm=comm_world,rc=status)
-      _VERIFY(status)
+      call initialize_global_time_profiler(comm=comm_world)
+      call start_global_time_profiler()
```
tclune, line 43: No - we need a pure initialize() and finalize() procedure in the profiling package.  Those can start/stop the global timer.   This gives us maximum flexibility for the future and also makes the packages consistent.

weiyuan-jiang: We may need a counter class to trace how many profilers are created and the last one releases the types and op ?

tclune: That's what the finalize() procedure is for.  It is up to the user to call it "after" all the profilers are done.    I don't think we need a counter (or profiler set) for this purpose, but could be considered for later.  For now, just do the simple thing which is all-but-equivalent to what you've done above.  It allows the profiler to evolve without changing interfaces.

weiyuan-jiang: Do you mean put start and stop inside the initialize and finalize? We probably can't

tclune: No.  Create 2 procedures for the profiling package:   initialize() and finalize().      Initialize() will call initialize_global_time_profiler(...)  and finalize() will call finalize_global_time_profiler().   Then in Application support just call initialize() and finalize().  This one level of indirection lets us add complexity later in the profiling package without having to touch application_support() again.

weiyuan-jiang: Where should we put these two subroutines?

tclune: https://github.com/GEOS-ESM/MAPL/blob/main/profiler/MAPL_Profiler.F90

weiyuan-jiang: I have pushed my changes

base/ApplicationSupport.F90
```diff
@@ -39,8 +39,8 @@ subroutine MAPL_Initialize(unusable,comm,logging_config,rc)
       call initialize_pflogger(comm=comm_world,logging_config=logging_configuration_file,rc=status)
       _VERIFY(status)
 #endif
-      call start_global_profiler(comm=comm_world,rc=status)
-      _VERIFY(status)
+      call initialize_profiler(comm=comm_world)
```
tclune, line 42: I think this will need an rc, because the global profiler might return an invalid status.

base/ApplicationSupport.F90
```diff
@@ -39,8 +39,8 @@ subroutine MAPL_Initialize(unusable,comm,logging_config,rc)
...
+      call start_global_time_profiler()
```
tclune, line 43: This line should be inside initialize() of the profiler package.

tclune (changes requested):
close

base/ApplicationSupport.F90
tclune: I may be wrong about this.  Push back pleas.

profiler/MAPL_Profiler.F90
```diff
@@ -39,5 +39,15 @@ module mapl_Profiler
    use mapl_SeparatorColumn
    
    implicit none
-   
+contains 
+
+   subroutine initialize_profiler(comm)
```
tclune, line 44: Should be called `initialize()`  Use Fortran rename in the other layer to call it with the longer name.

profiler/MAPL_Profiler.F90
```diff
@@ -39,5 +39,15 @@ module mapl_Profiler
...
+      integer, optional, intent(in) :: comm
+      call initialize_global_time_profiler(comm = comm)
+   end subroutine initialize_profiler
+
+   subroutine finalize_profiler()
```
tclune, line 49: Same as with initialize - should be the short name here.

tclune (dismissed):

base/ApplicationSupport.F90
weiyuan-jiang: there is no rc in initialize and finalize. I added rc to start and stop

profiler/MAPL_Profiler.F90
weiyuan-jiang: changed

profiler/MAPL_Profiler.F90
weiyuan-jiang: changed

weiyuan-jiang:
@mathomp4 , Any special g5_module for me to test if those types and op get freed successfully?

tclune (dismissed):

weiyuan-jiang:
This message " [WARNING] yaksa: 3 leaked handle pool objects" disappears.  So the types are freed at the end

PR #1279, opened 2022-01-06
Add retry to CI build

mathomp4:
## Description

Every so often, the Intel CircleCI build will fail. My current guess is this is due to memory pressure on the `large` VM which has 8 GB of RAM. Since we build with `make -j$(nproc)` this could happen. So what this PR will try and do is add an extra `make -j$(nproc) install` step when the first one fails. This is in the hope that maybe the memory pressure will be less when running the second time (since it might take a different path through the make graph).

Of course, this is impossible to test as the failure is sporadic, but the CI in MAPL is at a much higher frequency than other repos, so this is the best place to start.

## Motivation and Context

Less time needed to re-run CI jobs.

## Types of changes

## Checklist:


mathomp4:
Annnnnd...of course it didn't trigger this time.

tclune (approved):

PR #1293, opened 2022-01-11
Fixes #1292. Fix for Logger Message

### LINKED ISSUE #1292: Issue with MOM6 on macOS: Logger Message
When trying to test out the C12 MOM6 on Parcel (my MacBook), the model crashed with:
```
 Character Resource Parameter: MOM6_GEOSPLUG:libMOM6_GEOSPlug.so
 FormatString.F90         173
 format_vector() - additional unprocessed arguments.
1
```
Working with @tclune we found this was due to a logger message call:

https://github.com/GEOS-ESM/MAPL/blob/f98fae5c158838b991eeda51ebe3e7d93f4d6fea/generic/MAPL_Generic.F90#L4799-L4800

Namely the single quotes. @tclune will look at pFlogger but for now, I'll make a PR for a workaround.

mathomp4:
## Description

Small changes to logger message due to unexpected single quote behavior in pFlogger.

## Related Issue

Closes #1292 

## Motivation and Context

Let's us run MOM6 (really any `AddChildDSO`) on non-.so systems.

## How Has This Been Tested?

I ran on my laptop with Intel. I want to test with GNU as well.

## Types of changes

## Checklist:

### DIFF
--- a/generic/MAPL_Generic.F90
+++ b/generic/MAPL_Generic.F90
@@ -759,7 +759,7 @@ contains
 
    ! !INTERFACE:
    recursive subroutine MAPL_GenericInitialize ( GC, import, EXPORT, CLOCK, RC )
-      
+
       !ARGUMENTS:
       type(ESMF_GridComp), intent(INOUT) :: GC     ! Gridded component
       type(ESMF_State),    intent(INOUT) :: IMPORT ! Import state
@@ -1656,7 +1656,7 @@ contains
          if (state%requested_services%size()>0) then
             call FillRequestBundle(state%requested_services, state%get_internal_state(), __RC__)
          end if
-         
+
          ! process any service connections
          call MAPL_ProcessServiceConnections(state, __RC__)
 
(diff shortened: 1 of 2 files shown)


generic/MAPL_Generic.F90
```diff
@@ -4796,7 +4796,7 @@ recursive integer function AddChildFromDSO(gc, name, userRoutine, grid, sharedOb
 
       if (.not. is_valid_dso_name(SharedObj)) then
          lgr => logging%get_logger('MAPL.GENERIC')
-         call lgr%warning("AddChildFromDSO: changing shared library extension '%a~' to system specific extension '%a~'.", &
+         call lgr%warning("AddChildFromDSO: changing shared library extension from %a~ to system specific extension %a~", &
               extension, SYSTEM_DSO_EXTENSION)
```
tclune, line 4800:
```suggestion
         call lgr%warning("AddChildFromDSO: changing shared library extension from %a~ to system specific extension %a~", &
              "'"//extension//"'", "'"//SYSTEM_DSO_EXTENSION//"'")
```

mathomp4:
@tclune Well, I get the quotes now:
```
        GENERIC: WARNING: AddChildFromDSO: changing shared library extension from '.so   ' to system specific extension '.dylib'
```
Should we perhaps add a `trim()` to the `extension`? Or maybe we should trim elsewhere?

tclune:
Ah - extensions is bizarrely declared as len=6.  it should be a deferred length string and that will fix the issue.

mathomp4:
> Ah - extensions is bizarrely declared as len=6. it should be a deferred length string and that will fix the issue.

Done. Seems to work.

atrayano (approved):

PR #1305, opened 2022-01-13
fixes #1294

### LINKED ISSUE #1294: Input regridding does not identify file-defined missing values
I recently investigated a reported crash in GCHP when using an optional emissions inventory. In looking closer at the issue I stumbled upon a bug in MAPL. The emissions file contains a field with missing values defined in the NetCDF file as -1e-31. Rather than treat them as missing, MAPL regridded them as if they were valid. The resulting negative values for some grid cells caused HEMCO to crash.

The proper behavior should be recognition of missing values via the missing values attribute per field in the NetCDF file, and then appropriate handling during regridding.

bena-nasa:
This fixes #1294 
When a file is read in GriddedIO (used by ExtData and my standalone bundle read routine used by Regrid_Util.x) it checks what the value of _FillValue is.
If this differs from MAPL_UNDEF, the _FillValue from the file is replace with MAPL_UNDEF after it is read but before anything else is done with the data.
Note that for now I'm only looking for _FillValue, we may want to expand this, for example what is the file also has missing_value, what is that differs from the _FillValue who wins?
I replied to the issue asking this but no one said anything. In any case this is zero-diff for GEOS and I took an existing file, changed the _FillVlaue with NCO, then confirmed if I regridded the original file and the new file with the different fill value with Regrid_Util.x I got the same answer so seems to be working to me..

## Types of changes

## Checklist:

### DIFF
--- a/base/FileMetadataUtilities.F90
+++ b/base/FileMetadataUtilities.F90
@@ -21,6 +21,11 @@ module MAPL_FileMetadataUtilsMod
       procedure :: get_level_name
       procedure :: is_var_present
       procedure :: get_file_name
+      procedure :: var_get_missing_value
+      procedure :: var_has_missing_value
+      procedure :: var_has_attr
+      procedure :: get_var_attr_real32
+      procedure :: get_var_attr_string
    end type FileMetadataUtils
 
    interface FileMetadataUtils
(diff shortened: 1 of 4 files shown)


base/FileMetadataUtilities.F90
```diff
@@ -46,6 +51,110 @@ subroutine create(this,metadata,fname)
...
+
+      var => this%get_variable(var_name,_RC)
+      _ASSERT(associated(var),"no variable named "//var_name//" in file")
+      var_has_missing_value = var%is_attribute_present("_FillValue")
+
+      _RETURN(_SUCCESS)
+   end function var_has_missing_value
+
+   logical function var_has_attr(this,var_name,attr_name,rc)
+      class(FileMetadataUtils), intent(inout) :: this
```
tclune, line 89: I'm slightly surprised that this functionality does not already exist.

base/FileMetadataUtilities.F90
```diff
@@ -46,6 +51,110 @@ subroutine create(this,metadata,fname)
...
+      _RETURN(_SUCCESS)
+   end function var_has_attr
+
+   function get_var_attr_real32(this,var_name,attr_name,rc) result(attr_real32)
+      class(FileMetadataUtils), intent(inout) :: this
+      character(len=*), intent(in) :: var_name
+      character(len=*), intent(in) :: attr_name
+      integer, optional, intent(out) :: rc
+
+      real(REAL32) :: attr_real32
```
tclune, line 109: I prefer the declaration of the function result to be before the dummy arguments.   But only make a change if there is something else substantive in the review.

base/FileMetadataUtilities.F90
```diff
@@ -46,6 +51,110 @@ subroutine create(this,metadata,fname)
...
+      _RETURN(_SUCCESS)
+   end function get_var_attr_real32
+
+   function get_var_attr_string(this,var_name,attr_name,rc) result(attr_string)
+      class(FileMetadataUtils), intent(inout) :: this
+      character(len=*), intent(in) :: var_name
+      character(len=*), intent(in) :: attr_name
+      integer, optional, intent(out) :: rc
+
+      character(len=:), allocatable :: attr_string
```
tclune, line 138: Same placement preference as above.

tclune (approved):
A few issues that might prompt one more iteration.  But approving otherwise.

base/FileMetadataUtilities.F90
bena-nasa: Nope, didn't exist.

base/FileMetadataUtilities.F90
bena-nasa: I'm going to keep it things as they are rather than futz with checking for missing_value. I'll remember this in the future

base/FileMetadataUtilities.F90
```diff
@@ -21,6 +21,11 @@ module MAPL_FileMetadataUtilsMod
       procedure :: get_level_name
       procedure :: is_var_present
       procedure :: get_file_name
+      procedure :: var_get_missing_value
+      procedure :: var_has_missing_value
+      procedure :: var_has_attr
+      procedure :: get_var_attr_real32
```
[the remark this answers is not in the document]
bena-nasa: I might as well add this (as well as INT32 and INT64).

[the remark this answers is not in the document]
bena-nasa: Ok, added REAL64,INT32,INT64 and fixed the declaration the return variable.

tclune (approved):

lizziel:
@bena-nasa, the current implementation only works if the input is float (see code [here](https://github.com/GEOS-ESM/MAPL/blob/91f33407f1e10281f7790c5f5ad4537fc6429993/base/FileMetadataUtilities.F90#L69-L71)). This is a problem for GCHP since some of our input files use double precision. Do you plan on extending the _FillValue handling to inputs beyond REAL32?

PR #1313, opened 2022-01-25
Hotfix/bmauer/fixes #1311

### LINKED ISSUE #1311: Issue with ungridded dimension of size 1 in History Output
A user running GOCART2G was finding a problem with the History output. Essentially History was saying a variable was being defined as 3D when creating the variable metadata, but there was no "lev" dimension in the metadata already.
On inspection, the issue appears to be that there is some field created when running GOCART2D as a 3D field with with an ungridded dimension whose size is the number of wavelengths specified here:
```
wavelengths_for_vertically_integrated_aop_in_nm: 550 # must be included in LUT
```
and so if there is 1, the variable has an ungridded dimension of size 1, if two it is two etc ...
The actual variable in question is this in the History.rc file:
```
'TOTEXTTAU'   , 'GOCART2G'
```
and here is the top of the particular collection:
```
  tavg2d_aer_x.format:     'CFIO',
  tavg2d_aer_x.template:   '%y4%m2%d2_%h2%n2z.nc4',
  tavg2d_aer_x.archive:    '%c/Y%y4',
  tavg2d_aer_x.mode:       'time-averaged',
  tavg2d_aer_x.frequency:  010000,
  tavg2d_aer_x.ref_time:   210000,
  tavg2d_aer_x.grid_label: PC96x49-DC,
  tavg2d_aer_x.splitField: 1,
  tavg2d_aer_x.fields:         'LWI'         , 'SURFACE'     ,
        'TOTEXTTAU'   , 'GOCART2G'    ,
                               'TOTSCATAU'   , 'GOCART2G'    ,
                               'TOTANGSTR'   , 'GOCART2G'    ,
                               'DUEM'        , 'DU'      ,
```
It looks like the logic I have for creating the vertical or ungridded dimension in the metadata was getting bypassed because the ungridded size was 1 and was saying that was no "lev" dimension but the pointer for the variable is 3d so the variable definition includes the "lev" dimension.
This is a super trivial fix so that if the ungridded dim size is 1 you get a "lev" dimension of size 1.

bena-nasa:
Fixes #1311 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - io profiler is fixed due to the change of profiler finalize
 - Files read by GriddedIOMod now respect the file defined _FillValue rather than assuming it is MAPL_UNDEF
+- Fix bug so that fields with an ungridded dimension of size 1 can be written via History
 
 ### Added
 
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -3684,14 +3684,9 @@ contains
             end if
          end do
          if (count == 0) then
-            if (n > 1) then
-               do i=1,n
-                  write(splitNameArray(i),'(A,I3.3)') trim(aliasName), i
-               end do
-            else
-               count = count+1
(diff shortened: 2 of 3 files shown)


atrayano (approved):
Looks OK

PR #1329, opened 2022-01-31
Revert "Merge pull request #1326 from GEOS-ESM/feature/wjiang/exit_af…

mathomp4:
…ter_connected"

This reverts commit 852cd94145d2078b02d755142fa73311c31c283c, reversing
changes made to 96f4916c4bd322845d9a81b113cc0893e68f58b2.

## Description

As said above, reverts #1326

This was created by:
```
git revert -m1 852cd94145d2078b02d755142fa73311c31c283c
git checkout -b feature/mathomp4/revert-1326
```

~Currently testing.~ Tested and works with GEOSgcm

## Related Issue

Closes #1328 

## Motivation and Context

Let's MAPL `develop` work.

## How Has This Been Tested?

Ran a 1-day run of GEOSgcm at C24 and C24-Replay. Both worked.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [x] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [x] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #1330, opened 2022-01-31
bug fix. remove cycle

weiyuan-jiang:
remove buggy cycle
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed issue in `CMakePresets.json` where Ninja presets were broken
-- Removed cycle in MAPL_Generic.F90
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed issue in `CMakePresets.json` where Ninja presets were broken
+- Removed cycle in MAPL_Generic.F90
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed issue in `CMakePresets.json` where Ninja presets were broken
+- Removed cycle in MAPL_Generic.F90
```
tclune, line 13: This is not a very useful note for the ChangeLog.   Users can look at "diff" to see the deleted line.   Here we should explain the why?  What was breaking what were the symptoms or consequences.

tclune (changes requested):
See inline comment.

CHANGELOG.md
weiyuan-jiang: That line was added by me. After thinking twice, I believe that "I" and "J" are from different sources and should not be cycled when they are the same.

mathomp4:
I've put a blocker on my #1329 PR for now because this PR is here.

My question is: did you run GEOSgcm with this fix? I'd rather not merge it in if you haven't (the checkbox is not checked).

weiyuan-jiang:
> 

Yes, I have tested GEOSgcm and it runs

CHANGELOG.md
mathomp4: @weiyuan-jiang You can probably remove this line and then edit line 19:
```
- A small performance improvement. cycle => exit in MAPL_Generic.F90
```
to be whatever you think this PR plus the first on should reflect.

mathomp4: Yeah. I'll just remove this and add "skip changelog".

mathomp4 (approved):

Issue #1334, opened 2022-02-04
Missing gFTL v2 target for MAPL.profiler prevents MAPL from building


rmontuoro:
The following `MAPL.profiler` file includes a template from gFTL v2:
https://github.com/GEOS-ESM/MAPL/blob/28a874ee60734690d0bbdc3fb266a64146fe3ce3/profiler/MeterNodeStack.F90#L9

However, the `GFTL::gftl-v2` target is missing in:
https://github.com/GEOS-ESM/MAPL/blob/28a874ee60734690d0bbdc3fb266a64146fe3ce3/profiler/CMakeLists.txt#L52

mathomp4:
Hmm. I thought `GFTL::gftl` brought in both versions...

tclune:
I'm also curious why this is not a problem on our end.

tclune:
Checking - no GFTL::gftl is just an alias for v1.    It will switch to v2 when gFTL 2.0 is released.   Still confused why we don't get the error though.

tclune:
@weiyuan-jiang Can you do a hotfix off of main to add GFTL::gftl-v2 in the esma_add_library call in the profiler directory.

@rmontuoro Please let us know if you need a patch to any older release.

weiyuan-jiang:
I saw oomph has replaced GFTL::gftl with  GFTL::gftl-v2. So I removed GFTL::gftl in profiler. It builds successfully.

tclune:
No - both are needed.  Some containers still use the older v1 (which is what "gftl" brings in).   We don't yet know why our build seems to work either way.

weiyuan-jiang:
> 

Oops, I remove it.  Let me see if I can have a fresh build

weiyuan-jiang:
@tclune Since there is no problem building profiler without v1 dependency, I prefer to removing it now

rmontuoro:
@mathomp4 -- here are the CMake settings we're using to build MAPL standalone:

```
cmake .. \
      -DCMAKE_INSTALL_PREFIX=$prefix \
      -DCMAKE_MODULE_PATH="${ESMA_CMAKE_ROOT};${CMAKEMODULES_ROOT}/Modules;${ECBUILD_ROOT}/share/ecbuild/cmake" \
      -DCMAKE_BUILD_TYPE=Release \
      -DBUILD_WITH_FLAP=OFF \
      -DBUILD_WITH_PFLOGGER=OFF \
[...]

PR #1335, opened 2022-02-04
fixed profiler dependency on new gftl-v2

weiyuan-jiang:
Fix profiler dependency on gftl-v2
## Types of changes

## Checklist:

### DIFF
--- a/generic/CMakeLists.txt
+++ b/generic/CMakeLists.txt
@@ -56,7 +56,7 @@ find_package(GFTL_SHARED REQUIRED)
 esma_add_library(${this}
   SRCS ${srcs}
   DEPENDENCIES MAPL.oomph MAPL.shared MAPL.profiler MAPL.base
-               PFLOGGER::pflogger GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1 GFTL::gftl GFTL::gftl-v2 GFTL::gftl-v1
+               PFLOGGER::pflogger GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1  GFTL::gftl-v2 GFTL::gftl-v1
   TYPE ${MAPL_LIBRARY_TYPE}
   )
 target_include_directories (${this} PUBLIC
--- a/generic/CMakeLists.txt
+++ b/generic/CMakeLists.txt
@@ -56,7 +56,7 @@ find_package(GFTL_SHARED REQUIRED)
 esma_add_library(${this}
   SRCS ${srcs}
   DEPENDENCIES MAPL.oomph MAPL.shared MAPL.profiler MAPL.base
-               PFLOGGER::pflogger GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1 GFTL::gftl
+               PFLOGGER::pflogger GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1 GFTL::gftl GFTL::gftl-v2 GFTL::gftl-v1
   TYPE ${MAPL_LIBRARY_TYPE}
   )
 target_include_directories (${this} PUBLIC
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add explicit dependence of gFTL whenerver it is used
+
 ### Changed
 
 ### Removed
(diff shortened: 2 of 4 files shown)


weiyuan-jiang:
This PR addresses issue #1334

mathomp4:
Blocking for just a second so I can fix up the branch for release.

mathomp4:
Actually, let me keep blocked as I run a GEOSgcm test.

mathomp4:
Zero-diff and builds.

weiyuan-jiang:
Add label DNA back. I want to understand what happens here

weiyuan-jiang:
I guess I figure out what happened here. The include path can be inherited from dependence.  Since profiler depends on mapl.shared, it inherits its include path. Mapl.shared depends on pflogger which has all gFTL's paths.  In our GEOSgcm, we build and use pFlogger but in UFS, only pFlogger_stub is used.  That is why in UFS, v2 path is missing. This PR should work for UFS, but we may add fake dependence to pfloger_stub to make it really mimic pFlogger.

tclune:
@weiyuan-jiang Well done!

And this speaks directly to why it is so important for us to list all the explicit dependencies of each layer.    Otherwise we get surprising results when unrelated layers are changed elsewhere in the model.

mathomp4:
> I guess I figure out what happened here. The include path can be inherited from dependence. Since profiler depends on mapl.shared, it inherits its include path. Mapl.shared depends on pflogger which has all gFTL's paths. In our GEOSgcm, we build and use pFlogger but in UFS, only pFlogger_stub is used. That is why in UFS, v2 path is missing. This PR should work for UFS, but we may add fake dependence to pfloger_stub to make it really mimic pFlogger.

Dang. Good job figuring this out. CMake is...fun this way!. The question for @tclune is: should @weiyuan-jiang add the fake dependence in pflogger_stub?

mathomp4:
@weiyuan-jiang Okay. I talked with @tclune. I think what you should do is this. In *both* `generic/CMakeLists.txt` and in `profiler/CMakeLists.txt`, we should have:

```
GFTL::gftl-v1 GFTL::gftl-v2
```

The reason is that both are currently using both! You can see all the v2 uses:
```
❯ rg 'template.inc'
generic/VarConnVector.F90
8:#include "vector/template.inc"

generic/VarSpecVector.F90
8:#include "vector/template.inc"

profiler/MeterNodeStack.F90
9:#include "vector/template.inc"
```
but of course they use v1 in many more places!

weiyuan-jiang:
> @weiyuan-jiang Okay. I talked with @tclune. I think what you should do is this. In _both_ `generic/CMakeLists.txt` and in `profiler/CMakeLists.txt`, we should have:
> 
> ```
> GFTL::gftl-v1 GFTL::gftl-v2
> ```
> 
> The reason is that both are currently using both! You can see all the v2 uses:
> 
> ```
> ❯ rg 'template.inc'
> generic/VarConnVector.F90
> 8:#include "vector/template.inc"
> 
> generic/VarSpecVector.F90
> 8:#include "vector/template.inc"
> 
> profiler/MeterNodeStack.F90
> 9:#include "vector/template.inc"
> ```
> 
> but of course they use v1 in many more places!

Changed

mathomp4:
Thanks. I'll do a build-and-run test to be sure all is well.

mathomp4:
Works for me. Approving.

mathomp4 (approved):

generic/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ find_package(GFTL_SHARED REQUIRED)
 esma_add_library(${this}
   SRCS ${srcs}
   DEPENDENCIES MAPL.oomph MAPL.shared MAPL.profiler MAPL.base
-               PFLOGGER::pflogger GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1 GFTL::gftl
+               PFLOGGER::pflogger GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1 GFTL::gftl GFTL::gftl-v2 GFTL::gftl-v1
```
tclune, line 59: We want to be thorough, but not redundant.
```suggestion
               PFLOGGER::pflogger GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1  GFTL::gftl-v2 GFTL::gftl-v1
```

weiyuan-jiang: Your suggestion seems fine to me. Why do you take it back?

tclune:
GFTL::gftl and GFTL::gftl-v1 are the same thing.  So one should not specify them both.    

Someday GFTL::gftl will be an alias for gftl-v2.    It sort of provides a future-safe way of upgrading, except that the Fortran files still need to change for it to help.    Probably a waste of effort on my part.

mathomp4 (approved):

PR #1345, opened 2022-02-07
Add option to make time an integer variable if needed for backwards c…

bena-nasa:
Add a global option to the History.rc to make time an integer variable for backwards compatibility if desired (by default the time variable is a floating point variable as it should be). If this is not possible with the time interval specified throws an error. Also this option is not used in collections specifying monthly output as I'm worried about breaking this feature.

## Motivation and Context

Because the the existing GEOS file spec says the time variable is an integer with units of minutes since date/time although this needs to change as GEOSgcm.x runs operationally at a 7.5 minute timestep. Note a similar change was made in the MAPL v2.8.0.1 patch, this gives the option of keeping integer time if somebody really wants that.

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_TimeMethods.F90
+++ b/base/MAPL_TimeMethods.F90
@@ -131,17 +131,17 @@ contains
        time_increment = packed_hms
     case('minutes')
        if (this%integer_time) then
-          ASSERT(mod(this%frequency,60)==0,"Requested output frequency not representable as an integer minute")
+          _ASSERT(mod(this%frequency,60)==0,"Requested output frequency not representable as an integer minute")
        end if
        time_increment = packed_hms
     case('hours')
        if (this%integer_time) then
-          ASSERT(mod(this%frequency,3600)==0,"Requested output frequency not representable as an integer hour")
+          _ASSERT(mod(this%frequency,3600)==0,"Requested output frequency not representable as an integer hour")
        end if
        time_increment = packed_hms
     case('days')
        if (this%integer_time) then
-          ASSERT(mod(this%frequency,86400)==0,"Requested output frequency not representable as an integer day")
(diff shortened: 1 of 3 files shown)


base/MAPL_TimeMethods.F90
```diff
@@ -85,52 +93,74 @@ function define_time_variable(this,rc) result(v)
     integer :: status
     character(len=ESMF_MAXSTR) :: startTime,timeUnits
     type(ESMF_Time) :: currTime
-    integer :: i1,i2,i3,i123,ipos1,ipos2,isc,imn,ihr,ifreq
+    integer :: i1,i2,i3,ipos1,ipos2,isc,imn,ihr
+    logical :: integer_representable
+    integer :: begin_date, begin_time, time_increment, packed_hms
 
-    call ESMF_CLockGet(this%clock,currTime=currTime,rc=status)
+    _ASSERT(this%frequency/=TimeData_uninit_int,"TimeData object was not created before use")
```
tclune, line 100: Shouldn't the message be "frequency component was not set before use." ?

base/MAPL_TimeMethods.F90
```diff
@@ -85,52 +93,74 @@ function define_time_variable(this,rc) result(v)
...
+       integer_representable = mod(this%frequency,86400) == 0
+       time_increment = this%frequency/86400
     case default
        _ASSERT(.false., 'Not supported yet')
     end select
-    call v%add_attribute('time_increment',ifreq)
+
+    if (this%integer_time) then
+        _ASSERT(integer_representable,"time interval requested not integer representable")
+    end if
```
tclune, line 149: Would it be clearer to eliminate `integer_representable` and just put a tailored ASSERT inside each CASE statement?

tclune (changes requested):
A suggested change and a question in the inline comments.

tclune (approved):

PR #1352, opened 2022-02-08
add mapL_stubcomponent to the top level MAPL package

bena-nasa:
This adds MAPL_StubComponent into the over MAPL package defined by MAPL.F90
This is needed as I am adding the ability to run a history instantiation in the predictor phase of the gcm gridded component and to do this, I have to follow a similar pattern that is used in the capGridComp which requires mapl_stubcomponent

## Types of changes

## Checklist:


tclune (approved):
Amusingly - this should go away when we do DSOs ...

atrayano (approved):

PR #1383, opened 2022-02-18
Fixes #1382. Update NRL solar table read code

### LINKED ISSUE #1382: Is there a way to avoid these broadcast calls?
https://github.com/GEOS-ESM/MAPL/blob/0dbe0f14adf059e53d29112d0ee2da9f4d560275/base/MAPL_sun_uc.F90#L2748-L2753

Can we do some kind of parallel read here instead of having the root node broadcast everything.  This is trapping a load imbalance from the moist physics leading to a significant overhead in Surface times.

Every PE is opening and closing the file currently as it is written

mathomp4:
## Description

This PR updates the `MAPL_SunGetSolarConstantFromNRLFile` procedure to:

1. Open and read the file only on root
2. Broadcast the *tables* instead of just the three output values
3. Does interpolation on all processes to avoid broadcast at every step

## Related Issue

Closes #1382 

## Motivation and Context

Reduces some blocking MPI calls on every run step.

## Types of changes

## Checklist:


mathomp4:
As far as my testing can see, this is zero-diff. I'll undraft and let @dr0cloud and @wmputman take a look.

base/MAPL_sun_uc.F90
```diff
@@ -2475,54 +2475,45 @@ subroutine MAPL_SunGetSolarConstantFromNRLFile(CLOCK,filename_in,SC,MG,SB,Persis
...
+         ! number of lines so we can allocate our tables
+         ! ---------------------------------------------
 
          if (amIRoot) then
 
+            ! Open the file
+            ! -------------
+            filename = trim(filename_in)
+            open(newunit=unit, file=filename, form="formatted", status="old", iostat=status)
+            _ASSERT(status==0,'Could not find NRL data file '// filename )
```
tclune, line 2488: Need a `trim()` on filename here.
```suggestion
            _ASSERT(status==0,'Could not find NRL data file '// trim(filename ))
```

tclune (approved):

dr0cloud:
@mathomp4 Sorry, I left early Friday so didn't get a chance to review. I have reviewed your changes unofficially and they seem fine. I don't think I can review in git anymore, right? Re the other two issues I mentioned, "belt and suspenders check" and INDX1&2 ASSERTs, let's discuss on Teams chat Tuesday.

mathomp4:
> @mathomp4 Sorry, I left early Friday so didn't get a chance to review. I have reviewed your changes unofficially and they seem fine. I don't think I can review in git anymore, right? Re the other two issues I mentioned, "belt and suspenders check" and INDX1&2 ASSERTs, let's discuss on Teams chat Tuesday.

I've pushed a reloop of the code, @dr0cloud. Please take a look.

mathomp4:
@dr0cloud The two asserts are now in as well. Seems to work.

tclune (approved):

PR #1385, opened 2022-02-23
Fixes #1322 Feature/tclune/#1332 toggle timers on off

tclune:
## Description

Added switches to activate global profiling.    Much cleanup ensued.

## Related Issue

#1322
## How Has This Been Tested?

C12 GEOSgcm

## Types of changes

## Checklist:


tclune:
Hmm.  Build and test worked on my laptop and discover with ifort ...  investigating.

tclune:
Ah - there are obsolete unit tests.   Fixing.

tclune:
Argh.  Bits of the ADAS use some of the MAPL interfaces that I thought maybe were not used externally ... had to reintroduce some files (and unit tests).

mathomp4:
@tclune The CHANGELOG conflict was a bit crazy. Can you take a look and make sure the CHANGELOG here looks good?

tclune:
I only spotted a duplicate entry which is now removed. (Forcing CI to run again.)

mathomp4:
Okay. I tested this and it works, but there is also some weird output. I think something isn't quite right with the SetServices. For example, the `Initialize` timers look like:
```
--Initialize                                              2    13.814   7.69     0.009   0.01
----GCM                                                   1    10.325   5.75     1.328   0.74
------AGCM                                                1     8.931   4.97     0.131   0.07
--------SUPERDYNAMICS                                     1     0.935   0.52     0.017   0.01
----------DYN                                             1     0.917   0.51     0.917   0.51
--------PHYSICS                                           1     7.324   4.08     0.098   0.05
----------GWD                                             1     0.018   0.01     0.018   0.01
...
```
but the `SetServices` timers look like:
```
--SetService                                              1     0.057   0.03     0.057   0.03
----GCM                                                   1     0.000   0.00     0.000   0.00
----HIST                                                  1     0.000   0.00     0.000   0.00
----EXTDATA                                               1     0.000   0.00     0.000   0.00
--AGCM                                                    1     0.000   0.00     0.000   0.00
--AIAU                                                    1     0.000   0.00     0.000   0.00
--ADFI                                                    1     0.000   0.00     0.000   0.00
--OGCM                                                    1     0.000   0.00     0.000   0.00
```
It's like the hierarchy isn't there anymore.

Is this the expected output?

tclune:
I need to look into it, but I think that is pretty much the way we expect it now.  Set services of children are now registered and called _after_ the parent set services completes.  So it flattens everything in that sense.   There is still a nesting of calls to the _generic_ setservices though, so I'm a bit confused.   It may just be the timers are not where the would be best located.

mathomp4 (approved):
Seems to work. Zero diff. If @tclune is happy, let's pull it in!

PR #1406, opened 2022-03-04
Fixes #1405. Marks connections with specific 'reserved' keyword as used

### LINKED ISSUE #1405: MAPL_Terminate all imports from a particular child fails
When we exercised the wave model, where we call MAPL_DoNotConnectAnyImport, we get the following error
        GENERIC: ERROR: SRC_NAME: <MAPL_AnyChildImport>  DST_NAME: <MAPL_AnyChildImport> is not satisfied
 DO_NOT_CONNECT ERRORS FOUND in OGCM

atrayano:
## Types of changes

## Checklist:


tclune (approved):
Hmm. This line looks like the one you showed me yesterday.   Must be in a different place?

atrayano:
Yes, it is very similar to what I showed you yesterday, but this line is a different procedure (this on is in APPEND, which puts the items in the container). Yesterday, I had similar logic inside varIsListed

mathomp4:
Running tests with this now after merging in #1411

mathomp4:
This looks to be zero-diff as well. Tested at C48, AMIP and Replay.

mathomp4 (approved):

PR #1408, opened 2022-03-05
Fixes #1407. Adjusted the ring time to work properly even when they a…

### LINKED ISSUE #1407: Averaging couplers are not running correctly when called from gridded components
When History calls the couplers, they behave as expected. But when a gridded component calls them (for example by call to MAPL_RunCouplers), the alarms ring at the wrong time(s). The issue is that History is called after ESMF_ClockAdvance, and this assumption is build in the alarms of the couplers

atrayano:
…re called before ESMF_ClockAdvance

## Types of changes

## Checklist:


atrayano:
This issue was discovered when working with @mmanyin on providing support for 24-hour averaged imports for OH component. The original fix went to the CVS version of the code, this is just a port of the same fix.

tclune (approved):

PR #1439, opened 2022-03-17
Fixes from M1 work

mathomp4:
## Description

This PR brings over the M1 Mac changes found in concert with @natalie-perlin. These include removing a (currently not supported by M1 GNU) `REAL128` variable as well as a CMake change for building when Baselibs is not found.

## Motivation and Context

It should let MAPL build on an M1. If nothing else, it doesn't break building on other chips.

## How Has This Been Tested?

I've built and ran with the GEOSgcm and zero-diff.

## Types of changes

## Checklist:


mathomp4:
No reason adas should fail. I'll re-run.

tclune (approved):

PR #1456, opened 2022-03-24
Fixes #1454. Now MAPL_Shmem memory is deallocated properly

### LINKED ISSUE #1454: GenericCplComp.F90 doest not deallocate tilemask properly
@sdrabenh reported a crash when outputting monthly averages from 'tile' collection.

29593  NOT using buffer I/O for file: geosgcm_landice_checkpoint
29594 forrtl: severe (173): A pointer passed to DEALLOCATE points to an object that cannot be deallocated
29595 Image              PC                Routine            Line        Source·············
29596 GEOSgcm.x          00000000034090D1  Unknown               Unknown  Unknown
29597 libMAPL.generic.s  00002AAAB633A755  mapl_genericcplco        1535  GenericCplComp.F90
29598 libesmf.so         00002AAABB6969E4  _ZN5ESMCI6FTable1     Unknown  Unknown

The problem is the tilemask is allocated conditionally using MAPL_Shmem or on the heap, but not deallocated accordingly

atrayano:
## Types of changes

## Checklist:

### DIFF
--- a/generic/GenericCplComp.F90
+++ b/generic/GenericCplComp.F90
@@ -1,5 +1,14 @@
-#define DEALOC_(A) if(associated(A))then;if(MAPL_ShmInitialized)then;call MAPL_SyncSharedMemory(rc=STATUS);call MAPL_DeAllocNodeArray(A,rc=STATUS);else;deallocate(A,stat=STATUS);endif;_VERIFY(STATUS);NULLIFY(A);endif
-
+#define _DEALLOC(A) \
+    if(associated(A))then; \
+          if(MAPL_ShmInitialized)then; \
+              call MAPL_SyncSharedMemory(rc=STATUS); \
+              call MAPL_DeAllocNodeArray(A,rc=STATUS); \
+           else; \
+              deallocate(A,stat=STATUS); \
+           endif; \
+       _VERIFY(STATUS); \
+       NULLIFY(A); \
+    endif
 #include "MAPL_Generic.h"
 #include "unused_dummy.H"
 
@@ -1340,7 +1349,7 @@ contains
           case default
              _ASSERT(.false., "Unsupported rank")
           end select
-          DEALOC_(mask)
+          _DEALLOC(mask)
        end do
 
        if (am_i_root) call Free_File(unit = UNIT, rc=STATUS)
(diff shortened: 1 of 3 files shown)


generic/GenericCplComp.F90
```diff
@@ -1,3 +1,4 @@
+#define DEALOC_(A) if(associated(A))then;if(MAPL_ShmInitialized)then;call MAPL_SyncSharedMemory(rc=STATUS);call MAPL_DeAllocNodeArray(A,rc=STATUS);else;deallocate(A,stat=STATUS);endif;_VERIFY(STATUS);NULLIFY(A);endif
```
tclune, line 1: If we are going to make hideous macros, let's at least make them readable.  You'll need to check that I did not break anything in the edits though.
```suggestion
#define DEALOC_(A) \
    if(associated(A))then \
          if(MAPL_ShmInitialized)then \
              call MAPL_SyncSharedMemory(rc=STATUS) \
              call MAPL_DeAllocNodeArray(A,rc=STATUS) \
           else \
              deallocate(A,stat=STATUS) \
           endif \
       _VERIFY(STATUS) \
       NULLIFY(A) \
    endif
```

generic/GenericCplComp.F90
```diff
@@ -1338,7 +1340,7 @@ subroutine ReadRestart(CC, SRC, DST, CLOCK, RC)
           case default
              _ASSERT(.false., "Unsupported rank")
           end select
-          if(associated(mask)) deallocate(mask)
+          DEALOC_(mask)
```
tclune, line 1343: I also suggest renaming the macro to have a leading underscore `_DEALOC` unless that is too different than existing practice.

generic/GenericCplComp.F90
atrayano: This is much cleaner! I wish we have done very similar macros the same way.

tclune: I only recently became aware that this was portable.

atrayano: The intel compiler did not like the new style. Let me try with ; at the end of each line

tclune: Sorry about that.

atrayano: ";" works. I am adopting the new style, and about to run a quick test, and push the changes

generic/GenericCplComp.F90
atrayano: _DEALLOC works

tclune (dismissed):

mathomp4:
Let me do a sanity check before I issue the release. For now, I'm going to hijack this PR to prepare for a release. @tclune might need to re-approve.

mathomp4 (approved):
Looks good!

PR #1466, opened 2022-04-01
Enable GCM run in CI

mathomp4:
I am using this PR to test running the GCM with a MAPL PR. I'm wanting to know the costs, etc for this. 

Change 1: Need to use the `compiler-bcs` executor for GCM runs. Also, I don't need *all* the `install` directories, just `bin`, `etc` and `lib` → orb v1.2.0

---

Okay. Finally got it to work. In run-credit terms this goes from 1779 → 2134. Now, on top of that are the storage costs. It looks like an Intel run has 678 MiB of storage and a GNU run has 354 MiB. So a total of 1 GiB

---

Next up, I've set the workplace retention setting in CircleCI to be 1 day instead of 15. Not sure this matters. Still, before I re-triggered the CI, we had 3.7 GB "storage-months" of workspace.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,7 @@ workflows:
           develop_repos: "cmake GEOSana_GridComp" # GEOSadas needs some extra branches to work with mainline MAPL
           rebuild_procs: 8
 
-      # Run GCM (note uses bcs executor inside)
+      # Run GCM (1 hour, no ExtData)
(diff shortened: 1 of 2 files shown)


mathomp4:
You know what. We have credits to burn for a bit. Let's turn on the GCM Run here and see what happens!

tclune (approved):

PR #1475, opened 2022-04-11
ExtData2G updates, bugs fixes and multiple filestream feature.

bena-nasa:
This is more development of ExtData 2.0
This fixes multiple bugs uncovered when testing with the real GEOSgcm system rather than my standalone tester.

This also implements the requested feature to use different rules for different time periods

## Types of changes

## Checklist:


Tests/ExtDataDriverGridComp.F90
```diff
@@ -603,6 +608,24 @@ subroutine run_MultipleTimes(gc, rc)
...
+          call ESMF_ClockAdvance(cap%clock,rc=status)
+          _VERIFY(status)
+          call cap%run_one_step(status)
+          _VERIFY(status)
+       enddo
+       do n=1,cap%nsteps
+          call ESMF_ClockSet(cap%clock,direction=ESMF_DIRECTION_REVERSE,_RC)
+          call ESMF_ClockAdvance(cap%clock,rc=status)
+          call ESMF_ClockSet(cap%clock,direction=ESMF_DIRECTION_FORWARD,_RC)
+       enddo
```
tclune, line 622: Can't this be done in one step?  Or are the separate substeps necessary exercise for ESMF?

base/MAPL_NewArthParser.F90
```diff
@@ -742,6 +744,136 @@ SUBROUTINE CopyScalarToField(ptrs,rn,rc)
...
+             j = j+in-1
+             IF (j > lFunc) EXIT
+             c = Func(j:j)
+          end if
+       END IF
+       DO WHILE (c == ')')                                   ! Check for closing parenthesis
+          ParCnt = ParCnt-1
+          IF (ParCnt < 0) THEN
+             CALL ParseErrMsg (j, FuncStr, ipos, 'Mismatched parenthesis')
+             _ASSERT(.FALSE.,'needs informative message')
```
tclune, line 836: IF the "informative message" is generated by the ParseErrMsg() call, shouldn't that go in now?   Otherwise, this seems a bit verbose.

gridcomps/ExtData2G/ExtDataDerived.F90
```diff
@@ -51,6 +55,57 @@ function new_ExtDataDerived(config,unusable,rc) result(rule)
       _RETURN(_SUCCESS)
    end function new_ExtDataDerived
 
+   function get_variables_in_expression(this,rc) result(variables_in_expression)
```
tclune, line 58: I think it might be good to work on refactoring this long procedure before we move on.   Just try to make it as simple as possible.  I could work with you on it, though I don't immediately see specific things I want done.

gridcomps/ExtData2G/ExtDataUpdatePointer.F90
```diff
@@ -36,56 +42,91 @@ subroutine create_from_parameters(this,update_time,update_freq,update_offset,tim
...
+                       this%first_time_updated = .false.
+                       use_time = this%last_ring + this%offset
+                    end if
+               ! otherwise we land on a time when the alarm would ring and we would update
+               else if (this%last_ring == current_time) then
+                  do_update =.true.
+                  this%first_time_updated = .false.
+                  use_time = current_time+this%offset
+               end if
+            end if
```
tclune, line 122: As above - I'd like to see if there are ways to simplify this.

gridcomps/ExtData2G/ExtDataDerived.F90
bena-nasa: It doesn't seem that long, the only thing I would do is perhaps moving all the masking procedures to a separate class.

gridcomps/ExtData2G/ExtDataUpdatePointer.F90
bena-nasa: I have to handle multiple special cases. 
1. If this is the first time
2. If the clock moves forward and the alarm rings while moving forward, then you externally rewind the clock
3. If the clock moves forward and the alarm does not ring while moving forward, then you externally rewind the clock
4. If the clock moves forward and the alarm does not ring while moving forward, then you externally rewind the clock and this the last update was the first time running
5. Oh, and I don't trust ESMF alarms at all with the currently implementation so I have to do this without ESMF Alarms.

base/MAPL_NewArthParser.F90
bena-nasa: Yes, you are right. This was just a copy and paste from CheckSyntax (which I have not looked at in probably 5 years or more). CheckSyntax has the same issue

gridcomps/ExtData2G/ExtDataUpdatePointer.F90
tclune: Understood.     But we should still be able to introduce helper functions so that the large scale structure is simpler.  Thee key is coming up with a good name for what happens in each block.  A purist might even argue that it is worth procedures along the lines of  `this%is_now()` `this%is_past()` and `this%is_future()` or something like that.  Mostly just want it to be considered; not pushing hard.

gridcomps/ExtData2G/ExtDataDerived.F90
bena-nasa: The masking is a bit of a mess as this was refactor of the old Extdata rather than a clean slate... so the new one just inherited that. I'll work on abstracting the masking into a separate class although might not happen by the end of today (4/12).

gridcomps/ExtData2G/ExtDataUpdatePointer.F90
bena-nasa: I'll try to introduce helper functions

base/MAPL_NewArthParser.F90
bena-nasa: I push an update that eliminated the redundant parseErrMsg calls

gridcomps/ExtData2G/ExtDataDerived.F90
bena-nasa: I moved all the masking to it's own class, should be cleaner.

gridcomps/ExtData/ExtDataGridCompMod.F90
```diff
@@ -3254,15 +3254,15 @@ subroutine MAPL_ExtDataInterpField(item,time,field,vector_comp,rc)
...
-        end if
+        !if (lgr%isEnabledFor(DEBUG) .and. .not. item%doInterpolate) then
+           !call lgr%debug('   MAPL_ExtDataInterpField: Uninterpolated field %a set to sample L %i0.8 %i0.6', trim(item%name),  nymd1, nhms1)
+        !else if (time == item%interp_time1) then
+           !call lgr%debug('   MAPL_ExtDataInterpField: Interpolated field %a set to sample L %i0.8 %i0.6', trim(item%name),  nymd1, nhms1)
+        !else if (time == item%interp_time2) then
+           !call lgr%debug('   MAPL_ExtDataInterpField: Interpolated field %a set to sample R %i0.8 %i0.6', trim(item%name),  nymd2, nhms2)
+        !else
+           !call lgr%debug('   MAPL_ExtDataInterpField: Interpolated field %a between %i0.8 %i0.6 and %i0.8 %i0.6 (%f10.6 fraction)', trim(item%name), nymd1, nhms1, nymd2, nhms2, alpha)
+        !end if
```
tclune, line 3265: Can this commented out code be deleted?   Or why is ic commented out?

bena-nasa: I was debugging and I remember comment that out as the output was overwhelming, I'll undelete it.

tclune:
Note - there are levels of debugging even lower than "DEBUG".   The named levels are DEBUG, INFO, WARN, ...  but they correspond to integers 10, 20, 30, ...

So you could make those "DEBUG" something like "9" and then they would only show up when you set the logger threshold to "9".

base/MAPL_NewArthParser.F90
```diff
@@ -780,26 +901,22 @@ SUBROUTINE CheckSyntax (FuncStr,Var,needed,ExtVar,rc)
        IF (c == '-' .OR. c == '+') THEN                      ! Check for leading - or +
           j = j+1
           IF (j > lFunc) THEN 
-             CALL ParseErrMsg (j, FuncStr, ipos, 'Missing operand')
-             _ASSERT(.FALSE.,'needs informative message')
+             _ASSERT(.FALSE.,'Missing operand in '//trim(funcstr))
```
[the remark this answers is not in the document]
bena-nasa: Just pushed something to remove all the _ASSERT(.falses. every in ExtData2G

tclune (approved):

PR #1498, opened 2022-05-03
Closes #1420. Extend profiler width

### LINKED ISSUE #1420: Extend format width for inclusive and exclusive times in profile reporting
The current `(f9.2)` may not be sufficient for runs that span more than 10 wall clock days (GCHP).   `f(10.2)` should suffice for all known use cases.

mathomp4:
## Description

This change extends the profiler width per #1420 for use with GCHP.

## Related Issue

Closes #1420 

## Motivation and Context

GCHP runs with loooong wall times and this lets the profiler work with that.

## How Has This Been Tested?

I built the change in GEOS and it didn't crash. Should be safe.

## Types of changes

## Checklist:


mathomp4:
This seems to work. Before:
```
 Times for component <EXTDATA>
                                       Min                          Mean                         Max                    PE         # cycles
                            ==========================   ==========================   ==========================   =============   ========
Name                          %    inclusive exclusive     %    inclusive exclusive     %    inclusive exclusive    max    min
                          | ------ --------- --------- | ------ --------- --------- | ------ --------- --------- | ------ ------ |
EXTDATA                   |   0.00     22.17      0.00 |   0.00     22.19      0.00 |   0.00     22.21      0.00 |  00012  00094 |       99
--SetService              |   0.00      0.00      0.00 |   0.00      0.00      0.00 |   0.00      0.00      0.00 |  00001  00091 |        1
```
After:
```
 Times for component <EXTDATA>
                                        Min                            Mean                           Max                     PE         # cycles
                            ============================   ============================   ============================   =============   ========
Name                          %    inclusive  exclusive      %    inclusive  exclusive      %    inclusive  exclusive     max    min
                          | ------ ---------- ---------- | ------ ---------- ---------- | ------ ---------- ---------- | ------ ------ |
EXTDATA                   |   0.00      20.97       0.00 |   0.00      20.99       0.00 |   0.00      21.01       0.00 |  00069  00094 |       99
--SetService              |   0.00       0.00       0.00 |   0.00       0.00       0.00 |   0.00       0.00       0.00 |  00078  00091 |        1
```

I'll let @tclune take a look at this and see if he thinks I did it right.

tclune (approved):
Good enough for now.   We probably want a named parameter for the width so that these repeated expressions can be derived and then modified more easily in the future.

OTOH, the long term plan is to have a yaml config file for all this ...

PR #1516, opened 2022-05-17
Automatic restart skip

bena-nasa:
If a field makes it to ExtData, go through the hierarchy and make any component with that field in it's import MAPL_RestartSkip. Added new routine in MAPL_Generic to do this. This was requested by the chemistry folks so they don't have to remember to add restartskip and if something may or may not be provided by ExtData depending on the configuration of the system.

## Types of changes

## Checklist:

### DIFF
--- a/generic/MAPL_Generic.F90
+++ b/generic/MAPL_Generic.F90
@@ -11368,12 +11368,10 @@ contains
          call ESMF_AttributeSet(field,name=att_name,value=att_val,_RC)
       end if
       nc = state%get_num_children()
-      if (nc > 0) then
-         do i=1,nc
-            child_gc => state%get_child_gridcomp(i)
-            call MAPL_AddI4AttributeToFields(child_gc,field_name,att_name,att_val,_RC)
-         enddo
-      end if
+      do i=1,nc
+         child_gc => state%get_child_gridcomp(i)
+         call MAPL_AddI4AttributeToFields(child_gc,field_name,att_name,att_val,_RC)
+      enddo
       
       _RETURN(_SUCCESS)
    end subroutine MAPL_AddI4AttributeToFields
--- a/generic/MAPL_Generic.F90
+++ b/generic/MAPL_Generic.F90
@@ -343,7 +343,7 @@ module MAPL_GenericMod
    end interface MAPL_GetLogger
 
    interface MAPL_AddAttributeToFields
-      module procedure MAPL_AddI4AttributeToFields
+      module procedure MAPL_AddAttributeToFields_I4
    end interface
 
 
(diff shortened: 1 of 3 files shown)


atrayano (approved):

generic/MAPL_Generic.F90
```diff
@@ -11334,4 +11339,43 @@ subroutine warn_empty(string, MPL, rc)
       _RETURN(ESMF_SUCCESS)
    end subroutine warn_empty
 
+   recursive subroutine MAPL_AddI4AttributeToFields(gc,field_name,att_name,att_val,rc)
```
tclune, line 11342:
```suggestion
   recursive subroutine MAPL_AddAttributeToFields_I4
   (gc,field_name,att_name,att_val,rc)
```
Nominal preference to put the disambguators at the end to make them easier to spot.  I think this is more consistent with other cases, though I'm sure it is far from uniform in MAPL.  No need to change if this is the only issue I find.

generic/MAPL_Generic.F90
```diff
@@ -11334,4 +11339,43 @@ subroutine warn_empty(string, MPL, rc)
...
+         end if
+         call ESMF_AttributeSet(field,name=att_name,value=att_val,_RC)
+      end if
+      nc = state%get_num_children()
+      if (nc > 0) then
+         do i=1,nc
+            child_gc => state%get_child_gridcomp(i)
+            call MAPL_AddI4AttributeToFields(child_gc,field_name,att_name,att_val,_RC)
+         enddo
+      end if
```
tclune, line 11376: Conditional is redundant.
```suggestion
      do i=1,nc
         child_gc => state%get_child_gridcomp(i)
         call MAPL_AddI4AttributeToFields(child_gc,field_name,att_name,att_val,_RC)
      enddo
```

gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -690,8 +690,9 @@ subroutine initialize_history(cap, rc)
   end subroutine initialize_history
 
 
-  subroutine initialize_extdata(cap , rc)
+  subroutine initialize_extdata(cap , root_gc, rc)
```
tclune, line 693: This will definitely give me some thought about how to treat in MAPL 3

tclune (changes requested):
Does any of this require an update as ESMF_Info on the V3 branch?

Minor changes requested.

mathomp4:
> Does any of this require an update as ESMF_Info on the V3 branch?

@tclune Yes, it will. The ESMF_AttributeGet call will need to be an infoget. 

Probably a good time to do a sweep of MAPL3 to see if any ESMF_Attribute calls have snuck in. (Maybe there is a way at ESMF_init time to pass a flag that doesn't allow any ESMF_Attribute calls? That way we can be strict.)

generic/MAPL_Generic.F90
bena-nasa: Made the change

tclune (approved):

PR #1520, opened 2022-05-18
Handmerge develop into MAPL3 2022-May-18 (2nd of the day)

mathomp4:
This is a handmerge of `develop` into `release/MAPL-v3`. We are doing this by hand to make sure we got the conversion of ESMF_Attribute to ESMF_Info correct.

I did the merge with @bena-nasa looking over my shoulder, so we should be okay... we hope.


mathomp4:
Looks like we'll need @bena-nasa and/or @tclune to look at this. Something is different between MAPL 2 and 3 here.

PR #1531, opened 2022-05-27
Hotfix 1530: Initialize lperp

### LINKED ISSUE #1530: Uninitialized logical in Cap possibly leads to crash with MAPL-in-UFS using GCC 11.3
Our UFS colleagues have found what we think is a bug in MAPL (see https://github.com/GEOS-ESM/GOCART/issues/127#issuecomment-1139885184). It exists in 2.12.2 (which they are running) and even in current `main` MAPL. 

To wit, in the cap object is a logical called `lperp` which is only initialized in this block:

https://github.com/GEOS-ESM/MAPL/blob/642108c7f1a6163b120e3a4bdc0d12a973ff99f4/gridcomps/Cap/MAPL_CapGridComp.F90#L336-L352

GEOS normally runs in the `cap_clock_is_present=.false.` case, and so we enter this code and `cap%lperp` is set.

But, if `cap_clock_is_present=.true.`, which I think is how UFS runs, we never set that. Then, eventually, you get to either:

https://github.com/GEOS-ESM/MAPL/blob/642108c7f1a6163b120e3a4bdc0d12a973ff99f4/gridcomps/Cap/MAPL_CapGridComp.F90#L1134-L1138

or:

https://github.com/GEOS-ESM/MAPL/blob/642108c7f1a6163b120e3a4bdc0d12a973ff99f4/gridcomps/Cap/MAPL_CapGridComp.F90#L1256-L1261

(not sure which path they are hitting). No matter what, it looks like GCC 11.3 has the behavior such that some how it's resolving as `.true.` and UFS is going into the perpetual clock code.

mathomp4:
## Description

Initialize `cap%lperp` to `.false.` in the data type declaration. Our UFS friends found a bug in MAPL when you run in a way where you pass the clock into Cap. This is not how GEOS runs, but it is how UFS runs. In our case, `cap%lperp` is always set to something, but for UFS it's uninitialized.

## Related Issue

Fixes #1530 

## Motivation and Context

Fixes a bug found by UFS with GCC 11.3

## How Has This Been Tested?

- [x] GEOSgcm.x run
- [ ] UFS GOCART run

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 
+## [2.21.2] - 2022-05-27
+
+### Fixed
+
+- Initialize `cap%lperp` in `MAPL_CapGridComp.F90` (see Issue #1530)
+
 ## [2.21.1] - 2022-05-24
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 
-## [2.21.2] - 2022-05-27
+## [2.21.2] - 2022-05-31
 
 ### Fixed
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_policy (SET CMP0054 NEW)
 
 project (
   MAPL
-  VERSION 2.21.1
+  VERSION 2.21.2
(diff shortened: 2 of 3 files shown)


mathomp4:
Okay. I have tested with GEOSgcm and it is zero-diff. I'll wait until hearing back from @jkbk2004 on https://github.com/GEOS-ESM/GOCART/issues/127 before undrafting (and making a MAPL 2.12.3).

jkbk2004:
@mathomp4 UFS with GCC 11.3 ran ok with the fix.

mathomp4:
> @mathomp4 UFS with GCC 11.3 ran ok with the fix.

Thanks. I'll undraft this. I'll also work on MAPL 2.12.3.

PR #1541, opened 2022-06-03
Just brings the code for the new sorting. Intentionally have not yet …

atrayano:
…changed the names to avoild name collision, nor included in the build system

## Types of changes

## Checklist:

### DIFF
--- a/shared/new_sort/MAPL_Sort.F90
+++ b/shared/new_sort/MAPL_Sort.F90
@@ -1,0 +1,468 @@
+
+#define ASSERT_(A) if(.not.(A))call exit(1)
+
+
+!  $Id: MAPL_Sort.F90,v 1.12 2014-12-11 21:08:00 atrayano Exp $
+
+!=============================================================================
+!BOP
+
+! !MODULE: MAPL_Sort   -- A utility to sort integers
+
+! !INTERFACE:
+
+module MAPL_SortMod
+
+  implicit none
+  private
+
+! !PUBLIC ROUTINES:
+
+  public MAPL_Sort
+
+! !DESCRIPTION:
+! 
+!   {\tt MAPL\_Sort} is a utility to do a quicksort on integers. The general
+!   interface is:
+!\bv       
+!       subroutine MAPL_Sort(A[,Bi,Bl,Br,Bd,Ci,Cl,Cr,Cd,Di,Dl,Dr,Dd,DIM])
+!         integer(kind=[4,8]),           intent(INOUT) :: A(:)
+!         integer(kind=4)    , optional, intent(INOUT) :: Bi(:,:)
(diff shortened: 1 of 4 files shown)


shared/new_sort/MAPL_Sort.F90
```diff
@@ -0,0 +1,468 @@
...
+   module procedure SORT2SS
+   module procedure SORT2SR
+   module procedure SORT2SL
+   module procedure SORT2SD
+end interface
+
+contains
+
+subroutine SORTL(A,Bi,Bl,Br,Bd,Ci,Cl,Cr,Cd,Di,Dl,Dr,Dd,DIM)
+  integer(kind=8),           intent(INOUT) :: A(:)
```
mathomp4, line 91: I imagine we'll need to generalize these for NAG, since `kind=8` doesn't work there...

shared/new_sort/MAPL_Sort.F90
```diff
@@ -0,0 +1,468 @@
...
+  len = size(A)
+
+  if    (present(Bi)) then
+     call QSORTL44(A,Bi,Dm,len,1,0)
+  elseif(present(Bl))then
+     call QSORTL84(A,Bl,Dm,len,1,0)
+  elseif(present(Br))then
+     call QSORTL44(A,Br,Dm,len,1,0)
+  elseif(present(Bd)) then
+     call QSORTL84(A,Bd,Dm,len,1,0)
```
tclune, line 118: Yeah - no, this is a terrible/unacceptable style.    In theory all 4 of these args could be present.   The proper way todo this is to have separate interfaces for each case.

shared/new_sort/quicksort.c
```diff
@@ -0,0 +1,203 @@
...
+}
+
+
+#define MAX_SORT_LEN 400000000
+
+
+
+// If the length of the sort exceeds a critical value (MAX_SORT_LEN),
+// it halves it repeatedly until it is below this value, does
+// order LEN/MAX_SORT_LEN sorts, and shuffles these in ascending order.
```
tclune, line 146: This comment leaves out the "why".   Is it to give better performance?  To avoid some memory limitation?

atrayano: Only Max could answer "why". My guess is for better performance... I do not think we exceed the threshold currently. Which brings the question, if the "swap-and-shuffle" scheme was well tested

weiyuan-jiang: I remember one thing. I changed the sort so it pickd the middle item as pivot point. Before that, it chose the first one as the pivot point. So if an array is already sorted, it takes O( n^2). If the array is very large, it would be very slow. It helps if it limits the max size.

tclune:
Probably should be replaced with radix-sort in any event.

mathomp4:
Marking as long-term

PR #1550, opened 2022-06-07
More Updates for Spack/non-Baselibs Compatibility

mathomp4:
## Description

This is a set of changes needed for building MAPL in a non-Baselibs (aka Spack) situation. The issue is that there have been recent changes to ESMA_cmake where:

1. We no longer do a `find_package(MPI)` call for everything using `esma.cmake`
2. We no longer carry around a `FindESMF.cmake` file and instead use the one from ESMF itself.

Because of these changes, we need to change how the "if not Baselibs" side of MAPL works. So we do a few things:

1. We explicitly do a `find_package(MPI)`
2. We do a `find_package(ESMF MODULE REQUIRED)` 
3. We let CMake know that (at least for MAPL) ESMF requires MPI::MPI_Fortran
~4. We change all the `esmf` in our CMake to `ESMF` as that is what the `FindESMF.cmake` from ESMF provides~
5. We add an alias library back to `esmf` as the rest of GEOS still uses that.

NOTE 1: This now means non-Baselibs builds will need to append to `CMAKE_MODULE_PATH` where their `FindESMF.cmake` file is installed/located.

ETA: It turns out we can't move from `esmf` to `ESMF` for GEOS-ESM CMake reasons. Fixtures like the ADAS and LDAS aren't seem to fail with transitive CMake issues...in MAPL.

```
CMake Error at @cmake/@ecbuild/cmake/ecbuild_add_executable.cmake:218 (add_dependencies):
  The dependency target "ESMF" of target "cub2latlon.x" does not exist.
Call Stack (most recent call first):
  src/Shared/@MAPL/base/CMakeLists.txt:83 (ecbuild_add_executable)


CMake Error at @cmake/@ecbuild/cmake/ecbuild_add_executable.cmake:218 (add_dependencies):
  The dependency target "ESMF" of target "Regrid_Util.x" does not exist.
Call Stack (most recent call first):
  src/Shared/@MAPL/griddedio/CMakeLists.txt:25 (ecbuild_add_executable)
```

This shouldn't happen (the GCM seems fine) but I'm guessing perhaps I need updates to ESMA_cmake in the upstream fixtures to enable this. For now, let's go back to `esmf`. Don't rock the boat more than needed.

## Motivation and Context

Allows MAPL to build built in environments without Baselibs.

## How Has This Been Tested?

I will try to build/test on discover. I can easily test the "did I break Baselibs builds". It's a bit harder to test the non-Baselibs builds. I can say that MAPL does build (with a few hacks in the MAPL Spack package), so this should work with non-Baselibs builds. But our friends at Harvard (ping @lizziel) and NOAA (ping @rmontuoro) might eventually come back to us with more changes needed.

## Types of changes

## Checklist:


mathomp4:
This seems zero-diff to me. I'll undraft

mathomp4:
Stupid random ADAS build test. Let's try again!

mathomp4:
Okay. Thanks to the merge of #1572 I think this should now all be happy in CI. Let's find out!

tclune (approved):

PR #1556, opened 2022-06-10
Add ExtData Testing Framework

bena-nasa:
I spent some time cleaning up the ExtData testing framework I maintain on my own personal git repo that I use to make sure I have not broken ExtData (and to some extent History as it requires History to function with the most basic functionality). I'm making this PR here, more for feedback.

The idea is that I have a set of test cases that exercise various ExtData data cases by having History generate the input for ExtData. Each case consists of input files only, no data as that is generated in each case on the fly.

I then have a set of python scripts to run each case, by moving the RC files for the specified cases to a temp directory, running ExtDataDriver.x, checking that the program finishes successfully, and then deleting the temp directory

I thought eventually we would want this in MAPL, but not sure how it fits into any other tests (its not PFUNIT!) or we would make use (is it something user has to explicitly run, that cmake can run or that CI can run like we run the model for example). Maybe it doesn't belong here yet but I figured since I took the time to clean things up for my own sake I would see if there are thoughts about this and how we might incorporate it, if that is something that anyone besides me would have interest in. 

## Types of changes

## Checklist:


tclune (dismissed):

tclune:
CMake can add tests that for things that are not pFunit.    Just need to define a command that should return 0 (success), and then a cmake `add_test()` command.  (Have not done it in a while, so I'm rusty on the details.)  Then the existing CI will exercise this as well.

The challenge usually is to get relevant input data into the CMake test environment.    If the test generates its own inputs, all the better.

bena-nasa:
> CMake can add tests that for things that are not pFunit. Just need to define a command that should return 0 (success), and then a cmake `add_test()` command. (Have not done it in a while, so I'm rusty on the details.) Then the existing CI will exercise this as well.
> 
> The challenge usually is to get relevant input data into the CMake test environment. If the test generates its own inputs, all the better.

Thanks, I'll take a look around and see how it is done elsewhere in MAPL see how this could fit in with cmake

Tests/ExtData_Testing_Framekwork/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+     set(num_procs "1")
+  endif()
+  if (${TEST_EXTDATA1G})
+     add_test(
+        NAME "ExtData1G_${TEST_CASE}"
+        COMMAND ${CMAKE_COMMAND} 
+          -DTEST_CASE=${TEST_CASE} 
+          -DMPIEXEC_EXECUTABLE=${MPIEXEC_EXECUTABLE} 
+          -DMPIEXEC_NUMPROC_FLAG=${MPIEXEC_NUMPROC_FLAG}
+          -DMY_BINARY_DIR=${CMAKE_BINARY_DIR}/bin
```
tclune, line 28:
```suggestion
          -DMY_BINARY_DIR=${CMAKE_CURRENT_BINARY_DIR}/bin
```
BINARY is top of tree, CURRENT_BINARY is relative to source.   When the test is built inside full geos this makes a big difference.

Not sure `bin` is the best name for the subdir.

Tests/ExtData_Testing_Framekwork/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+          -DIS_EXTDATA2G=NO
+          -P ${CMAKE_CURRENT_SOURCE_DIR}/run_extdata.cmake
+     if (${num_procs} LESS ${cutoff})
+        set_tests_properties ("ExtData1G_${TEST_CASE}" PROPERTIES LABELS "EXTDATA1G_SMALL_TESTS")
+     else()
+        set_tests_properties ("ExtData1G_${TEST_CASE}" PROPERTIES LABELS "EXTDATA1G_BIG_TESTS")
+     endif()
+        )
+  endif()
+  if (${TEST_EXTDATA2G})
```
tclune, line 39: This line is suspicious.  Usually  you test the var directly not its expansion.

tclune (commented):
Added inline comments/questions.

Tests/ExtData_Testing_Framekwork/CMakeLists.txt
bena-nasa: I'll try this in the model.

bena-nasa: I tested this in a full model build. The tests worked so what Matt and I ended up doing here appears to be correct. We didn't choose bin, that's the binary directory in the build so no choice there...

Tests/ExtData_Testing_Framekwork/CMakeLists.txt
bena-nasa: Changed this

bena-nasa:
> If @bena-nasa finds the test useful, I would not want to lose it. It could still be exercised by the nightly scripts.

I'm not deleting it, we updated the ci to skip that test but certainly be hand run or if I could figure out a way to test this on fewer cores might be the best bet

mathomp4:
Woooo! All pass! 🎉

PR #1557, opened 2022-06-12
Refactored FieldDictionary and tests.

tclune:
Changes from code review.

## Types of changes

## Checklist:

### DIFF
--- a/generic3g/FieldDictionary.F90
+++ b/generic3g/FieldDictionary.F90
@@ -26,7 +26,6 @@ module mapl3g_FieldDictionary
    private
 
    public :: FieldDictionary
-   public :: GEOS_Field_Dictionary
 
    type :: FieldDictionary
       private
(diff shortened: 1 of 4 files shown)


tclune:
During development, I think it is confusing to have ChangeLog entries for each bit of MAPL 3 as it comes in.   When the ultimate merge happens, there does need to be a sizable entry in the log that describes the main differences, but to reiterate each piece of old  MAPL as it is reimplemented - possibly with just cleaner code becomes counterproductive.

So I'd suggest disabling for this branch if possible.   Otherwise I'll try to get better about setting the label.

mathomp4:
> During development, I think it is confusing to have ChangeLog entries for each bit of MAPL 3 as it comes in. When the ultimate merge happens, there does need to be a sizable entry in the log that describes the main differences, but to reiterate each piece of old MAPL as it is reimplemented - possibly with just cleaner code becomes counterproductive.
> 
> So I'd suggest disabling for this branch if possible. Otherwise I'll try to get better about setting the label.

I'll see if I can figure that out; it's been a while since I've looked at the label enforcer... 

Though, if at some point we do make a commit that seems to be non-zero-diff, it would be good to be able to track down when that happened in MAPL 3.

I've labeled and pressed the "Update branch" button.

atrayano (approved):

PR #1559, opened 2022-06-13
Fix bundleio tests

bena-nasa:
This fixes these bundleio tests that a were added by Gian. The issues is that he calls two tests with the same executable. One exercises create a lat-lon file, the next exercises creating a cubed-sphere file. The problem is the executable uses a hard coded name for the temporary file created. So after say the lat-lon test runs, when the cube test runs the old lat-lon file from the previous test is still there. Since we do not clobber on creation (might want to revisit this) it tries to use the old existing lat-lon file and goes kaboom since it tries to write to a variable that does not exist.
So I just added the temporary file as an option to the command line for the testing program and use different files for the two tests.

## Types of changes

## Checklist:


atrayano (approved):

tclune (approved):

mathomp4:
Now here's a new question (maybe for @tclune): Should we try enabling these in CI? Or are they not quite tests for CI? Still, I'll pull this in and maybe do an experiment to see if CircleCI is happy when I do.

PR #1570, opened 2022-06-15
Updates for FORD Testing

### LINKED ISSUE #1567: Bad ASSERTs in pfio_io_demo
Testing with Ford exposed some bad _ASSERT calls in `pfio_io_demo`, a la:

https://github.com/GEOS-ESM/MAPL/blob/d46c2ed476db26b4d21ba52da5cb4256b1a6cf0e/pfio/pfio_io_demo.F90#L52

In our code `_ASSERT` requires two arguments. Now, apparently we don't compile this code because it should have broke instantly.

I suppose the question for @weiyuan-jiang (or @tclune) is "Do we use this code anymore?" If yes, these should be fixed, if not, we should remove the code.

mathomp4:
## Description

Testing with the FORD Documentation maker found a few changes needed to support FORD. These included adding a `program` statement to the top of a few programs (that only had an `end`), removing `pfio/pfio_io_demo.F90` (see #1567) and fixing a [macro redefinition](https://github.com/Fortran-FOSS-Programmers/ford/issues/421#issuecomment-1156217398).

## Related Issue

Closes #1567 

## Motivation and Context

Let's me try out FORD with MAPL!

## How Has This Been Tested?

I don't see how any of these changes should break the model. Pretty trivial

## Types of changes

## Checklist:


tclune (dismissed):

mathomp4:
Re-requesting @tclune review. Apparently I pushed after the review (I removed that pf test that wasn't being run)

tclune (approved):

PR #1576, opened 2022-06-18
by pass missing value of NaN

weiyuan-jiang:
If the missing values is Nan, all the calculation will be Nan. It is not necessary to swap or test
## Types of changes

## Checklist:

### DIFF
--- a/griddedio/GriddedIO.F90
+++ b/griddedio/GriddedIO.F90
@@ -1091,29 +1091,29 @@ module MAPL_GriddedIOMod
         else
            allocate(ptr2d(0,0))
         end if
+
+        if (isnan(fill_value)) then
+           where(isnan(ptr2d)) ptr2d=MAPL_UNDEF
+        else
+           where(ptr2d==fill_value) ptr2d=MAPL_UNDEF
+        endif
+
      else if (fieldRank==3) then
         if (hasDE_in) then
            call ESMF_FieldGet(field,farrayPtr=ptr3d,_RC)
         else
            allocate(ptr3d(0,0,0))
         end if
-     else
-        _FAIL('rank not supported')
-     end if
 
-     if (associated(ptr2d)) then
-        if (isnan(fill_value)) then
-           where(isnan(ptr2d)) ptr2d=MAPL_UNDEF
-        else
-           where(ptr2d==fill_value) ptr2d=MAPL_UNDEF
-        endif
-     else if (associated(ptr3d)) then
         if (isnan(fill_value)) then
            where(isnan(ptr3d)) ptr3d=MAPL_UNDEF
         else
            where(ptr3d==fill_value) ptr3d=MAPL_UNDEF
         endif
-     endif
+
+     else
+        _FAIL('rank not supported')
+     end if
 
      _RETURN(_SUCCESS)
 
--- a/griddedio/GriddedIO.F90
+++ b/griddedio/GriddedIO.F90
@@ -29,10 +29,9 @@ module MAPL_GriddedIOMod
   
   private
 
-  character(len=20), parameter :: fill_value_label = "GriddedIO_Fill_Value"
-
   type, public :: MAPL_GriddedIO
      type(FileMetaData) :: metadata
+     type(fileMetadataUtils), pointer :: current_file_metadata
      integer :: write_collection_id
      integer :: read_collection_id
      integer :: metadata_collection_id
@@ -939,12 +938,10 @@ module MAPL_GriddedIOMod
      type(ESMF_Grid) :: output_grid
      logical :: hasDE
      class(AbstractGridFactory), pointer :: factory
-     type(fileMetadataUtils), pointer :: metadata
      real(REAL32) :: missing_value
 
      collection => Datacollections%at(this%metadata_collection_id)
-     metadata => collection%find(filename, __RC__)
-     this%metadata = metadata
+     this%current_file_metadata => collection%find(filename, __RC__)
      filegrid = collection%src_grid
      factory => get_factory(filegrid)
      hasDE=MAPL_GridHasDE(filegrid,rc=status)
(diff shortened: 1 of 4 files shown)


weiyuan-jiang:
This PR is addressing the issue https://github.com/ufs-community/ufs-weather-model/issues/1259

mathomp4:
I can say this is zero-diff for a C24 run of GEOSgcm. I'll let @tclune or someone more knowledgable than I comment on the code change.

weiyuan-jiang:
> I don't see how this change has anything to do with NaN. The change merely verifies that if a fill is specified then it is of a numeric type. (I.e., it will ignore fills that are strings or bools.) What am I missing?

This line crashes without the fix. 
https://github.com/GEOS-ESM/MAPL/blob/d46c2ed476db26b4d21ba52da5cb4256b1a6cf0e/griddedio/GriddedIO.F90#L1092

I suspect this is a bug from ESMF.  Just two lines above,  I test the return type of the '_FillValue',  it is ESMF_TYPEKIND_R4 , but  ESMF crashes when trying to get it.

weiyuan-jiang:
>

tclune:
I don't see how your change would affect that case.   R4 matches so `has_missing_value` should be `.true.` before and after your change.   If the code change affects results, it seems to suggest memory corruption.

weiyuan-jiang:
> @bena-nasa , I think it is fine to bypass the Nan values even there is Nan. Any calculation relults that Nan is involved would be Nan too.

Our Metadata does not take "Nan" as a numerical type. So there is no missing data if it is Nan. I think it is fine because the results would be Nan if Nan is involved in the calculation

weiyuan-jiang:
I have another theory. When ESMF checks the type of attribute '_FillValue', it actually checks the type of the variable. They are supposed to be the same type anyway. Whey trying to get it,  it crashes because it is a string ( Nan or Nanf)

tclune:
Your theory is plausible, but I still don't see how the changes here have any effect.

weiyuan-jiang:
pfio can set and get Nan attribute,  but ESMF does not seem to work.  But I think it is safe to by pass Nan.

griddedio/GriddedIO.F90
```diff
@@ -1013,6 +1013,10 @@ subroutine request_data_from_file(this,filename,timeindex,rc)
              this%read_collection_id, fileName, trim(names(i)), &
              & ref, start=localStart, global_start=globalStart, global_count=globalCount)
         deallocate(localStart,globalStart,globalCount)
+        ! if is Nan, do not set '_FillValue'.
+        ! The pair ESMF_AttributeSet and ESMF_AttributeGet cannot handle Nan
+        if (missing_value /= missing_value) cycle
```
tclune, line 1018: Ok.   

1. Perhaps we should return an error if it is found to be NaN in this manner?   Better to fail here then when doing a regrid with such values later on.  
2. Please open a ticket with esmf support to see if this is something that they can fix on their end.
3. This mechanism for detecting NaN is just obscure enough that it is probably worth it to introduce a helper function `is_NaN(x)`  and then call that instead.   I would not ask, except you have to modify the code still to handle the error condition.

tclune (changes requested):
See inline comment.

griddedio/GriddedIO.F90
mathomp4: If it is a "real" nan, `ieee_is_nan()` could do it for us. We use it in various places in GEOS. (Though for historical purposes we often rename it as `isnan`, a la https://cs.github.com/GEOS-ESM/GEOSgcm_GridComp/blob/3e52659a4baa55aea5b0c937afcbfe90f0fafd92/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_CatchCNRestarts.F90#L199)

weiyuan-jiang:
If  regrid gets trouble when there is Nan, this PR won't work. Nan needs to be replaced somewhere.

griddedio/GriddedIO.F90
weiyuan-jiang: The Nan now is changed to MAPL_UNDEF. But I think it is fine not to replace it.  The results will be Nan if a Nan is involved in calculation

tclune: @weiyuan-jiang please elaborate on what you wrote above.    NaN is never MAPL_UNDEF, but I think you meant that the FillValue is now changed to MAPL_UNDEF.   This is not a change we are permitted to make though.

bena-nasa: I just don't know what the right thing to do here is, we could fail and return an error if the missing value is nan. On the other hand if we do not, the latest commit from weiyuan seems like all we can do. We can't let the nan's propagate to a real calculation. At least if the Nan's are replaced with MAPL_UNDEF which we using internal no matter what the missing value is those points will be skipped without leading to an error. I would approve but I will defer to @tclune.

weiyuan-jiang: @tclune I am not sure about "This is not a change we are permitted to make though".  I think we simply replace Nan in the memory with MAPL_UNDEF for the convenience of calculation

tclune: ok - trying to look at the code to see where this has an impact.  The CYCLE seems to suggest we don't need to tell ESMF the fill value in the first place.

weiyuan-jiang: How about saving the metadata for swap call when request_data_from_file. Is it safe?  @bena-nasa  , please see my new change

weiyuan-jiang:
I have verified this PR resolves the Nan issue from ufs @bena-nasa @tclune

tclune (dismissed):

weiyuan-jiang:
@mathomp4 This CMakeLists.txt is working on Orion.  If it is working here, lets merge it.

tclune (approved):

mathomp4:
Okay. Let me make an update to CMakeLists.txt with a comment.

mathomp4 (approved):
@tclune has approved. This is just a "re-approval" after my CMake comment.

jkbk2004:
@mathomp4 are you going to release a new patched version of the mapl-2.21.3 out of this pr?

mathomp4:
> @mathomp4 are you going to release a new patched version of the mapl-2.21.3 out of this pr?

@jkbk2004 I think once we get in #1580, develop should be in a good place. But let me ping the team to see if anything else needs in. If not, I'll work on getting something out today (probably MAPL 2.22).

PR #1586, opened 2022-06-28
Tutorial Proposal

bena-nasa:
I've made the start of a tutorial here for MAPL from a code exercise sense and put this here for feedback. As far as the PP that gives an overall view of MAPL that's something else. 

Here's the idea. I introduced a tutorial directory. I have some GridComps, that increase in "complexity", from doing nothing from creating a grid to creating a grid, an export, and adding two children and connecting.
I then have a single driver executable that I get away with by adding a one more MAPL_AddChild overload to add a child from a dso. I then added a new command argument to that when you create a cap object passing a setservices is optional. If you don't you can specify a dso library on the command line to use as root. I guess another way would have been to specify the name of the library in the CAP.rc.

In any case CapGridComp calls the right interface to MAPL_AddChild to add the root depending whether it was delivered a procedure pointer or a dso library name.

I then have a series of "tutorial" cases that use these (and are driven through cap like the regular model) as well as a simple bash driver where you can pass the installation path and path to the tutorial directory you want to run.
So far are 4 + 5th-ish
They are
Create a single gridcomp, all it does it print hello world in run
Create a single gridcomp, adds an export, writes out in history
Create parent+child (child has export)
Create parent+2 siblings one exports other imports field
Create parent+ (child has import via Extdata, but to do this I need to commit a netcdf file..., that's not text or source code ...)

So questions,
If people are generally ok with what I've done here. We need documentation to go with this. Some top level about how to run teach and for each tutorial.
Should it live in each tutorial, each one could have a README.md describing it. This could be elsewhere such as on the wiki as well.

I assume whatever the documentation for each tutorial would describe what is going on, what is is trying to demonstration, as well as maybe propose something further the user could try.

After these synthetic demonstrations, the next assignment could be like, ok now try such and such in the real model.

Also, I do not "install" the actual tutorial RC files as I'm not sure where we should put this under the installation. Should this be under etc for example?

## Types of changes

## Checklist:


atrayano (dismissed):

mathomp4 (approved):
Cmake good.

PR #1587, opened 2022-07-01
Update develop branch to use ESMA_env v4.2.0

bena-nasa:
This updates MAPL to use ESMA_env v4.2.0, this uses yaFYAML v1.0.3 which has new interfaces relative to what develop was using.
This is zero-diff. It required updating ExtData2G to use the new interfaces here (previously only the MAPL v3 branch was). But as a bonus since yafYaml v1.0.3 has proper return codes during parsing rather than just stops in the code, I can check the error status and gracefully with a message about what file was actually being parsed when the error occurred.

## Types of changes

## Checklist:


mathomp4:
Putting a blocker on this until I can fix up the CI

mathomp4:
I'm blocking this again. I want to discuss some GFE/CMake things with @tclune before we move on.

tclune (dismissed):

PR #1595, opened 2022-07-08
Fixes #1562  ISO8601 time format support

### LINKED ISSUE #1562: Update support for ISO 8601 time format
The logic in the file: 
  https://github.com/GEOS-ESM/MAPL/blob/main/gridcomps/ExtData2G/TimeStringConversion.F90
is a good start at implementing this standard, but several updates are needed/desired.

1. The file should be moved to `./base` as many layers will need to use this.
2. The names must be improved for clarity.
3. Possibly the procedures should be decomposed further.
4. Further checking of error codes should be done (e.g., READ statements without IOSTAT checks)
5. We should have a swath of simple unit tests to verify the implementation. (1st or 2nd step moving forward.)

tclune:
## Types of changes

## Checklist:


base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+   use MAPL_KeywordEnforcerMod
+   use MAPL_ExceptionHandling
+   use ESMF
+   implicit none
+
+!   private
+   public :: DD
+   public :: TD
+   public :: TP
+   public :: TZ
```
tclune, line 18: If these really need to be PUBLIC (hope not), then they will need long informative names.   `ASSOCIATE` can be used to give them shorter names when actually used in expressions.

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+   integer, parameter :: UB_MILLISECOND = 1000
+
+   integer, parameter :: TIMEZONE_DEFAULT = 0
+
+   integer, parameter :: ID_WIDTH = 100
+   integer, parameter :: ID_DAY = 1
+   integer, parameter :: ID_MONTH = ID_WIDTH * ID_DAY
+   integer, parameter :: ID_YEAR = ID_WIDTH * ID_MONTH
+
+   type :: date_fields
```
tclune, line 72: Preference is to have type names in CamelCase.    Also types are usually singular nouns, though more thought would be needed to come up with something that is not plural here.

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+      type(ISO8601DateTime) :: end_datetime
+      integer :: repetitions
+   end type ISO8601TimeInterval
+
+contains
+
+
+! SIMPLE CONSTRUCTORS
+
+   pure function construct_date_fields(fields) result(df)
```
tclune, line 147: Bonus points for using `pure`.

We'll see if it survives the  requirements for error handling ...

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+      integer, dimension(DIM_DF), intent(in) :: fields
+      type(date_fields) :: df
+      df % year = fields(1)
+      df % month = fields(2)
+      df % day = fields(3)
+      df % valid = is_valid_date(fields)
+   end function construct_date_fields
+
+   pure function construct_time_fields(fields) result(tf)
```
tclune, line 156: I will be interested in your reasoning for the signature of this constructor and the preceding one.  Usually one would simply have named scalars rather than an array.  But maybe the use case makes this preferred?  Makes me squeamish about sensitivity to the order of the array elements.

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+
+! LOW-LEVEL UTILITY FUNCTIONS FOR STRING PARSERS
+
+   pure function make_char_array(s) result(a)
+      character(len=*), intent(in) :: s
+      character, dimension(len(s)) :: a
+      integer :: i
+      do i=1, len(s)
+         a(i) = s(i:i)
+      end do
```
tclune, line 178:
```suggestion
            integer :: i
            a = [(s(i:i), i=1,len(s))]
```

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+   pure logical function are_digits(s)
+      character(len=*), intent(in) :: s
+      are_digits = all(is_digit(make_char_array(s)))
+   end function are_digits
+
+   pure logical function is_delimited_date(s)
+      character(len=*), intent(in) :: s
+      is_delimited_date = len(s)==10 .and. are_digits(s(1:4)) &
+         .and. s(5:5)==DD .and. are_digits(s(6:7)) .and. s(8:8)==DD &
+         .and. are_digits(s(9:10))
```
tclune, line 197: Would be nice to avoid the magic numbers here.    Of course using named parameters will make the expression far worse, so may need a different approach.  Maybe something along the lines of:
```
is_delimited_date = (len(s)==???) .and.   are_digits(year_pos(s)) .and. is_separator(separator_pos(s)) .and. ...
```

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+      select case(len(s))
+         case(10)
+            is_delimited_time = s(1:1)==TP .and. are_digits(s(2:3)) &
+               .and. s(4:4)==TD .and. are_digits(s(5:6)) .and. s(7:7)==TD &
+               .and. are_digits(s(8:9)) .and. s(10:10)==TZ
+         case(14)
+            is_delimited_time = s(1:1)==TP .and. are_digits(s(2:3)) &
+               .and. s(4:4)==TD .and. are_digits(s(5:6)) .and. s(7:7)==TD &
+               .and. are_digits(s(8:9)) .and. s(10:10)=='.' &
+               .and. are_digits(s(11:13)) .and. s(14:14)==TZ
```
tclune, line 217: Enough duplication with the previous block that it suggests introduction of another helper procedure?

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+            is_undelimited_time = .FALSE.
+      end select
+
+   end function is_undelimited_time
+
+   pure function split_datetime_string(s) result(parts)
+      character(len=*), intent(in) :: s
+      character(len=len(s)), dimension(2) :: parts
+      parts = split(s, TP)
+   end function split_datetime_string
```
tclune, line 244: I'm wondering if this is clearer than just using `split` directly?

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+      integer, intent(in) :: n
+      is_between = n > lower .and. n < upper
+   end function is_between
+
+   ! Return true if y is a leap year, false otherwise
+   pure logical function is_leap_year(y)
+      integer, intent(in) :: y
+      ! Leap years are years divisible by 400 or (years divisible by 4 and not divisible by 100)
+      is_leap_year = is_factor(y,400) .or. ( is_factor(y, 4) .and. is_not_factor(y, 100) )
+   end function is_leap_year
```
tclune, line 355: This function occurs in other contexts.   Not sure where it is, but probably best that we find it and use it to avoid duplication.   Your version is probably better than the existing one, as it uses `pure`.

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+!      integer, intent(in) :: d
+!      integer :: month_end
+!
+!      if(is_valid_year(y) .and. is_valid_month(m)) then
+!         month_end = get_month_end(y, m)
+!         is_valid_day_ints = is_between(LB_DAY, month_end+1, d) 
+!      else
+!         is_valid_day_ints = .FALSE.
+!      endif
+!   end function is_valid_day_ints
```
tclune, line 408: Reminder to delete if not used.  (This comment is probably more visible than the comment characters - just trying to help.)

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+
+   pure logical function is_valid_hour(hour)
+      integer, intent(in) :: hour
+      is_valid_hour = is_between(LB_HOUR, UB_HOUR, hour)
```
tclune, line 412: If this is the only use of the LB and UB bounds, maybe they could be defined locally here instead?

base/MAPL_ISO8601_Time.F90
```diff
@@ -0,0 +1,720 @@
...
+      integer, dimension(DIM_DF) :: fields
+      integer :: stat = -1
+      
+      call parse_datestring(date_string, fields, stat)
+      if(stat == 0) then
+         df = date_fields(fields)
+         if(df % valid) then
+            date = ISO8601Date(df)
+         else
+            stat == -1
```
tclune, line 630: stat will need to be a return value ...

base/tests/test_MAPL_ISO8601_Time.pf
```diff
@@ -0,0 +1,486 @@
...
+      m = 02
+      s = 27
+      ms = 37
+     
+      tf = time_fields(h, m, s, ms)
+
+      @assertEqual(h, tf % hour)
+      @assertEqual(m, tf % minute)
+      @assertEqual(s, tf % second)
+      @assertEqual(ms, tf % millisecond)
```
tclune, line 64: Is this just testing that the constructor copies the data?   If so, probably not worth the effort.

base/tests/test_MAPL_ISO8601_Time.pf
```diff
@@ -0,0 +1,486 @@
...
+      character, dimension(len(char_string)) :: char_array
+      integer :: i = -1
+
+      char_array = make_char_array(char_string)
+
+      @assertEqual(len(char_string), size(char_array))
+      
+      do i=1, len(char_string)
+         @assertEqual(char_string(i:i), char_array(i))
+      end do
```
tclune, line 126: I think this can just be:
```suggestion     
         @assertEqual(char_string, char_array)
```
But it is possible that this rarely used overload is missing.

tclune (commented):
Impressive!

PR #568, opened 2020-10-08
Fixes #563. Fixes for Leap Year Extrapolation

mathomp4:
## Description

Fixes the issue with leap year extrapolations where the code could look for, say, 2019-02-29.

## Related Issue

Fixes #563 

## Motivation and Context

If you try and extrapolate during leap years, badness happens.

## How Has This Been Tested?

I ran GEOS, but:

1. My usual test is in Y2000, so no extrapolation is done. Even if there is extrapolation, by default it's not on (but since the run didn't crash, must not be used).
2. This does let the 2020-02-27 21z run referenced in #563 work, but...is it right?

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [X] I have tested this change with a run of GEOSgcm (if non-trivial)
- [X] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [ ] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #1601, opened 2022-07-15
Hotfix to Fix LDAS PR #568

mathomp4:
## Description

In testing MAPL 2.23.0 with the LDAS it was found that it ran into issues with binary History output. History kept saying that the binary output file existed before writing. @bena-nasa could only really see this in the GCM when more than one file was written at the same time step, but the LDAS could do it with one.

So, for now we only do the "file exists" check with netCDF4 output. A reading of `GETFILE` makes me think it should do something similar for binary...but not sure. 

## Related Issue

https://github.com/GEOS-ESM/GEOSldas/pull/568

## Motivation and Context

Let's the LDAS use MAPL 2.23

## How Has This Been Tested?

This has been tested by @biljanaorescanin with the LDAS and it fixes their issue. We will test with the GEOSgcm before merging.

## Types of changes

## Checklist:


mathomp4:
I'm going to block this until it can be tested with GEOSgcm. Should be a trivially zero-diff fix for the GCM (we don't do binary History) but better safe than sorry.

mathomp4:
Sigh. Stupid GitHub Actions. I might as well fix it in `main` too.

Issue #1607, opened 2022-07-21
Emit good failure message with duplicates in HISTORY.rc


mathomp4:
This is something @sdrabenh encountered recently and I thought it would make a good update for MAPL. Namely, if you have two of the same field being output in the same collection, like:
```
...
                               'QLTOT'    , 'AGCM'         , 'QL'  ,
                               'RH2'      , 'MOIST'        , 'RH'  ,
                               'RH2'      , 'MOIST'        , 'RH'  ,
                               'O3'       , 'CHEMISTRY'    ,
```
then the model crashes in a not-very-intuitive way:
```
pe=00033 FAIL at line=02766    Base_Base_implementation.F90             <status=13>
pe=00033 FAIL at line=02295    MAPL_HistoryGridComp.F90                 <status=13>
pe=00033 FAIL at line=01806    MAPL_Generic.F90                         <status=13>
pe=00033 FAIL at line=00688    MAPL_CapGridComp.F90                     <status=13>
pe=00033 FAIL at line=00633    MAPL_CapGridComp.F90                     <status=13>
pe=00033 FAIL at line=00934    MAPL_CapGridComp.F90                     <status=13>
pe=00033 FAIL at line=00259    MAPL_Cap.F90                             <status=13>
pe=00033 FAIL at line=00218    MAPL_Cap.F90                             <status=13>
pe=00033 FAIL at line=00152    MAPL_Cap.F90                             <status=13>
pe=00033 FAIL at line=00129    MAPL_Cap.F90                             <status=13>
pe=00033 FAIL at line=00034    GEOSgcm.F90                              <status=13>
Abort(0) on node 33 (rank 33 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 0) - process 33
```

Put in enough prints and you'll figure it out.

However, we should be able to figure out you have duplicates before you crash and emit a nice `_FAIL("You have duplicated "//field)` or whatever.

Issue #1376, opened 2022-02-16
Change default format and protect against illegal values in History.rc


bena-nasa:
A user mistakenly tried to use the settings that were appropriate for the old pre-MAPL2 oserver in History with a post MAPL2 version of MAPL by specifying the "format" of each collection as "CFIOasync". Surprisingly the code did not crash but rather defaulted to binary output. This should be cleaned up. Detect if the user makes an incorrect option, either make CFIO the default and specify binary is needed (the LDAS still uses this for tiled output). More than that, the format options should really be "binary" or "NetCDF" as those are the best descriptions of the two options.

mathomp4:
Oh. Huh. I think this is the same as #1374. Commenting here to keep open

mathomp4:
Probably best strategy is:

1. Add option for `binary` and `nc4` or whatever in `develop` and deprecate it.
2. in MAPL 3 *only* allow `binary and `nc4`

mathomp4:
MAPL2 → Make `CFIOasync` do `CFIO`
MAPL3 → Add `nc4` and `geosbinary` and change default of History to be `nc4`
MAPL4 → No more `CFIO` and `flat`

Issue #1374, opened 2022-02-16
CFIOasync triggers bad path in MAPL History


mathomp4:
A fun bug found by @megandamon. If you ask for a `CFIOasync` format in History (like in the old days), MAPL will happily give you...binary.

Turns out the "default" in history is flat binary. Anything other than a format of `CFIO` falls over into binary.

So, questions for @tclune, @atrayano, @bena-nasa ...

1. Should the default be `CFIO` instead of flat binary?
2. Should we rename the formats from `flat` and `CFIO` to `binary` and `nc4` (or whatever)?

Obviously, we'd have to add in the new words and still allow the old formats and then maybe in MAPL 3 we could enforce `nc4` and `flat` and make `nc4` the default?

mathomp4:
@bena-nasa Is this still an issue in modern MAPL? Or was this fixed at some point?

mathomp4:
Marking long-term as this still might be an issue.

Issue #1285, opened 2022-01-10
Bad chunking and bad write in pFIO isn't trapped correctly


mathomp4:
When looking at issue #626 with @tclune I decided to run a test. I set this for `geosgcm_prog`:
```
...
  geosgcm_prog.grid_label: PC96x49-DC
  geosgcm_prog.chunksize:   500 200 1 1,
  geosgcm_prog.deflate:   1,
...
```
This is an obviously bad chunksize for a 96x49 file. The issue is that the model does not crash when you do this:
```
...
 AGCM Date: 2000/04/14  Time: 23:30:00  Throughput(days/day)[Avg Tot Run]:   2194.5   2601.6   2664.1  TimeRemaining(Est) 000:00:35   26.9% :  23.2% Mem Comm:Used
 AGCM Date: 2000/04/14  Time: 23:45:00  Throughput(days/day)[Avg Tot Run]:   2234.1   2730.8   2799.8  TimeRemaining(Est) 000:00:34   26.9% :  23.1% Mem Comm:Used

 Writing:    483 Slices to File:  mapl12151-gcm-2022Jan06-1day-c24-Institution.geosgcm_prog.20000415_0000z.nc4
 Writing:    153 Slices to File:  mapl12151-gcm-2022Jan06-1day-c24-Institution.geosgcm_surf.20000414_2230z.nc4
 Writing:      5 Slices to File:  mapl12151-gcm-2022Jan06-1day-c24-Institution.geosgcm_budi.20000415_0000z.nc4
 Writing:     21 Slices to File:  mapl12151-gcm-2022Jan06-1day-c24-Institution.geosgcm_buda.20000414_2230z.nc4
pe=00000 FAIL at line=00701    NetCDF4_FileFormatter.F90                <status=-127>
pe=00000 FAIL at line=00297    NetCDF4_FileFormatter.F90                <status=-127>
pe=00000 FAIL at line=00063    HistoryCollection.F90                    <status=-127>
pe=00000 FAIL at line=00029    NetCDF4_put_var.H                        <status=-33>
pe=00000 FAIL at line=00029    NetCDF4_put_var.H                        <status=-49>
pe=00000 FAIL at line=00029    NetCDF4_put_var.H                        <status=-49>
pe=00000 FAIL at line=00029    NetCDF4_put_var.H                        <status=-49>
pe=00000 FAIL at line=00029    NetCDF4_put_var.H                        <status=-49>
[...]

PR #1608, opened 2022-07-22
Fix multiple history issues

bena-nasa:
Fixes: 
#1607
#1285
#1376
#1374
#626 

## Types of changes

## Checklist:

### DIFF
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -3330,9 +3330,9 @@ ENDDO PARSER
 !      check for duplicates
        do i=1,field_set%nfields-1
           do j=i+1,field_set%nfields
-             match_short_name = trim(field_set%fields(1,i)) == trim(field_set%fields(1,j))
+             match_short_name = field_set%fields(1,i) == field_set%fields(1,j)
              match_alias = trim(field_set%fields(3,i)) == trim(field_set%fields(3,j))
-             match_component = trim(field_set%fields(2,i)) == trim(field_set%fields(2,j))
+             match_component = field_set%fields(2,i) == field_set%fields(2,j)
              if (match_short_name) then
                 if (match_component) then
                    _FAIL("Caught collection with duplicate short name: "//trim(field_set%fields(1,i))//" and duplicate component")
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -3331,7 +3331,7 @@ ENDDO PARSER
        do i=1,field_set%nfields-1
           do j=i+1,field_set%nfields
              match_short_name = field_set%fields(1,i) == field_set%fields(1,j)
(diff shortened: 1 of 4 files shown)


atrayano (commented):
@bena-nasa Are we sure the new alias checking works well with the field splitting. My concern is the timing of the check - before we do the splitting, the alias field might "look the same' until is expanded during the splitting

bena-nasa:
> @bena-nasa Are we sure the new alias checking works well with the field splitting. My concern is the timing of the check - before we do the splitting, the alias field might "look the same' until is expanded during the splitting

I'll check that I haven't messed up the field splitting

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3323,6 +3327,22 @@ subroutine parse_fields(cfg, label, field_set, items, rc)
           if(present(items)) call items%push_back(item)
        enddo
        field_set%nfields = m
+!      check for duplicates
+       do i=1,field_set%nfields-1
+          do j=i+1,field_set%nfields
+             match_short_name = trim(field_set%fields(1,i)) == trim(field_set%fields(1,j))
```
mathomp4, line 3333: I'm not sure the `trim()` are needed here. @tclune would know for sure, but I think when you compare strings, like `x == y` then it just goes character by character to the right and as long as all characters in `x` are the same as `y`, it's equal. 

And I think if you have unequal strings, then the shorter is padded with spaces?

bena-nasa:
> @bena-nasa Are we sure the new alias checking works well with the field splitting. My concern is the timing of the check - before we do the splitting, the alias field might "look the same' until is expanded during the splitting

A split collection was zero-diff when I checked a split collection vs the control.

atrayano (approved):

griddedio/GriddedIO.F90
```diff
@@ -254,6 +257,35 @@ subroutine set_default_chunking(this,rc)
...
+        integer, intent(in) :: lev_size
+        integer, optional, intent(out) :: rc
+
+        integer ::  global_dim(3)
+        integer :: status
+
+        call MAPL_GridGet(this%output_grid,globalCellCountPerDim=global_dim,rc=status)
+        _VERIFY(status)
+        if (global_dim(1)*6 == global_dim(2)) then
+           _ASSERT(this%chunking(1) <= global_dim(1), "Bad chunk size")
```
mathomp4, line 271: Should we have more "explicit" asserts here? Like "chunk for X dim, this%chunking(1), must be less than or equal to global_dim(1)"? Not sure.

gridcomps/History/MAPL_HistoryGridComp.F90
bena-nasa: Yep, appear to not need the trims

griddedio/GriddedIO.F90
bena-nasa: Added informative messages

mathomp4 (approved):

PR #1610, opened 2022-07-26
just some cleanup of ExtData2G

bena-nasa:
I was staring at ExtData2G as I'm  prototyping breaking it up into 2 components. I noticed the main grid comp still uses the __RC__ macro in many places so I thought I should just change that, because we are trying to not use the macros that require IAM in new code so why not replace it and get that into develop since this is the code going forward.

I just replaced it with _RC and got rid of all the no longer needed IAMs.

I also noticed there was some redundant code in the run method and also found another completely unused routine I must have forgotten to get rid of so got rid of that while I was at it.

In my testing zero-diff. Should have no functional effect on the code.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Replaced depreciated __RC__ macro with _RC and remove unsed code  in ExtData2G
 - Moved to `checkout@v3` action due to git safe directory issue
 - Added tutorials to CI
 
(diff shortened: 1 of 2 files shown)


mathomp4:
Hmm. Looks like Extdata Cases 22 and 23 failed:
```
The following tests FAILED:
         54 - ExtData1G_case22 (Failed)
         55 - ExtData2G_case22 (Failed)
         56 - ExtData1G_case23 (Failed)
         57 - ExtData2G_case23 (Failed)
Errors while running CTest
```

I'm not sure why, but I'll attach the logfile from the test here:
[ctest.log](https://github.com/GEOS-ESM/MAPL/files/9198102/ctest.log)

It's rather...enormous!

mathomp4 (approved):

PR #1622, opened 2022-08-05
Fixes #1621

### LINKED ISSUE #1621: Make History check for file existence optional
The change that came it at v2.22.0 to check if a file already exists when History decides it is time for a new file (and thus indicating improper cleanup in the driving scripting) needs to be made optional for NOAA as they apparently rerun but do no clean the temporary work space they are running in. This will be a trivial change.

bena-nasa:
Fixes #1621

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Check for duplicate entries in the History.rc file
 - Check that a user provided chunking in the History.rc is compatible with the output grid
 - If a user request CFIOasync in the History.rc print warning and set to CFIO
-- Make it optional for History to check for pre-existing files before writing
+- Added option allow writing to pre-existing files with History
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Check for duplicate entries in the History.rc file
 - Check that a user provided chunking in the History.rc is compatible with the output grid
 - If a user request CFIOasync in the History.rc print warning and set to CFIO
+- Make it optional for History to check for pre-existing files before writing
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


atrayano (dismissed):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Check for duplicate entries in the History.rc file
 - Check that a user provided chunking in the History.rc is compatible with the output grid
 - If a user request CFIOasync in the History.rc print warning and set to CFIO
+- Make it optional for History to check for pre-existing files before writing
```
tclune, line 20:
```suggestion
- Added option allow writing to pre-existing files with History
```

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -116,6 +116,7 @@ module MAPL_HistoryGridCompMod
      logical                             :: integer_time
      integer                             :: collectionWriteSplit
      integer                             :: serverSizeSplit
+     logical                             :: check_if_file_exists
```
tclune, line 119:
```suggestion
     logical                             :: fail_if_file_exists
```

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -537,6 +538,9 @@ subroutine Initialize ( gc, import, dumexport, clock, rc )
     call ESMF_ConfigGetAttribute(config, value=cFileOrder,         &
                                          label='FileOrder:', default='ABC', rc=status)
     _VERIFY(STATUS)
+    call ESMF_ConfigGetAttribute(config, value=intState%check_if_file_exists,  &
+                                         label='Check_If_File_Exists:', default=.true., _RC)
```
tclune, line 542:
```suggestion
                                         label='Fail_If_File_Exists:', default=.true., _RC)
```
This is the name that really matters.   Probably should have a bit of discussion.   In general, the default for new options should be `.false.` if we can arrange it.
Variants:

-    `NO_CLOBBER`  (default true)
-    `CLOBBER` (default false).

tclune (changes requested):

gridcomps/History/MAPL_HistoryGridComp.F90
bena-nasa: @tclune Made all the changes, what does no_clobber/clobber have to do with this PR? This has nothing to do with clobber here in history. And the PFIO Server is not clobbering and has never been clobbering now if the file is there, it's just writing whether the file is good or not, if it's bad well it will crash without more fixes in pFIO which I think I see how to do at least track and take action, modifying the API is more working since there are a lot.

tclune (approved):

PR #1631, opened 2022-08-10
Feature/bmauer/fixes #1620

### LINKED ISSUE #1620: PFIO server needs to be smarter and remember who created a file
@tclune here's about what PFIO should probably do with clobbering and History

Recently a change was added in v2.22.0 of History with the following behaviour.
History has logic each time it runs that decides if I am writing, will it be to a new file or an existing file (say you have multiple times per file).
However, it did not check if it decides it is time for a new file, does it exist already for some reason. Maybe you had a previous run in the same working space, maybe you have scripting that copied a file in of the same name like in this issue: [Das Scripting Issue](https://github.com/GEOS-ESM/GEOSadas/issues/188). This can be a problem if said file is not consistent with what you are trying to write.

So the simple solution is that if History decides, hey time for a new file, if somehow the file is already there, just fail which I did. This apparently broke NOAA. Now the "right" thing here is to make sure your run scripting cleans up before running. But I suppose we could make it so that hey if a file is there, just clobber it in a NetCDF sense.

**However, this is not as simple as just making the default when you call the create method in the NETCDF4_FileFormatter (which is just a hook into the NetCDF create) to be clobber for the following reason.**

The PFIO server is not that smart. Basically PFIO server has this logic. You make a request to write some data to some file. PFIO checks, does the file exist already which if you have multiple times per file is perfectly reasonable. If the file does exist, is says, I must have created it already so I"ll just open it and write the hyperslice you told me to. If it doesn't it says, ok, I'll create the file. Even if the create clobbers this doesn't solve the problem.

Because what the file is there and it shouldn't be and PFIO DIDN'T create it. Well, it will still just open it and who knows what's in the file.

So PFIO needs to keep track of, did I create this file, if not, I shouldn't be trying to write to so rather the OPEN the file, I need to call create with clobber.

weiyuan-jiang:
The users now can decide if they want to overwrite the existing history output
## Types of changes

## Checklist:

### DIFF
--- a/pfio/NetCDF4_FileFormatter.F90
+++ b/pfio/NetCDF4_FileFormatter.F90
@@ -155,7 +155,7 @@ contains
       if (present(mode)) then
          pfio_mode=mode
       else
-         pfio_mode=PFIO_CLOBBER
+         pfio_mode=PFIO_NOCLOBBER
       end if
 
       select case (pfio_mode)
@@ -193,7 +193,7 @@ contains
       if (present(mode)) then
          pfio_mode=mode
       else
-         pfio_mode=PFIO_CLOBBER
+         pfio_mode=PFIO_NOCLOBBER
       end if
 
       select case (pfio_mode)
--- a/pfio/NetCDF4_FileFormatter.F90
+++ b/pfio/NetCDF4_FileFormatter.F90
@@ -184,7 +184,7 @@ contains
       integer, optional, intent(in) :: info
       integer, optional, intent(out) :: rc
 
-      integer :: comm_,
+      integer :: comm_
       integer :: info_
       integer :: status
       integer :: mode_
--- a/pfio/NetCDF4_FileFormatter.F90
+++ b/pfio/NetCDF4_FileFormatter.F90
@@ -150,12 +150,21 @@ contains
 
       integer :: status
       integer :: mode_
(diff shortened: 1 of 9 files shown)


pfio/NetCDF4_FileFormatter.F90
```diff
@@ -175,17 +184,25 @@ subroutine create_par(this, file, unusable, mode, comm, info, rc)
...
       integer :: status
       integer :: mode_
+      integer :: pfio_mode
 
       if (present(mode)) then
-         mode_=mode
+         pfio_mode=mode
       else
-         mode_=NF90_CLOBBER
+         pfio_mode=PFIO_CLOBBER
```
weiyuan-jiang, line 196: To be consistent throughout, the default should be PFIO_NOCLOBBER

pfio/NetCDF4_FileFormatter.F90
```diff
@@ -150,12 +150,21 @@ subroutine create(this, file, unusable, mode, rc)
...
```
weiyuan-jiang, line 158: Same, should be PFIO_NOCLOBBER

bena-nasa: changed

bena-nasa (approved):

PR #1643, opened 2022-08-23
Fixes #1636. Add History regrid_method attribute

### LINKED ISSUE #1636: Add regridding method as a per-variable attribute to History output?
In looking at the new-ish `regrid_method` for History (see https://github.com/GEOS-ESM/GEOSgcm_App/pull/344) I thought of something. We can put in `HISTORY.rc` that collection `foo` is regridded conservatively or bilinear or whatever, but as far as I can see, we have no way of knowing that in the history output itself.

ETA: This was rewritten after consulting @tclune and @bena-nasa to say we should instead maybe add it to the *variable* attributes as we could eventually regrid each variable in a collection in a different way.

So now we have, say:
```
        float T(time, lev, lat, lon) ;
                T:_FillValue = 1.e+15f ;
                T:add_offset = 0.f ;
                T:fmissing_value = 1.e+15f ;
                T:long_name = "air_temperature" ;
                T:missing_value = 1.e+15f ;
                T:scale_factor = 1.f ;
                T:standard_name = "air_temperature" ;
                T:units = "K" ;
                T:valid_range = -1.e+15f, 1.e+15f ;
                T:vmax = 1.e+15f ;
                T:vmin = -1.e+15f ;
```
and then we could add:
```
                T:regrid_method = "bilinear" ;
```
or whatever. (This does seem [legal per the CF Conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#_attributes) for "extra" attributes:

> This standard describes many attributes (some mandatory, others optional), but a file may also contain non-standard attributes. Such attributes do not represent a violation of this standard. Application programs should ignore attributes that they do not recognise or which are irrelevant for their purposes.

I guess the question is: would this be desirable? I mean, in a way it isn't needed if the `HISTORY.rc` that made the collection is available or if the File Spec notes it. Or maybe it doesn't actually *matter*? I mean, we've decided how to regrid the collection that we'll present to the users, the end user I suppose shouldn't care since it's been regridded for them.

Of course, it's also not hard for someone to do this after the fact with `ncatted`, say, either. But if it would be desired, it would be simple for MAPL to do this.

I suppose I now summon @tclune and @bena-nasa for their thoughts.

mathomp4:
## Description

This PR adds a new variable metadata to history corresponding to the regrid method used. Testing shows it to be zero-diff and all seems to work.

Also, this fixes a bug where in NEAREST_STOD was not supported by mistake.

## Related Issue

Closes #1636 

## Motivation and Context

We could eventually have different regridding methods per variable. Now we at least have per-collection. By adding some extra metadata, we can at least tell how each variable was regridded:

```
        float BCDP001(time, lat, lon) ;
                BCDP001:_FillValue = 1.e+15f ;
                BCDP001:add_offset = 0.f ;
                BCDP001:fmissing_value = 1.e+15f ;
                BCDP001:long_name = "Carbonaceous Aerosol Dry Deposition (Bin %d)" ;
                BCDP001:missing_value = 1.e+15f ;
                BCDP001:regrid_method = "conserve_2nd" ;
                BCDP001:scale_factor = 1.f ;
                BCDP001:standard_name = "Carbonaceous Aerosol Dry Deposition (Bin %d)" ;
                BCDP001:units = "kg m-2 s-1" ;
                BCDP001:valid_range = -1.e+15f, 1.e+15f ;
                BCDP001:vmax = 1.e+15f ;
                BCDP001:vmin = -1.e+15f ;
```

## How Has This Been Tested?

Ran GEOS and it built and ran and was zero-diff

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add option to build source tarfile when building MAPL standalone. By default this is `OFF`, but can be enabled with
   `-DINSTALL_SOURCE_TARFILE=ON`
+- Added `regrid_method` metadata to History output
 
 ### Changed
 
--- a/base/RegridMethods.F90
+++ b/base/RegridMethods.F90
@@ -19,6 +19,7 @@ module mapl_RegridMethods
    public :: UNSPECIFIED_REGRID_METHOD
    public :: TILING_METHODS
    public :: get_regrid_method
+   public :: translate_regrid_method
 
    enum, bind(c)
       enumerator :: REGRID_METHOD_IDENTITY
@@ -59,7 +60,7 @@ module mapl_RegridMethods
       case ("VOTE")
          int_regrid_method = REGRID_METHOD_VOTE
       case ("FRACTION")
-         int_regrid_method = REGRID_METHOD_FRACTION   
+         int_regrid_method = REGRID_METHOD_FRACTION
       case ("CONSERVE_2ND")
          int_regrid_method = REGRID_METHOD_CONSERVE_2ND
       case ("PATCH")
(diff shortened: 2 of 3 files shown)


mathomp4:
Just realized. I need to test this *without* https://github.com/GEOS-ESM/GEOSgcm_App/pull/344 as they might come in at different times. I mean...it should work, but let's be careful!

mathomp4:
> Just realized. I need to test this _without_ [GEOS-ESM/GEOSgcm_App#344](https://github.com/GEOS-ESM/GEOSgcm_App/pull/344) as they might come in at different times. I mean...it should work, but let's be careful!

I have done this and all seems to work and I get `regrid_method`. So...I'll undraft!

tclune (approved):

PR #1650, opened 2022-08-31
Feature/aoloso/load balance

aoloso:
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Added
+
+-Added an overloaded interface for MAPL_BalanceWork to handle both REAL32 and REAl64
 ## [Unreleased]
 
 ### Fixed


atrayano (commented):
Should we change line 87 in MAPL_LoadBalance.F90 from real to real(MAPL_r4). My concern is that if we ever compile MAPL with -r8 flag (probably many other pieces will break too. Fortunately, we never do this), the overload won't work

tclune (approved):
Definitely a case that screams for templates in the language.

tclune:
> Should we change line 87 in MAPL_LoadBalance.F90 from real to real(MAPL_r4). My concern is that if we ever compile MAPL with -r8 flag (probably many other pieces will break too. Fortunately, we never do this), the overload won't work

I am not worried about anyone using `-r8` on MAPL; we'll never support that in our CMake, and I cannot imagine anyone rolling their own build.

But a closely related question on which we (I) never really come to a firm conclusion:  what is the right/best way to declare such things.  We have these choices:
 1. `REAL*4` vs `REAL*8`
 2. `REAL32` vs `REAL64`
 3. `SELECT_REAL_KIND(...)`  (of which `MAPL_R4` and `MAPL_R8` are proxies)
 4. `C_FLOAT` vs `C_DOUBLE`

None of these is strictly guaranteed to work with all compilers.   (1) is not standard conforming.  For (2), compilers can return a value of -1 if they don't support a given precision.   (3) may return the same KIND for both precision and therefore make the overload illegal.  And (4) the compiler may not have a "companion" C compiler.     Since all extant machines are IEEE and have companion compilers,  (2-4) are effectively equivalent (assuming smart choices for the args in (3).)

@mathomp4 and I have tended towards choosing (2) as it standardizes where the named constants come from rather than each package defining their own (MAPL, ESMF, FMS, ...) ala (3).    But (4) has that and in theory is necessary for C interoperability, which is not a high priority.  (And in practice is equivalent to (2) anyway.)

Ugh.

PR #1655, opened 2022-09-06
Feature/aoloso/hybrid openmp

aoloso:
## Description

Several changes to support component level hybrid MPI/OpenMP.

## Types of changes

## Checklist:


base/Base/Base_Base_implementation.F90
```diff
@@ -3021,6 +3020,7 @@ module subroutine MAPL_GetHorzIJIndex(npts,II,JJ,lon,lat,lonR8,latR8,Grid, rc)
     real(ESMF_KIND_R8), allocatable :: target_lons(:),target_lats(:)
     real(ESMF_KIND_R8), allocatable :: corner_lons(:,:),corner_lats(:,:),center_lats(:,:),center_lons(:,:)
     type(ESMF_CoordSys_Flag) :: coordSys
+    character(len=ESMF_MAXSTR) :: grid_type
```
tclune, line 3023: It is not obvious to me how this change relates to OpenMP.

base/MaplGrid.F90
```diff
@@ -144,9 +145,130 @@ subroutine set(this, grid, unusable, rc)
...
+
+     do I=1, counts(1)
+        do J=1, counts(2)
+           if (crdOrder == 1) then
+              coord(i,j) = i !ALT + I1 - 1
+           else 
+              coord(i,j) = j !ALT + J1 - 1
+           end if
+        end do
+     end do
```
tclune, line 225:
```suggestion
     do concurrent (i=1:counts(1),j=1:counts(2)
           coord(i,j) =  merge(i,j,crdorder==1)
        end do

```
Trying out advanced syntax.  Needs testing though, as I could easily have made a mistake.

base/MaplGrid.F90
```diff
@@ -144,9 +145,130 @@ subroutine set(this, grid, unusable, rc)
...
+     _FAIL('Unsupported coordinate system:  ESMF_COORDSYS_CART')
+  end if
+
+  if (tk == ESMF_TYPEKIND_R4) then
+     if (coordDimCount(crdOrder)==2) then
+        call ESMF_GridGetCoord(grid, localDE=0, coordDim=crdOrder, &
+             staggerloc=location, &
+             computationalCount=COUNTS,  &
+             farrayPtr=R4D2, rc=status)
+        _VERIFY(STATUS)
```
tclune, line 246: New code should use `_RC` rather than the more verbose `_VERIFY`

generic/MAPL_Generic.F90
```diff
@@ -1830,6 +1839,100 @@ recursive subroutine MAPL_GenericWrapper ( GC, IMPORT, EXPORT, CLOCK, RC)
...
+      type(ESMF_GridComp) :: thread_gc
+      integer :: i, me
+      integer :: userRC
+      character(len=ESMF_MAXSTR) :: comp_name
+      integer :: phase
+
+      !call start_global_time_profiler('run1()')
+      !call ESMF_UserCompGetInternalState (GC, 'GOCART_State', wrap, status)
+      !VERIFY_(status)
+      !self => wrap%ptr
```
tclune, line 1877: Eliminate commented out code.

generic/MAPL_Generic.F90
```diff
@@ -1830,6 +1839,100 @@ recursive subroutine MAPL_GenericWrapper ( GC, IMPORT, EXPORT, CLOCK, RC)
...
+         allocate(statuses(num_threads), __STAT__)
+         allocate(user_statuses(num_threads), __STAT__)
+         statuses=0
+         user_statuses=0
+         !$omp parallel default(none), &
+         !$omp& private(thread, subimport, subexport, thread_gc), &
+         !$omp& shared(gc, statuses, user_statuses, clock, PHASE, MAPL)
+
+         thread = 0
+         !$ thread = omp_get_thread_num()
```
tclune, line 1907: Do we maybe want to bury this construct into something like `MAPL_get_thread_id()` or something like that (not sure of best name).   It's otherwise a bit jarring to realize the "commented" line overrules the line before it.

generic/MaplComponent.F90
```diff
@@ -63,12 +63,12 @@ subroutine run_child(this, name, clock, phase, unusable, rc)
...
-      type(ESMF_State), pointer :: internal_state
-
-      internal_state => this%internal_state
-   end function get_internal_state
+   !function get_internal_state(this) result(internal_state)
+   !   class(MaplComponent), target, intent(in) :: this
+   !   type(ESMF_State), pointer :: internal_state
+!
+   !   internal_state => this%internal_state
+   !end function get_internal_state
```
tclune, line 71: Eliminate commented code.
```suggestion
```

generic/BaseFrameworkComponent.F90
```diff
@@ -43,11 +43,11 @@ function get_component(this) result(component)
...
-      class(BaseFrameworkComponent), target, intent(in) :: this
-
-      state => this%component%get_internal_state()
-   end function get_internal_state
+   !function get_internal_state(this) result(state)
+   !   type(ESMF_State), pointer :: state
+   !   class(BaseFrameworkComponent), target, intent(in) :: this
+   !
+   !  state => this%component%get_internal_state()
+   !end function get_internal_state
```
tclune, line 51: Eliminate commented code.
```suggestion
```

generic/MAPL_Generic.F90
tclune:
```suggestion
```

generic/MaplGenericComponent.F90
```diff
@@ -167,4 +219,254 @@ subroutine set_logger(this, lgr)
...
+     end if
+
+     do i = 1, num_children
+        child => this%get_child(i) 
+        SELECT TYPE (child)
+        CLASS IS (MaplGenericComponent)
+           call child%activate_threading(num_threads)
+        CLASS DEFAULT
+           _FAIL('illegal type for child')
+        END SELECT
```
tclune, line 324: capitalization change
```suggestion
        select type (child)
        class is (MaplGenericComponent)
           call child%activate_threading(num_threads)
        class default
           _FAIL('illegal type for child')
        end select
```

generic/MaplGenericComponent.F90
```diff
@@ -167,4 +219,254 @@ subroutine set_logger(this, lgr)
...
+     
+     num_children = this%get_num_children()
+     do i = 1, num_children
+        child => this%get_child(i) 
+        SELECT TYPE (child)
+        CLASS IS (MaplGenericComponent)
+           call child%deactivate_threading()
+        CLASS DEFAULT
+           _FAIL('illegal type for child')
+        END SELECT
```
tclune, line 369: capitalization

generic/MaplGenericComponent.F90
```diff
@@ -167,4 +219,254 @@ subroutine set_logger(this, lgr)
...
+         gridcomp => this%gridcomp
+     else
+        thread = 0
+        !$ thread = omp_get_thread_num()
+        gridcomp => this%subcomponents(thread+1)%gridcomp
+     end if
+
+  end function get_gridcomp
+
+  function make_subgridcomps(this, num_grids, unusable, rc) result(subgridcomps)
```
tclune, line 394: I'm sure there is a reason that this procedure cannot go into `OpenMP_Support.F90` with the other "sub" creator procedures.  But please double check just in case it can.

profiler/BaseProfiler.F90
```diff
@@ -136,8 +141,10 @@ subroutine start_name(this, name, rc)
       type(MeterNodePtr), pointer :: node_ptr
       class(AbstractMeterNode), pointer :: node
 
+      !$omp master
       if (this%stack%empty()) this%status = INCORRECTLY_NESTED_METERS
-      _ASSERT_RC(.not. this%stack%empty(),"Timer <"//name// "> should not start when empty.",INCORRECTLY_NESTED_METERS)
+      !_ASSERT_RC(.not. this%stack%empty(),"Timer <"//name// "> should not start when empty.",INCORRECTLY_NESTED_METERS)
```
tclune, line 146: Eliminate commented out code.

tclune (changes requested):
Mostly lots of tiny suggestions.   But I do want to do another pass after you have made these.

atrayano:
I also have some suggestions for changes. I'll write a separate comment after I finish my lunch

base/Base/Base_Base_implementation.F90
aoloso: This was introduced to get MAPL_GetHorzIJIndex to work correctly for mini grids. The original test using 
`if (im_world*6==jm_world) then`
failed for mini grids. Therefore a new test using 
`if(trim(grid_type) == "Cubed-Sphere") then`
was introduced.

generic/MAPL_Generic.F90
aoloso: We might right a `MAPL_get_thread_id()` but won't the same two lines be present in that function too?

tclune: Yes, but ... then at least it is only in _one_ place instead of scattered in a few places.   Plus one can document the odd construct in the helper procedure, whereas documenting each 2 line bit in the code would further undermine legibility.

aoloso: Got it. Should this be part of MetaComp?

tclune: Ideally, we want to push as much logic into the `OpenMP_Support` layer as possible.   Meta is already too large.

aoloso: Okay.

mathomp4:
Uh oh. Need to fix up the GEOSadas-for-MAPL-develop branch...a moment please...

mathomp4:
The UFS failure might need @tclune help:
```
2/7 Test  #8: MAPL.profiler.tests ..............***Failed  Error regular expression found in output. Regex=[Encountered 1 or more failures/errors during testing]  1.13 sec
................ Timer <timer_1> does not match start timer <timer_2>
F....................
Time:         0.005 seconds
  
Failure
 in: 
test_TimeProfiler_suite.test_stop_wrong_meter
  Location: 
[test_TimeProfiler.pf:41]
Failed to throw exception: <Timer <timer_1> does not match start timer <timer_2>>
  
 FAILURES!!!
Tests run: 36, Failures: 1, Errors: 0
, Disabled: 0
*** Encountered 1 or more failures/errors during testing. ***

```

base/MaplGrid.F90
```diff
@@ -144,9 +135,115 @@ subroutine set(this, grid, unusable, rc)
   
     deallocate(jms, ims)
 
+    call GridCoordGet(   this%ESMFGRID, this%LATS       , &
+         Name     = "Latitude"              , &
+         Location = ESMF_STAGGERLOC_CENTER  , &
+         Units    = 99                      , & !ESMFL_UnitsRadians
```
atrayano, line 141: Could we change '99' to a predefine constant, for example ESMFL_UnitsRadians. This unfortunately needs adding 'use ESMFL_Mod' at the top. Similar change for line 147
```suggestion
         Units    = ESMFL_UnitsRadians, &
```

generic/MAPL_Generic.F90
aoloso: Since it will be in `OpenMP_Support` instead of MAPL per se, I am about to name it `get_current_thread_id`. Is this okay?

generic/MaplGenericComponent.F90
```diff
@@ -12,16 +12,56 @@ module mapl_MaplGenericComponent
...
+   use mapl_OpenMP_Support
+   use mapl_MaplGrid
+   !$ use omp_lib
    implicit none
    private
 
    public :: MaplGenericComponent
+   public :: get_grid
+
+   procedure(), pointer :: user_method1 => null()
```
atrayano, line 24: There are several places in the layer where we assume the gridded component has 2 run phases (this is only the first place where we have this assumption). We could easily lift this assumption by looping over phases from 1 to size(this%run_entry_points) and calling ESMF_GridCompSetEntryPoint (instead of set_services)

generic/MAPL_Generic.F90
aoloso: I will add another function to get num_threads for the same reason.

base/MaplGrid.F90
aoloso: Will change, commit , test, and push.

generic/MaplGenericComponent.F90
aoloso: I will try this.

aoloso: @atrayano Do you mind we look at this together?

base/MaplGrid.F90
```diff
@@ -63,32 +65,26 @@ subroutine set(this, grid, unusable, rc)
...
+    call ESMF_VmGet(VM, localPet=this%MYID, petCount=ndes, __RC__)
 
 ! Vertical coordinate must exist and be THE THIRD DIMENSION
 ! ---------------------------------------------------------
 
     this%VERTDIM = 3
 
-    call MAPL_GridGet(this%ESMFGRID, localCellCountPerDim=COUNTS, RC=STATUS)
-    _VERIFY(STATUS)
+    call MAPL_GridGet(this%ESMFGRID, localCellCountPerDim=COUNTS, __RC__)
```
tclune, line 87: Please do a global replace of `__RC__` with `_RC`.   Inside MAPL this is the preferred form.
```suggestion
    call MAPL_GridGet(this%ESMFGRID, localCellCountPerDim=COUNTS, _RC)
```

base/MaplGrid.F90
```diff
@@ -144,9 +136,115 @@ subroutine set(this, grid, unusable, rc)
...
+  else if (tk == ESMF_TYPEKIND_R8) then
+     if (coordDimCount(crdOrder)==2) then
+        call ESMF_GridGetCoord(grid, localDE=0, coordDim=crdOrder, &
+             staggerloc=location, &
+             computationalCount=COUNTS,  &
+             farrayPtr=R8D2, __RC__)
+        allocate(coord(counts(1), counts(2)), __STAT__)
+        coord = conv2rad * R8D2
+     else
+        _RETURN(ESMF_FAILURE)
```
tclune, line 240: Feel free to improve the error message.  Just my best guess from a quick scan.
```suggestion
        _FAIL('Must have 2 gridded dimensions.')
```

base/MaplGrid.F90
```diff
@@ -144,9 +136,115 @@ subroutine set(this, grid, unusable, rc)
...
+     endif
+  else
+     _RETURN(ESMF_FAILURE)
```
tclune, line 243: Change to `_FAIL(msg)` with some interesting msg.

generic/MAPL_Generic.F90
```diff
@@ -3909,13 +4010,16 @@ subroutine MAPL_GenericStateGet (STATE, IM, JM, LM, VERTDIM,                &
...
            ORB2B_LAMBDAP_REF, ORB2B_LAMBDAP_RATE
+      type(MaplGrid), pointer :: temp_grid
 
       if(present(IM)) then
          IM=STATE%GRID%IM
       endif
 
       if(present(JM)) then
-         JM=STATE%GRID%JM
+         !JM=STATE%GRID%JM
```
tclune, line 4020:
```suggestion
```

generic/MaplComponent.F90
```diff
@@ -27,7 +27,7 @@ module mapl_MaplComponent
 
       procedure :: run_child
 
-      procedure :: get_internal_state
+      !procedure :: get_internal_state
```
tclune, line 30:
```suggestion
```

profiler/CMakeLists.txt
```diff
@@ -52,6 +52,11 @@ set (srcs
 esma_add_library (${this} SRCS ${srcs} DEPENDENCIES GFTL_SHARED::gftl-shared GFTL::gftl-v1 GFTL::gftl-v2 MAPL.shared MPI::MPI_Fortran TYPE ${MAPL_LIBRARY_TYPE})
 target_include_directories (${this} PRIVATE ${MAPL_SOURCE_DIR}/include)
 
+#if (NOT CMAKE_Fortran_COMPILER_ID MATCHES "NAG")
```
tclune, line 55: This line either needs to be deleted, or there needs to be a matching `#endif`

tclune, line 55: Should this line (and others like it) be deleted?

mathomp4: I think NAG still has issues with OpenMP and CMake: https://gitlab.kitware.com/cmake/cmake/-/issues/21280

Though I haven't tested in a while.

One question I have is: Where is `OMP` set? I don't see any `option()` or anything for it. And it should probably have a better name as we have code in GEOS that we *always* compile with OpenMP (like, `GEOSgcm.x`), and so this might need a more descriptive name.

aoloso: OMP is set in respective CMakeLists.txt files. Some of the files have if(OMP)/endif() constructs thus OpenMP is enable in those cases by adding -DOMP option to the cmake step. We may enable OpenMP by default in those cases.

tclune (approved):

profiler/CMakeLists.txt
mathomp4: @aoloso I guess that is good for now, as it is a bit of a "hidden" feature, but if this ever goes to production we should have a much more explicit name for an option and it should be an actual CMake `option()`.

mathomp4:
Note: There still seems to be an issue as the CI does not want to build MAPL:

```
/root/project/MAPL/base/MaplGrid.F90:46:9:

   46 |     use ESMFL_Mod, only : ESMFL_UnitsRadians
      |         1
Fatal Error: Cannot open module file 'esmfl_mod.mod' for reading at (1): No such file or directory
compilation terminated.
```
I'm a bit baffled by this as it looks like both `MaplGrid.F90` and `ESMFL_Mod.F90` are both in `base/`

@tclune Could this be a weird dependency issue? MaplGrid depends on ESMFL, ESMFL depends on MAPL_Base, and MAPL_Base depends on MaplGrid?

mathomp4:
Yep. Ninja lays it out:
```
ninja: build stopped: dependency cycle: include/MAPL.base/mapl_abstractgridfactorymod.mod -> include/MAPL.base/mapl_basemod.mod -> include/MAPL.base/mapl_maplgrid.mod -> include/MAPL.base/esmfl_mod.mod -> include/MAPL.base/mapl_abstractgridfactorymod.mod.
```

tclune:
OK - presumably something in this PR triggered this.   Presumably it is MaplGrid.  Looking...

tclune:
OK - this is perverse.    The "units" argument is not even used by ESMFL, so there is no reason to specify it!

In any even the named constant should be moved to `./shared/Constants/InternalConstants.F90` if it is truly needed.

ESMFL should be treated as deprecated, but we can't delete it just yet.

profiler/BaseProfiler.F90
```diff
@@ -100,10 +100,13 @@ subroutine start_self(this, unusable, rc)
       class(KeywordEnforcer), optional, intent(in) :: unusable
       integer, optional, intent(out) :: rc
 
+      !$omp master
       if (this%stack%size()/= 0) this%status = INCORRECTLY_NESTED_METERS
-      _ASSERT_RC(this%stack%size()== 0,"Timer "//this%root_node%get_name()// " is not a fresh self start",INCORRECTLY_NESTED_METERS)
+      !_ASSERT_RC(this%stack%size()== 0,"Timer "//this%root_node%get_name()// " is not a fresh self start",INCORRECTLY_NESTED_METERS)
+      if(this%stack%size() .ne. 0) print*, "Timer "//this%root_node%get_name()// " is not a fresh self start"
```
tclune, line 106: The unit tests rely on an exception being thrown in the bad case, so the PRINT is not an acceptable replacement.    I need to better understand why GFortran rejects the macro inside the OMP region.   Is it the use of any FPP at all?  Or is it the expanded logic and calls?  

If we really can't do the ASSERT then the relevant unit test should be disabled.    I'll have to look up the syntax, but I think  it is something like
`@test`
becomes
`@test(disabled)`

tclune: Short term fix.  Change the failing `@test` to

`@test(ifndef= __GFORTRAN__)`

(@mathomp4 please check that I've spelled the predefined token correctly.)

atrayano: I agree with @tclune , but in addition  it would be better to uncomment line 176 in profiler/BaseProfile.F90 and comment out line 177 (i.e. we need to throw the correct exception in general)

profiler/CMakeLists.txt
aoloso: @tclune : GFortran error message is

    _ASSERT_RC(this%stack%size()== 1,"Stack not empty when timer stopped.",INCORRECTLY_NESTED_METERS)
      |                                                                                                                                                                                                                       ^
Error: invalid branch to/from OpenMP structured block

profiler/BaseProfiler.F90
aoloso: @tclune GFortran error message is:
    _ASSERT_RC(this%stack%size()== 1,"Stack not empty when timer stopped.",INCORRECTLY_NESTED_METERS)
      |                                                                                                                                                                                                                       ^
Error: invalid branch to/from OpenMP structured block

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Removed ESMFL_UnitsRadians from ESMFL_Mod.F90 and replaced it with MAPL_UnitsRadians in InternalConstants.F90. Did a global search/replace. This avoid the circular dependency build error when ESMF_UnitsRadians is used in MaplGrid.F90
+
+## [hybrid-openmp] - 2022-09-14
```
mathomp4, line 13: @aoloso Could you remove this line?

And, you probably want to add something in the Unreleased-Added section about all the work you did here! 😄

profiler/BaseProfiler.F90
```diff
@@ -100,10 +100,17 @@ subroutine start_self(this, unusable, rc)
       class(KeywordEnforcer), optional, intent(in) :: unusable
       integer, optional, intent(out) :: rc
 
+      logical :: empty_stack = .true.
+
+      !$omp master
       if (this%stack%size()/= 0) this%status = INCORRECTLY_NESTED_METERS
-      _ASSERT_RC(this%stack%size()== 0,"Timer "//this%root_node%get_name()// " is not a fresh self start",INCORRECTLY_NESTED_METERS)
+      empty_stack = this%stack%size()== 0
```
tclune, line 107: I'm a little wary that `empty_stack` is a saved variable.   A client would have to do something fairly unusual for this to become an issue, but I'd be happier to have a line _before_ the master section that sets it to `.true.` in each call.

aoloso: Fixed as suggested.

CHANGELOG.md
aoloso: Done!

tclune (approved):

mathomp4:
I tested this (along with #1683) with the GCM and it seems zero-diff and good to go.

mathomp4 (approved):
Forgot to approve. I suppose if @tclune is happy, we can fold this in!

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Removed ESMFL_UnitsRadians from ESMFL_Mod.F90 and replaced it with MAPL_UnitsRadians in InternalConstants.F90. Did a global search/replace. This avoid the circular dependency build error when ESMF_UnitsRadians is used in MaplGrid.F90
+
+### Added
+Several changes to support component level hybrid MPI/OpenMP.
```
mathomp4, line 14: Ooh. Actually, could you fix this up? You need a dash before this line...and you might need to do something with the next. Maybe something like:

```
- Several changes to support component level hybrid MPI/OpenMP.
  - See files OpenMP_Support.F90, EntryPointVector.F90, RunEntryPoint.F90, MaplGenericComponent.F90, MAPL_Generic.F90, and BaseProfiler.F90. The first three are new and the rest  are modified.
```

which renders as:

- Several changes to support component level hybrid MPI/OpenMP.
  - See files OpenMP_Support.F90, EntryPointVector.F90, RunEntryPoint.F90, MaplGenericComponent.F90, MAPL_Generic.F90, and BaseProfiler.F90. The first three are new and the rest  are modified.

mathomp4 (approved):

PR #1658, opened 2022-09-06
fixes # 1656

### LINKED ISSUE #1656: Unused keywords should be removed from History source code
@rtodling was asking what blocksize did today. That was an old tuning option that was used before the switch in History to use ESMF regridding, the PFIO server etc. Although it is still read from the History.rc file if there, when I looked through the code it it not used. Nor in fact were a couple of other options that are global to history (prepost, markdone). These should be removed.

bena-nasa:
Fixes #1656 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


mathomp4 (approved):
Always good to remove dead code. Approve

PR #1659, opened 2022-09-06
fixes #1657

### LINKED ISSUE #1657: In ExtData2G, do not check for file file variable if collection is /dev/null
I checked out main of the GEOSgcm fixture as I saw some of the repos were moved to newer version and did my usual does ExtData2G still work test. I found it was failing as some new variables were added to the TR grid comps ExtData files that are set to /dev/null for now. But those rules did not have a "variable" keyword as that's not needed in this case but the lack of this was causing a crash.
I went back and looked at my test case 19 that tests /dev/null and indeed it looks like I just took an existing file and changed the collection to /dev/null and left the variable there so that's why my test case 19 was working. But when I looked at the logic in ExtData2G, it was a bit screwy in that if the collection was /dev/null I still insisted the variable keyword be there. This is not necessary in that case and should be fixed.

bena-nasa:
Fixes #1657 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed bug that required a /dev/null ExtData entry to still have a file variable name
+
 ### Added
 
 ### Changed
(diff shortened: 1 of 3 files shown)


atrayano (approved):
Looks OK

PR #1683, opened 2022-09-22
add all exports from a direct child

weiyuan-jiang:
In GEOSldas, the exports in Catchment or CatchmentCN need to be bubbled up to Land grid comp and to GEOSens grid comp for average. It is inconvenient to add and repeat all exports.  
## Types of changes

## Checklist:


mathomp4:
I tested this (along with #1655) with the GCM and it seems zero-diff and good to go.

PR #1684, opened 2022-09-22
check userRC in MAPL_GenericInitialize

bena-nasa:
Really trivial change, we were not checking the userRC from the ESMF_GridCompInitialize in MAPL_GenericInitialize

## Types of changes

## Checklist:


tclune (approved):
I assume that this does not fix the other issue though?

bena-nasa:
> I assume that this does not fix the other issue though?

No, but we are missing it, and when Matt saw the original problem it let the code run longer than it should have so I thought lets just get it in there.

PR #1696, opened 2022-09-29
Feature/bmauer/fixes #1682

### LINKED ISSUE #1682: Request that ExtData2G support a masked sum, in a Derived Export
It would be very helpful if ExtData2G could combine masking and algebraic computation of fields within the specification of Derived Exports. As a use-case, the EDGAR methane (CH4) emissions dataset is split into source-specific files; we would like to combine subsets of these files and then apply geographic masks to the results, without creating new (custom tailored) emissions files.  Currently ExtData2G supports masking, or algebraic expressions in derived exports, but not both together.  If ExtData2G could combine the two operations, something like MASK( field_A + field_B + field_C, <masking parameters> ), it would be very useful for tagged tracer experiments.

bena-nasa:
Implements feature request in issue #1682 

## Types of changes

## Checklist:


mathomp4:
@bena-nasa Can you add a CHANGELOG entry? If not, I can attempt it when I go for the merge.

mathomp4 (approved):

PR #1697, opened 2022-09-29
Feature/bmauer/complete tripolar grid factory

bena-nasa:
While I was trying to profile some timings transformations involving the tripolar grid, I realized that the grid factory was not fully implemented. In particular the createFromFileMetadata factory was not implemented.

I've rectified this. This is needed so that I can ingest tripolar files into Regrid_Util.x, ExtData, etc...

What I've done is that if I do create from filemetdata I set that as the "grid_file" it reads the coordinates from. Of course we pass metadata, so I had to add something so that when we use the PFIO file formatter to open and read the file metadata I store the name of the file that the metadata was created from. It was either that or change the makefromfilemetadata interface in every factory.

Also the names that myself or Yuri or Santha those for the grid coordinate file are maybe not what we would want in the output. The grid files Santa and/or Yuri made have:
lons_centers,lat_centers, Lon_corners, lat_corners

But when we write files we write
lons,lats,corner_lons,corner_lats
This is consistent with what we do with the cube.
So I have logic that says if created from metadata the names of the corner/center coordinates are such and such, but if we create from a cf it is what we chose maybe a year ago.

We could reconcile these..., I don't know if it is worth it. I suppose the other choice would be to detect this in the routine where we create the coordinates, query the metadata. For now what I have works.

## Types of changes

## Checklist:


sanAkel:
@bena-nasa next please _mention_ me _explicitly_ with "@sanAkel" that way I know of any issues ... I was looking for something that @mathomp4 pointed me to and found _this_ PR of yours ... or else, I would not have known of the following:

> output. The grid files Santa and/or Yuri made have:
lons_centers,lat_centers, Lon_corners, lat_corners

>But when we write files we write
lons,lats,corner_lons,corner_lats
This is consistent with what we do with the cube.
So I have logic that says if created from metadata the names of the corner/center coordinates are such and such, but if we create from a cf it is what we chose maybe a year ago.

I would suggest that if all okay "at your end", in MAPL, go ahead with _this_ PR, but I prefer to follow up and _fix_ my end- I just had a chat with @atrayano and so it is on me now to generate an `ocean_hgrid.nc` (ocean mosaic):
-  That has `centers` **AND** `corners` (so far, I wrote former only) and that [`MAPL_Tripolar.nc`](https://github.com/GEOS-ESM/GEOS_OceanGridComp/tree/develop/MOM6_GEOSPlug/mom6_app/mk_MAPL_Tripolar) used to generate `corners`. Since I generate the `ocean_hgrid.nc` and I know the centers, I can write out corners and hence eliminate the need for another utility, namely, `mk_MAPL_Tripolar` in our workflow.
- @atrayano also suggested that I write everything in `real*8` as opposed to `real*4`.

Thank you!

base/MAPL_GridManager.F90
```diff
@@ -457,8 +457,10 @@ function make_factory_from_file(this, file_name, unused, force_file_coordinates,
          attr => file_metadata%get_attribute('grid_type')
          attr_value => attr%get_value()
          select type (attr_value)
-         type is (StringWrap)
-            grid_type => attr_value%value
+         type is (character(*))
+            grid_type => attr_value
+         class default
+            _FAIL("grid_type attribute must be stringwrap")
```
tclune, line 463: We might still need the StringWrap  case for gfortran?  Are you sure this can be changed?   I guess if it passes CI ...

tclune (approved):

base/MAPL_GridManager.F90
mathomp4: It seems to have passed CI, but all that means is that GNU can compile it. Maybe @bena-nasa knows if any of our unit tests or integration tests would hit this code?

bena-nasa: @I'll try running Regrid_Util.x on this branch with gfotran with tripolar input or output to exercise this and see.

PR #1702, opened 2022-10-03
Fixes #1194 (fix return macros)

### LINKED ISSUE #1194: return without abort
We may add a conditional in the "RETURN_" macro for it to do the right thing when `I_AM_MAIN` is in effect.
https://github.com/GEOS-ESM/MAPL/blob/0d3a347ec87e3b39e1052be403f1fde6898267ed/include/MAPL_ErrLog.h#L89

Without this, the main program will abort even if `rc` is 0 (or success).

tclune:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `_RETURN` and `RETURN_` macros are no longer defined `I_AM_MAIN` is defined.  These macros were only meant to be used in procedures and would erroneously abort when used in a successful main program.
+- 
 ### Added
 
 ### Changed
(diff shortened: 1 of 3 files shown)


tclune:
The intent was precisely so that one could not use `_RETURN` inside a main program.    Instead one should use `_FAIL` is the examples you sited.   Even better if we can come up with a good error message.

Without this change, my expectation is that all of these would have called `MPI_Abort()`.  If that was not the case, we need to study the macros again to see why that did not happen.

tclune:
Hmm.   I suppose an even better fix would be to have `_RETURN` generate the string "Do not use _RETURN in main programs."   which will generate a nice syntax error that the user might be able to understand.

We'll discuss when you are back online tomorrow.

mathomp4:
Close in favor of #1760

PR #1713, opened 2022-10-04
Feature/wdboggs/extend auto code gen

darianboggs:
## Description

* Update MAPL_GridCompSpecs_ACG.py with additional columns
* Fix bugs
* Update comments
* Clean up files

## Motivation and Context

Some fields (columns) in MAPL_Add________Spec calls present in MAPL client (scientific) code were not supported by the python script. This PR adds those fields. Also, it fixes some edge case bugs, updates comments and docstrings, and cleans up files.

## How Has This Been Tested?

This branch was tested by running GEOSgcm with the versions of the python script in the develop branch and this branch and compared with cmpscratch. The result was zero-diff.

## Types of changes

## Checklist:


Apps/MAPL_GridCompSpecs_ACG.py
```diff
@@ -184,19 +187,25 @@ def dataframe(reader, columns):
...
+        'dims'       : 'dims',
+        'ungridded'  : 'ungridded_dims',
+        'prec'       : 'precision',
+        'cond'       : 'condition',
+        'default'    : 'default',
+        'restart'    : 'restart',
+        'friendlyto' : 'friendlyto',
+        'add2export' : 'add2export',
+        'num_subtiles' : 'num_subtiles,',
+        'averaging_interval' : 'averaging_interval'
```
tclune, line 208: It might be worth trying to shorten some of these longer aliases.   Perhaps

- avg intvl
- n_subtiles
- add2exp

Apps/MAPL_GridCompSpecs_ACG.py
```diff
@@ -249,6 +256,7 @@ def open_with_header(filename):
     f.write(header())
     return f
 
+categories = ("IMPORT","EXPORT","INTERNAL")
```
tclune, line 259: Is there a pythonic way to make this appear more as a global const?   E.g., should it maybe be in all caps?   Not a recommendation, just a question.

tclune (dismissed):
See inline comments.   Also unrelated to this change, but caught my eye:

`optional_options` is a bit redundant.   Perhaps `optional_arguments`?

Apps/MAPL_GridCompSpecs_ACG.py
darianboggs: Change made

darianboggs:
`optional_options` is the options from `all_options` that are optional. (OK, that sounded better in my head.) So, it has to match. One could reasonably posit that both names should change, and one would not be wrong. but I am not one today, so it will have to wait for a future update.

Apps/MAPL_GridCompSpecs_ACG.py
darianboggs: New aliases merged after running zero-diff test.

darianboggs:
After running another zero-diff test (passed), I push up the changes that @tclune suggested.

tclune (approved):

mathomp4:
@darianboggs Could you maybe add a Changelog entry for this? You did add new functionality, so it seems more than a trivial clean up and should probably be reflected in that.

darianboggs:
Closes #1713

mathomp4 (approved):
Seems to work for me as well.

PR #1735, opened 2022-10-14
Fix incorrect value for _FAILURE

### LINKED ISSUE #1734: incorrect value for `_FAILURE`
The existing logic in MAPL error handling interprets a return code of "1" as a very specific type of failure.  Then, because `_FAILURE` is defined as "1" we do not get the `<unknown error>` that is appropriate.

Probably fix is just to change the value to "-1", but another fix would be to edit the list of failures to make the first item be `<unknown error>`.

darianboggs:
Fixes #1734 incorrect value for _FAILURE

## Description

This PR defines the `_FAILURE` macro as -1, which is consistent with `shared/MAPL_ErrorHandling.F90`.

## Motivation and Context

Previously the `_FAILURE` macro was defined as 1 in `include/MAPL_ErrLog.h`, but in `shared/MAPL_ErrorHandling.F90` another constant has this value. This resulted in incorrect error descriptions. 

## How Has This Been Tested?

I tested the PR with a cmpscratch zero-diff test, which passed successfully.

## Types of changes

## Checklist:


darianboggs:
For now, I just changed `_FAILURE` to -1. A fuller modification of the error codes in `shared/MAPL_ErrorHandling.F90` may be prudent, but that is a modification, and it would be part of a broader feature.

atrayano (approved):
Looks good

darianboggs:
Return codes in *nix operating systems are nonnegative, so if `_FAILURE` is passed by the executable (with the current value of -1), it would violate the standard. If `_FAILURE` is caught by MAPL error/exception handling, which then sets the return code, the value of `_FAILURE` must only be internally consistent. On the other hand, if the codes in `shared/MAPL_ErrorHandling.F90` are changed so that `_FAILURE` can be defined as 1, the changes in the internal error codes, including the current error code = 1, may have side effects.

tclune:
@darianboggs makes a good point about the *nix requirements on return codes.   The "larger" fix is really still a rather small fix.

PR #1743, opened 2022-10-17
Update error codes to match _FAILURE macro

darianboggs:
Update error codes so that the `_FAILURE = 1` matches the error codes
## Description

`_FAILURE = 1` and `MAPL_UNKNOWN_ERROR = 1` (enumerator) for consistency. The remain error codes (enumerators) are incremented to accommodate `MAIL_UNKNOWN_ERROR`. `MAPL_SUCCESS = 0` is unchanged.
## Related Issue

#1734

## Motivation and Context

Previously, `_FAILURE = 1` was not consistent with error code `MAPL_UNKNOWN_ERROR = -1` while `MAPL_NO_SUCH_PROPERTY = 1`. This change makes them consistent.

## How Has This Been Tested?

I ran a cmpscratch diff-0 test between this branch (pull request) and main successfully.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Error codes in `shared/MAPL_Error_Handling.F90` are now consistent with `_FAILURE = 1` in `include/MAPL_ErrLog.h`
 
 ### Removed
 
--- a/shared/MAPL_ErrorHandling.F90
+++ b/shared/MAPL_ErrorHandling.F90
@@ -13,31 +13,31 @@ module MAPL_ErrorHandlingMod
    public :: MAPL_abort
 
 
-   public :: MAPL_UNKNOWN_ERROR
    public :: MAPL_SUCCESS
 
+   public :: MAPL_UNKNOWN_ERROR
    public :: MAPL_NO_SUCH_PROPERTY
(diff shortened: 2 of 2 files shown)


mathomp4:
@darianboggs you did this against `main`. I think it should be a PR against `develop`.

darianboggs:
Hmm... I based it off `develop` originally. There were a bunch of diffs related to URL's that were dropped. So @tclune asked me to rebase it off `main`. I started with a fresh checkout from `main` rather than rebase to make it cleaner. So I am unsure how to proceed, now. The actual changes are minor, so it is easy to base it off of either branch, but the build-run-compare cycle is sufficiently long that I'd like to only do it once more. It's a bug fix, so `main` is a plausible base for the change,  but it's really matter of our policy as to where bug fixes are based and applied to. Without the fix, the error handling mishandles error codes of -1, since `_FAILURE = 1` in `include/MAPL_ErrorLog.h`, but `MAPL_NO_SUCH_PROPERTY = 1` in `shared/MAPL_ErrorHandling.F90`. So I guess we ask @GitJudge 👩🏿‍⚖️? Maybe we apply it to `develop` and then only merge that commit into `main`.

tclune:
I don't really have a strong opinion, but I do tend to think of this as a (harmless) bug.  But it is technically breaking backward compatibility for _broken_ runs of the model, so best to have it as a separate PR on main.

I don't think NOAA would really care though, so I don't see the need for a fresh semver release.

mathomp4 (approved):

PR #1748, opened 2022-10-20
Fixes #1695. Fix build failure with macOS and Intel

### LINKED ISSUE #1695: Build failure on macOS and Intel with MAPL3
I think some odd behavior with Intel + macOS has changed recently. Code that is pretty dang old has now stopped working. To wit:
```
[ 15%] Linking Fortran shared library ../lib/libMAPL.profiler.dylib
... delete a lot of warnings ...
Undefined symbols for architecture x86_64:
  "_mallinfo", referenced from:
      _mapl_mallocgauge_mp_get_measurement_ in MallocGauge.F90.o
ld: symbol(s) not found for architecture x86_64
```

Thing is `MallocGauge.F90` hasn't changed...ever? So this must be an Intel issue? Or something? 

I'm going to assign @tclune as this is his code. I suppose a simple "if darwin and intel don't do this" is all that's needed, but I'm not exactly sure how to correctly do that here.

mathomp4:
## Description

This PR adds `-DsysDarwin` to the MAPL build. For reasons not yet known, this now seems to be required with Intel oneAPI on macOS. The code needing this hasn't changed in months, but ... 🤷🏼 

This echoes what is done in the GCM and is also why the *GCM* with MAPL3 on macOS can build but MAPL itself doesn't. 

Note: I'm putting this on `develop` just in case we add some `#ifdef sysDarwin` code on develop and are baffled why it's failing.

I will test with GEOSgcm to be sure, but as the GCM already defines this, it's just sort of belt-and-suspenders on that.

## Related Issue

Closes #1695 

## Motivation and Context

Allows MAPL3 to build with Intel oneAPI on macOS

## How Has This Been Tested?

I was able to build MAPL on macOS with Intel with this fix.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Add define for `-Dsys${CMAKE_SYSTEM_NAME}` to fix build issue with macOS and Intel (#1695)
+
 ### Added
 
 ### Changed
+
 - Error codes in `shared/MAPL_Error_Handling.F90` are now consistent with `_FAILURE = 1` in `include/MAPL_ErrLog.h`
 
 ### Removed
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,8 @@ else ()
   add_subdirectory (pflogger_stub)
 endif ()
 
+add_definitions(-Dsys${CMAKE_SYSTEM_NAME})
+
 # Special case - MAPL_cfio is built twice with two different precisions.
 add_subdirectory (MAPL_cfio MAPL_cfio_r4)
 add_subdirectory (MAPL_cfio MAPL_cfio_r8)


mathomp4:
> Looks Ok. I am concerned that just adding a definition flag "fixes" the issue. This looks like a bandaid fix, not a cure. OTH, I do not see how this would hurt

Well, @tclune fixed it long ago by protecting code with an `#ifdef sysDarwin` and its opposite:

https://github.com/GEOS-ESM/MAPL/blob/0df2432c67b26500bb51d296994731afa9cf38f8/profiler/MallocGauge.F90#L36-L43

https://github.com/GEOS-ESM/MAPL/blob/0df2432c67b26500bb51d296994731afa9cf38f8/profiler/MallocGauge.F90#L67-L72

The issue is that for some reason, `-DsysDarwin` seems to be needed now when it didn't before!

The GCM never saw this issue with MAPL3 because we do that `add_definitions()` call in the root `CMakeLists.txt`

bena-nasa (approved):

PR #1782, opened 2022-11-03
add subroutine to calculate global IJ index for cubed sphere grid

weiyuan-jiang:
1)This subroutines avoids "search". For a given lat lon, it returns the global index
2)All the processors can calculate the indexes , even the lat lon is not in its domain

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added assert to NetCDF4_get_var.H to print variable name if data retrieval fails
 
-- Added MAPL_GetGlobalHorzIJIndex subroutine
+- Added subroutine to calculate IJ indexes of given lats lons for cubed-sphere grid
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added MAPL_GetGlobalHorzIJIndex subroutine
+
 ### Changed
 
 ### Removed
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -3187,13 +3187,12 @@ contains
     real(ESMF_KIND_R8), allocatable, dimension (:,:) :: xyz
     real(ESMF_KIND_R8), allocatable, dimension (:)   :: x,y,z
     real(ESMF_KIND_R8), allocatable :: max_abs(:)
-    real(ESMF_KIND_R8)              :: sqr2, alpha, dalpha, error, shift
+    real(ESMF_KIND_R8)              :: sqr2, alpha, dalpha, error, shift, Japan_shift
     real(ESMF_KIND_R8), allocatable :: lons(:), lats(:)
+    real(ESMF_KIND_R8), allocatable :: corner_lons(:,:), corner_lats(:,:)
+    integer :: I1, I2, J1, J2
+    logical, save :: grid_is_ok = .false.
 
-!
-! TODO
-! 
-!   make sure the grid is generated with gnomonic_ed, i.e., equal distance of the edge on the sphere surface
 
     if (npts == 0 ) then
       _RETURN(_SUCCESS)
@@ -3211,6 +3210,22 @@ contains
     sqr2   = sqrt(2.0d0)
     alpha  = asin(1.d0/sqrt(3.d0)) ! range [-alpha, alpha]
     dalpha = 2.0d0*alpha/IM_WORLD
+    Japan_shift = MAPL_PI_R8/18
+
+!   make sure the grid is generated with gnomonic_ed with Japan mountain shift
+    if ( .not. grid_is_ok ) then
+       call MAPL_GridGetInterior(GRID,I1,I2,J1,J2)
+       if ( I1 <=1 .and. 1<=I2 .and. &
+            J1 <=1 .and. 1<=J2 ) then
+          allocate(corner_lons(I2-I1+2, J2-J1+2))
+          allocate(corner_lats(I2-I1+2, J2-J1+2))
+          call MAPL_GridGetCorners(Grid,corner_lons,corner_lats,rc=status)
+          _ASSERT( abs((1.750d0*MAPL_PI_R8 - Japan_shift) - corner_lons(1,1)) <= epsilon(1.0), "Grid should have pi/18 shift")
+          _ASSERT( abs(dalpha - (corner_lats(1,2) - corner_lats(1,1))) <= epsilon(1.0), "Grid should be gnomonic_ed")
+          deallocate(corner_lons, corner_lats)
+       endif
+       grid_is_ok = .true.
+    endif
 
     allocate(lons(npts),lats(npts))
     if (present(lon) .and. present(lat)) then
@@ -3222,7 +3237,7 @@ contains
     end if
 
     !shift the conner away from Japan Fuji Mt.
-    lons = lons + MAPL_PI/18
+    lons = lons + Japan_shift
 
     ! get xyz from sphere surface
     allocate(xyz(3, npts), max_abs(npts))
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -3255,8 +3255,6 @@ contains
        where (abs(z + 1.0d0) <=error) z = z+shift
        II = ceiling((atan(y/sqr2) + alpha)/dalpha)
        JJ = ceiling((atan(z/sqr2) + alpha)/dalpha)
-       where (JJ == 0) JJ =1
-       where (JJ == IM_WORLD+1) JJ = IM_WORLD
     endwhere
 
     ! face = 2
@@ -3267,8 +3265,6 @@ contains
        where (abs(z + 1.0d0) <=error) z = z+shift
        II = ceiling((atan(-x/sqr2) + alpha)/dalpha)
        JJ = ceiling((atan(z/sqr2) + alpha)/dalpha)
-       where (JJ == 0) JJ =1
-       where (JJ == IM_WORLD+1) JJ = IM_WORLD
        JJ = JJ + IM_WORLD
     endwhere
 
@@ -3279,8 +3275,6 @@ contains
        where (abs(x +  1.0d0) <=error) x = x+shift
        II = ceiling((atan(-x/sqr2) + alpha)/dalpha)
        JJ = ceiling((atan(-y/sqr2) + alpha)/dalpha)
-       where (JJ == 0) JJ =1
-       where (JJ == IM_WORLD+1) JJ = IM_WORLD
        JJ = JJ + IM_WORLD*2
     endwhere
 
@@ -3291,8 +3285,6 @@ contains
        where (abs(z +  1.0d0) <=error) z = z+shift
        II = ceiling((atan(-z/sqr2) + alpha)/dalpha)
        JJ = ceiling((atan(-y/sqr2) + alpha)/dalpha)
-       where (JJ == 0) JJ =1
-       where (JJ == IM_WORLD+1) JJ = IM_WORLD
        JJ = JJ + IM_WORLD*3
     endwhere
 
@@ -3302,8 +3294,6 @@ contains
        where (abs(z +  1.0d0) <=error) z = z+shift
        II = ceiling((atan(-z/sqr2) + alpha)/dalpha)
        JJ = ceiling((atan(x/sqr2) + alpha)/dalpha)
-       where (JJ == 0) JJ =1
-       where (JJ == IM_WORLD+1) JJ = IM_WORLD
        JJ = JJ + IM_WORLD*4
     endwhere
 
@@ -3311,13 +3301,8 @@ contains
     where (abs(z+1.0d0) <= error)
        II = ceiling((atan(y/sqr2) + alpha)/dalpha)
        JJ = ceiling((atan(x/sqr2) + alpha)/dalpha)
-       where (JJ == 0) JJ =1
-       where (JJ == IM_WORLD+1) JJ = IM_WORLD
        JJ = JJ + IM_WORLD*5
     endwhere
-
-    where(II == 0) II = 1
-    where(II == IM_WORLD+1) II = IM_WORLD
 
      _RETURN(_SUCCESS)
   end subroutine MAPL_GetGlobalHorzIJIndex
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -3205,7 +3205,7 @@ contains
     _VERIFY(STATUS)
     IM_World = dims(1)
     JM_World = dims(2)
-    _ASSERT( IM_WORLD*6 == JM_WORLD, "it only works for cubed-sphere grid")
+    _ASSERT( IM_WORLD*6 == JM_WORLD, "It only works for cubed-sphere grid")
 
     ! define constants
     sqr2   = sqrt(2.0d0) ! distance from center to the mid of an edge for a 2x2x2 cube
@@ -3214,6 +3214,7 @@ contains
     Japan_shift = MAPL_PI_R8/18    
 
 !   make sure the grid is generated with gnomonic_ed with Japan mountain shift
+!   
     if ( .not. grid_is_ok ) then
        call MAPL_GridGetInterior(GRID,I1,I2,J1,J2)
        ! edge of face 1 along longitude
@@ -3222,16 +3223,16 @@ contains
           allocate(corner_lats(I2-I1+2, J2-J1+2))
           call MAPL_GridGetCorners(Grid,corner_lons,corner_lats,rc=status)
           if ( J1 == 1) then
-             print*, "acurate corner lat lon: ", -alpha,  1.750d0*MAPL_PI_R8 - Japan_shift
-             print*, "from getCorner lat lon: ", corner_lats(1,1), corner_lons(1,1)
+             !print*, "acurate corner lat lon: ", -alpha,  1.750d0*MAPL_PI_R8 - Japan_shift
+             !print*, "from getCorner lat lon: ", corner_lats(1,1), corner_lons(1,1)
              _ASSERT( abs((1.750d0*MAPL_PI_R8 - Japan_shift) - corner_lons(1,1)) <= epsilon(1.0), "Grid should have pi/18 shift")
           endif
           do j = J1, J2+1
              c_lat = -alpha + (j-1)*dalpha 
-             if ( abs(c_lat - corner_lats(1,j-J1+1)) > epsilon(1.0)) then
-               print*, "Grid should be gnomonic_ed: ", c_lat, corner_lats(1,j-J1+1) , j
-             endif
-             !_ASSERT( abs(c_lat - corner_lats(1,j-J1+1)) <= epsilon(1.0), "Grid should be gnomonic_ed")
+             _ASSERT( abs(c_lat - corner_lats(1,j-J1+1)) <= 10.0*epsilon(1.0), "could be 1)Grid is NOT gnomonic_ed, 2)lats lons from MAPL_GridGetCorners are NOT accurate (single precision from ESMF)")
+             !if ( abs(c_lat - corner_lats(1,j-J1+1)) > epsilon(1.0)) then
+             !  print*, "Grid should be gnomonic_ed: ", c_lat, corner_lats(1,j-J1+1) , j
+             !endif
           enddo
           deallocate(corner_lons, corner_lats)
        endif
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -3255,107 +3255,104 @@ contains
 
     ! The edge points are assigned in the order of face 1,2,3,4,5,6
 
-    tolerance = epsilon(1.0d0)
-
-    ! face = 1
-    where ( abs(x-1.0d0) <= tolerance)
-       II = ceiling((atan(y/sqr2) + alpha)/dalpha)
-       JJ = ceiling((atan(z/sqr2) + alpha)/dalpha)
-       where (JJ == 0 ) JJ = 1
-       where (JJ == IM_WORLD+1 ) JJ = IM_WORLD
-
-    ! face = 2
-    elsewhere (abs(y-1.0d0) <= tolerance)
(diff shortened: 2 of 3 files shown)


base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,144 @@ end function ijsearch
...
+! TODO
+! 
+!   make sure the grid is generated with gnomonic_ed, i.e., equal distance of the edge on the sphere surface
+
+    if (npts == 0 ) then
+      _RETURN(_SUCCESS)
+    endif
+
+    if ( .not. present(grid)) then
+      _ASSERT(.false., "need cubed-sphere grid")
```
mathomp4, line 3203: Hmm. If this is true, should we have a more robust check here? Something that would trigger if a user passed in a lat-lon, tripolar, or EASE grid?

Or is `grid` something that only happens with cube-sphere?

weiyuan-jiang:
@bena-nasa , I compared the old subroutine with the new subroutine. The results are different.  There are 1 or 2 index shift. More importantly,  sometimes the faces are different. ( bold one). Old routine  is on face 6, new routine is on face 2.  I have not done the thorough tests, but I suspect the old subroutine is not accurate. Two reasons:

1) In my C12 gcm run, no subroutine can calculate  the point lat=0, lon=0. It is excluded by all processors.
2) The points here I show should be in face 2.

I will keep test.
 

          lons          lats                  II           JJ

 Bens :  0.3936991      -1.373447               7          68
 Bens :  0.7863982      -1.177097               9          68
 Bens :   1.179097     -0.9807477              11          67
 **Bens :   1.571796     -0.7843981              12          65**
 Bens :   1.964495     -0.5880486              11          14
 Bens :   2.357195     -0.3916990              10          38
 Bens :   2.749894     -0.1953495               8          41
 Bens :   3.142593      1.0000467E-03           6          44
 Bens :   3.535292      0.1973496               5          47
 Bens :   3.927991      0.3936991               3          50
 Bens :   4.320690      0.5900487               2          53
 Bens :   4.713389      0.7863982               5          36
 Bens :   5.106088      0.9827478               4          34
 Bens :   5.498787       1.179097               4          32
 Bens :   5.891486       1.375447               5          31

 Jiang:   0.3936991      -1.373447               7          68
 Jiang:   0.7863982      -1.177097               8          68
 Jiang:    1.179097     -0.9807477              11          68
 **Jiang:    1.571796     -0.7843981               7          13**
 Jiang:    1.964495     -0.5880486               9          14
 Jiang:    2.357195     -0.3916990              10          37
 Jiang:    2.749894     -0.1953495               8          40
 Jiang:    3.142593      1.0000467E-03           6          43
 Jiang:    3.535292      0.1973496               5          45
 Jiang:    3.927991      0.3936991               3          49
 Jiang:    4.320690      0.5900487               2          52
 Jiang:    4.713389      0.7863982               6          36
 Jiang:    5.106088      0.9827478               5          35
 Jiang:    5.498787       1.179097               5          32

base/Base/Base_Base_implementation.F90
weiyuan-jiang: Yes. Actually, we need more details here. Not all cubed-sphere grid can be done in this way.

bena-nasa:
@weiyuan-jiang 
For that point on the different face, just based on the values, that looks to be very near or practically on the edge of the face.
Sounds like maybe there is a problem if the point is exactly on the cell boundary, like I suspect 0,0 is. I assume it something like on the bottom and left edge of the cell the comparison needs to be a >= and on top and right needs to be <, but as it is now I have just > on. the left and bottom edge so in the case where you are exactly on the edge it is possible the point does not get assigned to either

weiyuan-jiang:
> @weiyuan-jiang
> For that point on the different face, just based on the values, that looks to be very near or practically on the edge of the face.
> Sounds like maybe there is a problem if the point is exactly on the cell boundary, like I suspect 0,0 is. I assume it something like on the bottom and left edge of the cell the comparison needs to be a >= and on top and right needs to be <, but as it is now I have just > on. the left and bottom edge so in the case where you are exactly on the edge it is possible the point does not get assigned to either

lat=0 and lon =0 is exactly on the center of the face 1. It is surrounded by cell (6,6), (6,7,), (7,6), (7,7) in C12.  Would you please help to test the subroutine too? You may find a better way to torture this algorithm.

bena-nasa:
> > @weiyuan-jiang
> > For that point on the different face, just based on the values, that looks to be very near or practically on the edge of the face.
> > Sounds like maybe there is a problem if the point is exactly on the cell boundary, like I suspect 0,0 is. I assume it something like on the bottom and left edge of the cell the comparison needs to be a >= and on top and right needs to be <, but as it is now I have just > on. the left and bottom edge so in the case where you are exactly on the edge it is possible the point does not get assigned to either
> 
> lat=0 and lon =0 is exactly on the center of the face 1. It is surrounded by cell (6,6), (6,7,), (7,6), (7,7) in C12. Would you please help to test the subroutine too? You may find a better way to torture this algorithm.

Sure, I'll pull this branch, a bit surprised the existing routine had trouble with 0,0 if it is in the center of the cell, how did you get those values in the numbers you showed, did you just randomly generate coordinates

weiyuan-jiang:
> > > @weiyuan-jiang
> > > For that point on the different face, just based on the values, that looks to be very near or practically on the edge of the face.
> > > Sounds like maybe there is a problem if the point is exactly on the cell boundary, like I suspect 0,0 is. I assume it something like on the bottom and left edge of the cell the comparison needs to be a >= and on top and right needs to be <, but as it is now I have just > on. the left and bottom edge so in the case where you are exactly on the edge it is possible the point does not get assigned to either
> > 
> > 
> > lat=0 and lon =0 is exactly on the center of the face 1. It is surrounded by cell (6,6), (6,7,), (7,6), (7,7) in C12. Would you please help to test the subroutine too? You may find a better way to torture this algorithm.
> 
> Sure, I'll pull this branch, a bit surprised the existing routine had trouble with 0,0 if it is in the center of the cell, how did you get those values in the numbers you showed, did you just randomly generate coordinates

It is not on the center of the cell. It is the center of the face. It is the conner of the 4 cellsl

weiyuan-jiang:
> 

I will send you the codes I generate the lat lons

atrayano:
I remembered it slightly wrong: the actual shift is 10 degrees (pi/18). I will send you a code reference shortly...

atrayano:
Check the code below

https://github.com/GEOS-ESM/GFDL_atmos_cubed_sphere/blob/geos/main/tools/fv_grid_tools.F90#L694-L700

tclune:
Another thing to keep in mind is support for stretched grid.   We either need to make the procedure fail under stretched grid (and then have to fix when someone uses it), or the algorithm needs to adapt.

One approach is to "undo the stretch" which will bring the point back to a uniformly spaced grid.  This relies on the stretch function being easily invertible - which I think it is, but memory is poor.

Ideally the algorithm can also "derive" the pi/18 shift as that could in theory change someday.   Indeed,  do we know if NOAA has that shift?    

All this might require some initialization logic that runs once at the beginning to check stretch factors and rotation and saves the info for subsequent IJ index requests.  Not trivial, and I don't like the idea of using save'd variables.   Needs thought.

atrayano:
Good point about the stretched grid!

The shift came from GFDL (nobody wants a big terrain change near any corners of the cube), so most likely NOAA is using the same grid.

weiyuan-jiang:
@atrayano , Thanks a lot. Your information is really helpful. @bena-nasa After the shift, now our results match. I will test more points.  We need to use "plus" instead of "minus" because we shift the sphere not the cube.  https://github.com/GEOS-ESM/MAPL/blob/929e124f7a1af8aea98c0e8170b6a5c4b6fbe0f6/base/Base/Base_Base_implementation.F90#L3226

weiyuan-jiang:
I compared 1000 points randomly generated on C12 surface. The results match with the old program.  I am going to test other resolutions.  Since this is for gnomonic_ed grid and with Japan Fuji mountain shift, can we get those information from the grid passed into this subroutine? If it is not the grid we want, we should report failure. @bena-nasa @atrayano

atrayano:
@weiyuan-jiang This is excellent news. Testing for other resolutions makes sense, although I do not anticipate surprises. Please, make sure you tested C720, since this is the resolution we run operationally (plus any other resolutions Bill might run, like c1440, c2880, etc.). Many folks run c90 or higher. 

As far as getting the info from the grid, this might be a tricky issue: fist, the corner shift is done "silently" deep in the fv routines, second, the UFS case where NOAA runs GOCART, they pass us a grid and we do not have any control. (Of course, as a very temporary kludge, we could attach some attributes to the grid, in both our cubed-sphere grid factory and in the NOAA cap)

tclune:
I know this is a bit harder, but I really think we need a test that purposefully chooses points just inside the corners of cells (some or all) and verifies that it gets the right index.

To do this work backwards.   Given a index (I,J) get the corners and the center.   Create new points near each corner:
1. Convert lat/lon to  cartesian x,y,z
2. Calculate 0.99 * corner  + 0.01 * center
3. Convert back to lat/lon

Now you can call the procedure and see if the new I,J match the original I,J.

weiyuan-jiang:
I tested 1000 points randomly on C720, all the I J  indexes match.

weiyuan-jiang:
To check if the grid is "right" for the new subroutine, we can compare the points along the longitude of the face 1 since it is easy to calculate.  The tests pass for C12 but not C720.  The reason i guess  is they lost the double precision. Here is the results of the print
[
](https://github.com/GEOS-ESM/MAPL/blob/b5444d8d4243621c1893b4039164c987136ccc56/base/Base/Base_Base_implementation.F90#L3225-L3226)

 acurate corner lat lon:  -0.615479708670387        5.32325421858271     
 from getCorner lat lon:  -0.615479767322540        5.32325410842896   

Two failed point from C720                
                        c_lat                  from getCorners    J
Grid should be gnomonic_ed:   0.610350711098134        0.610350847244263   718
Grid should be gnomonic_ed:    -0.610350711098134      -0.610350847244263  4

I can relax the comparison criteria, which I think is not a good idea. So the question here is where do we lose the accuracy? @tclune @bena-nasa @atrayano

bena-nasa:
@weiyuan-jiang I did some playing around, certainly my algorithm fails for points exactly on a corner or edge, since it does no communication and not global, I like @tclune  idea for a test

As far as corners and your comparison to what the corner should be ideally, ESMF stores them in double precision, certainly we retrieve them in double precision on our end we are not doing anything to "lose" precision. But you would have to look in ESMF itself, I don't know if Peggy did everything in double precision, maybe she did the calculations in single, or used a single precision definition of PI, remember the corners in the grid, we pointed her to what FV3 does and said here is the implementation, then she reimplemented it in ESMF so you would have to look there to answer this question.

weiyuan-jiang:
> 
That sounds a good place to ask. Do you have Peggy's contact ? @bena-nasa

atrayano:
@weiyuan-jiang , Peggy retired last year from JPL. I can can the relevant code in the ESMF and will report back here

weiyuan-jiang:
I think it is probably ESMF that messes up the accuracy. In ESMF_Grid.F90, I saw a block that assigns R8 centers and corners with R4 values.  @atrayano @bena-nasa @tclune 

         if (coordTypeKindLocal .eq. ESMF_TYPEKIND_R8) then
           if (docenter) then
             if (coordSysLocal .eq. ESMF_COORDSYS_SPH_DEG) then
               lonPtr = lonPtrR4 * ESMF_COORDSYS_RAD2DEG
               latPtr = latPtrR4 * ESMF_COORDSYS_RAD2DEG
             else
               ! extra copy, can be optimized
               lonPtr = lonPtrR4
               latPtr = latPtrR4
             endif
             deallocate(lonPtrR4, latPtrR4)
           endif
           if (docorner) then
             if (coordSysLocal .eq. ESMF_COORDSYS_SPH_DEG) then
              lonCornerPtr = lonCornerPtrR4 * ESMF_COORDSYS_RAD2DEG
              latCornerPtr = latCornerPtrR4 * ESMF_COORDSYS_RAD2DEG
             else
              lonCornerPtr = lonCornerPtrR4
              latCornerPtr = latCornerPtrR4
             endif
             deallocate(lonCornerPtrR4, latCornerPtrR4)
           endif
         else ! ESMF_TYPE_KIND_R4
            ....
         endif

weiyuan-jiang:
Aha, all the lat lon coming from ESMF are single precision. They didn't really copy GFDL. Here is ESMF's codessubroutine gnomonic_grids(grid_type, im, lon, lat)
 integer, intent(in):: im, grid_type
 real, intent(out):: lon(im+1,im+1)
 real, intent(out):: lat(im+1,im+1)
 integer i, j

  if(grid_type==0) call gnomonic_ed(  im, lon, lat)
  if(grid_type==1) call gnomonic_dist(im, lon, lat)
  if(grid_type==2) call gnomonic_angl(im, lon, lat)


  if(grid_type<3) then
     call symm_ed(im, lon, lat)
     do j=1,im+1
        do i=1,im+1
           lon(i,j) = lon(i,j) - pi
        enddo
     enddo
!    call van2_init(lon, lat, im+1, im+1)
  endif

!   gnomonic_grid = .true.

 end subroutine gnomonic_grids

Here is GFDL's codes

https://github.com/GEOS-ESM/GFDL_atmos_cubed_sphere/blob/8c4707a93fe23a99500f38c601101aac535ee37b/model/fv_grid_utils.F90#L1268-L1289

base/Base/Base_Base_implementation.F90
weiyuan-jiang: add the test of the grid at the very begining

tclune:
@weiyuan-jiang We have asked ESMF to improve the precision.   They said they would.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added assert to NetCDF4_get_var.H to print variable name if data retrieval fails
 
+- Added MAPL_GetGlobalHorzIJIndex subroutine
```
tclune, line 15: Git diff can provide this sort of information.   What is needed here is a description of the capability that the procedure provides.   Still useful to name the actual procedure of course, and yes, the name is somewhat indicative, but ...

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+    endif
+    call MAPL_GridGet(grid, globalCellCountPerDim=dims,rc=status)
+    _VERIFY(STATUS)
+    IM_World = dims(1)
+    JM_World = dims(2)
+    _ASSERT( IM_WORLD*6 == JM_WORLD, "It only works for cubed-sphere grid")
+
+    ! define constants
+    sqr2   = sqrt(2.0d0) ! distance from center to the mid of an edge for a 2x2x2 cube
+    alpha  = asin(1.d0/sqrt(3.d0)) ! angle between the two lines(center to mid and center to end of an edge)
```
tclune, line 3212: I think these might already be available as MAPL constants.   If not, they should still be PARAMETERs in this procedure, rather than variables.

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+    integer :: status
+    integer :: dims(3), IM_WORLD, JM_WORLD
+    real(ESMF_KIND_R8), allocatable, dimension (:,:) :: xyz
+    real(ESMF_KIND_R8), allocatable, dimension (:)   :: x,y,z
+    real(ESMF_KIND_R8), allocatable :: max_abs(:)
+    real(ESMF_KIND_R8)              :: sqr2, alpha, dalpha, error, shift, Japan_shift, c_lat
+    real(ESMF_KIND_R8), allocatable :: lons(:), lats(:)
+    real(ESMF_KIND_R8), allocatable :: corner_lons(:,:), corner_lats(:,:)
+    integer :: I1, I2, J1, J2, j
+    logical, save :: grid_is_ok = .false.
```
tclune, line 3194: No - cannot use a saved variable here.  If it is needed, please explain.    Grid is an input variable, so presumably "ok" potentially changes in each run.

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+             c_lat = -alpha + (j-1)*dalpha 
+             _ASSERT( abs(c_lat - corner_lats(1,j-J1+1)) <= 10.0*epsilon(1.0), "could be 1)Grid is NOT gnomonic_ed, 2)lats lons from MAPL_GridGetCorners are NOT accurate (single precision from ESMF)")
+             !if ( abs(c_lat - corner_lats(1,j-J1+1)) > epsilon(1.0)) then
+             !  print*, "Grid should be gnomonic_ed: ", c_lat, corner_lats(1,j-J1+1) , j
+             !endif
+          enddo
+          deallocate(corner_lons, corner_lats)
+       endif
+       grid_is_ok = .true.
+    endif
```
tclune, line 3240: Make this a separate function that can instead be called at this location.

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+    ! project onto cube
+    max_abs = maxval(abs(xyz), dim=1)
+
+    xyz(1,:) = xyz(1,:)/max_abs
+    xyz(2,:) = xyz(2,:)/max_abs
+    xyz(3,:) = xyz(3,:)/max_abs
+
+    allocate(x, source=xyz(1,:))
+    allocate(y, source=xyz(2,:))
+    allocate(z, source=xyz(3,:))
```
tclune, line 3269:
```suggestion
    x = xyz(1,:))
    y = xyz(2,:))
    z = xyz(3,:))
```

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+
+    II = -1
+    JJ = -1
+
+    ! Here the error and shift are used for edge points.
+    ! the edge points are assigned in the order of face 1,2,3,4,5,6
+    error = epsilon(1.0d0)
```
tclune, line 3276: And should be a PARAMETER
```suggestion
    tolerance = epsilon(1.0d0)
```

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+    shift = 10.0d0*error
+
+    ! face = 1
+    where ( abs(x-1.0d0) <= error )
+       ! Four edges assigned to face 1
+       where (abs(y - 1.0d0) <=error) y = y-shift
+       where (abs(y + 1.0d0) <=error) y = y+shift
+       where (abs(z - 1.0d0) <=error) z = z-shift
+       where (abs(z + 1.0d0) <=error) z = z+shift
```
tclune, line 3285: I am interested in what these lines are for.

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+
+    ! face = 2
+    where (abs(y-1.0d0) <= error)
+       ! Three edges assigned to face 2
+       where (abs(x + 1.0d0) <=error) x = x+shift
+       where (abs(z - 1.0d0) <=error) z = z-shift
+       where (abs(z + 1.0d0) <=error) z = z+shift
+       II = ceiling((atan(-x/sqr2) + alpha)/dalpha)
+       JJ = ceiling((atan(z/sqr2) + alpha)/dalpha)
+       JJ = JJ + IM_WORLD
```
tclune, line 3298: I think it is probably good to make this expression a function of x and z.   You would then just pass different values in each of the blocks for the 6 faces.    "FACE_ID" would be one of the parameters used to do the fake JJ

base/Base/Base_Base_implementation.F90
```diff
@@ -3169,6 +3169,174 @@ end function ijsearch
...
+       II = ceiling((atan(-z/sqr2) + alpha)/dalpha)
+       JJ = ceiling((atan(x/sqr2) + alpha)/dalpha)
+       JJ = JJ + IM_WORLD*4
+    endwhere
+
+    ! face = 6
+    where (abs(z+1.0d0) <= error)
+       II = ceiling((atan(y/sqr2) + alpha)/dalpha)
+       JJ = ceiling((atan(x/sqr2) + alpha)/dalpha)
+       JJ = JJ + IM_WORLD*5
```
tclune, line 3334: Do we really still need the fake JJ?    I would love to find a way to work instead with a proper `(I,J,face)` tuple.

tclune (changes requested):
As I said - I'm going to want this procedure to be extremely high quality.  It is a very important procedure and we want to be able to adapt it in the future.

CHANGELOG.md
weiyuan-jiang: changed

base/Base/Base_Base_implementation.F90
weiyuan-jiang: Ok. Good news is even it checks every time, it is cheap.

base/Base/Base_Base_implementation.F90
weiyuan-jiang: now they become parameters. No calculations any more

base/Base/Base_Base_implementation.F90
weiyuan-jiang: Done

base/Base/Base_Base_implementation.F90
weiyuan-jiang: Cannot use function inside a where struct

base/Base/Base_Base_implementation.F90
weiyuan-jiang: I am open to this. if we use (I, J, face), we need to change a little bit on MAPL_GridGetInterior

base/Base/Base_Base_implementation.F90
weiyuan-jiang: I realize we don't need these shift. With these shift, the edge points are assigned to the face in the order of 1,2,3,4,5,6. If these shift are remove,  the assignment simply reverse the order, i.e., in the order of 6,5,4,3,2,1

tclune: True, but that would be very confusing to someone trying to follow what happens for a given point.    And slightly inefficient as some calculations are repeated.    Blocks that explicitly do not overlap are much preferred, as discussed offline.

weiyuan-jiang:
All looks good to me @tclune

mathomp4 (approved):
Tested and zero-diff

tclune (approved):
I think there is still room to improve the structure here, but we need to get it into a release.

Issue #1747, opened 2022-10-19
UFS/GOCART failure with MAPL 2.28.0 - GridType attribute not present in external grid objects


rmontuoro:
MAPL 2.28.0 causes GOCART to fail when calling `MAPL_GetHorzIJIndex` within the UFS since the `GridType` attribute is not necessarily present in externally provided grid objects.

MAPL fails here:
https://github.com/GEOS-ESM/MAPL/blob/6d6869ab4475aa57c196a14389ef14cf300a6d51/base/Base/Base_Base_implementation.F90#L3052-L3053

mathomp4:
I'm also adding @bena-nasa to this as well as he is the Grid Guru.

tclune:
This might require some work, unless we kludge some assumption that unknown grids are canonical CS just coming in from "outside".

Ideally we could determine the indices through purely ESMF mechanisms, but those don't really have any concept of "face", which constrains things a bit.      Maybe it is something the core team could provide.

rmontuoro:
We could also provide `GridType` as an input argument to `set_grid()`. Not robust, but it's a start.

bena-nasa:
@rmontuoro 
By far the simplest solution is to add an attribute to the grid after this line in your Aerosol_Cap after the set_grid "creates" the grid.
https://github.com/GEOS-ESM/GOCART/blob/main/ESMF/UFS/Aerosol_Cap.F90#L342
```
call ESMF_AttributeSet(grid,name='GridType',value='Cubed-Sphere',_RC)
```
In MAPL we decorate the grid with a "GridType" attribute because we need to make decisions in the "binning" routine based on the class of grid that are unavoidable so we must know what it is. For the 3 most common grid types these are the attribute values:
"Cubed-Sphere"
"Tripolar"
"LatLon"
Obviously since here we are just being provided the already created grid externally, we do not have this set.
If you do no wish to hard code this (although for UFS is there any use case that is not Cubed-Sphere now?), the next simplest option would be to read this from a resource file, then set the attribute. Either way in this case, someone made the grid, and had to know what it was when it was made so within MAPL we need to be told what type of grid it is within our conventions.

tclune:
Note that in MAPL3 we hope to protect the namespace of such attributes with "MAPL_".   But there are lots of ways in which MAPL3 will be not backwards compatible.

mathomp4:
@rmontuoro Did the suggested solution by @bena-nasa work for you?

[...]

PR #1794, opened 2022-11-08
Add option to set_grid to set GridType explicitly.

darianboggs:
Add option to `set_grid` to set `GridType` explicitly.
Move initialization of GridManager prototypes to new subroutine.
Add function to determine if a prototype is present in the GridManager.

## Description

This PR fixes the bug #1747 by adding an option to `set_grid` to set `GridType` explicitly for external grids.
## Related Issue

#1747
## Motivation and Context

Currently, it is not possible to set `GridType` for external grid, which results in an error when `GridType` cannot be determined.
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Fix bug #1747: when an external grid is set, the GridType is incorrect.
-This PR adds an option to set_grid to explicitly set the GridType
+Add an option to set_grid to set GridType explicitly.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Fix bug #1747: when an external grid is set, the GridType is incorrect.
+This PR adds an option to set_grid to explicitly set the GridType
+
 ### Removed
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Fix bug #1747: when an external grid is set, the GridType is incorrect.
+-  When a grid is received from outside of MAPL (e.g., NOAA UFS), MAPL must provide a mechanism to specify a grid-type for internal processing.
 Add an option to set_grid to set GridType explicitly.
 
 ### Removed
--- a/base/MAPL_GridManager.F90
+++ b/base/MAPL_GridManager.F90
@@ -1,4 +1,5 @@
 #include "MAPL_Generic.h"
+#include "unused_dummy.H"
 
 !!!  NOTE: This class implements the Singleton pattern - there should
 !!!        be only one GridManager for the application.  However,
@@ -28,7 +29,6 @@ module MAPL_GridManager_private
    type :: GridManager
       private
       logical :: keep_grids = .false.
-      logical :: has_uninitialized_prototypes = .true.
       integer(kind=ESMF_KIND_I8) :: counter = 0
       type (StringGridFactoryMap) :: prototypes
       type (Integer64GridFactoryMap) :: factories
@@ -60,7 +60,6 @@ module MAPL_GridManager_private
       procedure :: make_clone
       procedure :: get_id
       procedure :: add_factory
-      procedure, private :: intialize_prototypes
       procedure :: is_valid_prototype
 
    end type GridManager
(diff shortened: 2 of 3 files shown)


mathomp4:
Is this a hotfix? If so, we'll need to prepare the CHANGELOG and CMakeLists for that.

darianboggs:
It's a bugfix, definitely. I branched off `main` for that reason. I don't know how important fixing #1747 bug is. I believe the user has a workaround, so this may not be a hotfix.

tclune:
We have generally said that anything which is a "bug" that has already been exhibited/experienced should be treated as a hotfix from a gitflow perspective.    You don't need to rename the branch of course.  Just adding clarity (I hope).  And I'm sure I'll be inconsistent on this at some point.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Fix bug #1747: when an external grid is set, the GridType is incorrect.
```
tclune, line 16:
```suggestion
-  When a grid is received from outside of MAPL (e.g., NOAA UFS), MAPL must provide a mechanism to specify a grid-type for internal processing.
```

base/MAPL_GridManager.F90
```diff
@@ -78,14 +79,83 @@ subroutine add_prototype(this, grid_type, prototype)
...
    end subroutine add_prototype
 
+   ! Is prototype_name present in the prototypes map keys?
+   logical function is_valid_prototype(this, prototype_name, unusable, rc)
+      class (GridManager), intent(inout) :: this
+      character(len=*), intent(in) :: prototype_name
+      class (KeywordEnforcer), optional, intent(in) :: unusable
+      integer, optional, intent(out) :: rc
+      integer :: status
+      logical, save :: initialized = .false.
```
tclune, line 89: Possibly not threadsafe?   Might be ok though.

base/MAPL_GridManager.F90
```diff
@@ -78,14 +79,83 @@ subroutine add_prototype(this, grid_type, prototype)
       
    end subroutine add_prototype
 
+   ! Is prototype_name present in the prototypes map keys?
```
tclune, line 82: I presume that the current logic completely skipped this check?   Good catch if so.

base/MAPL_GridManager.F90
```diff
@@ -78,14 +79,83 @@ subroutine add_prototype(this, grid_type, prototype)
...
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+      type (LatLonGridFactory) :: latlon_factory
+      type (CubedSphereGridFactory) :: cubed_factory
+      type (TripolarGridFactory) :: tripolar_factory
+      type (LlcGridFactory) :: llc_factory
+      type (ExternalGridFactory) :: external_factory
+ 
+      ! This is a local variable to prevent the subroutine from running
```
tclune, line 135: It is late in the day, but ... I know this logic must already exist somewhere.   I am concerned that something has been duplicated.   You probably need to walk me through this PR.    (Not complaining - just want to understand.)

tclune (changes requested):
Have some questions.  Want to discuss.

base/MAPL_GridManager.F90
darianboggs: In theory, it is not threadsafe. I _think_ in practice, it's OK.

`initialized` is a flag to prevent initializing the GridManager `prototypes` StringGridFactory map multiple times. The `make_clone` function was using for this purpose in existing code. In the new function `is_valid_prototype` and subroutine `initialize_prototypes`, I use it, as well. Originally, I added a `initialized` attribute to the GridManager for all three routines to set and get, but that definitely was not threadsafe. With the saved local variable, it's possible the value could vary by thread, which means the initialization could run extra times, but the state of the flag only changes from `.false.` to `.true.`.

The map itself would have greater thread safety issues, but since the initialization sets the entries in the map, multiple initializations should add the same entry to the map multiple times, overwriting each time. The initialization of the map is from constants. I don't know if the map ever changes after the initialization, but it appears to be a constant singleton-ish type for lookups. To be safe, we should probably limit a single thread to the initialization. We can discuss further.

base/MAPL_GridManager.F90
darianboggs: I agree. In the current code, it exists in the `make_clone` function. I moved it to a new auxiliary subroutine to be more accessible, but I don't think that's the best place for it. Having said that, I don't know where the best place is.

tclune: The `make_clone()` procedure explicitly listed the 5 concrete factories I highlighted?   That seems surprising.

tclune: I was just tired and missed the corresponding deletes.

darianboggs:
Fixes #1747

PR #1799, opened 2022-11-09
define real64 constants more consistently

weiyuan-jiang:
define real64 with real64 values. Hopefully it won't cause non-zero diff
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Defined real64 constants in consitent way
+
 ### Added
 - Added assert to NetCDF4_get_var.H to print variable name if data retrieval fails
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Defined real64 constants in consitent way
```
mathomp4, line 12:
```suggestion
- Defined real64 constants in consistent way
```

mathomp4:
Well, as far as I can tell this is zero-diff. Actually sort of surprises me. I guess I figured something would get tiny different with this PR...

mathomp4:
> > Well, as far as I can tell this is zero-diff. Actually sort of surprises me. I guess I figured something would get tiny different with this PR...
> 
> I am a little bit surprised too. I found the inconsistency here when I chased the in-accuracy of lat-lon corners and centers in cubed-sphere grid. maybe we should put on a label "possible non-zero-diff"?

Maybe I should try C720? I suppose if that works, then it's probably as zero-diff as we can test?

mathomp4 (approved):

PR #1810, opened 2022-11-14
Add netCDF quantize support

### LINKED ISSUE #1779: Add quantize support in pFIO
Soon we'll be moving to a Baselibs with the latest netCDF which has support for the quantize updates from @czender, @edwardhartnett, and colleagues. So, we should add support to that here in MAPL/pFIO.

Once we have it in, we would need to do testing to compare it to our current bit-shaving routine (which I think is sort of equivalent bitgrooming with all zeroes?)

mathomp4:
## Description

First, this is all zero-diff as it can't affect the state. Obviously it is non-zero-diff for History if you enable it.

This PR adds quantize support to pFIO. See #1779 for more about this.

This PR has some requirements:

1. Baselibs v7.6.0+ (aka ESMA_env 4.7.0+) as this has the updated netCDF
2. ESMA_cmake v3.20.0 for an update CMake variable for netCDF

If you don't have these, I *think* it should still compile and run, it'll just crash if you try to use quantize.

Now, assuming you do have all the requirements, then you can enable it by adding to History things like:
```
  geosgcm_prog.quantize_level: 4,
  geosgcm_prog.quantize_algorithm: 'BitGroom',
...
  geosgcm_surf.quantize_level: 4,
  geosgcm_surf.quantize_algorithm: 'GranularBR',
...
  geosgcm_moist.quantize_level: 10,
  geosgcm_moist.quantize_algorithm: 'BitRound',
```

Note that for `BitGroom` and `GranularBR`, the `quantize_level` is number of significant _digits_, but for `BitRound` it is the number of significant _bits_. 

## Related Issue

Closes #1779 

## Motivation and Context

netCDF introduced some cool new quantization options thanks to @czender, @edwardhartnett, @rkouznetsov, and more I have no doubt. MAPL does have its own bit-shaving code, but these routines are probably more advanced or, at the least, different and since they are part of netCDF, then the the shaving is more "standard".

## How Has This Been Tested?

I've tested quite a bit on discover and tried out all options. Note I only tried with Intel so I'm sort of using this PR to see if/how I broke GNU!

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ parameters:
 
 # Anchors to prevent forgetting to update a version
 os_version: &os_version ubuntu20
-baselibs_version: &baselibs_version v7.5.0
+baselibs_version: &baselibs_version v7.6.0
 bcs_version: &bcs_version v10.23.0
 tag_build_arg_name: &tag_build_arg_name maplversion
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ parameters:
 
 # Anchors to prevent forgetting to update a version
 os_version: &os_version ubuntu20
-baselibs_version: &baselibs_version v7.6.0
+baselibs_version: &baselibs_version v7.6.1
 bcs_version: &bcs_version v10.23.0
 tag_build_arg_name: &tag_build_arg_name maplversion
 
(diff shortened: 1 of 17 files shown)


mathomp4:
I have tested this and it is zero-diff and runs with old baselibs if you don't enable quantize

tclune (approved):

PR #1829, opened 2022-11-22
Add capability to check MAPL Spec options (fields) to Automatic Code Generator (ACG)

### LINKED ISSUE #1738: Could the ACG do basic sanity checking of the state specs
A suggestion for the auto code generator. 

I just helped a gocart developer. They have some crash in MAPL generic they didn't understand in the MAPL_GenericInitialize sequence. Turned out the issue was that they added new exports in the statespec file. However, one of the the dims was xyz, but they accidentally said the vloc was n (no vlocation) which made the field createfromspec just say no you don't and die at runtime.

Could this be intercepted at the python acg when the spec file is parsed to construct the include file and return a failure there when cmake is run during the build if there are basic inconsistencies like this rather than waiting all the way until runtime deep in the initialization sequence.

darianboggs:
## Description

This PR adds the capability to ACG to check the options of a MAPL Spec against prescribed conditions. This includes the consistency of the options with each other.

## Related Issue

Fixes #1738

## Motivation and Context

This was prompted by linked issue where a runtime error occurred due to an error that could be detected a build time.

Could ACG intercept inconsistencies in MAPL spec options  
> 'when the spec file is parsed to construct the include file and return a failure there when cmake is run during the build if there are basic inconsistencies like this rather than waiting all the way until runtime deep in the initialization sequence.?'

## How Has This Been Tested?

Tested with a successful zero-diff run.

## Types of changes

## Checklist:


mathomp4 (approved):
Model ran for me, and trust in @darianboggs for all the new features. Approve!

PR #1849, opened 2022-12-07
Fixes #1846

### LINKED ISSUE #1846: ExtData2G broken with GNU after testing with v2.32.0
Just making a comment here for so I can reference in the PR.
Here is the issue. After v2.32.2 was released I thought I should try the full GEOS model with ExtData2G and GNU. I was not too worried as all my ExtData2G test cases pass CI which is GNU. But when I ran the model with GNU and ExtData2G, bamn, the model crashed in ExtData. I traced it back to an issue that I thought had gone away. The issue is that in ExtData2G I allow a file to have a block like this:
```
subconfigs:
   - file1
   - file2
```
In the routine that parses the extdata.yaml file, the file is loaded in
class(YAML_Node), allocatable :: config
object.

If this is found in the routine that parses the config,  each entry in that list is another file to process, and recursively calls the same subroutine to read that file (now I wish I had not done this but seemed like reasonable and elegant thing at the time to allow for nested hierarchy of files) and fill in the various dictionaries that get filled from the configuration file(s). This should instantiate a new copy of the routine load that file and return. But somehow with Fortran this corrupts the config object in the caller when using gfortran, intel seems to be fine, @tclune may have some ideas why this may occur.

In any case I will have to restore the trick I had removed (which was keeping a stack of saved yaml_nodes) in a separate type and using those.

bena-nasa:
Fixes #1846 

## Types of changes

## Checklist:


tclune (changes requested):
Should there be a line in the CHANGELOG about adding in all of these RC files?

bena-nasa:
> Should there be a line in the CHANGELOG about adding in all of these RC files?

Added a line about the new files for the new test case

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed bug with mutliple ExtData2G files when using gfortran
```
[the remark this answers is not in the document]
bena-nasa: Updated the comment, this is what I told you about yesterday. Yafyaml + recursiveness + gfortran apparently is bad for reason I don't understand.

tclune (dismissed):

gridcomps/ExtData2G/ExtDataYamlNodeStack.F90
```diff
@@ -0,0 +1,12 @@
+module MAPL_ExtDataYamlNodeStack
+   use yaFyaml
+   use MAPL_ExtDataYamlNodeWrapper
+   implicit none
+
+   public :: stack_depth
+   public :: yaml_node_stack
+
+   integer, save :: stack_depth = 0
+   type(YamlNodeWrapper), save :: yaml_node_stack(20)
```
mathomp4, line 10: I assume this stack is "big enough" at 20? Want to make sure it's not a magic number we should document.

mathomp4:
Looks like the MAPL tests are failing for both Intel and GNU and it's on Case 17 (the new one). Should this be a test ignored by CI (because it's too big) or is this a real failure?

gridcomps/ExtData2G/ExtDataConfig.F90
```diff
@@ -74,6 +80,11 @@ recursive subroutine new_ExtDataConfig_from_yaml(ext_config,config_file,current_
       inquire(file=trim(config_file),exist=file_found)
       _ASSERT(file_found,"could not find: "//trim(config_file))
 
+#ifdef __GFORTRAN__
+      stack_depth = stack_depth + 1
```
tclune, line 84: Need conditional here to trap if `stack_depth` exceeds `MAX_DEPTH`

tclune:
> Looks like the MAPL tests are failing for both Intel and GNU and it's on Case 17 (the new one). Should this be a test ignored by CI (because it's too big) or is this a real failure?

ruh-roh

gridcomps/ExtData2G/ExtDataYamlNodeStack.F90
bena-nasa: It should be for now... hopefully @tclune can fix Yafyaml or at worst case I could try to rewrite this procedure without recursion

bena-nasa:
> Looks like the MAPL tests are failing for both Intel and GNU and it's on Case 17 (the new one). Should this be a test ignored by CI (because it's too big) or is this a real failure?

My bad, I was monkeying with the test case so it would not generate the files as I kept rerunning it and had the files there. But in the actual CI need to run both steps and make the files otherwise can't test

mathomp4 (dismissed):
I'll approve as @tclune approved before. But he is the ultimate arbiter.

gridcomps/ExtData2G/ExtDataYamlNodeStack.F90
```diff
@@ -0,0 +1,14 @@
...
+   use MAPL_ExtDataYamlNodeWrapper
+   implicit none
+
+   public :: max_file_depth 
+   public :: stack_depth
+   public :: yaml_node_stack
+
+   integer, save :: stack_depth = 0
+   integer, parameter :: max_file_depth = 20
+   type(YamlNodeWrapper), save :: yaml_node_stack(max_file_depth)
```
tclune, line 12:
```suggestion
   integer, parameter :: MAX_FILE_DEPTH = 20
   type(YamlNodeWrapper), save :: yaml_node_stack(MAX_FILE_DEPTH)
```

gridcomps/ExtData2G/ExtDataConfig.F90
```diff
@@ -74,6 +80,12 @@ recursive subroutine new_ExtDataConfig_from_yaml(ext_config,config_file,current_
...
+      _ASSERT(stack_depth <=max_file_depth,"yaml config stack too small")
```
tclune, line 85:
```suggestion
      _ASSERT(stack_depth <= MAX_FILE_DEPTH,"yaml config stack too small")
```

Issue #1855, opened 2022-12-08
Build issue with MAPL 2.26.0 on AWS 


lizziel:
I have been successfully using MAPL 2.26.0 in GCHP but am running into a build error when using the GEOS-Chem benchmark pipeline on AWS. I am getting build errors in the new ExtData section of the code, such as this:
```
/tmp/tmp.1eQRnmBr8I/GCHP/src/MAPL/gridcomps/ExtData2G/ExtDataGridCompNG.F90(1134):
error #6632: Keyword arguments are invalid without an explicit interface.   [RC]
```
The line in question for this example is a simple call to ESMF:
`call ESMF_InfoGetFromHost(grid,infoh,rc=status)`

The same error is occurring for calls to `ESMF_InfoIsPresent`. As far as I can see the interface for this in ESMF has not changed recently so I am confused about what the problem is. Have you seen this before?

yantosca:
@lizziel and I have been looking at  the [ESMF_EsmfInfoSync.F90](https://github.com/esmf-org/esmf/blob/c2f82972f6cf47af0b83b998fc07962722677e65/src/Superstructure/InfoAPI/interface/ESMF_InfoSync.F90) module.  One thing I noticed is this interface: 

https://github.com/esmf-org/esmf/blob/c2f82972f6cf47af0b83b998fc07962722677e65/src/Superstructure/InfoAPI/interface/ESMF_InfoSync.F90#L82-L95

The error seems to be that whatever is getting passed as the `host` argument to `call ESMF_InfoGetFromHost(grid,infoh,rc=status)` is an object of a type other than what are listed in the interface.  We are looking into the ESMF version used on our AWS environment.  It may be that we need to update.  

Any other thoughts would be appreciated.

mathomp4:
First, I'll ping @bena-nasa since this is his code. It is odd that it's the one place with ESMF_Info in MAPL 2...

That said, this code has been pretty stable for a while (since MAPL 2.23 or so) and ESMF_Info has been in ESMF for a long time now. I'd be interested to what version of ESMF you are running.

That is definitely an allowed use of `ESMF_InfoGetFromHost` otherwise our MAPL3 would never compile!

mathomp4:
Ah. MAPL 2.23.0 coincided with a move to Baselibs 7.5.0. That means ESMF 8.3.0. 

[...]

PR #1857, opened 2022-12-09
Enforce ESMF version

mathomp4:
## Description

This PR adds a version to the `find_package(ESMF)` call in our CMake. This is help prevent issues like #1855 in the future.

I have tested this in a spack build and it seemed to work just fine there as well.

## Motivation and Context

Make sure users of MAPL have the right version of ESMF.

## Types of changes

## Checklist:


mathomp4:
With some *very* ugly hacks to a MAPL spack package, I was able to test this. It worked in Spack, so it is good.

tclune (approved):

PR #1871, opened 2022-12-13
Add stricter debugging flags when building MAPL as Debug

mathomp4:
## Description

This PR adds some CMake code to apply additional flags when building MAPL as Debug.

This also updates the ESMA_cmake in the `components.yaml` to 3.22.0 which defines these new flags. Note that if you don't have this version of ESMA_cmake, then it's just a no-op as all the variables are empty.

## Motivation and Context

The extra flags are good since Intel is often "too loose" and at least it will throw a lot of warnings and not allow things like `if(integer)` or setting a logical to an integer.

## How Has This Been Tested?

Built and ran GEOS. If it builds...that's good enough!

I've also checked and the flags only appear if Debug.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Update `components.yaml`
+  - ESMA_cmake v3.22.0 (defines stricter debug flags for Intel)
+
 ### Fixed
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-Added `MAPL_find_bounds => find_bounds` and `MAPL_Interval => Interval` to `MAPL.F90` for use when doing component level OpenMP
+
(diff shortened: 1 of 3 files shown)


tclune (approved):
There has got to be a cleaner way to add a flag.  When you have a chance you need to remind me why de-duplication is involved.

mathomp4:
> There has got to be a cleaner way to add a flag. When you have a chance you need to remind me why de-duplication is involved.

The reason is that we have flags like:
```
-diag-error 6188 -diag-error 6192
```
and when I tried it out, many of the CMake methods made them into:
```
-diag-error 6188 6192
```
per CMake:

> The final set of options used for a target is constructed by accumulating options from the current target and the usage requirements of its dependencies. The set of options is de-duplicated to avoid repetition.
> 
> New in version 3.12: While beneficial for individual options, the de-duplication step can break up option groups. For example, -option A -option B becomes -option A B. One may specify a group of options using shell-like quoting along with a SHELL: prefix. The SHELL: prefix is dropped, and the rest of the option string is parsed using the [separate_arguments()](https://cmake.org/cmake/help/latest/command/separate_arguments.html#command:separate_arguments) UNIX_COMMAND mode. For example, "SHELL:-option A" "SHELL:-option B" becomes -option A -option B.

I suppose my question was, should I "hide" this in a file. In the end, probably not worth it...

tclune (approved):

PR #1889, opened 2022-12-20
Extend OMP layer to support ESMF callback states

aoloso:
## Types of changes

## Checklist:


generic/CMakeLists.txt
```diff
@@ -50,7 +50,11 @@ set (srcs
 
   RunEntryPoint.F90
   EntryPointVector.F90
+
+  ESMF_Interfaces.F90
```
tclune, line 54: I recommend we rename this to  `MAPL_ESMF_Interfaces.F90` to avoid confusion.    Minor.

generic/MaplGenericComponent.F90
```diff
@@ -16,11 +16,14 @@ module mapl_MaplGenericComponent
...
+   use mapl_ESMF_Interfaces
+   use mapl_CallbackMap
    implicit none
    private
 
    public :: MaplGenericComponent
    public :: get_grid
+   public :: MAPL_AddMethod
 
    procedure(), pointer :: user_method => null()
```
tclune, line 28:
```suggestion
   procedure(), pointer :: user_method => null()
```
I can't remember why this is a module variable.  Unrelated to this PR, but should be fixed I think.

generic/MaplGenericComponent.F90
```diff
@@ -380,4 +383,25 @@ function get_gridcomp(this) result(gridcomp)
...
+      call get_callbacks(state, callbacks, _RC)
+      call callbacks%insert(label, wrap(userRoutine)
+
+   contains
+
+      function wrap(proc) result(userRoutine)
+         type(CallbackMethodWrapper) :: wrapper
+         procedure(I_CallBackMethod) :: userRoutine
+         wrapper%userRoutine => proc
+      end function wrap
```
tclune, line 403:
```suggestion
      function wrap(userRoutine) result(wrapper)
         type(CallbackMethodWrapper) :: wrapper
         procedure(I_CallBackMethod) :: userRoutine
         wrapper%userRoutine => userRoutine
      end function wrap
```

generic/OpenMP_Support.F90
```diff
@@ -519,20 +523,22 @@ function make_subgridcomps(GridComp, run_entry_points, num_grids, unusable, rc)
         character(len=ESMF_MAXSTR) :: comp_name
         character(len=:), allocatable :: labels(:)
         integer :: phase
+        type(ESMF_Config) :: CF
```
tclune, line 526: I am not aware of why CF is involved in this PR.  Please explain.

generic/OpenMP_Support.F90
```diff
@@ -519,20 +523,22 @@ function make_subgridcomps(GridComp, run_entry_points, num_grids, unusable, rc)
...
 
         allocate(subgridcomps(num_grids))
 
         call ESMF_VMGetCurrent(vm, _RC)
         call ESMF_VMGet(vm, localPET=myPET, _RC)
 
-        call ESMF_GridCompGet(GridComp, name=comp_name, _RC)
+        call ESMF_GridCompGet(GridComp, name=comp_name, CONFIG=CF, _RC)
```
tclune, line 533: CONFIG should not be all caps

generic/OpenMP_Support.F90
```diff
@@ -570,6 +576,67 @@ subroutine set_services(gc, rc)
...
+
+       integer :: n_multi, i
+       integer :: status
+       type(CallbackMethodWrapper), pointer :: wrapper
+       type(CallbackMap), pointer :: callbacks
+       type(CallbackMapIterator) :: iter
+
+       n_multi = size(multi_states)
+       call get_callbacks(state, callbacks, _RC)
+       if (associated(callbacks)) then
```
tclune, line 597: There should be an assertion that callbacks is associated.  Even if the dictionary is empty, we should have a pointer to a valid (albeit empty) dictionary.

generic/OpenMP_Support.F90
```diff
@@ -570,6 +576,67 @@ subroutine set_services(gc, rc)
...
+               call iter%next()
+            end do
+          end associate
+       end if
+
+       _RETURN(ESMF_SUCCESS)
+
+    end subroutine copy_callbacks
+
+    subroutine get_callbacks(state, callbacks, rc)
```
tclune, line 614: `get_callbacks()` should maybe me moved to some other location.   MAPL_Generic ought not to depend so explicitly on this layer.   But if no obvious choice presents itself, it can stay here for now.

generic/OpenMP_Support.F90
aoloso: CF is retrieved in MOIST run method and therefore it is being set for the mini GCs. However it appears not to be used at all therefore it can and will be removed.

generic/CMakeLists.txt
aoloso: File renamed.

generic/OpenMP_Support.F90
aoloso: `CONFIG` has been removed because it is not needed.

mathomp4:
I think I need to look at the label enforcer. It seems a bit...noisy. 😄

generic/MaplGenericComponent.F90
aoloso: It appears not to be needed. The line can and will be removed.

generic/OpenMP_Support.F90
aoloso: Will do!

generic/OpenMP_Support.F90
```diff
@@ -570,6 +574,74 @@ subroutine set_services(gc, rc)
...
+       use mapl_ESMF_Interfaces
+       use mapl_CallbackMap
+       type(ESMF_State), intent(inout) :: state
+       type(CallbackMap), pointer, intent(out) :: callbacks
+       integer, optional, intent(out) :: rc
+
+       integer :: status
+       integer, allocatable :: valueList(:)
```
tclune, line 619:
```suggestion
       integer(kind=ESMF_KIND_I4), allocatable :: valueList(:)
```
Paranoid mode ...

generic/OpenMP_Support.F90
```diff
@@ -570,6 +574,74 @@ subroutine set_services(gc, rc)
...
+       type CallbackMapWrapper
+          type(CallbackMap), pointer :: map
+       end type
+       type(CallbackMapWrapper) :: wrapper
+
+       call ESMF_AttributeGet(state, name='MAPL_CALLBACK_MAP', isPresent=isPresent, _RC)
+       if (.not. isPresent) then ! create callback map for this state
+          allocate(callbacks)
+          wrapper%map => callbacks
+          valueList = transfer(wrapper, i)
```
tclune, line 632:
```suggestion
          valueList = transfer(wrapper, [integer::])
```

generic/OpenMP_Support.F90
```diff
@@ -570,6 +574,74 @@ subroutine set_services(gc, rc)
...
+          call ESMF_AttributeSet(state, name='MAPL_CALLBACK_MAP', valueList=valueList, _RC)
+       end if
+
+       ! Ugly hack to decode ESMF attribute as a gFTL map
+       valueList = transfer(wrapper, i)
```
tclune, line 637:
```suggestion
       valueList = transfer(wrapper, [integer::])
```

tclune:
@aoloso I created a small test program based on these ideas and quickly spotted a mistake in the use of `TRANSFER`.   The 2nd argument needs to be an array when it is an integer, otherwise `transfer()` truncates and gives us just a single integer.     I've put in GitHub "suggestions" to make the necessary changes.   Please try these and let me know how it goes.

generic/OpenMP_Support.F90
aoloso: I thought CF is not needed but it seems to be needed in some of the MOIST modules

generic/OpenMP_Support.F90
```diff
@@ -593,16 +593,17 @@ subroutine copy_callbacks(state, multi_states, rc)
...
+          iter = callbacks%begin()
+          do
+             wrapper => iter%second()
+             do i = 1, n_multi
+                call ESMF_MethodAdd(multi_states(i), label=iter%first(), userRoutine=wrapper%userRoutine, _RC)
+             end do
+             call iter%next()
+             if (iter == callbacks%end()) exit
+          end do
+       end if
```
tclune, line 606: The usual pattern is:
```suggestion
       associate( e => callbacks%end())
          iter = callbacks%begin()
          do while (iter /= e)
             wrapper => iter%second()
             do i = 1, n_multi
                call ESMF_MethodAdd(multi_states(i), label=iter%first(), userRoutine=wrapper%userRoutine, _RC)
             end do
             call iter%next()
          end do
       end associate
```
I.e., the containers are designed to do the "right" thing when empty. So no need to test for that.

generic/OpenMP_Support.F90
```diff
@@ -629,14 +630,15 @@ subroutine get_callbacks(state, callbacks, rc)
        if (.not. isPresent) then ! create callback map for this state
           allocate(callbacks)
           wrapper%map => callbacks
-          valueList = transfer(wrapper, i)
+          !valueList = transfer(wrapper, i)
```
tclune, line 633:
```suggestion
```

generic/OpenMP_Support.F90
```diff
@@ -629,14 +630,15 @@ subroutine get_callbacks(state, callbacks, rc)
...
+          valueList = transfer(wrapper, valueList)
           call ESMF_AttributeSet(state, name='MAPL_CALLBACK_MAP', valueList=valueList, _RC)
        end if
 
        ! Ugly hack to decode ESMF attribute as a gFTL map
-       valueList = transfer(wrapper, i)
+       !valueList = transfer(wrapper, i)
```
tclune, line 639:
```suggestion
```

generic/OpenMP_Support.F90
aoloso: I thought `do while` is deprecated.

tclune (dismissed):

PR #1899, opened 2023-01-04
Fixes #1878. The adds ability to expand %d in the long name, when we …

### LINKED ISSUE #1878: History: Detemplate %d in long name
Per @amdasilva, he would like the `%d` seen in the ACG here:

image

to be expanded to the Bin Number when the fields are split on bins:

> Now, when history expands the bin number for the ungridded dimension, it is supposed to expand the %d above with the bin number; right now, I believe the output file shows “%d” which is not right. Another possibility is to omit the “(Bin %d)” from the rc file and have history add (Bin 001), (Bin 002), etc as it flattens out the ungridded dimension.  Atanas knows what I am talking about.

Note to @atrayano, I guess if only one bin just use `Bin 001`??

atrayano:
…split a field in History

## Description

Expand %d in the long name when splitting field in History

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I have tested this change with a run of GEOSgcm (if non-trivial)

### DIFF
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -3857,7 +3857,7 @@ contains
          i1 = index(longName, "%d")
          _ASSERT(i1>0, "Nothing to expand")
          i2 = i1 + 2 ! size of "%d"
-         tlen = len(trim(longName))
+         tlen = len_trim(longName)
          _ASSERT(tlen + 1 <= len(longName),'LONG_NAME would exceed MAX length after expansion')
          write(tmp,'(i3.3)') i
          newLongName = longName(1:i1-1)//tmp//trim(longName(i2:tlen))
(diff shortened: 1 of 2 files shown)


base/Base/Base_Base_implementation.F90
```diff
@@ -3828,9 +3829,46 @@ module subroutine MAPL_FieldSplit(field, fields, aliasName, rc)
...
+      character(len=ESMF_MAXSTR) :: longName
+      character(len=3) :: tmp
+      character(len=ESMF_MAXSTR) :: newLongName
+
+      do i = 1, size(fields)
+         call ESMF_AttributeGet(fields(i), NAME='LONG_NAME', VALUE=longName, _RC)
+         i1 = index(longName, "%d")
+         _ASSERT(i1>0, "Nothing to expand")
+         i2 = i1 + 2 ! size of "%d"
+         tlen = len(trim(longName))
```
tclune, line 3860:
```suggestion
         tlen = len_trim(longName)
```

tclune (approved):

PR #1902, opened 2023-01-05
Move some sun orbital system parameters to PhysicalConstants

weiyuan-jiang:
Move some sun orbital system parameters to PhysicalConstants 
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Moved some orbital system properties to Constants module
 - Update `components.yaml`
   - ESMA_cmake v3.22.0 (defines stricter debug flags for Intel)
 - Reduced amount of CI tests to reduce cost
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Moved some orbital system properties to Constants module
+
 ### Fixed
 
 ### Removed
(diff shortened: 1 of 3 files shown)


weiyuan-jiang:
@dr0cloud , Hi Peter, I have moved some orbital system properties to PhysicalConstants as part of an effort to cleanup our GEOSldas. Any comment to this PR? Thanks

dr0cloud:
Weiyuan, Lets discuss these changes with Tom before proceeding. I have some concerns. Thanks, Peter.

tclune (changes requested):
@weiyuan-jiang 

After discussion with @dr0cloud  we would prefer that this logic and the physical parameters be relocated instead to the module in `./base/MAPL_sun_uc.F90`.   This keeps everything close to where it is used.

@dr0cloud  also has some feedback about preserving some important comments when the literals are relocated.  He'll provide that separately.

We also may want to rename some of these to be more explicit. This is because things like "year length" have lots of different definitions due to real world perturbations.

Please take a first crack at relocating in the file above, and then add @dr0cloud as a reviewer.

dr0cloud:
@tclune @weiyuan-jiang I'm actually wondering if it wouldn't be better for me to have a crack at this myself? Since I'm not great at git, I would probably start a new feature branch to do this and out you two as reviewers. P.

tclune:
> @tclune @weiyuan-jiang I'm actually wondering if it wouldn't be better for me to have a crack at this myself? Since I'm not great at git, I would probably start a new feature branch to do this and out you two as reviewers. P.

No problem.

tclune:
I will close this PR then.

dr0cloud:
@tclune @weiyuan-jiang Sorry was in a meeting. Yes, I will work open this and submit a new PR, probably early next week. Thanks both. P.

PR #1911, opened 2023-01-09
correct values in the halo corner of LatLon grid

weiyuan-jiang:
The corner values of halo region of latlon grid are filled.
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Added the correct values to halo corner of LatLon grid  
 - Fixed range in halo of LatLonGridFactory
 
 ### Removed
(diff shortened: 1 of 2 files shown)


weiyuan-jiang:
It should be non-zero diff. But we don't have such tests so I tag it with zero-diff

mathomp4:
> It should be non-zero diff. But we don't have such tests so I tag it with zero-diff

@weiyuan-jiang What do you mean it's non-zero-diff but we can't test it? Is it because we can't run with the lat-lon grid?

weiyuan-jiang:
> > It should be non-zero diff. But we don't have such tests so I tag it with zero-diff
> 
> @weiyuan-jiang What do you mean it's non-zero-diff but we can't test it? Is it because we can't run with the lat-lon grid?

I mean we don't have test  of halo of latlon and we have not used it . So we cannot see the non-zero-diff

mathomp4:
> I mean we don't have test of halo of latlon and we have not used it . So we cannot see the non-zero-diff

Ah. Okay. I'll update the branch and get it in.

Issue #1930, opened 2023-01-13
Feature request: always report model throughput after finalize


lizziel:
Reporting the model throughput to log at the end of the run is currently tied to pFlogger and requires MAPL logger level INFO (see [here](https://github.com/GEOS-ESM/MAPL/blob/694c8fb26f9c24dcc3cdde9c45efb00391a03738/gridcomps/Cap/MAPL_Cap.F90#L309-L310)).

It would be useful if it was always output regardless of logging settings. Currently both the throughput estimation per timestep and the model timing info are always printed and are not tied to pFlogger. End-of-run model throughput reporting would fall into the same category of information as these, and always doing it would not impact performance or clutter the log.

The code has a comment that it is going to be removed soon. We use this in GCHP as a handy performance metric so request that it stay. It also seems to be much more accurate than the throughput reported per timestep. For example, my 6-hr run has the following stats. If you do the math with the total run time the correct estimation is 48 days/day.
```
2019/07/01  Time: 06:00:00 Throughput(days/day)[Avg Tot Run]:         63.3         73.4        219.6
   TOTAL                      449.3732       0.0000     449.3732       0
           MAPL: INFO: Model Throughput:       48.008 days per day
```

mathomp4:
@lizziel Huh. You are right. There is a comment from William that it will be removed. I'm not sure why. I depend on it quite a bit (of course, I wrote it so, well... 😄). 

Obviously as you say, the reason it is more accurate overall is that the during-run timers never see the Finalize actions (and perhaps any History writing still going on). When I wrote that, I used the brain dead method of just wall time divided by model time. Nothing fancy.

Frankly, I think that comment probably should go away or maybe change. Perhaps @tclune had a thought to revamp how we calculate or report that a few years ago?

I guess the other question for @tclune is is there a "logger" that always exists? That is, even if a user doesn't provide a `logging.yaml`, is there still a way to use pFlogger to write it "always"?

tclune:
Even if we revamp, it is clear the comment is wrong.   People now rely on it, so it cannot be deleted.

The whole point of pflogger is to be able to control whether some outputs happen and also to control where they go.   I want the timing profile to go through pflogger as in theory we really should want that in a separate file rather than hidden at the bottom of the console log.  (Can be both, of course.)

Isn't INFO the default, which means this in "always on" unless you don't want it to be?    Could be I've not thought it through.

mathomp4:
[...]

PR #1932, opened 2023-01-13
Remove comment saying throughput report will be removed

mathomp4:
Per #1930, remove comment sayings we'll remove a useful throughput timer output.

Note: skipping the changelog because, this doesn't need to be noted in there. 😄


mathomp4:
Closing this in favor of #1938 where I removed the comment there as well. That way @tclune need only approve one PR. :)

PR #1938, opened 2023-01-19
Create MAPL.profiler logger

### LINKED ISSUE #1930: Feature request: always report model throughput after finalize

mathomp4:
## Description

This PR creates a new `MAPL.profiler` logger and moves some current writes to use logger:

* Model throughput
* Per-component timers
* Global timers

For GEOS, this will create pretty much the same log output as long as:

```yaml
   MAPL.profiler:
       handlers: [console_plain]
       propagate: FALSE
       level: WARNING
       root_level: INFO
```
is in `logging.yaml`. As such, an PR will be needed for GEOSgcm_App (see https://github.com/GEOS-ESM/GEOSgcm_App/pull/391)

## Related Issue

Closes #1930 

## Motivation and Context

This will allow users to have most of MAPL at WARNING LEVEL but have the timers always output.

## How Has This Been Tested?

Tested with GEOS. Zero-diff.

## Types of changes

## Checklist:


mathomp4:
@tclune Please take a look at this. It was a pretty one-for-one swap of writes to logger. But perhaps I'm now violating how logger should be used with things like logger-on-one-process:

```fortran
      if (my_rank == 0) then
         report_lines = reporter%generate_report(t_p)
         lgr => logging%get_logger('MAPL.profiler')
         call lgr%info('Report on process: %i0', my_rank)
         do i = 1, size(report_lines)
            call lgr%info('%a', report_lines(i))
         end do
      end if
```

tclune (approved):
Ooh nice.  I had no idea that this was being worked on!

lizziel:
Looks good to me! Thanks for putting this together.

Issue #1918, opened 2023-01-10
Refactor MAPL_Generic.F90 GetResource


tclune:
There are a couple of issues with the existing logic for `MAPL_GetResource` in `generic/MAPL_Generic.F90`.

1. The code (or at least most of it) should be pulled into a separate module.   Something in Base, like `MaplResource.F90`
2. The procedures that have  `MetaComp` arguments should be very thin, but remain in MAPL_Generic.   These procedures just call to the new layer mentioned above, but first extracting the bits of the `MetaComp` that are needed.   I think this is just the config and the comp name.  But I did not dwell on it.
3. Unfortunately, the code is "almost" already organized the way described in (2), but ... not quite.   I think things would be improved if additional procedures were introduced that handle such things as searching through labels:

https://github.com/GEOS-ESM/MAPL/blob/14b9789f91f72f022d0cf037256c7f2375b0c80d/generic/MAPL_Generic.F90#L8459-L8476

With these changes we can bring most of this code over to MAPL3 (separate step).   It also will improve our ability to expand for using YAML when ESMF supports it. (Though only slightly.)

Mentioning @atrayano and @bena-nasa in case they have concerns.

darianboggs:
Got it. This dovetails with the work I am finishing on MAPL_GetResource (#1453). To clean it up. I am refactoring already. I may merge the two.

mathomp4:
I think we can close this now

Issue #1453, opened 2022-03-23
Extend MAPL_GetResource with a config array overload


mathomp4:
At the moment `MAPL_GetResource` has three overloads:

https://github.com/GEOS-ESM/MAPL/blob/8edb2d0d72e663cce3a35b34d2f5703e25f86101/generic/MAPL_Generic.F90#L314-L318

They are:

1. scalar from MAPL_MetaComp
2. scalar from ESMF_Config
3. array from MAPL_MetaComp

This should be extended to have "array from ESMF_Config". Should be fairly simple to do.

I suppose `MAPL_GetResource_array` should be renamed as `MAPL_GetResourceFromMAPL_array` and then we can have a `MAPL_GetResourceFromConfig_array` to match.

I'll assign this to @darianboggs as it might be a nice simple MAPL change to get used to editing MAPL. (And myself, @atrayano or @bena-nasa can help out as needed. Heck, this is the sort of MAPL change I think *I* have a shot at making 😄 )

mathomp4:
@darianboggs This could be a good one

darianboggs:
PR #1453 merged

darianboggs:
Close

PR #1943, opened 2023-01-23
Refactor MAPL_GetResource

darianboggs:
## Description

This PR moves much of the code for MAPL_GetResource generic subroutine from `MAPL_GenericMod` to a new module `MAPL_ResourceMod` in `base`

## Related Issue

#1918

## Motivation and Context

Previously, `MAPL_GetResource` accessed `MAPL_MetaComp` and `ESMF_Config` types, and all the code was located in `MAPL_GenericMod` (`generic/MAPL_Generic.F90`). Most of the code does not need to access code in that directory. The refactored code places most of the code in a new module `MAPL_ResourceMod` in `base`. The specific subroutines that need access to `MAPL_MetaComp` type remain in `MAPL_GenericMod` (**MAPL_GetResourceFromMAPL_scalar** and **MAPL_GetResource_array**) in much abbreviated form. The **MAPL_GetResourceFromConfig_scalar** specific subroutine for `MAPL_GetResource` does not depend on `MAPL_MetaComp` types, so it could be located in `MAPL_ResourceMod`. However, `MAPL_GenericMod` contains the `MAPL_GetResource` interface for the generic subroutine. To avoid putting the **MAPL_GetResourceFromConfig_scalar** specific subroutine for `MAPL_GetResource` in a different module, an abbreviated **MAPL_GetResourceFromConfig_scalar** remains in `MAPL_GenericMod`. All three specific subroutines for  the `MAPL_GetResource` subroutine ( **MAPL_GetResourceFromConfig_scalar**, **MAPL_GetResourceFromMAPL_scalar**, and **MAPL_GetResource_array**) call subroutines in `MAPL_ResourceMod`.

## How Has This Been Tested?

0-diff test

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_Resource.F90
+++ b/base/MAPL_Resource.F90
@@ -9,46 +9,46 @@
 #  undef SET_VAL
 #endif
 
-#define SET_VAL(T) \
+#define SET_VAL(T, VAL) \
 type is (T) ;\
    if (default_is_present .and. .not. label_is_present) then ;\
       select type(default) ;\
       type is(T) ;\
-         val = default ;\
+         VAL = default ;\
       class default ;\
-         _FAIL("Type of 'default' does not match type of 'val'.") ;\
+         _FAIL("Type of 'default' does not match type of 'VAL'.") ;\
       end select ;\
    else ;\
-      call ESMF_ConfigGetAttribute(config, val, label = actual_label, _RC) ;\
+      call ESMF_ConfigGetAttribute(config, VAL, label = actual_label, _RC) ;\
    end if
 
 #ifdef SET_VALS
 #  undef SET_VALS
 #endif
 
-#define SET_VALS(T) \
+#define SET_VALS(T, VALS) \
 type is (T) ;\
    if (default_is_present .and. .not. label_is_present) then ;\
       select type(default) ;\
       type is(T) ;\
-         vals = default ;\
+         VALS = default ;\
       class default ;\
-         _FAIL("Type of 'default' does not match type of 'vals'.") ;\
+         _FAIL("Type of 'default' does not match type of 'VALS'.") ;\
       end select ;\
    else ;\
-      call ESMF_ConfigGetAttribute(config, valuelist = vals, count = count, label = actual_label, _RC) ;\
+      call ESMF_ConfigGetAttribute(config, valuelist = VALS, count = count, label = actual_label, _RC) ;\
    end if
 
 #ifdef SET_STRINGS
 #  undef SET_STRINGS
 #endif
 
-#define SET_STRINGS(T, TS, TF) \
+#define SET_STRINGS(T, TSTR, TFMT) \
 type is (T) ;\
-   type_str = TS ;\
-   val_str = intrinsic_to_string(val, TF) ;\
+   type_str = TSTR ;\
+   val_str = intrinsic_to_string(val, TFMT) ;\
    if (present(default)) then ;\
-      default_str = intrinsic_to_string(default, TF) ;\
+      default_str = intrinsic_to_string(default, TFMT) ;\
    end if
 
 !=============================================================================
@@ -171,12 +171,12 @@ contains
       end if
 
       select type(val)
-      SET_VAL(integer(int32))
-      SET_VAL(integer(int64))
-      SET_VAL(real(real32))
-      SET_VAL(real(real64))
-      SET_VAL(character(len=*))
-      SET_VAL(logical)
+      SET_VAL(integer(int32), val)
+      SET_VAL(integer(int64), val)
+      SET_VAL(real(real32), val)
+      SET_VAL(real(real64), val)
+      SET_VAL(character(len=*), val)
+      SET_VAL(logical, val)
       class default
          _FAIL( "Unupported type")
       end select
@@ -232,12 +232,12 @@ contains
       count = size(vals)
 
       select type(vals)
-      SET_VALS(integer(int32))
-      SET_VALS(integer(int64))
-      SET_VALS(real(real32))
-      SET_VALS(real(real64))
-      SET_VALS(character(len=*))
-      SET_VALS(logical)
+      SET_VALS(integer(int32), vals)
+      SET_VALS(integer(int64), vals)
+      SET_VALS(real(real32), vals)
+      SET_VALS(real(real64), vals)
+      SET_VALS(character(len=*), vals)
+      SET_VALS(logical, vals)
       class default
          _FAIL( "Unsupported type")
       end select
--- a/base/MAPL_Resource.F90
+++ b/base/MAPL_Resource.F90
@@ -20,6 +20,7 @@ module MAPL_ResourceMod
    use MAPL_CommsMod
    use MAPL_Constants, only: MAPL_CF_COMPONENT_SEPARATOR
    use MAPL_ExceptionHandling
+   use MAPL_KeywordEnforcerMod
    use, intrinsic :: iso_fortran_env, only: REAL32, REAL64, int32, int64
 
    ! !PUBLIC MEMBER FUNCTIONS:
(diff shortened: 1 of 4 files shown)


mathomp4:
I changed the base to `develop`. I think it was accidentally on main

mathomp4:
Also, this is just the refactor, right?  Not the "clean up the crazy duplication" we noticed with the array resource

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,411 @@
...
+      end if
+
+      ! If compname is present, find label in some form in config. Else search
+      ! for exact label
+      if (present(compname)) then
+         call get_label_to_use(config, label_to_find, compname, label_is_present, label, _RC)
+      else
+         label = label_to_find
+         call ESMF_ConfigFindLabel(config, label = label, isPresent = label_is_present, _RC)
+      end if
```
tclune, line 107: I think that `get_label_to_use` should make the `compname` argument optional.   This will simplify the higher level logic here by eliminating the conditional block.
```suggestion
      call get_label_to_use(config, label_to_find, compname, label_is_present, label, _RC)
```
You will unfortunately need to reorder the arguments as well, as I would prefer that the optional arguments be after the non-optional.  Not a public interface, se we have that flexibility.   Indeed, probably should throw in a `KeywordEnforcer` argument for good form.

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,411 @@
...
+         return
+      end if
+
+      select type(val)
+      type is(integer(int32))
+         if (default_is_present .and. .not. label_is_present) then
+            select type(default)
+            type is(integer(int32))
+               val = default
+            end select
```
tclune, line 122: There should be a check that val and default have the same type.   Unfortunately the handy intrinsic `SAME_TYPE_AS()` does not quite do what we want (long story of confusion in the standard committee), so you may need a DEFAULT block in each of these.  Ugh.

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,411 @@
...
+         end if
+      type is(integer(int64))
+         if (default_is_present .and. .not. label_is_present) then
+            select type(default)
+            type is(integer(int64))
+               val = default
+            end select
+         else
+            call ESMF_ConfigGetAttribute(config, val, label = label, _RC)
+         end if
```
tclune, line 134: This might be the occasion where a small include file is preferable.   You would use the FPP preprocessor to define the type and then include the logic  which is otherwise identical in each block.

We can justify it as a first step towards when Fortran supports templates.

tclune: Actually, we could try it first with a (very long) FPP macro?
Something like:
```
#define TYPE_BLOCK(T) \
     type is (T) \
     if (default_is_present .and. .not. label_is_present) then \
            select type(default) \
            type is(T) \
               val = default \
            class default
               _FAIL("Type of 'default' does not match type of 'val'.")
            end select \
         else \
            call ESMF_ConfigGetAttribute(config, val, label = label, _RC) \
         end if

File: base/MAPL_Resource.F90 [hunk 4]
```diff
@@ -0,0 +1,411 @@
...
+         call print_resource(printrc, label_to_print, val, default=default, _RC)
+      end if
+
+      _RETURN(ESMF_SUCCESS)
+
+   end subroutine MAPL_GetResource_config_scalar
+
+   ! Find value of array variable in config
+   subroutine MAPL_GetResource_config_array(config, compname, vals, label_to_find, default, rc)
```
tclune, line 192: Why are the arguments in this interface in a different order than the scalar case?    And shouldn't compname be optional here as well?   (May need to consult @atrayano and/or @bena-nasa for history)

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,411 @@
...
+            call ESMF_ConfigGetAttribute(config, valuelist = vals, count = count, label = label_to_use, _RC)
+         end if
+      type is(integer(int64))
+         if (default_is_present .and. .not. label_is_present) then
+            select type(default)
+            type is(integer(int64))
+               vals = default
+            end select
+         else
+            call ESMF_ConfigGetAttribute(config, valuelist = vals, count = count, label = label_to_use, _RC)
```
tclune, line 238: Same comments on this massive block as for the scalar case.

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,411 @@
...
+         if (present(default)) then
+            default_str = intrinsic_to_string(default, type_format)
+         end if
+      type is(integer(int64))
+         type_str = "'Integer*8 '"
+         type_format = '(i0.1)'
+         val_str = intrinsic_to_string(val, type_format)
+         if (present(default)) then
+            default_str = intrinsic_to_string(default, type_format)
+         end if
```
tclune, line 319: The FPP macro here will probably need 3 arguments:  `type, type_str, and type_format`.

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,411 @@
...
+
+      if (vector%size() /= 0) then
+         do while (iter /= vector%end())
+            if (trim(string) == iter%get()) then
+               vector_contains_str = .true.
+               return
+            end if
+            call iter%next()
+         end do
+      end if
```
tclune, line 382:
```suggestion
         do while (iter /= vector%end())
            if (trim(string) == iter%get()) then
               vector_contains_str = .true.
               return
            end if
            call iter%next()
         end do
```
The check on size is not necessary.   An empty vector starts the iterator equal to end.

I'd also like to update this to gFTL2 while we are looking at it.   Should just need to change the USE statement to reference `gftl2` and change `iter%get()` to `iter%of()`

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,411 @@
+#include "MAPL_Exceptions.h"
+#include "MAPL_ErrLog.h"
+#include "unused_dummy.H"
+
+!=============================================================================
+
+
+module MAPL_ResourceMod
```
tclune, line 8: My big question is whether we can come up with a better name for this module and associated procedures.  "Resource" is a bit vague.   We should debate on Thu.

darianboggs: Sounds good. We can discuss what this module should encompass beyond `GetResource`

base/MAPL_Resource.F90
darianboggs: Done

base/MAPL_Resource.F90
darianboggs: Fixed. Component name is required for the array version but not for the scalar version. So `compname` occurred before `default` which is `optional`. Now it occurs after `compname` and both are `optional`. Functionally, `compname` is still required, so I added an assert that `compname` is present. I added `unusable` before `default` for keyword enforcement.

base/MAPL_Resource.F90
darianboggs: Done. Changed `use gFTL_StringVector` to `use gFTL2_StringVector`, eliminated size check, and changed `iter%get()` to `iter%of()`.

base/MAPL_Resource.F90
darianboggs: Done.

base/MAPL_Resource.F90
darianboggs: Done.

base/MAPL_Resource.F90
darianboggs: Done.

base/MAPL_Resource.F90
darianboggs: Done. Separate macro because of slight difference.

base/MAPL_Resource.F90
darianboggs: Per MAPL refactoring meeting: Leaving the name the same for now.

darianboggs:
@tclune Changes to branch made locally. Testing before push up.

darianboggs:
All conversations resolved, bugs fixed. @tclune

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,343 @@
...
+#include "unused_dummy.H"
+
+!=============================================================================
+!FPP macros for repeated (type-dependent) code
+
+#ifdef SET_VAL
+#  undef SET_VAL
+#endif
+
+#define SET_VAL(T) \
```
tclune, line 12: Not strictly necessary, but maybe it would be best to put `val` as an argument to the macro?

I'm also wondering if `GET` might be more indicative than `SET`, but I can see both arguments.   The procedure is "getting" values but the macro is "setting".  But it is also getting from ESMF ...

Give it some thought, but don't spend time on that.

base/MAPL_Resource.F90
```diff
@@ -0,0 +1,343 @@
...
+      end select ;\
+   else ;\
+      call ESMF_ConfigGetAttribute(config, valuelist = vals, count = count, label = actual_label, _RC) ;\
+   end if
+
+#ifdef SET_STRINGS
+#  undef SET_STRINGS
+#endif
+
+#define SET_STRINGS(T, TS, TF) \
```
tclune, line 46: Macros are confusing enough.   Consider making the template arguments a bit move explicit like "T_fmt" and "T_str"?

And maybe `TO_STRING` as the macro name?  Should certainly be singular, not plural if I'm reading this correctly.

base/MAPL_Resource.F90
darianboggs: I will include `val` to make it more funtionish. I'll keep `Set` since the macro ends up setting `val` locally for return.

base/MAPL_Resource.F90
darianboggs: Good idea for the names. The macro actually sets two strings, so I'll leave the name as is.

tclune: Ah, I see.  One solution would be to bring the ultimate `print` statement into the macro.   But we really do want to minimize the amount of logic in our macros.  (This is supposed to be Fortran after all.)

But maybe put both output strings into the macro as with `SET_VAL` to make the action a bit clearer at the point of invocation.

darianboggs:
The UFS build tests were failing. Don't know why, but I copied the GFTL lines from `generic/CMakeLists.txt` to `base/CMakeLists.txt`. After I make a few changes, I'll update the PR later tonight.

tclune:
Could be V1 vs V2 issues.  Library/mod dependencies and such.

 There is a long-open PR to switch MAPL to v2, but @weiyuan-jiang encountered a few obstacles.  Should have at least updated those that could.

base/MAPL_Resource.F90
darianboggs: Done.

base/MAPL_Resource.F90
darianboggs: I agree. The print block is rather ghastly. I updated Matt's branch for the #1453 issue with better logic. I think it's time for me to manually pull that branch in. Originally, I was planning a second PR for Matt's branch, but the refactoring made the two branches incompatible, so I might as well fold Matt's branch in manually. That should clean up the logic.

darianboggs: I'll leave this unresolved until I do that.

darianboggs: Merging will be simpler if this PR is merged into `develop` before the PR for issue #1686. Please approve.

tclune (approved):

PR #1947, opened 2023-02-01
fixes #1941

### LINKED ISSUE #1941: Bit shaving results in non-layout independent History?
The GMAO-OPS @rlucches  team reported that the GEOS-IT system was giving different diagnostic output at different layouts (but the model itself is layout independent as far as the checkpoints) but @lltakacs traced it down to the bit shaving as the culprit.

Indeed, I've been able to reproduce with "main" branch of the GEOSgcm fixture as of 1/20/2023 (that's when model I just ran was cloned and built). At C48, I have collection that is output on the native grid of the fields, the fields are 2D, so there's no processing. If I turn on bit shaving I get different results at 1x12 and 4x48 for the collection at the first write. If I turn off bit-shaving, the collection is layout independent.

I'll investigate; I'm perplexed as the bit shaving should be an element wise operation, just not seeing how the layout could possibly matter, but apparently it does.

bena-nasa:
Fixes #1941 
This implements the bit shaving as a layout independent algorithm when used.
This is obviously non-zero diff for History diagnostic output when using the nbits option in History unless your application is only being run on a single MPI processes but also obviously has no effect on the module checkpoints and MODEL non-zero diff.

I only marked this as non-zero diff because it will change history diagnostic output. I don't know if that makes zero-diff or non-zero diff, i.e. does the non 0-diff label only apply to restarts.

note per @tclune request I also moved the code to Shared from PFIO, hence why there seems to be a lot of code changes here for a simple algorithmic change.

Also note I had to go against the official MAPL style guide and prepend the c routines in shavemantissa with MAPL_ as there is a copy of this c code in little cfio so if I did not do this, we have name collisions

## Types of changes

## Checklist:

### DIFF
--- a/shared/DownBit.F90
+++ b/shared/DownBit.F90
@@ -1,20 +1,23 @@
-module pFIO_DownbitMod
+#include "MAPL_Generic.h"
+module MAPL_DownbitMod
    use, intrinsic :: iso_c_binding, only: c_f_pointer, c_loc, c_ptr
+   use mpi
+   use MAPL_ExceptionHandling
 
    implicit none
    private
 
-   public :: pFIO_DownBit
+   public :: DownBit
 
-   interface pFIO_DownBit
-      module procedure pFIO_DownBit1D
-      module procedure pFIO_DownBit2D
-      module procedure pFIO_DownBit3D
-   end interface pFIO_DownBit
+   interface DownBit
+      module procedure DownBit1D
+      module procedure DownBit2D
+      module procedure DownBit3D
+   end interface DownBit
 
 contains
 
-   subroutine pFIO_DownBit3D ( x, xr, nbits_to_keep, undef, flops, rc )
+   subroutine DownBit3D ( x, xr, nbits_to_keep, undef, flops, mpi_comm, rc )
 
      implicit NONE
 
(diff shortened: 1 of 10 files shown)


shared/DownBit.F90
```diff
@@ -208,8 +223,38 @@ subroutine pFIO_DownBit1D ( x, xr, nbits_to_keep, undef, flops, rc )
...
+      local_max_value = maxval(array,mask=undef_mask)
+      if (mpi_comm /= MPI_COMM_NULL) then
+         call MPI_AllReduce(local_min_value,min_value,1,MPI_FLOAT,MPI_MIN,mpi_comm,status)
+         _VERIFY(status)
+         call MPI_AllReduce(local_max_value,max_value,1,MPI_FLOAT,MPI_MAX,mpi_comm,status)
+         _VERIFY(status)
+      else
+         min_value = local_min_value
+         max_value = local_max_value
+      end if
```
tclune, line 256: Are you certain that `MPI_AllReduce()` does not do the "right thing" with a null communicator?

tclune: Well, I could not find a definite answer, and did see some comments on the internet that suggest NULL does not work in this context.  So closing this.

tclune (approved):

PR #1970, opened 2023-02-13
Handmerge develop into MAPL-v3 2023-Feb-13

mathomp4:
Supersedes #1969 

This is a handmerge of `develop` into `release/MAPL-v3`. Keeping draft until I can test on discover. I *think* it's right, but profiler/logger code moved and so it might not work.


mathomp4:
It works! I'll merge in!

PR #1972, opened 2023-02-14
Fixes #1957

### LINKED ISSUE #1957: Possible ExtData2G issue in v10.25.0 of the GEOSgcm fixture
@sdrabenh found that when doing a non-standard replay setup that results in a 5 hour predictor segment, the model was not passing the 0-inc replay vs straight run test.
I was able to reproduce in v10.25.0 of the GEOSgcm fixture and confirm if I just turn off ExtData the problem goes away.

The interesting thing is that the SAME MAPL version, but a slightly older fixture does not have this problem so it seems hard to believe that it is the ExtData2G code as I can run with the 5 hour predictor and it passes this test in the older fixture.

The other weird thing is that the standard Merra2 type replay with a 3 hour predictor passes this test in v10.25.0

I am trying to get to the bottom of this.

The other suspicious thing is that go-cart has all these 3 hour hard coded frequencies in it, which have not been fixed despite multiple pleas, so I am still leary that there is not something with that going on...

bena-nasa:
This fixes issue #1957 
In addition I have done two things.
I added two new test cases to the ExtData/History test framework that "simulate" replay.
I needed to do some work on the ExtDataDriver.x application to do all the tests I wanted.
Finally I have not looked at this in a long time and while I was monkeying with it, I gave the grid comp associated with this application a major cleanup.
I removed a lot of pointless comments, probably because I coped this from some other grid comp long ago, changed all the _VERIFY macros to the _RC, and cleaned up a bunch of unused variables so I would not have to look at them every time I compiled.

## Types of changes

## Checklist:


tclune (approved):
Quieter builds appreciated!

mathomp4:
I made a push to changelog and cmakelists so that I can make a release.

mathomp4 (approved):

PR #1984, opened 2023-02-23
Add new routine to spherical geometry package

bena-nasa:
…package (needed for a utility I created related to my topography generation workflow)

## Motivation and Context

Just a little PR to add a routine to our spherical geometry package so I can have my cube-sphere grid scrip file generation utility not need fv3 linkage, I can get the cubed-sphere grid from the ESMF call, but ESMF does not have a method to get the area of the grid cells (at least not without making a Regridding handle), so I just added this, just the fv3 implementation.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


atrayano (approved):
Look good

PR #2002, opened 2023-03-06
Fix/wjiang/set grid fix

weiyuan-jiang:
When set_grid is call, the cf_root is not created. So the GridType should come from cap.
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


weiyuan-jiang:
This fix is necessary for issue https://github.com/GEOS-ESM/GOCART/issues/207

mathomp4:
I might need some @tclune thoughts on this. This might be better as a hotfix than a feature if you are saying the previous code wasn't right.

tclune (changes requested):
Yes - make this a hotfix.

weiyuan-jiang:
closed. Replaced by #2003

Issue #2019, opened 2023-03-16
Lambert conformal grid output needed for stretched-grid initiative


bena-nasa:
We need to output on this lamber grid. We have the coordinates, so the easiest way is to make a new grid factory that an make an arbitrary non-periodic ESMF Grid given a file with 2D Lons and lats.

PR #2020, opened 2023-03-16
Feature/bmauer/general xy grid

bena-nasa:
This addresses issue #2019 
Note this overlaps I think with the swath grid work @metdyn is doing, but @wmputman needs this ASAP and I needed to be sure we could generate this Lambert Conformal grid output needed for the stretched grid runs he must make. None of the existing ones were adequate and we needed to know now if we could get the output needed for this run, so I no choice but to do this. There was simply no grid factory where I could give it a set of 2D Lons and 2D lats and construct a non-periodic ESMF grid.

This is a grid factory to make non-periodic ESMF grid, given an input file that has the Lons and Lats as 2D arrays. Now Bill can output on this Lambert Conformal grid as I have the coordinates for that grid in a file can be used in the construction of this factory.

To create this, I basically took the Tripolar factory, changed the grid definition from 1d periodic/monopole to a simple non-periodic grid and made adding the corner stagger location optional based on whether they were in the input file to begin with. The factory must be provided a file that contains at least the grid cell centers.

## Types of changes

## Checklist:

- X] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [X ] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Implemented new generic XY grid factory to create regional grids on any input set of 2D lons and lats
+
 ### Changed
 
 ### Fixed
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -54,6 +54,7 @@ set (srcs
   MAPL_ISO8601_DateTime_ESMF.F90
   FieldUtilities.F90
   MAPL_Resource.F90
+  MAPL_XYGridFactory.F90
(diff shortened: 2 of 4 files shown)


bena-nasa:
I just realized there is an incompleteness in this code. In append meta-data, I forgot to put in an if has corners define xcdim and ycdim. Note those are commented out but should be restored with an if this%has_corners block. None of this matters for what Bill needs since I don’t have corners for this grid he needs but just looking here on my iPad and noticed

tclune (approved):

PR #2022, opened 2023-03-16
Hotfix/bmauer/fixes #2018

### LINKED ISSUE #2018: another climatology bug found in extdata2g
user @sdrabenh found a bug with extdata2g and the new wsub inputs which are a climatology. Testing indeed reproduced this bug which was causing it to use the wrong time when search the file when wrapping around to the previous or next year.

bena-nasa:
fixes #2018 
This fixes another bug found with climatologies. I also add a new option to the ExtDataDriver.x testing framework to be able to produce a variable that is the day relative to the start of the year, rather than an absolute reference time which I needed to confirm that the climatology was doing what it should.

I have not added a new test case yet as I'm not sure how to implement a case to test this.

I confirmed it was working by essentially using ExtDataDriver.x to create a climatology set for a year where I put the day of the year relative to the start of the year in the files.

I could then look, that indeed when the files were read and time interpolated into the ExtDataDriver.x component as a climatology, they indeed had the right values over a several year run of ExtDataDriver.x

Tried with both a climatology representing a leap year and non-leap year for both a single file climatology, a monthly file climatology and a climatology that has a daily file.

I'll try to think of a way to update the tests, but want this fix

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ parameters:
 # Anchors to prevent forgetting to update a version
 os_version: &os_version ubuntu20
 baselibs_version: &baselibs_version v7.7.0
-bcs_version: &bcs_version v10.23.0
+bcs_version: &bcs_version v11.00.0
 tag_build_arg_name: &tag_build_arg_name maplversion
 
 orbs:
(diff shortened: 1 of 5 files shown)


mathomp4 (approved):
All my testing shows zero-diff (as I guess I don't trigger the bug).

PR #2040, opened 2023-03-29
Feature/bmauer/cleanup cpl alarms

### LINKED ISSUE #2039: Allow accumulation to be not reset by writing to file in History
A suer @sdrabenh was asking if History could perform accumulation, I.E. do the accumulation for the time average but not divide the count. The motivation for this is that the stretched hazardous weather experiments effort needs accumulated precipitation (basically precipitation from 0-1z, 0-2z, etc...  for each day) as one of the products we must submit. I looked at it and indeed that in addition to min/max, a straight accumulation was allowed by the coupler.

However, there was a one problem. The coupler essentially has two alarms and receiver two frequencies. The history collection has two keywords "frequency" which controls how often the collection writes, and one I was unaware of "acc_interval" which defaults to the frequency and does not appear to be used, but does exist.

Both frequencies are passed to the coupler, and two alarms are made. One is a "time to couple" alarm which controls basically when the coupler puts the result in the output. The 2nd is "time to clear" which clears the accumulators and counts. Currently both alarms are set to the frequency of "frequency", passed to the coupler, even though it gets a 2nd frequency which is the acc_interval from history.

Essentially for all practical purposes the way we run history, the "time to couple" and "time to clear" run on the same alarm. So when you write, you clear, and if you wanted something like hourly accumulated precipitation and that reset once a day, you are out of luck.

The 2nd thing I discover is that History has this notion of a reference time for the collection which defaults to 0z, so the "frequency" of writing is relative to this ref time, but this ref time is never passed to the coupler, so if you do time average, unless you set the ref_time to the start time you intend to run, the time-averaging behavior is not right. I will make a separate issue for that.

However, the coupler looks like it was designed to do this, but almost but not quite there. I made some minor tweaks and I think I've fixed it so @sdrabenh could do what he wanted. 
The changes that have to take place are:
1. the "time to zero" alarm in the coupler should be set to the acc_interval passed from history
2. some sort of reference time for this alarm needs to be passed. Since you can't pass this I highjacked the offset of the varspec which is not passed via history. I added an acc_ref_tinme, then I compute an offset so that the time-to-zero alarm can be set to ring at the time the user wants which may not be the start time which the time to couple alarm is hardwired to.
3. I removed 2 bits of logic the coupler, one was that when the couple is called it stops accumulation for that item even if you don't zero, but @atrayano could not think of a use case for that. 2nd for the time to zero alarm I removed the restriction that is has to start at or later than the model start time.

bena-nasa:
fixes #2039 

Note that if the History.rc files are not changed, for all practical purposes these modifications do not change behaviors and are zero-diff in the testing I was able to do. This enables new functionality that must be enabled.

## Types of changes

## Checklist:

### DIFF
--- a/generic/GenericCplComp.F90
+++ b/generic/GenericCplComp.F90
@@ -442,11 +442,12 @@ contains
 
           if (TCLR < TS) TCLR = TS
 
-          rTime = TM0 + TOFF - TCLR
+          !rTime = TM0 + TOFF - TCLR
+          rTime = TM0 + TOFF
 
-          do while (rTime < currTime) 
-             rTime = rTime + TCPL
-          end do
+          !do while (rTime < currTime) 
+             !rTime = rTime + TCPL
+          !end do
 
           STATE%TIME_TO_CLEAR(J) = ESMF_AlarmCreate(NAME='TIME2CLEAR_' // trim(COMP_NAME) &
                // '_' // trim(NAME),   &
--- a/generic/GenericCplComp.F90
+++ b/generic/GenericCplComp.F90
@@ -451,7 +451,7 @@ contains
           STATE%TIME_TO_CLEAR(J) = ESMF_AlarmCreate(NAME='TIME2CLEAR_' // trim(COMP_NAME) &
                // '_' // trim(NAME),   &
                clock        = CLOCK,   &
-               ringInterval = TCPL,    & 
+               ringInterval = TCLR,    & 
                ringTime     = rTime,   &
                sticky       = .false., &
                rc=STATUS   )
@@ -1208,7 +1208,7 @@ contains
 
           end select
 
-          STATE%ACCUM_COUNT(J) = -1
+          !STATE%ACCUM_COUNT(J) = -1
 
        end if
 
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -765,6 +765,7 @@ contains
        list(n)%regex = (useRegex /= 0)
        call ESMF_ConfigGetAttribute ( cfg, list(n)%frequency, default=060000, &
                                       label=trim(string) // 'frequency:',_RC )
+
        call ESMF_ConfigGetAttribute ( cfg, list(n)%acc_interval, default=list(n)%frequency, &
                                       label=trim(string) // 'acc_interval:',_RC )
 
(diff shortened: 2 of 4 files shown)


generic/GenericCplComp.F90
```diff
@@ -1208,7 +1209,7 @@ subroutine COUPLE(SRC, STATE, RC)
 
           end select
 
-          STATE%ACCUM_COUNT(J) = -1
+          !STATE%ACCUM_COUNT(J) = -1
```
atrayano, line 1212:
```suggestion
          STATE%ACCUM_COUNT(J) = 0
```

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -5070,13 +5083,41 @@ subroutine CopyStateItems(src, dst, rc)
...
+     call ESMF_TimeGet(current_time,yy=year,mm=month,dd=day,h=hour,m=minute,s=second,_RC)
+     call MAPL_UnpackTime(ref_time,hour,minute,second)
+     call ESMF_TimeSet(new_time,yy=year,mm=month,dd=day,h=hour,m=minute,s=second,_RC)
+     t_int = new_time - current_time
+
+     call ESMF_TimeIntervalGet(t_int,s=diff_sec,_RC)
+     if (diff_sec == 0) then
+        acc_offset = 0
+     else if (diff_sec > 0) then
+        acc_offset = diff_sec - 86400
```
atrayano, line 5114: Why we subtract only a single day?

atrayano (changes requested):

gridcomps/History/MAPL_HistoryGridComp.F90
bena-nasa: So the logic here is I have this acc_ref_time. I want to make an alarm that rights at that time in the future starting from the model start time.

Either is it 
1. the same as the model start time
2. in the future relative to the current time, so the alarm should ring that day
3. in the past, so the alarm should not ring until the next day.

Since all I can pass to the coupler this offset, for case 2, I actually need first ring time to be the PREVIOUS day, otherwise, the alarm won't ring during the first day. So that is what that is doing. A more elegant solution would be to pass the ref_times to the coupler varspec does not allow that now.

generic/GenericCplComp.F90
bena-nasa: @atrayano is setting to zero, really the right behavior there. Shouldn't that be the providence of the "time_to_zero" alarm. All the "time_to_couple" says is take whatever I have and fill the destination arrays if I understand right, why reset anything at all? Shouldn't that be responsibility of the time to zero alarm?

atrayano: My mistake, please ignore the suggested change. I was thinking this is happening during initialization

gridcomps/History/MAPL_HistoryGridComp.F90
atrayano: My concern was if somehow the offset is more that 1 day. Probably not likely/impossible to occur

atrayano (approved):

PR #2059, opened 2023-04-06
Component Testing Framework

nataliepatten:
## Description

This PR creates a new feature to test individual components as code is refactored. This testing framework consists of a driver, utility python script, and a run script. The latter gets a subset of the grid if requested, calls the driver, and calculates differences between states captured from GEOS and the driver's results. Changes were also made within current MAPL files to allow for capturing of states and restoring export variables.

## Motivation and Context

This testing framework allows for data within fields in individual components to be verified as code is refactored and/or changed.

## How Has This Been Tested?

This was tested with Sea Salt, Nitrate, Dust, and Gravity Wave Drag components at different phases, with and without subsetting the restart states.

## Types of changes

## Checklist:

### DIFF
--- a/Apps/CMakeLists.txt
+++ b/Apps/CMakeLists.txt
@@ -30,3 +30,11 @@ target_include_directories (time_ave_util.x PRIVATE $<BUILD_INTERFACE:${MAPL_SOU
 if (NOT CMAKE_Fortran_COMPILER_ID MATCHES "NAG")
   target_link_libraries(time_ave_util.x PRIVATE OpenMP::OpenMP_Fortran)
 endif ()
+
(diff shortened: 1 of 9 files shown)


nataliepatten:
The testing framework seems to be working fine after rebasing. I just tested it quickly with Nitrates.

Apps/COMP.rc
```diff
@@ -0,0 +1,57 @@
+COMPONENT_TO_RECORD: NI
```
tclune, line 1: No need to change this now, but I would like to find a better name in the long run.

tclune (dismissed):

mathomp4 (approved):
I have tested this in an AMIP run and nothing broke...

PR #2070, opened 2023-04-13
Add "geom" framework

darianboggs:
Add the `geom` framework (and directory).
## Description

The framework will be middleware for **MAPL3**. Also, it will a MAPL geometric type.

## How Has This Been Tested?

This is new functionality. It has been tested internally, though some tests are disabled because some of the framework has not been implemented.
## Types of changes

## Checklist:

### DIFF
--- a/geom/FieldBLAS_IntrinsicFunctions.F90
+++ b/geom/FieldBLAS_IntrinsicFunctions.F90
@@ -1,0 +1,60 @@
+module FieldBLASIntrinicFunctions
+
+   implicit none
+
+   public :: IntrinsicReal64Function
+   public :: IntrinsicReal64BiFunction
+   public :: Sin
+
+   private 
+
+   abstract interface
+      function IntrinsicReal64Function(x) result(fx)
+         real(real64), intent(in) :: x
+         real(real64) :: fx
+      end function IntrinsicReal64Function
+
+      function IntrinsicReal64BiFunction(x, y) result(fx)
+         real(real64), intent(in) :: x, y
+         real(real64) :: fx
+      end function IntrinsicReal64BiFunction
+   end abstract interface
+
+contains
+
+   subroutine intrinsic_real64_func(func, f, func_f, rc)
+      procedure(IntrinsicReal64Function), pointer, intent(in) :: func
+      type(ESMF_Field), intent(inout) :: f
+      type(ESMF_Field), intent(out) :: func_f
+      integer, optional, intent(out) :: rc
+      
+      ! Apply func to f.
+      ! Set rc based on errors from func. 
+      ! Probably a lot of generic framework to apply func to field and set rc.
+   end subroutine intrinsic_real64_func
+
+   subroutine intrinsic_real64_bifunc(bifunc, f, bifunc_f, rc)
+      procedure(IntrinsicReal64BiFunction), pointer, intent(in) :: bifunc
+      type(ESMF_Field), intent(inout) :: f1, f2
+      type(ESMF_Field), intent(out)  :: bifunc_f
+      integer, optional, intent(out) :: rc
+      integer :: status
+      
+      ! Apply bifunc to f.
+      ! Set rc based on errors from bifunc. 
+      ! Probably a lot of generic framework to apply bifunc to field and set rc.
+   end subroutine intrinsic_real64_bifunc 
+
+   function Sin(f, rc) result(sinf)
+      type(ESMF_Field), intent(inout) :: f
+      integer, optional, intent(in) :: rc
+      type(ESMF_Field), intent(out) :: sinf
+      procedure(IntrinsicReal64Function), pointer :: func 
+      integer :: status
+
+      func => dsin ! Is this right?
+      call intrinsic_real64_func(func, f, func_f=sinf, _RC)
+      _RETURN(_SUCCESS)
+   end function sin_field
+
+end module FieldBLASIntrinicFunctions
--- a/geom/tests/Test_FieldBLAS.pf
+++ b/geom/tests/Test_FieldBLAS.pf
@@ -1,0 +1,962 @@
+#include "MAPL_Generic.h"
+
+! MACROS FOR TROUBLESHOOTING
+
+#ifdef SUCCESSFUL 
+#  undef SUCCESSFUL 
+#endif
+
+#define SUCCESSFUL 0
+
+#ifdef SUCCESS
+#  undef SUCCESS
+#endif
+
+#define SUCCESS(RC) (RC == SUCCESSFUL)
+
+#ifdef MSG
+#  undef MSG
+#endif
+
+#define MSG(DETAIL, STATUS) DETAIL // ': ' // rc_to_string(STATUS)
+
+#ifdef SET_RC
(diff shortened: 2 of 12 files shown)


geom/FieldBLAS_IntrinsicFunctions.F90
```diff
@@ -0,0 +1,60 @@
+module FieldBLASIntrinicFunctions
```
tclune, line 1: This file is not being compiled currently.  Just wanted to be sure @darianboggs is aware.

geom/tests/Test_FieldBLAS.pf
```diff
@@ -0,0 +1,926 @@
...
+      
+      call ESMF_FieldGet(x, farrayPtr = x_ptr, rc = status)
+      _ASSERT(SUCCESS(status), MSG('Failed to get source data pointer', status))
+
+      call ESMF_FieldGet(y, farrayPtr = y_ptr, rc = status)
+      _ASSERT(SUCCESS(status), MSG('Failed to get destination data pointer', status))
+
+      @assertEqual(x_ptr, y_ptr)
+
+      PASSED(TEST_NAME)
```
tclune, line 274: Note - the print statements are not necessary.    If you run the tests with `-d` it emits a message as each test starts and stops.      I'll approve with it as is, but it is definitely something to be fixed.  Successful tests should be silent.

geom/tests/Test_FieldBLAS.pf
```diff
@@ -0,0 +1,926 @@
...
+      _ASSERT(SUCCESS(status), MSG('Call failed', status))
+      @assertTrue(are_same_typekind)
+      _ASSERT(SUCCESS(status), MSG('Call failed', status))
+
+      y = YR4
+      are_same_typekind = .TRUE.
+      are_same_typekind = FieldsAreSameTypeKind(x, y, rc = status)
+      _ASSERT(SUCCESS(status), MSG('Call failed', status))
+      @assertFalse(are_same_typekind)
+      _ASSERT(SUCCESS(status), MSG('Call failed', status))
```
tclune, line 634: Recently I've taken to introducing an `_RC` macro for the unit tests to avoid the tedious checking during unit tests and focus on the "meat".  Requires making an explicit pFUnit call rather than using the `@assert` mechanism.

OTOH, your approach of relying on MAPL's own error handling suggests that maybe even the regular macro would work this way.

tclune (dismissed):
I had several comments, but so long as this compiles and runs, I want to get it into develop so that I can start exercising and extending the layer.

geom/tests/Test_FieldBLAS.pf
```diff
@@ -0,0 +1,926 @@
...
+   ! Basic test of FieldCOPY subroutine (REAL32 -> REAL64)
+   subroutine test_FieldCOPY_R4R8()
+      character(len=*), parameter :: TEST_NAME = 'test_FieldCOPY_R4R8'
+      type(ESMF_Field) :: x
+      type(ESMF_Field) :: y
+      real(kind=ESMF_KIND_R4), dimension(:,:), pointer :: x_ptr
+      real(kind=ESMF_KIND_R8), dimension(:,:), pointer :: y_ptr
+      integer :: status, rc
+      
+      RUNNING(TEST_NAME)
```
tclune, line 288: Note that pFUnit knows the name of its tests, so you don't need to define `TEST_NAME` in each test.

geom/FieldBLAS_IntrinsicFunctions.F90
darianboggs: Correct. Not in `CMakeLists`. I'll leave the extra files for now.

geom/tests/Test_FieldBLAS.pf
darianboggs: Removed. All messages come through _ASSERT and @assert now.

geom/tests/Test_FieldBLAS.pf
darianboggs: Removed.

geom/tests/Test_FieldBLAS.pf
```diff
@@ -0,0 +1,926 @@
...
+      _ASSERT(SUCCESS(status), MSG('Call failed', status))
+
+      y = YR4
+      are_same_typekind = .TRUE.
+      are_same_typekind = FieldsAreSameTypeKind(x, y, rc = status)
+      _ASSERT(SUCCESS(status), MSG('Call failed', status))
+      @assertFalse(are_same_typekind)
+      _ASSERT(SUCCESS(status), MSG('Call failed', status))
+
+      are_same_typekind = .TRUE.
```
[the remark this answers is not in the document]
darianboggs: I eliminated the `_ASSERT(SUCCESS`... statements. The `are_same_typekind` are necessary for functions with rc codes and the `_RC` macro. I'm testing after removing unnecessary code including `_ASSERT` macro calls, `TEST_NAME` constants. Will post updates to branch when I am finished with tests.

darianboggs:
There are some tests that I am working on. Will post updates to branch as I finish the tests.

darianboggs:
Disabled the `@assertEqual` in test_FieldAXPY_R8 temporarily.

darianboggs:
**Intel** build/test works. **GNU** build/tests fails on `test_FieldCOPY_R8R4`. The conversion from _REAL64_ to _REAL32_ generates a difference below threshold for **Intel**, but the difference is above threshold for **GNU**.

The test formula is:

$\frac{\textrm{abs}(x-y)}{\textrm{abs}(y)} < \textrm{epsilon}(\textrm{real}(1.0, \textrm{kind}=\verb|ESMF_KIND_R4|))$

$x$ is a _REAL64_ `ESMF_Field` and $y$ is a _REAL32_ `ESMF_Field`. $x$ is copied to $y$, which should cause a loss of precision, but the loss of precision should be less be less than the precision of the _REAL32_ `ESMF_Field`.

I'll look at this over the weekend.

tclune:
@darianboggs   The test does not need to be so strict.   The failure you sight is not a failure in your implementation but in how the individual compiler is doing its arithmetic, which is beyond the scope of a unit test.   If I see a simple relaxation, I'll make the change and approve.

tclune:
OK - I'm flumoxed.   The values in the test look like they should be integers.  Roundoff should not be happening and you should be getting exact equality.

Side note.   pFUnit has tolerance options (absolute and relative) for `assertEqual`.     (This is part of why I encouraged you to work with me once you had sketched out a few tests.    You've done some good work, but ... could have saved yourself some difficulty.)

mathomp4:
Note: I updated the `CHANGELOG.md` in this branch because it looked like a merge got a bit...wonky.

tclune (dismissed):

mathomp4:
Also: Lots of the modules have `mapl3g` in them but the PR is to `develop`. Is that right?

tclune:
Only one module is activated. The others should not have been added, but were due to insufficient instructions on my part.
Not harming anything for now though.

mathomp4:
GNU seems to have failed one of the new unit tests.

PR #2073, opened 2023-04-13
Fixes #2072

### LINKED ISSUE #2072: Gridded IO needs to output files with no explicit time
For the swath sampling GriddedIO needs to output variables that do not depend on time. I'm just making an issue as I'm going to make a PR to enable the and want to reference the issue.

bena-nasa:
Fixes #2072

This adds an option to the MAPL_TimeUtilities class with a logical to know if the instantiation has been initialized via the constructor. If not, this can be detected by a layer like griddedIO and griddedIO can not add time as a dimension to variables for the swath sampler use case.
I've modified GriddedIO to act in this manner. If the TimeInfo object which is required to passed was not initialized, then it is simply ignored.
@metdyn can then use this for his swath sampling stuff and when he writes those files, don't pass an initialized time object that that will be that. No time as a dimension in the file and no variables can therefore depend on it as well.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add option to GriddedIO class so that if an unitilized time object is passed, the resulting file will not depend on time
+
 ### Fixed
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Add option to GriddedIO class so that if an unitilized time object is passed, the resulting file will not depend on time
+- Add logic to GriddedIO class so that if an unitialized time object is passed, the resulting file will not depend on time
 
 ### Fixed
 
--- a/base/MAPL_TimeMethods.F90
+++ b/base/MAPL_TimeMethods.F90
@@ -246,10 +246,8 @@ contains
           tdim = this%ntime
        end if
        call metadata%add_dimension('time',tdim)
-       v = this%define_time_variable(rc=status)
-       _VERIFY(status)
-       call metadata%add_variable('time',v,rc=status)
-       _VERIFY(status)
+       v = this%define_time_variable(_RC)
+       call metadata%add_variable('time',v,_RC)
     end if
     _RETURN(_SUCCESS)
 
--- a/base/MAPL_TimeMethods.F90
+++ b/base/MAPL_TimeMethods.F90
@@ -20,6 +20,7 @@ module MAPL_TimeDataMod
      type(ESMF_TimeInterval) :: offset
      character(len=:), allocatable :: funits
      logical :: integer_time
+     logical :: is_initialized = .false.
    contains
      procedure :: add_time_to_metadata
      procedure :: define_time_variable
@@ -27,6 +28,7 @@ module MAPL_TimeDataMod
      procedure :: get_start_time
      procedure :: get
      procedure :: setFrequency
+     procedure :: am_i_initialized
   end type timeData
 
   interface timeData
@@ -34,6 +36,12 @@ module MAPL_TimeDataMod
   end interface timeData
 contains
 
+  function am_i_initialized(this) result(logical_temp)
+     logical  :: logical_temp
+     class(TimeData), intent(inout) :: this
+     logical_temp = this%is_initialized
+  end
+     
   function new_time_data(clock,ntime,frequency,offset,funits,integer_time,rc) result(tData)
     type(timeData) :: tData
     type(ESMF_Clock),intent(inout) :: clock
@@ -44,6 +52,7 @@ contains
     logical, optional, intent(in) :: integer_time
     integer, optional, intent(Out) :: rc
 
+    tData%is_initialized = .true.
     tdata%clock=clock
     tdata%ntime=ntime
     tdata%frequency=frequency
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add option to GriddedIO class so that if an unitilized time object is passed, the resulting file will not depend on time
+
```
tclune, line 19: I did not see any changes to interfaces, so I suspect that this is not an "option" per se, but rather a change to the logic to improve treatment of uninitialized time objects.    Please clarify the text.

base/MAPL_TimeMethods.F90
```diff
@@ -44,6 +52,7 @@ function new_time_data(clock,ntime,frequency,offset,funits,integer_time,rc) resu
     logical, optional, intent(in) :: integer_time
     integer, optional, intent(Out) :: rc
 
+    tData%is_initialized = .true.
```
tclune, line 55: Is there a path where one does not go through this constructor?  Otherwise it looks like `is_initialized` is always `.true.` despite having a different default value in the constructor.  I'm probably missing something basic here.

CHANGELOG.md
bena-nasa: @tclune, yes small change in logic to detect an uninitialized time object. That way @metdyn for his swath work, can just pass an uninitialized time object into the class, and that's fine, it will not be used, and griddedIO will produce files where there is no time dimension and the variables do not depend on time.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add option to GriddedIO class so that if an unitilized time object is passed, the resulting file will not depend on time
```
tclune, line 18:
```suggestion
- Add logic to GriddedIO class so that if an unitialized time object is passed, the resulting file will not depend on time
```

tclune (approved):

CHANGELOG.md
metdyn: @bena-nasa  Thanks for the un_initialized time option! I am not there yet, but look forward to use this griddedIO and this functionality.

metdyn: @bena-nasa  I think you should merge this PR. I will test it next week. Thanks!

griddedio/GriddedIO.F90
```diff
@@ -443,15 +461,22 @@ subroutine bundlepost(this,filename,oClients,rc)
...
 
-        tindex = size(this%times)
-        if (tindex==1) then
-           call this%stage2DLatLon(filename,oClients=oClients,_RC)
+           tindex = size(this%times)
+           if (tindex==1) then
+              call this%stage2DLatLon(filename,oClients=oClients,_RC)
+           end if
+        else
+           tindex = -1
```
metdyn, line 479: I see this tindex=-1 is an addition in bundlepost. I will give it try in swath to double check no effect on my code.

griddedio/GriddedIO.F90
```diff
@@ -403,14 +413,18 @@ subroutine modifyTime(this, oClients, rc)
...
-        call this%metadata%modify_variable('time',v,rc=status)
-        _VERIFY(status)
-        if (present(oClients)) then
-           call var_map%insert('time',v)
-           call oClients%modify_metadata(this%write_collection_id, var_map=var_map, rc=status)
+        if (this%timeInfo%is_initialized) then
+           v = this%timeInfo%define_time_variable(rc=status)
+           _VERIFY(status)
+           call this%metadata%modify_variable('time',v,rc=status)
            _VERIFY(status)
```
metdyn, line 420: I have some thoughts on the naming:  timeInfo%define_time_variable  from MAPL_TimeMethods.F90.   This subroutine shows:
```
subroutine modifyTime(this, oClients, rc)
           v = this%timeInfo%define_time_variable(rc=status)
           call this%metadata%modify_variable('time',v,rc=status)
```
Initially I thought  function define_time_variable  should appear only once. 
Then I see  the `define_time_variable`   in MAPL_TimeMethods.F90 did the following:
 - defines the time variable like long_name
 - read the clock to get the current time value to define begin_date, time_increment

I wonder if this should be called:  extract_time_variable_attribute_from_clock

CHANGELOG.md
metdyn: I tested this code by merging to my local branch and find no effect on existing output.

metdyn (dismissed):
Tested with current swath code, showing zero difference

base/MAPL_TimeMethods.F90
```diff
@@ -230,16 +239,19 @@ subroutine add_time_to_metadata(this,metadata,rc)
...
+       if (this%ntime==-1) then
+          tdim = pFIO_UNLIMITED
+       else
+          tdim = this%ntime
+       end if
+       call metadata%add_dimension('time',tdim)
+       v = this%define_time_variable(rc=status)
+       _VERIFY(status)
+       call metadata%add_variable('time',v,rc=status)
+       _VERIFY(status)
```
tclune, line 252:
```suggestion
       v = this%define_time_variable(_RC)
       call metadata%add_variable('time',v,_RC)
```

griddedio/GriddedIO.F90
```diff
@@ -410,14 +420,18 @@ subroutine modifyTime(this, oClients, rc)
...
-        v = this%timeInfo%define_time_variable(rc=status)
-        _VERIFY(status)
-        call this%metadata%modify_variable('time',v,rc=status)
-        _VERIFY(status)
-        if (present(oClients)) then
-           call var_map%insert('time',v)
-           call oClients%modify_metadata(this%write_collection_id, var_map=var_map, rc=status)
+        if (this%timeInfo%is_initialized) then
+           v = this%timeInfo%define_time_variable(rc=status)
+           _VERIFY(status)
```
tclune, line 425:
```suggestion
        if (this%timeInfo%is_initialized) then
           v = this%timeInfo%define_time_variable(_RC)
```

tclune (approved):

mathomp4 (approved):

tclune (approved):

tclune (approved):

Issue #2069, opened 2023-04-12
Complaint to subroutine MAPL_GetGlobalHorzIJIndex


metdyn:
@weiyuan-jiang:

I have a complaint on 
 module subroutine MAPL_GetGlobalHorzIJIndex
https://github.com/GEOS-ESM/MAPL/blob/main/base/Base/Base_Base_implementation.F90#L3173-L3253

The error is from my recent code addition:

    MAPL_HistoryGridComp.F90
     subroutine Run
        call list(n)%td_mask%get_mask(time_span,grid_in,_RC)  !  td_mask is an type obj. from TimeDependentMask defined below.

      MAPL_HistoryTdMask.F90
        module MAPL_TimeDependentMaskMod
            type :: TimeDependentMask
             contains
                 procedure :: get_mask

                 subroutine get_mask(this, time_span, grid, rc)
                    call  MAPL_GetGlobalHorzIJIndex(nx,II,JJ,lon=obs_lons,lat=obs_lats,grid=Grid,_RC)   !! code hangs
                   !! call MAPL_GetHorzIJIndex(nx,II,JJ,lon=obs_lons,lat=obs_lats,grid=grid,_RC)             !! not code hang
      
If I use  MAPL_GetHorzIJIndex,  there is no error.
If I use  MAPL_GetGlobalHorzIJIndex,  after a few  model time steps and subsequent history calls,  the code hangs forever. 

I suspect:
   -   elemental subroutine calculate(x, y, z, i, j)  is fancy and I see arrays not deallocated.  
  
I did the following tests:

    1. allocate(x(npts), y(npts), z(npts))
        deallocate (lons, lats, xyz, max_abs, x, y, z)
        it did not solve my specific problem. 

    2. inside my subroutine get_mask 
       if I make a do loop call
          do k=1, 2
              call  MAPL_GetGlobalHorzIJIndex(nx,II,JJ,lon=obs_lons,lat=obs_lats,grid=Grid,_RC)
          enddo 
       The code does not hang.  There is no problem.   This surprised me.  I donot know how to explain.

Additional information:   my input nx = 54810  e.g. 

It appears that MAPL_GetHorzIJIndex does not have element subroutines. So I think deallocate and element subroutines are the issues here. 

Will you please take a look?

Thanks!

weiyuan-jiang:
What is "k" 's role in original call ?

metdyn:
The "k" does not belong to the original code. 

[...]

PR #2079, opened 2023-04-14
Pulled call MAPL_GridGetCorners out of the condition to avoid hang in…

weiyuan-jiang:
… MAPL_GetGlobalHorzIJIndex

Pull MAPL_GridGetCorners out of conditional block to avoid hang
## Types of changes

## Checklist:

### DIFF
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -3307,17 +3307,16 @@ contains
        call MAPL_GridGetInterior(grid,I1,I2,J1,J2)
        OK = .true. 
        ! check the edge of face 1 along longitude
+       allocate(corner_lons(I2-I1+2, J2-J1+2))
+       allocate(corner_lats(I2-I1+2, J2-J1+2))
+       call MAPL_GridGetCorners(Grid,corner_lons,corner_lats)
        if ( I1 ==1 .and. J2<=IM_WORLD ) then
-          allocate(corner_lons(I2-I1+2, J2-J1+2))
-          allocate(corner_lats(I2-I1+2, J2-J1+2))
-          call MAPL_GridGetCorners(Grid,corner_lons,corner_lats)
           if (J1 == 1) then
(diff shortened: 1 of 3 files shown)


weiyuan-jiang:
This PR fixes issue #2069

metdyn (approved):
I tested with the swath code and the code hang problem is removed.

mathomp4:
I've pulled in the latest from `main` and have updated the CHANGELOG and CMakeLists for a release once I test this.

base/Base/Base_Base_implementation.F90
```diff
@@ -3307,17 +3307,16 @@ function grid_is_ok(grid) result(OK)
        call MAPL_GridGetInterior(grid,I1,I2,J1,J2)
        OK = .true. 
        ! check the edge of face 1 along longitude
+       allocate(corner_lons(I2-I1+2, J2-J1+2))
+       allocate(corner_lats(I2-I1+2, J2-J1+2))
+       call MAPL_GridGetCorners(Grid,corner_lons,corner_lats)
```
metdyn, line 3312: Having the call MAPL_GridGetCorners outside the if statement solves the run time error when calling  the subroutine MAPL_GetGlobalHorzIJIndex.   Would we still enforce the deallocate statement here?

mathomp4: I asked @tclune about this and I think since Fortran automatically deallocates when you leave a procedure, it's not necessary.

tclune: The only time an explicit deallocate is required for a _local_ ALLOCATABLE variable is when an explicit allocation happens multiple times.  E.g., inside a loop.

Sometimes you may do an explicit deallocate to restore memory early if you are doing something where large memory use is a concern.

Usually, you just let the compiler do it when the variable goes out of scope at the end of the procedure.

tclune (approved):

PR #2087, opened 2023-04-20
print files and writing time in MultiGroupServer

weiyuan-jiang:
The files and its writing time are printed out
## Types of changes

## Checklist:

### DIFF
--- a/pfio/MultiGroupServer.F90
+++ b/pfio/MultiGroupServer.F90
@@ -611,9 +611,13 @@ contains
        type (c_ptr) :: address
        type (ForwardDataAndMessage), target :: f_d_m
        type (FileMetaData) :: fmd
-
+       type(AdvancedMeter) :: file_timer
+       real(kind=REAL64) :: time
+       character(len=:), allocatable :: filename
+ 
        back_local_rank = this%rank
        thread_ptr => this%threads%at(1)
+       file_timer = AdvancedMeter(MpiTimerGauge())
        do while (.true.)
 
          ! 1) get collection id from captain
@@ -799,10 +803,17 @@ contains
             end select
             select type (q=>msg)
             class is (AbstractDataMessage)
+               filename =q%file_name
+               call file_timer%start()
                call thread_ptr%put_dataToFile(q, address, _RC)
+               call file_timer%stop()
             end select
             call msg_iter%next()
          enddo
+         time = file_timer%get_total()
+         write(*,"(A, F9.6, A)"), " It takes ", time, "s to write file "//filename
+         call file_timer%reset()
+
          call thread_ptr%clear_hist_collections()
          call thread_ptr%hist_collections%clear()
          deallocate (buffer_fmd)
--- a/pfio/MultiGroupServer.F90
+++ b/pfio/MultiGroupServer.F90
@@ -820,7 +820,7 @@ contains
          file_size = file_size*4./1024./1024. ! 4-byte integer, unit is converted to MB
          speed = file_size/time
          lgr => logging%get_logger('MAPL.pfio')
-         call lgr%info(" Writing time: %f12.3 s, speed: %f12.3 MB/s, size: %f12.3 MB, oserver node: %i0~, file: %a", time, speed, file_size, this%node_rank, filename)
+         call lgr%info(" Writing time: %f9.3 s, speed: %f9.3 MB/s, size: %f9.3 MB, at server node: %i0~, of rank: %i0~, file: %a", time, speed, file_size, this%node_rank, this%innode_rank, filename)
          call file_timer%reset()
 
          call thread_ptr%clear_hist_collections()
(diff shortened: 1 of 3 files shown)


pfio/MultiGroupServer.F90
```diff
@@ -799,10 +803,17 @@ subroutine start_back_writers(rc)
...
             class is (AbstractDataMessage)
+               filename =q%file_name
+               call file_timer%start()
                call thread_ptr%put_dataToFile(q, address, _RC)
+               call file_timer%stop()
             end select
             call msg_iter%next()
          enddo
+         time = file_timer%get_total()
+         write(*,"(A, F9.6, A)"), " It takes ", time, "s to write file "//filename
```
mathomp4, line 814: Query for @weiyuan-jiang or @tclune: At this point in pFIO, do we know the *size* of the file being written or could we estimate it? We could put in a MB/s number here as well.

tclune: @weiyuan-jiang we really need this in terms of GB/sec.   Absolute time is good as well, so do both.   I suggest a format like the following to make things line up:
```
<sec> <GB/s>  wrote var <var> to file <file> on node <node-rank>
```

weiyuan-jiang: The speed is not quite accurate. In one of my c24 experiment, with deflation level 1,  8.8 M file becomes 5.7 M.  Also, the size if not accurate. With 9M data ( excluding meta data) , the size of the file is 8.8M without deflation.  For small file, I think the speed is way off.  Output example:

Writing test_writing_prof.geosgcm_budi.20000415_1200z.nc4(size     0.094464 Mb) on node 0
time :     0.005555 s    test_writing_prof.geosgcm_budi.20000415_1200z.nc4
speed:    17.004751 Mb/s test_writing_prof.geosgcm_budi.20000415_1200z.nc4
Writing test_writing_prof.geosgcm_prog.20000415_1200z.nc4(size     9.088511 Mb) on node 0
time :     0.013254 s    test_writing_prof.geosgcm_prog.20000415_1200z.nc4
speed:   685.742554 Mb/s test_writing_prof.geosgcm_prog.20000415_1200z.nc4
Writing test_writing_prof.geosgcm_surf.20000415_1030z.nc4(size     2.860416 Mb) on node 0
time :     0.068756 s    test_writing_prof.geosgcm_surf.20000415_1030z.nc4
speed:    41.602604 Mb/s test_writing_prof.geosgcm_surf.20000415_1030z.nc4

mathomp4: @weiyuan-jiang You probably want `MB` instead of `Mb`...I think.

tclune: I think the raw data size for the variable is sufficient for our purposes.    Don't worry about what compression does on the far end.

pfio/MultiGroupServer.F90
```diff
@@ -799,10 +806,21 @@ subroutine start_back_writers(rc)
...
+               call file_timer%stop()
             end select
             call msg_iter%next()
          enddo
+         time = file_timer%get_total()
+         file_size = file_size*4./1000./1000. ! 4-byte integer, unit is converted to MB
+         speed = file_size/time
+         write(*, "(A, F12.6, A)")"Writing "//filename//"(size ", file_size, " Mb) on node "//i_to_string(this%node_rank)
+         write(*, "(A, F12.6, A)") "time : ", time,   " s    "//filename
+         write(*, "(A, F12.6, A)") "speed: ", speed,  " Mb/s "//filename
```
tclune, line 821: This needs to be one line of output or you need to include the process ID.  Otherwise the two lines might get a bit mixed in the log file.

pfio/MultiGroupServer.F90
```diff
@@ -820,7 +820,7 @@ subroutine start_back_writers(rc)
          file_size = file_size*4./1024./1024. ! 4-byte integer, unit is converted to MB
          speed = file_size/time
          lgr => logging%get_logger('MAPL.pfio')
-         call lgr%info(" Writing time: %f12.3 s, speed: %f12.3 MB/s, size: %f12.3 MB, oserver node: %i0~, file: %a", time, speed, file_size, this%node_rank, filename)
+         call lgr%info(" Writing time: %f9.3 s, speed: %f9.3 MB/s, size: %f9.3 MB, at server node: %i0~, of rank: %i0~, file: %a", time, speed, file_size, this%node_rank, this%innode_rank, filename)
```
tclune, line 823: Can we condense this a bit:
```suggestion
         call lgr%info(" Writing time: %f9.3 s, speed: %f9.3 MB/s, size: %f9.3 MB, at server node: %i0~:%i0~, file: %a", time, speed, file_size, this%node_rank, this%innode_rank, filename)
```
We'll know what   137:4 means in this context.

I'm also thinking we'll regret `i0` as it will make the columns jagged, but at least it is at the end of the line.   And any particular width is either wasteful or wrong for big nodes in the future.

tclune: In the interests of time,  I'll go ahead and commit my suggestion.

tclune (approved):

mathomp4 (approved):
Reapprove after changelog fixup

PR #2100, opened 2023-04-26
Feature/atrayano/depends on

atrayano:
## Motivation and Context
 This was requested by Arlindo. The new keywords added to the export specs automatically enable allocation of variables based on the content

## Types of changes

## Checklist:

- [ x] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-
+- Added support for "depends_on" and "depends_on_children" for export_specs-
 - Added support for use of pFlogger simTime in logging (only if `-DBUILD_WITH_PFLOGGER=ON`)
   - Note: Due to bug in pFlogger v1.9.3 and older, you *must* specify a `dateFmt` in your logging configuration file in the
     formatter when using `simTime` (see [pFlogger issue #90](https://github.com/Goddard-Fortran-Ecosystem/pFlogger/issues/90))
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed message in _ASSERT statement in ESMFL_Mod.F90
 - Fixed bug in CubedSphereGridFactory when constructing a grid from a file
 - Cleaned up cubed-sphere grid factory and NCIO to produce files with consistent capitalization and types for the stretching factor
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-
+- Added support for "depends_on" and "depends_on_children" for export_specs
```
tclune, line 11: This is an important and non-trivial feature.   I think the change log needs a few sentences to indicate at least when a user might want these.

generic/MAPL_Generic.F90
```diff
@@ -540,12 +540,86 @@ recursive subroutine MAPL_GenericSetServices ( GC, RC )
...
+       type (MAPL_MetaComp), pointer :: cmeta
+       type(ESMF_GridComp), pointer :: childgridcomp
+       logical :: found
+
+       ! get the export specs
+       call  MAPL_StateGetVarSpecs(meta, export=ex_specs, _RC)
+       ! allow for possibility we do not have export specs
+       if (.not. associated(ex_specs)) then
+          _RETURN(ESMF_SUCCESS)
+       end if
```
tclune, line 569:
```suggestion
       _RETURN_IF(.not. associated(ex_specs))
```

CHANGELOG.md
atrayano: Ok, I will add these shortly

atrayano: Done

atrayano:
The CHANGELOG has been changed. As far as testing: I did the following
1) Ran GEOS without the changes to produce a baseline results
2) Compiled GOES with the version of MAPL from this PR, but without the new keywords in any export (the no-harm test). This was 0-diff
3) Added 3 new, 2d exports (for testing purposes only), foo, bar1, and bar2, to GOCART2G. And specified DEPEDS_ON=['bar1','bar2'] for 'foo'. Then in the run method got the pointers to bar1 and bar2 and asserted that they were allocated. Similarly, asserted the those pointer were not allocated with the DEPENDS_ON being commented out for 'foo'
4) added 'DEPENDS_ON_CHILDREN=.true' for 'foo' without adding 'foo' to children. This failed as expected (this is a requirement as per Arlindo) with appropriate error message
5) Added 'foo' to some of the children of GOCART2G. This fails as expected (same requirement as above) with appropriate error message
6) Added 'foo' to all children of GOCART2G, and asserted that 'foo' for each child is allocated

tclune (approved):

PR #2148, opened 2023-05-23
fixes #2147

### LINKED ISSUE #2147: Little utility to set fields to a constant would be useful
I realized we have no utility to simply set a field (or a field in a state) to a constant for all supported rank/types. I'm going to make a little PR to fix this.

bena-nasa:
Fixes #2147 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added field utilities to set a field or field in a state to a constant real number
+
 ### Changed
 
 ### Fixed
--- a/base/FieldUtilities.F90
+++ b/base/FieldUtilities.F90
@@ -9,6 +9,18 @@ implicit none
 private
 
 public is_field_undef
+public set_field_to_constant
+public set_state_field_to_constant
+
+interface set_state_field_to_constant
+   module procedure set_state_field_to_constant_r4
+   module procedure set_state_field_to_constant_r8
+end interface
+
+interface set_field_to_constant
+   module procedure set_field_to_constant_r4
+   module procedure set_field_to_constant_r8
+end interface
 
 contains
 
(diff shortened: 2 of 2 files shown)


atrayano (approved):

tclune (commented):
Hmm.  I got to this ticket too late.    

This feature fits nicely with other things @darianboggs and I had been doing in the `geom` directory. 

No reason that this cannot be moved in a subsequent PR though.  

But ... I don't think we want the names here.  So ... please don't heavily exercise this feature until we've discussed the naming.

tclune (commented):
Hmm.  I got to this ticket too late.    

This feature fits nicely with other things @darianboggs and I had been doing in the `geom` directory. 

No reason that this cannot be moved in a subsequent PR though.  

But ... I don't think we want the names here.  So ... please don't heavily exercise this feature until we've discussed the naming.

tclune (commented):
Hmm.  I got to this ticket too late.    

This feature fits nicely with other things @darianboggs and I had been doing in the `geom` directory. 

No reason that this cannot be moved in a subsequent PR though.  

But ... I don't think we want the names here.  So ... please don't heavily exercise this feature until we've discussed the naming.

bena-nasa:
@tclune tell me what to change these to and I will change, I have no opinion what these procedures are called, you choose.
Also these are really geometry agnostic operations in terms of the geometry so should they go in the GEOM directory necessarily, I can move it there.

tclune:
1. I think with the existing style, we probably want something more like `FieldSet` for the pure field case.  (I'm still debating keeping our style of `MAPL_` for public interfaces, so this needs a bit of iteration.)
2. How much use is there for the StateField case - it's a bit of a clunky interface.  Should we maybe just get the user to pull the field out and then use the former?   Not wed to that.  Otherwise the name should be still be `FieldSet` and just overloaded with a state argument.   Please suggest alternatives - these are useful discussions.
3. Some of the stuff Darian already has inside FieldBLAS will simplify your implementation.   In particular, there is a way to extract a 1D pointer for a field, and then you can just set it to the constant.  See `FieldGetCptr` and then use `c_f_pointer` to cast as a 1D array.   I think you still need a `SELECT CASE` for the kind though.
4. Bonus points.  Overload the `assignment(=)` operator for R4 and R8 for the Field case (ignoring any failing rc's).

mathomp4:
Putting a blocker on this just until the @tclune concerns are met. (only because I might merge it in accidentally!)

PR #2153, opened 2023-05-25
Protect required pflogger in CMake

### LINKED ISSUE #2152: pflogger always used
Since 2.38.0, geom/CMakeLists.txt and pfio/CMakeLists.txt contain `find_package (PFLOGGER REQUIRED)` which are not conditional based on BUILD_WITH_PFLOGGER. This for example causes spack-stack builds with pflogger disabled (`mapl ~pflogger`) to fail.

mathomp4:
## Description

Add `BUILD_WITH_PFLOGGER` protections in CMake in pfio and geom

## Related Issue

Closes #2152 

## Motivation and Context

Lets our non-pflogger MAPL users build without pFlogger.

## How Has This Been Tested?

Well, it shouldn't affect with-pflogger builds. Perhaps @AlexanderRichert-NOAA can test for us in a non-pflogger build environment, or I'll try as soon as I can to do it in a spack instance.

## Types of changes

## Checklist:

### DIFF
--- a/pfio/CMakeLists.txt
+++ b/pfio/CMakeLists.txt
@@ -112,7 +112,9 @@ else ()
   message(STATUS "netCDF does not have quantize capability")
 endif ()
 
-find_package (PFLOGGER REQUIRED)
+if (NOT BUILD_WITH_PFLOGGER)
+  find_package (PFLOGGER REQUIRED)
+endif ()
 
 esma_add_library (${this} SRCS ${srcs} DEPENDENCIES MAPL.shared MAPL.profiler NetCDF::NetCDF_Fortran NetCDF::NetCDF_C TYPE ${MAPL_LIBRARY_TYPE})
 target_link_libraries (${this} PUBLIC GFTL_SHARED::gftl-shared PFLOGGER::pflogger PRIVATE MPI::MPI_Fortran)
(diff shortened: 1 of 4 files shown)


pfio/CMakeLists.txt
```diff
@@ -112,7 +112,9 @@ else ()
   message(STATUS "netCDF does not have quantize capability")
 endif ()
 
-find_package (PFLOGGER REQUIRED)
+if (NOT BUILD_WITH_PFLOGGER)
```
AlexanderRichert-NOAA, line 115: Shouldn't these be `if (BUILD_WITH_PFLOGGER)`?

mathomp4: sigh. Yes. Feelin' dumb... pushed change.

mathomp4: @AlexanderRichert-NOAA Please try again with, you know, the right logic 😄

tclune: Someone is not doing a good job of taking the day off!   (But at least they are invisible on Teams!)

AlexanderRichert-NOAA: Yep this looks good, thanks Matt. And if it's your day off then I better not see this updated till Tuesday :)

mathomp4:
Update: This also uses an update to the circleci-tools orb to allow us to *remove* pflogger from the UFS test. This should prevent such mistakes from happening in the future (at least until UFS moves to using pflogger).

tclune (approved):

PR #2159, opened 2023-06-05
create cubed-sphere grid from files split by face

weiyuan-jiang:
When a file is split by face, im = jm. But it is sill cube-sphere grid. 
## Types of changes

## Checklist:

### DIFF
--- a/generic/MAPL_Generic.F90
+++ b/generic/MAPL_Generic.F90
@@ -6068,8 +6068,6 @@ contains
       character(len=ESMF_MAXSTR)            :: FileType
       integer                               :: isNC4
       logical                               :: isPresent
-      class(AbstractGridFactory), pointer :: app_factory
-      class (AbstractGridFactory), allocatable :: file_factory
       character(len=ESMF_MAXSTR) :: grid_type
       logical :: empty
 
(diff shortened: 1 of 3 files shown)


generic/MAPL_Generic.F90
```diff
@@ -6306,13 +6306,26 @@ subroutine MAPL_ESMFStateReadFromFile(STATE,CLOCK,FILENAME,MPL,HDR,RC)
...
+                fname_by_face = get_fname_by_face(trim(fname), 1)
+                inquire(FILE = trim(fname_by_face), EXIST=fexist)
+                if(fexist) then
+                  allocate(file_factory,source=grid_manager%make_factory(fname_by_face))
+                else 
+                  allocate(file_factory,source=grid_manager%make_factory(trim(fname)))
+                endif
+                _ASSERT(file_factory%physical_params_are_equal(app_factory),"Factories not equal")
+              end if
+            end block
```
tclune, line 6328: Please replace this with a helper procedure.

tclune: I can't immediately think of a really good name, but maybe just `get_factory_from_filename()`.

generic/MAPL_Generic.F90
```diff
@@ -6306,13 +6306,26 @@ subroutine MAPL_ESMFStateReadFromFile(STATE,CLOCK,FILENAME,MPL,HDR,RC)
...
+                app_factory => get_factory(MPL%GRID%ESMFGRID)
+                ! at this point, arrdes%read_restart_by_face is not initialized
+                ! pick the first face
+                fname_by_face = get_fname_by_face(trim(fname), 1)
+                inquire(FILE = trim(fname_by_face), EXIST=fexist)
+                if(fexist) then
+                  allocate(file_factory,source=grid_manager%make_factory(fname_by_face))
+                else 
+                  allocate(file_factory,source=grid_manager%make_factory(trim(fname)))
+                endif
```
tclune, line 6325: One danger here is if `make_factory()` fails.   Just a note to myself that we need to ensure that any _polymorphic_ function that has an RC must always provide some concrete type in the return value for precisely this situation.
(Using `rc=status` won't work here because the `allocate()` will fail before the status can be checked.

generic/MAPL_Generic.F90
weiyuan-jiang: This is just sanity check.  I use grid_is_consistent function

tclune (approved):

PR #2160, opened 2023-06-05
Fixes #2147 field operations

### LINKED ISSUE #2147: Little utility to set fields to a constant would be useful

bena-nasa:
I'm opening this PR here now as I believe it will be controversial and I'm sure the names I've chosen will be need to be changed since I never seem to choose names people like and I'm sure it will need revision.

In the short term I really do need to be able to run ExtDataDriver.x with fields on tiles. This means I need be able to get the arithmetic parser working in a rank agnostic/type agnostic way 

So I've made a first attempt to implement the functionality I need NOW. This is not the long term proper operator overloading

What I need:

1. I need to be able to perform +,-,*,/,** and a wide variety of intrinsic functions on fields. The end game if of course true operator overloading but this a a first baby step.

2. I need these operations to respect an undefined value in the computation which will I think will be very controversial.

3. I need to take these and redo the parser so I can work on fields of any rank

What have I done:
1. I first move all the generic little utilities from FieldBlas into their own module. Those are related to getting a 1D pointer representation as well as other utilities from that module. 
2. I created 3 new modules:
3. One is a module that implements the four binary operations +-*/ that takes a field and adds a second field to the first field. This makes use of a template.
4. One is a module that implements all the intrinsic sin/cos, etc functions also uses a template
5. One is module that has all the stuff that doesn't fit in the first two that so far doesn't need a template. Take a field to a power, negate a field, set a field to a constant, test if a field is identically a constant
6. I have added more tests and split up the tests in the tests directory

I just want to get the ball rolling
## Types of changes

## Checklist:


bena-nasa:
I simply can't wait For ESMF take my branch get this localelementcount by fixed in a release. Until then I've made a kluge that will suffice. Removed DNA

geom/FieldBinaryOperatorTemplate.H
```diff
@@ -0,0 +1,51 @@
+#define _NAME field
+#include "function_overload.macro"
+
+   subroutine _SUB(field_a,field_b,undef,rc)
+      type(ESMF_Field), intent(inout) :: field_a
+      type(ESMF_Field), intent(inout) :: field_b
+      real, optional, intent(in) :: undef
+      integer, optional, intent(out) :: rc
```
tclune, line 8: I'm confused.   A binary operator should have 3 arguments:    `C = A .op. B`

geom/FieldFunctionTemplate.H
```diff
@@ -0,0 +1,41 @@
+#define _NAME field
+#include "function_overload.macro"
+
```
tclune, line 3: This file should probably hav `Unary` in the name to be consistent with the binary case.

geom/FieldFunctions.F90
```diff
@@ -0,0 +1,85 @@
+#include "MAPL_Generic.h"
+
+module MAPL_FieldFunctions
+   use ESMF
+   use MAPL_ExceptionHandling
+   use MAPL_FieldPointerUtilities
+   use, intrinsic :: iso_fortran_env, only: REAL32, REAL64
+   implicit none
+   private
+
```
tclune, line 10: In the MAP proposal I suggested a different approach that would rely less on templating.    Of course I did not have to take it as far, so may have missed something.   But @bena-nasa and @darianboggs  and myself should discuss to see if we want to aim in that direction instead.

geom/tests/geom_setup.F90
```diff
@@ -0,0 +1,162 @@
+#include "MAPL_Generic.h"
+
+module geom_setup
+
+   use ESMF
+   use funit
+   use MAPL_ExceptionHandling
+
```
tclune, line 8: Note to self:  in long term we may want to have a separate directory of test utilities that can be used in multiple MAPL directories.   These would create simple grids, fields, etc. for other tests to use.       Otherwise we get some duplication with varying quality for these sorts of things.

geom/FieldBinaryOperatorTemplate.H
bena-nasa: @tclune, what do you want me to name this. This is the place I put +-/*, I can call it whatever you want. What name collective describes those 4 operations?

geom/tests/geom_setup.F90
bena-nasa: Absolutely agree 100%

geom/FieldFunctions.F90
bena-nasa: @tclune, I'm not sure what you mean, seems like this was about the minimum amount of code I could have done to implement this.

geom/FieldFunctionTemplate.H
bena-nasa: @tclune
Maybe FieldUnaryTemplate.H? 

> Most comments are just thoughts. But I do need to understand about number of arguments for binary and unary operators.

@tclune, can you elaborate, is the "undef" value I'm passing along to all of these?

geom/FieldFunctions.F90
bena-nasa: I read the MAP proposal, section, now even more confused, to quote:
"The lowest layer will be a set of procedures that act on 1D arrays of data"
why would we need this? If I want to take the sin or cosine or any intrinsic fortran function of a 1D array, since they already intrinsically work on 1D arrays, why do we need such procedure, at some point you will get a 1D representation of the field, why would reimplement a "sin" procedure for example when Fortran's sin can already to a 1D array.

tclune:
> What arguments? The "undef" being passed in?

Discussed offline, but summarizing here.  My comment referred to the primary argument fields/arrays.  Fortran binary operators relate 3 quantities: 2 inputs and 1 output.  Our design must reflect that if we want to expos high-level overloaded operators for this purpose.

bena-nasa:
@tclune 
Just did a bit push:
To summarize
The binary operators are now subroutine with 3 arguments (field_out,field_a,field_b) ! field_out = field_a+field_b
The unary operators are now subroutines with 2 arguments (field_ouy,field_in) ! field_out =function(field_in)
I removed the undef from the interface. I'm now assuming it is a field attribute. Limitations here. Either a 2 (3) fields have a missing_value/undef or they don't, if some do, some don't I fail right now.
Nor am I handling the case if the field_out is of different type than field_in, for now if the type_in/=type_out I'm dying
So there's a ways to go for true, rich operator overloading, this is all I need in the near term and sufficient for the use cases I have now and near future.

Tests/ExtDataRoot_GridComp.F90
```diff
@@ -556,8 +556,8 @@ subroutine CompareState(State1,State2,tol,rc)
             call ESMF_StateGet(State2,trim(nameList(i)),field2,_RC)
             call ESMF_FieldGet(field1,rank=rank1,_RC)
             call ESMF_FieldGet(field2,rank=rank2,_RC)
-            all_undef1 = is_field_undef(field1,_RC)
-            all_undef2 = is_field_undef(field2,_RC)
+            all_undef1 = FieldIsConstant(field1,MAPL_UNDEF,_RC)
+            all_undef2 = FieldIsConstant(field2,MAPL_UNDEF,_RC)
```
tclune, line 560: Note for future extension of this layer.     We could overload `==` and re-express this as:
```
all_undef = all (field1 == MAPL_UNDEF)
```

geom/FieldBinaryOperatorTemplate.H
```diff
@@ -0,0 +1,70 @@
+#define _NAME field
+#include "function_overload.macro"
+
+   subroutine _SUB(field_out,field_a,field_b,rc)
+      type(ESMF_Field), intent(inout) :: field_out
+      type(ESMF_Field), intent(inout) :: field_a
+      type(ESMF_Field), intent(inout) :: field_b
```
tclune, line 7: Don't change this for now, but I think we want shorter names for these arguments.   Will make the procedure a bit easier to read, and helps with keywords.  Indicating the type in the name is unnecessary.

Probably  `C, A, B`  as in  `C = A + B`.   But I admit in subroutine form it's less obvious what the right choice is.  Hence - no change yet.

Similar comment for unary case.

tclune (approved):
I made one minor change.

geom/FieldBinaryOperatorTemplate.H
bena-nasa: ok, we can change as needed in a future PR, I don't think anyone but me will be making use if this in the near term so I have no problem going and changing in a future PR.

PR #2170, opened 2023-06-07
Update parsing for MAPL_NetCDF datetime with real datetime support and public subroutine return values

darianboggs:
## Description
1) Update the `intent(out)`/`intent(inout)` arguments for the public subroutines. Now they return the `ESMF_Time` for the base datetime in the NetCDF `units_string` + the NetCDF numeric value datetime interval, and the numeric value datetime interval from the `ESMF_Time` using the same units as the original NetCDF `units_string`.

2) Add support for `real(kind=ESMF_KIND_R8)` values for the numeric datetime intervals and fractional datetime units values in the base datetime in the NetCDF `units_string`.

## Motivation and Context
Yonggang needs to process datetime values in NetCDF. He asked that the outputs for the public subroutines be updated and that `real(kind=ESMF_KIND_R8)` datetime intervals be added.

## How Has This Been Tested?
This is an update to an existing new feature. I tested it successfully with a more extensive test suite including the public and private procedures.

## Types of changes

## Checklist:


tclune (commented):
Is "NetCDF datetime" really the right name for this?     The module and file name suggest that it is an interface into netcdf.    Let's discuss when you are back.

base/MAPL_NetCDF.F90
```diff
@@ -18,282 +14,247 @@ module MAPL_NetCDF
 
    implicit none
 
-   public :: convert_NetCDF_DateTime_to_ESMF
-   public :: convert_ESMF_to_NetCDF_DateTime
```
metdyn, line 22: I really need these two subroutines.  My code is now depending on the `convert_NetCDF_DateTime_to_ESMF` subroutine which provides a complete set of information that is very helpful.    Please keep these two subroutines.

darianboggs:
The name could be changed. It is an interface between datetimes in NetCDF on ESMF_Time and ESMF_TimeInterval instances.

metdyn:
Hi @darianboggs, 
Thank you for adding the new functions for units_string and R8 support.  
After I have used your code 
```
   public :: convert_NetCDF_DateTime_to_ESMF
   public :: convert_ESMF_to_NetCDF_DateTime
```
I find your code is very helpful, but thorough, giving all the information that users may need at some point.  
So can you please these tools? 
It does not hurt to keep them, I think. 

In addition, my recent PR using `convert_NetCDF_DateTime_to_ESMF` is merged into develop.  If you delete these two functions, my code and the develop branch will fail at the build stage, I think.

mathomp4:
Marking long-term for now. @darianboggs @metdyn Is this PR still needed?

base/MAPL_NetCDF.F90
```diff
@@ -18,282 +14,247 @@ module MAPL_NetCDF
...
+   character(len=*), parameter :: EMPTY_STRING = ''
+   integer, parameter :: MAX_CHARACTER_LENGTH = 64
 
 contains
 
-   ! Convert NetCDF_DateTime {int_time, units_string} to
-   ! ESMF time variables {interval, time0, time1} and time unit {tunit}
-   ! time0 is the start time, and time1 is time0 + interval
-   subroutine convert_NetCDF_DateTime_to_ESMF(int_time, units_string, &
-         interval, time0, unusable, time1, tunit, rc)
```
metdyn, line 61: Sorry @darianboggs, I gave you the wrong feedback about this subroutine.  Your first implementation is very advanced. It was ahead of me.   I am now using each of the options from your deleted line:

  subroutine convert_NetCDF_DateTime_to_ESMF(int_time, units_string, &
         interval, time0, unusable, time1, tunit, rc)
E.g. https://github.com/GEOS-ESM/MAPL/blob/feature/ygyu/traj_v2/gridcomps/History/MAPL_HistoryTrajectoryMod_smod.F90#L1212

Will you please keep those arguments (time0, time1)?   I find them very useful. 

When you try to build the code, you will find the MAPL_HistoryTrajectoryMod_smod.F90 needs to be changed.  You can change that to pass the build.

darianboggs: OK. I'll keep the interface. The only questions are with the arguments. I have read more about CF units and `UDUNITS`. So I better understand what the `units` string represents, and I understand better why the duration can be integer or real of various kinds.

So my subroutine:
1) Converts the `units` string into a base or reference datetime (**`time0`**). That is specific to NetCDF CF Units. It creates an intermediate representation of the base datetime before creating the `ESMF_Time`.
2) Pulls out the datetime unit from the beginning of `units` (**`tunit`**).
3) Converts the real or integer duration since the base datetime and `tunit` into an `ESMF_TimeInterval` (**`interval`**).
4) Creates an `ESMF_Time` from `time0` and `interval` (**`time1`**`= time0 + interval`)

`time0` and `interval` are the only required outputs. `time1` and `tunit` are optional.
* Are `time0` and `time1` the best names for the two `ESMF_Time` variables?
* Should `optional` change for the four output arguments (`time0`, `interval`, `time1`, `tunit`)?

It will return the same arguments as before. It's just a matter of the names, positions, and `optional` attributes for the four output arguments.

From the MAPL/ESMF perspective, `time1` is the most important. From the NetCDF perspective, `tunit`, `interval`, and `time0` are more fundamental.

To convert the duration and `units` to `time1`, an `ESMF_Time`, the base datetime `time0` and `interval` are created first and added together. To convert `time1` to a NetCDF datetime, you need to know:
`tunits` *and*
* the original duration *or*
* `interval` *or*
* `time0`

Internally, I use a toolbox of four procedures to achieve 1-4.

I can leave it the way it is. I just wanted to check before we finalize it.

Thoughts?

darianboggs:
I've marked this as draft and push up the latest to back it up. I'll look at the failing builds tomorrow.

base/MAPL_NetCDF.F90
metdyn: Hi @darianboggs : Thank you for your details description and questions.  Indeed it is matter of name convention and optional attribute for the interface. 

For the example Time =  2500 + "seconds since 1970-01-01 00:00:00"
I have a name option-2 for a reference (time_base may feel a bit strange, but overall it might be more literal).  Meanwhile I have been using your design and they work well for me already. Hence either way is good.


Name choice 1 | Name choice 2 | Values | Attribute
-- | -- | -- | --
Time0 | Time_base | 1970-01-01 00:00:00 | Optional
Interval | Time_increment | 2500 seconds | Optional
Time1 | Time_current | 2500 + "seconds since 1970-01-01 00:00:00" | Output
tunit | Unit_of_time | Seconds | Optional

metdyn:
I am on the task to run samplers with this new code.  Thank you!

metdyn:
Thank you so much, @darianboggs. I checked the trajectory sampler run and it all works well.

metdyn (commented):
The interface works for me and thank you for the work!

metdyn (approved):

mathomp4 (approved):

metdyn:
Hi @darianboggs ,

I just see an example of negative time relative to base time.  The statement somewhere in the code 
_FAIL ( time < 0)  
needs to be changed accordingly. 

I need to generate similar content as shown in the file below.  Thank you for your attention!

dimensions:
        nlocs = 68668 ;
// global attributes:
                :date_time = 2019080100 ;
variables:
        float time(nlocs) ;

 time = -0.78443, -0.78397, -0.78258, -0.78023, -0.78761, -0.78761, -0.8,
    -0.8, -0.75128, -0.75128, -0.75077, -0.74414, -0.73323, -0.72369,
    0.12228, 0.12228, 0.10805, 0.88627, 0.03291, 0.86278, 0.85712, 0.00835,
    0.85154, 0.82485, -0.02439, -0.02439, -0.04731, -0.054, 0.78588, 0.77452,
    -0.07542, -0.07954, 0.74864, -0.10124, -0.10502, -0.10874, 0.72843,
    0.72065, -0.11782, -0.11959, 0.71493, 0.71117, 0.71117, -0.12661,

darianboggs:
I'm working on implementing negative durations.

darianboggs:
Negative durations implemented and tested.

metdyn:
> Negative durations implemented and tested.
Great!  Thank you!

darianboggs:
@mathomp4 (or any other reviewer): @metdyn says he doesn't have any changes he needs. He's the one that is using it. So I think it's done.

tclune (approved):

PR #2188, opened 2023-06-12
fixes #2187

### LINKED ISSUE #2187: Bug with topo correction utility in MAPL_VerticalMethods.F90
The high resolution work at NAS exposed an issue. There was yet another place that was not being properly protected in the case when you have an output grid that can't be decomposed at sufficient core count. In this case it was the "correct_topo" method. For some reason I was checking the field completion status, rather than if the grid actually had a DE which causes a crash since it then tries to obtain the pointer but it doesn't exist...
Easy fix, just making the issue here to reference in the PR.

bena-nasa:
fixes #2187 

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ parameters:
 
 # Anchors to prevent forgetting to update a version
 os_version: &os_version ubuntu20
-baselibs_version: &baselibs_version v7.7.0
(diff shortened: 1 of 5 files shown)


atrayano (dismissed):

mathomp4 (approved):
Zero-diff in my testing at C24

PR #2191, opened 2023-06-15
ExtDataDriver.x updates for benchmarking

bena-nasa:
Just a few minor tweaks for using ExtDataDriver.x for benchmarking IO.

- I added a new run option so that one could fill the imports from the "menu" just like exports if you wanted to simply test benchmarking a checkpoint and turn off extdata
- I added new fill option. Since most of the options are filling with a constant, non-spatial varying data, that would lead to unrealistically good compression. I've added to the menu to fill fields with a random number.

## Types of changes

## Checklist:


mathomp4 (approved):
Trust in Ben as this is his code!

PR #2196, opened 2023-06-16
Feature/bmauer/fixes #2149

### LINKED ISSUE #2149: Arithmetic expression parser in MAPL needs to become GEOM/rank agnostic
The current MAPL arithmetic parser can perform element-wise operations on any combination of supported fields on the obviously same grid ( including layout, otherwise this wouldn't make sense!) given character expression involving the names of the fields.

Right now this is limited to 2D or 3D 32 bit fields on ESMF_Grids with the ability to broadcast (to borrow from numpy) 2D fields into 3D fields when doing an operation involving a pair of fields that have the same grid but one is 2D, other 3D with a single ungridded dimension.

Assuming the that +-/*^ are overloaded for fields on any GEOM/rank/type, as well as current supported Fortran intrinsic or single argument functions I see no reason why this should not be very easy to generalize the code to work on fields regardless of rank like @darianboggs has been going for other things.

Besides History, the arithmetic expression parser is an integral part of my ExtData/History testing framework to fill fields from a "menu" of available fields to the user. Indeed, without it my testing framework would not work. I was updating ExtDataDriver.x to be able to run on a "tile" grid to use as a tester when I update ExtData to handle tiled fields, but the one stumbling block is that it will fall flat when using the expression parser since the fields on tiles are 1D (but still on a grid remember, NOT a locstream), with some number of undistributed dimensions

I've just made a PR that should allow me to fix this issue.

bena-nasa:
This PR updates the arithemetic expression parser to use the new field utilities to work ons fields, rather than arrays. This means it can now work in a rank/type agnostic fashion which crucial for future development since ExtDataDriver.x makes use of this extensively and was limited to 2D/3D gridded fields without this.

In addition I found several bugs in the new field utilities. These were:

1. The binary operator template had a bug, and the test was not working to catch it which I quickly found when trying to use these in the expression parser.
2. The field clone function had been improperly implemented in the initial fieldBlas implementation and moreover the test that had been added for it that wasn't actually added with an @test nor even if it had been was it doing anything. This function was implemented and a proper test was added.
3. The initial testing framework in the geom directory never even added a way to create and test fields with ungridded dimensions for some bizarre reason. I added that.

This also requires 2 new features to the field geom class:

1. I added a "**" operator to the binary operators as this was necessary for the expression parser.

2. A "broadcast" copy, the idea is if you want to copy field x of rank x_r to field y of rank y_r, if y_r = x_r + 1 and the shape of the first x_r dimensions in field y matches field x, then copy the pointer from x to y and replicate x along the last dimension of y. This is 110% crucial for ExtDatDriver.x and without this I would be hosed and this is non-negotiable for inclusion.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added field utilities to perform basic numeric operations on fields
+- Update arithemetic parser to work with any rank and type of ESMF fields
 
 ### Changed
 
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -66,7 +66,7 @@ endif()
 
 esma_add_library(
   ${this} SRCS ${srcs}
-  DEPENDENCIES MAPL.shared MAPL.constants MAPL.profiler MAPL.pfio MAPL_cfio_r4 PFLOGGER::pflogger
+  DEPENDENCIES MAPL.shared MAPL.constants MAPL.profiler MAPL.pfio MAPL_cfio_r4 MAPL.geom PFLOGGER::pflogger
                GFTL_SHARED::gftl-shared-v2 GFTL_SHARED::gftl-shared-v1  GFTL::gftl-v2 GFTL::gftl-v1
                esmf NetCDF::NetCDF_Fortran MPI::MPI_Fortran
   TYPE ${MAPL_LIBRARY_TYPE})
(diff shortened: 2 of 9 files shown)


bena-nasa:
Yes I see the gfortran test is failing. I will see what is going on and fix.

bena-nasa:
ok, fixed gnu on Discover, lets see what the CI does.

atrayano (approved):
Looks reasonable

mathomp4 (approved):

PR #2225, opened 2023-07-07
Feature/bmauer/mapl3 hconfig

bena-nasa:
Convert generic3g to config

## Types of changes

## Checklist:


mathomp4:
I'm adding the skip changelog label to this. I mean, it's a big change, but it's a change that occurred in `develop` already and it's in the changelog.

I defer to @tclune if he'd prefer an entry in the MAPL3 bit of the changelog.

generic3g/GenericConfig.F90
```diff
@@ -1,14 +1,13 @@
...
+   use esmf, only: Esmf_HConfig, ESMF_Config
    implicit none
    private
   
    public :: GenericConfig
   
    type :: GenericConfig
       type(ESMF_Config), allocatable :: esmf_cfg
-      class(YAML_Node), allocatable :: yaml_cfg
+      type(ESMF_HConfig), allocatable :: yaml_cfg
```
tclune, line 10: Does not need to be fixed in this PR, but I think we can now just eliminate GenericConfig.   But maybe not just yet.

tclune (approved):

tclune:
> I'm adding the skip changelog label to this. I mean, it's a big change, but it's a change that occurred in `develop` already and it's in the changelog.
> 
> I defer to @tclune if he'd prefer an entry in the MAPL3 bit of the changelog.

I have thought about starting a separate CHANGELOG under generic3g  that will then become a massive entry when we merge into MAPL2.     But ... I think it will just be better to write a big entry from scratch that emphasizes the various changes.  It will be a big non-backward compatible change and everything else should be "see the new api" (which hopefully exists by then).

tclune (approved):

PR #2226, opened 2023-07-07
Fixes #2038. Cleaned the rules for split field names

atrayano:
## Description

This fixes the logic generating names for split fields

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 
+## [2.39.5] - 2023-07-10
+
+### Fixed
+- Fixed logic in generating the names of the split fields. If the alias field in the History.rc has separators (;), each substring is used to name the resulting fields. If there are no separators, this will be the exact name of the first split field
+
 ## [2.39.4] - 2023-06-23
 
 ### Fixed
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ endif ()
 
 project (
   MAPL
-  VERSION 2.39.4
+  VERSION 2.39.5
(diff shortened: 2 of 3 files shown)


atrayano:
@acollow If you have a chance, could please try my fix for the disobeying alias bug (issue #2038). Now the rule(s) for generating the names of the split fields is slightly different: if no alias is provided, we append the numbers 001, 002, etc to the original name. If we give 'alias' in the HISTORY.rc, then this is be the exact name if the first split field (i.e. no digits appended), unless we use semicolon separators (;). In that case, we match as many of the "substrings" (again, we do not append digits to these), and only then we append 00n to the original name for the rest of the split fields

acollow:
Yeah, I'll grab your branch and do a quick test. I likely won't be able to get this done before daycare pick up so I will report back on Monday.

tclune (approved):

acollow:
Good morning, I can confirm that the variable names look correct in my test run. Thanks for working on this!

mathomp4:
> Good morning, I can confirm that the variable names look correct in my test run. Thanks for working on this!

@acollow Sounds good. I'm going to do a test myself and if all works, I'll release a tag.

mathomp4:
My tests do show this is good to go. I see before:

```
DIFFER : VARIABLE : CA.bcEXTTAU001 : DOES NOT EXIST IN "mapl2395test-2023Jul10-1day-c24//scratch/mapl2395test-2023Jul10-1day-c24.tavg2d_aer_x.20000415_0900z.nc4"
...
DIFFER : VARIABLE : BCEXTTAU : DOES NOT EXIST IN "stock-2023Jul10-1day-c24//scratch/stock-2023Jul10-1day-c24.tavg2d_aer_x.20000415_0900z.nc4"
```
but if I diff the two they are the same.

I'll work on a release.

PR #2229, opened 2023-07-12
Updates to allow building MAPL with Spack instead of Baselibs

mathomp4:
## Description

This PR has changes needed to build MAPL with Spack as the libraries instead of Baselibs. Note that this is different than building MAPL with Spack (i.e., `spack install mapl`) but rather using spack as the provider of everything needed for Spack.

The main change is that we need to now include `FindESMF.cmake` inside of MAPL. The reason is that the spack ESMF package doesn't not readily expose where that file is in the installation of ESMF. It's not impossible to find, but Spack doesn't really lend itself to exposing its internals easily. I could work with @climbfuji and probably figure something out, but for now, just adding the file here is "easy". My guess is this won't interfere with building MAPL with Spack as the `FindESMF.cmake` file would be identical, but this will be needed to be tested. 

We also move to turn `BUILD_WITH_FLAP` to default `OFF` as that won't be in Spack. This is fine for GEOSgcm and GEOSldas but GEOSadas still needs FLAP. So in our CI, we explicitly turn it on. As our CI currently uses Baselibs, we can do this as FLAP is there. Eventually I'll need to add CI that uses spack instead of Baselibs. 

## Motivation and Context

This is a step to allowing us to build MAPL (and eventually GEOS) with using spack as our libraries and not Baselibs.

## How Has This Been Tested?

I've build MAPL. Need to do more testing.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -217,6 +217,7 @@ workflows:
           checkout_mapl_branch: true
           mepodevelop: false
           rebuild_procs: 1
+          extra_cmake_options: "-DBUILD_WITH_FLAP=ON"
 
   build-and-publish-docker:
     when:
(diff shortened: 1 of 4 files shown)


mathomp4:
I just tried this with develop (along with #2233 and #2056 ) and it built and runs. I suppose that's all I can say.

Issue #2228, opened 2023-07-10
Possible race conditions in PFIO multi group server


bena-nasa:
User @mmanyin has reported the following symptoms. When using the multi-group PFIO server they will occasionally find the that the run crashes and throws an error in the server when opening the file (in the NetCDF4_FileFormatter) for writing on the server side. Restarting the segment runs through just fine so does seem like a possible race condition.

The other piece is that this only seems to occur when the user has a collection that is writing multiple time steps to the file for example:
```
  bot_inst_1hr_g1440x721_x1.format:       'CFIO' ,
  bot_inst_1hr_g1440x721_x1.descr:        '2d,1-Hourly,Instantaneous,Single-Level,Bottom Layer Diagnostics' ,
  bot_inst_1hr_g1440x721_x1.conservative: 1
  bot_inst_1hr_g1440x721_x1.nbits:        12 ,
  bot_inst_1hr_g1440x721_x1.template:     '%y4%m2%d2_%h2%n2z.nc4' ,
  bot_inst_1hr_g1440x721_x1.archive:      '%c/Y%y4/M%m2' ,
  bot_inst_1hr_g1440x721_x1.mode:         'instantaneous' ,
  bot_inst_1hr_g1440x721_x1.grid_label: PC1440x721-DC
  bot_inst_1hr_g1440x721_x1.chunksize:    1440 721 1 1 ,
  bot_inst_1hr_g1440x721_x1.deflate:      1 ,
  bot_inst_1hr_g1440x721_x1.frequency:    010000 ,
  bot_inst_1hr_g1440x721_x1.duration:     240000 ,
  bot_inst_1hr_g1440x721_x1.ref_time:     000000 ,
  bot_inst_1hr_g1440x721_x1.levels:       72
  bot_inst_1hr_g1440x721_x1.fields:       'PS'          , 'DYN'     ,
```
One possibility is that if somehow the file were still being written due to disk slowness and this causes some sort of NetCDF error but this is all speculation.
I will attempt to reproduce this by running ExtDataDriver.x with the multi-group server and producing multi-step files. If this can't be reproduced it will be hard to fix.

In theory in the multi-group server is would be possible to have some sort of protection, when the "coordinator" process assigns work, if the file name in the request matches the file a worker is still writing, well something I suppose could be done although would require some thought.

bena-nasa:
[...]

PR #2230, opened 2023-07-12
Fixed a bug that a file is being written by more than one processor simultaneously

weiyuan-jiang:
This PR is addressing issue #2228. When a file is written at different time step, it may lead to crash because different processes may try to write to it simultaneously. 
## Description

The captain process keeps a record of the files being written. If a request to write a file which is still in the record, it waits until the completion of that process. To accomplish this, the front end now sends the file name to the captain when asking for a writer. The writer sends back the file name to the captain when it is free.
## Types of changes

## Checklist:


weiyuan-jiang:
@bena-nasa Could you please help to test my change?

bena-nasa:
> @bena-nasa Could you please help to test my change?

Yes, I will give this a workout

mathomp4:
@weiyuan-jiang Note that when you are happy with this code, it probably should be re-done as a hotfix against `main` rather than a bugfix against `develop`. Since this is the sort of fix that might need to be backported to older tags, it's probably good to have it isolated from all the other work on develop.

Plus as @tclune would say, it's a bug on main in released code so it's a hotfix!

.github/workflows/enforce-labels.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   require-label:
     runs-on: ubuntu-latest
     steps:
-      - uses: mheap/github-action-required-labels@v4
+      - uses: mheap/github-action-required-labels@v5
```
tclune, line 11: Something is amiss. This file should not have been touched.  I fear you started with the wrong baseline.

tclune:
This PR started from an incorrect branch.   Need to start over from fresh copy off `main` branch.

tclune (changes requested):
Start from `main` not `develop`

PR #2233, opened 2023-07-17
Fixes #2133. Add ability to set ESMF Logging mode at Run Time

### LINKED ISSUE #2133: Feature request: set ESMF logging mode at run-time via config file
The ESMF logging mode is currently part of `cap_options` set via a command line tool. We do not use the command line tool in GCHP and so set the logging mode at a very high level to pass when creating MAPL CAP (see [here](https://github.com/geoschem/GCHP/blob/0345abf5cb237a72e0fc33873263f6618ff3e16b/src/GCHPctm.F90#L29)). To change the logging mode, such as to disable the default `ESMF_LOGKIND_MULTI_ON_ERROR`, users need to change the code and recompile. 

It would be more convenient to set the logging mode at run-time via a config file. This is not possible with the current implementation of MAPL, however, because the logging mode is needed for ESMF initialization (see [here](https://github.com/GEOS-ESM/MAPL/blob/5545f08d47514b2af409966a80f7e67af9d26320/gridcomps/Cap/MAPL_Cap.F90#LL281C53-L281C53)) but ESMF initialization needs to happen prior to reading a config file.

The solution to this problem is to pass a config file rather than logging string to `ESMF_Initialize`. This is possible according to the [ESMF docs](http://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node4.html#SECTION04024100000000000000).

mathomp4:
## Description

This PR provides the ability to override the ESMF logging type at run time. To do so, create a file called `ESMF.rc` and add the flag there:
```
$ cat ESMF.rc
logKindFlag: ESMF_LOGKIND_MULTI_ON_ERROR
```
and if it's present, it will read it and change the log level.

Now, in this branch (at least) the file *MUST* be named `ESMF.rc` because, well, I couldn't figure anything else out. (Since `CAP.rc` is parsed by ESMF, you can't put anything in there to tell ESMF what this file will be called.)

## Related Issue

Closes #2133 

## Motivation and Context

Lets users override the ESMF logging flag.

## How Has This Been Tested?

Ran GEOS with it. Seemed happy.

## Types of changes

## Checklist:


atrayano (commented):
I have a mild concern about the inquire statement: this might present issues at NAS, especially at large CPU counts. How about only root queries the file, and broadcasts the result. The communicator is already available.

mathomp4:
> I have a mild concern about the inquire statement: this might present issues at NAS, especially at large CPU counts. How about only root queries the file, and broadcasts the result. The communicator is already available.

Ohh nuts you are right. Okay. Let me see if I can work on that.

tclune:
> I have a mild concern about the inquire statement: this might present issues at NAS, especially at large CPU counts. How about only root queries the file, and broadcasts the result. The communicator is already available.

Variant suggestion.  Introduce a `parallel_inquire()` procedure that wraps such logic.   This will allow us to readily update other occurrences as they are encountered.    Don't try to handle hard cases.  Just the basic.   (Might not be possible to cover all cases with a Fortran generic interface.)

mathomp4:
Of course, now I wonder: does `ESMF_Initialize` with a `configFileName` passed in open the file on every process? Paging @theurich...

theurich:
> Of course, now I wonder: does `ESMF_Initialize` with a `configFileName` passed in open the file on every process? Paging @theurich...

Yes, it does.

mathomp4:
Oh boy. I looked at the ESMF docs and this does not have to be an RC file, it can be a yaml file:

https://earthsystemmodeling.org/docs/nightly/develop/ESMF_refdoc/node4.html#SECTION04024000000000000000

But from that, I might need help from @tclune or @bena-nasa to translate the `configFile` and `configKey` sections.[^1]

[^1]: I'm ignoring `configFilenameFromArgNum` for now.

mathomp4:
> > Of course, now I wonder: does `ESMF_Initialize` with a `configFileName` passed in open the file on every process? Paging @theurich...
> 
> Yes, it does.

Aaaaaa! Well, I suppose with that it is moot to care about the inquire as we'll be doing 20000+ opens anyway if we did this at NAS. (Aka: don't do this at NAS.)

Still, now I wonder if we want to do this this way? It might be better to actually go old school and process `CAP.rc` or something with Fortran on the root and write our own code to decode the new log format and then just pass that integer to ESMF!

mathomp4:
Of course, we always have to do an ESMF_ConfigLoadFile anyway in MAPL at NAS and that hasn't gone belly up yet. So...

tclune:
We don't have to _pass_ the config file to every PET.    Not sure if that helps though.

tclune (approved):

PR #2234, opened 2023-07-17
fixes #2232

### LINKED ISSUE #2232: Bugs in time_ave_util.x identified
User @zhaobin74  was trying to use the time averaging utility in MAPL, time_ave_util.x which was developed to be a general time averaging utility that could work with all supported grids and file formats for said grids MAPL can support and replace the older time_ave.x that lived in GEOS_Util and used GFIO for IO. The user tried to give the code files that were
1. produced by History on a tripolar grid
2. Files have no vertical coordinate

This exposed 2 bugs, I'm making an issue here to reference when I made the PR to fix:
1. The tripolar grid factory has an assertion that the NX and NY chosen must evenly divide the grid size. This was probably an extra protection put in place so that when the user is running MOM5, which requires this, they can't screw this up. However, in this code or Regrid_Util.x, any code that just wants to make a tripolar grid for say regridding purposes, there's absolutely no reason for this. What happened was that since in this time_ave_util.x, the code chose a decomposition based on the number of processors, where as in the model this would be set by the user. Since what the code chose did not evenly divide, it failed, and for most cases it won't evenly divide. So this must be relaxed so this is useable in utilities such as this. It was kind of silly to put in at all since in the real model, the user controls, they should be responsible for ensuring they put the NX and NY that satisfies their application in AGCM.rc rather than crippling the entire factory for other uses.
2. The code was apparently not tested in a case where the input files did not have a vertical coordinate which was not properly handled, but a simple fix.

bena-nasa:
This fixes #2232

## Types of changes

## Checklist:

### DIFF
--- a/Apps/time_ave_util.F90
+++ b/Apps/time_ave_util.F90
@@ -1166,12 +1166,14 @@ contains
       basic_metadata=formatter%read(_RC)
       call metadata%create(basic_metadata,trim(filename))
       lev_name = metadata%get_level_name(_RC)
-      call metadata%get_coordinate_info(lev_name,coords=levs,coordUnits=lev_units,long_name=long_name,&
-            standard_name=standard_name,coordinate_attr=vcoord,_RC)
-      plevs => levs
-      vertical_data = VerticalData(levels=plevs,vunit=lev_units,vcoord=vcoord,standard_name=standard_name,long_name=long_name, &
-            force_no_regrid=.true.,_RC)
-      nullify(plevs)
+      if (lev_name /= '') then
+         call metadata%get_coordinate_info(lev_name,coords=levs,coordUnits=lev_units,long_name=long_name,&
+               standard_name=standard_name,coordinate_attr=vcoord,_RC)
+         plevs => levs
+         vertical_data = VerticalData(levels=plevs,vunit=lev_units,vcoord=vcoord,standard_name=standard_name,long_name=long_name, &
+               force_no_regrid=.true.,_RC)
+         nullify(plevs)
+      end if
 
       if (present(rc)) then
          rc=_SUCCESS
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Relaxed restriction in the tripolar grid factory so that grids can be made even when the decomposition deos not evenly divide the grid dimension so that the factory can be used in utilities where the core count makes such a condition impossible to satisfiy
```
tclune, line 13: This should not be combined with a hotfix ...

Apps/time_ave_util.F90
```diff
@@ -1166,12 +1166,14 @@ subroutine get_file_levels(filename,vertical_data,rc)
...
-            force_no_regrid=.true.,_RC)
-      nullify(plevs)
+      if (lev_name /= '') then
+         call metadata%get_coordinate_info(lev_name,coords=levs,coordUnits=lev_units,long_name=long_name,&
+               standard_name=standard_name,coordinate_attr=vcoord,_RC)
+         plevs => levs
+         vertical_data = VerticalData(levels=plevs,vunit=lev_units,vcoord=vcoord,standard_name=standard_name,long_name=long_name, &
+               force_no_regrid=.true.,_RC)
+         nullify(plevs)
+      end if
```
tclune, line 1176: Don't change,  but just a reminder/suggestion of how to avoid conditionals (almost) in these cases.

```suggestion
     _RETURN_IF(lev_name == '')

      call metadata%get_coordinate_info(lev_name,coords=levs,coordUnits=lev_units,long_name=long_name,&
            standard_name=standard_name,coordinate_attr=vcoord,_RC)
      plevs => levs
      vertical_data = VerticalData(levels=plevs,vunit=lev_units,vcoord=vcoord,standard_name=standard_name,long_name=long_name, &
            force_no_regrid=.true.,_RC)
      nullify(plevs)
```

mathomp4: Huh. I didn't know about `_RETURN_IF`. So this sort of "short-circuits" the code at that point and RETURNs?

tclune (approved):

PR #2237, opened 2023-07-18
Feature/bmauer/extdatadriver delay

bena-nasa:
@tclune, thoughts on this, naming, style, etc...

Implemented a simple "mapl_sleep" function so as to avoid using the vendor "sleep", it takes a single integer argument in milliseconds (happy to change seconds etc...) and does what the name says, for that number of milliseconds. The extra bits are just in case this is called and crosses midnight.

Could not really thing of anything to add to this, not like there's any sort of error checking I can even think of to do here or even how one could add some sort of unit test for this since it is not as if there's something this returns one can check...

I did play overriding the intrinsic date_and_time so I could convince myself that indeed this would work when the system crosses midnight.

Wanted this so I could add the ability in ExtDataDriver.x to introduce a tilmestep latency for testing purposes and not use "sleep" given it is non-standard, which I also did in this branch

## Types of changes

## Checklist:

### DIFF
--- a/Tests/ExtDataRoot_GridComp.F90
+++ b/Tests/ExtDataRoot_GridComp.F90
@@ -164,11 +164,10 @@ MODULE ExtDataUtRoot_GridCompMod
          synth => synthWrap%ptr
          call ESMF_ClockGet(Clock,currTime=currTime,_RC)
 
+         synth%delay = -1
          call ESMF_ConfigFindLabel(cf,label='delay:',isPresent=isPresent,_RC)
          if (isPresent) then
             call ESMF_ConfigGetAttribute(cf,label='delay:',value=synth%delay,_RC)
-         else
-            synth%delay = -1
          end if
 
          call ESMF_ConfigGetDim(cf,nrows,ncolumn,label="FILL_DEF::",rc=status)
--- a/Tests/ExtDataRoot_GridComp.F90
+++ b/Tests/ExtDataRoot_GridComp.F90
@@ -8,6 +8,7 @@
 MODULE ExtDataUtRoot_GridCompMod
       use ESMF
       use MAPL
+      use MAPLShared
       use VarspecDescriptionMod
       use VarspecDescriptionVectorMod
       use netcdf
@@ -38,6 +39,7 @@ MODULE ExtDataUtRoot_GridCompMod
          type(StringStringMap) :: fillDefs
          character(len=ESMF_MAXSTR) :: runMode
          type(timeVar) :: tFunc
+         integer :: delay
       end type SyntheticFieldSupport
 
       type :: SyntheticFieldSupportWrapper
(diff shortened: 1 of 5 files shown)


Tests/ExtDataRoot_GridComp.F90
```diff
@@ -38,6 +39,7 @@ MODULE ExtDataUtRoot_GridCompMod
          type(StringStringMap) :: fillDefs
          character(len=ESMF_MAXSTR) :: runMode
          type(timeVar) :: tFunc
+         integer :: delay
```
tclune, line 42: Are you certain you want integer and not real?   Can always change it later, but 1 second is a long time for a computer.

Tests/ExtDataRoot_GridComp.F90
```diff
@@ -161,6 +164,13 @@ SUBROUTINE Initialize_ ( GC, IMPORT, EXPORT, CLOCK, rc )
          synth => synthWrap%ptr
          call ESMF_ClockGet(Clock,currTime=currTime,_RC)
 
+         call ESMF_ConfigFindLabel(cf,label='delay:',isPresent=isPresent,_RC)
+         if (isPresent) then
+            call ESMF_ConfigGetAttribute(cf,label='delay:',value=synth%delay,_RC)
+         else
+            synth%delay = -1
+         end if
+
```
tclune, line 173:
```suggestion
         synth%delay = -1
         call ESMF_ConfigFindLabel(cf,label='delay:',isPresent=isPresent,_RC)
         if (isPresent) then
            call ESMF_ConfigGetAttribute(cf,label='delay:',value=synth%delay,_RC)
         end if
```

shared/MAPL_Sleep.F90
```diff
@@ -0,0 +1,44 @@
...
+   current_time = (t(5)*3600+t(6)*60+t(7))*1000+t(8)
+   temp_accumulation = current_time - start_of_day
+   if (current_time < previous_time) then
+      start_of_day = 0
+      previous_days_accumulated = previous_days_accumulated + previous_time
+      temp_accumulation = current_time - start_of_day
+   end if
+   total_accumulation = temp_accumulation + previous_days_accumulated
+   previous_time = temp_accumulation
+   if ( total_accumulation > wait_time ) exit
```
tclune, line 39: `system_clock()` is much easier to use.   Were you just unaware?  Or is there some reason you want actual time of day info?

Tests/ExtDataRoot_GridComp.F90
tclune: Ah I see now that this is in milliseconds.  I'll add  a different comment.

tclune, line 42:
```suggestion
         integer :: delay ! in milliseconds
```

Tests/ExtDataRoot_GridComp.F90
bena-nasa: changed

shared/MAPL_Sleep.F90
bena-nasa: OK redid with system_clock, now takes a real second, less code...

tclune (approved):

PR #2238, opened 2023-07-18
Fix a bug so that MultigroupServer does not allow a file written by m…

weiyuan-jiang:
…ultiple processes at the same time

The names of files being written are saved. Every time before sending an empty rpocess, it checks the file name first. If it is still being written, it would wait to avoid writing a file in multiple processors.
## Description

This PR replaces the PR #2230 
## Types of changes

## Checklist:


mathomp4:
Blocking this as I need to fix up main...

tclune (dismissed):

mathomp4:
Unblocking. My tests show it works!

mathomp4:
Uh oh. Bug in the UFS-like build. Testing incoming.

mathomp4:
I think I fixed it. We need an *explicit* dependency on gFTL v2. I suppose we were getting that somehow when not building like UFS.

Maybe we should just convert all of MAPL to use v2... but I leave that to @tclune

mathomp4 (approved):

PR #2246, opened 2023-07-19
Feature/bmauer/fixes #2149

### LINKED ISSUE #2149: Arithmetic expression parser in MAPL needs to become GEOM/rank agnostic

bena-nasa:
This allows ExtDataDriver.x to work on "tile" grids and create the locstream on an attached grid and then create 'tileonly' variables which is precursor to getting ExtData able to ingest tile data and retire MAPL_ReadForcing. This will allow me to develop without having to run the full model.

Unfortunately do to a bug or limitation (which has been reported to ESMF) when doing ESMF_FIeldCreate on grids with ESMF_USER_INDEX, I had to redo the "clone" function in the field operation utilities to have MAPL, rather than ESMF allocate the memory which of course of the ugly rank/type logic we wan to avoid. I'm sorry, very sorry about this but until ESMF either fixes the bug or removes the limitation this is what we must do and the way it must be, it just is.

I also updated and moved the field destroy that takes advantage of the fact ESMF now can report who allocated the memory so that I can destroy these fields no matter who allocated the memory.

## Types of changes

## Checklist:

### DIFF
--- a/Tests/ExtDataRoot_GridComp.F90
+++ b/Tests/ExtDataRoot_GridComp.F90
@@ -38,6 +38,7 @@ MODULE ExtDataUtRoot_GridCompMod
          type(StringStringMap) :: fillDefs
          character(len=ESMF_MAXSTR) :: runMode
          type(timeVar) :: tFunc
+         logical :: on_tiles
       end type SyntheticFieldSupport
 
       type :: SyntheticFieldSupportWrapper
@@ -74,13 +75,12 @@ MODULE ExtDataUtRoot_GridCompMod
          call MAPL_GridCompSetEntryPoint ( GC, ESMF_METHOD_INITIALIZE,  Initialize_, _RC)
          call MAPL_GridCompSetEntryPoint ( GC, ESMF_METHOD_RUN,   Run_, _RC)
 
-         call ESMF_ConfigFindLabel(cf,"tiling_file:",isPresent=on_tiles,_RC)
-
          allocate(synth)
          synthWrap%ptr => synth
          call ESMF_UserCompSetInternalState(gc,wrap_name,synthWrap,status)
          _VERIFY(status)
-         if (on_tiles) then
+         call ESMF_ConfigFindLabel(cf,"tiling_file:",isPresent=synth%on_tiles,_RC)
+         if (synth%on_tiles) then
             vloc = MAPL_DimsTileOnly
          else
             vloc = MAPL_DimsHorzOnly
@@ -163,8 +163,10 @@ MODULE ExtDataUtRoot_GridCompMod
          type(SyntheticFieldSupportWrapper) :: synthWrap
          type(SyntheticFieldSupport), pointer :: synth => null()
          character(len=ESMF_MaxStr) :: key, keyVal
+         type(MAPL_MetaComp), pointer :: MAPL
 
          call ESMF_GridCompGet( GC, name=comp_name, config=CF, _RC )
+         call MAPL_GetObjectFromGC ( GC, MAPL, _RC )
 
          call ESMF_UserCompGetInternalState(gc,wrap_name,synthWrap,status)
          _VERIFY(status)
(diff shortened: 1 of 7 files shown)


Tests/ExtDataRoot_GridComp.F90
```diff
@@ -509,40 +554,63 @@ subroutine FillState(inState,outState,time,grid,Synth,rc)
...
+         exPtr2=synth%tFunc%evaluate_time(Time,_RC)
+      end if
 
-      call MAPL_GetPointer(inState,exPtr2,'i_index',_RC)
-      do j = 1,ldims(2)
-         do i=1,ldims(1)
-            exPtr2(i,j)=i1+i-1
+      if (synth%on_tiles) then
+      
+      else
```
tclune, line 578:
```suggestion
      if (.not. synth%on_tiles) then
```

Tests/ExtDataRoot_GridComp.F90
```diff
@@ -509,40 +554,63 @@ subroutine FillState(inState,outState,time,grid,Synth,rc)
...
       seeds = mypet
       call random_seed(put=seeds)
-      call random_number(exPtr2)
+      if (synth%on_tiles) then
+         call MAPL_GetPointer(inState,exPtr1,'rand',_RC)
+         call random_number(exPtr1)
+      else
+         call MAPL_GetPointer(inState,exPtr2,'rand',_RC)
+         call random_number(exPtr2)
+      end if
```
tclune, line 613: ok - I'm curious.  What does randomness have to do with this?

geom/FieldPointerUtilities.F90
```diff
@@ -373,21 +378,56 @@ subroutine clone(x, y, rc)
...
-         ungriddedUBound=ungriddedUBound, name=name, _RC)
+      if (index_flag == ESMF_INDEX_USER) then
+         if (tk == ESMF_TYPEKIND_R4 .and. field_rank == 1) then
+            allocate(VR4_1d(lc(1)),_STAT)
+            y = ESMF_FieldCreate(grid,VR4_1d,gridToFieldMap=gridToFieldMap,name=name,_RC)
+         else if (tk == ESMF_TYPEKIND_R8 .and. field_rank == 1) then
+            allocate(VR8_1d(lc(1)),_STAT)
+            y = ESMF_FieldCreate(grid,VR8_1d,gridToFieldMap=gridToFieldMap,name=name,_RC)
+         else if (tk == ESMF_TYPEKIND_R4 .and. field_rank == 2) then
+            allocate(VR4_2d(lc(1),lc(2)),_STAT)
```
tclune, line 406: Will be nice in F2023 when we can use SELECT CASE on derived types ...

geom/FieldPointerUtilities.F90
```diff
@@ -859,4 +899,51 @@ subroutine GetFieldsUndef_r8(fields,undef_values,rc)
...
+    logical :: esmf_allocated
+
+    call ESMF_FieldGet(Field,typekind=tk,dimCount=rank,isESMFAllocated=esmf_allocated,_RC)
+    if (.not. esmf_allocated) then
+       if (tk == ESMF_TYPEKIND_R4 .and. rank == 1) then
+          call ESMF_FieldGet(Field,0,VR4_1d,_RC)
+          deallocate(VR4_1d,_STAT)
+       else if (tk == ESMF_TYPEKIND_R8 .and. rank == 1) then
+          call ESMF_FieldGet(Field,0,VR8_1d,_RC)
+          deallocate(VR8_1d,_STAT)
```
tclune, line 921: I'm tempted to ask you to try using the existing utilities to get a 1D pointer and deallocate that.   Not sure what the standard says about that though, so even if it works might be unsafe.

tclune: MAPL3 will almost always let ESMF do the allocation, so this block can maybe be removed then.

tclune (changes requested):
Minor change requested.  And some questions.

Tests/ExtDataRoot_GridComp.F90
bena-nasa: @tclune, a random number of one of the items in the "menu" of things you can use to fill exports, so that's what is has to do with it

bena-nasa:
> Minor change requested. And some questions.

@tclune, turns out some external repos call "MAPL_FieldDestroy" so I should not have change the name to just "FieldDestroy", as I broke the public API which was causing the GEOSgcm build to fail.

tclune:
> > Minor change requested. And some questions.
> 
> @tclune, turns out some external repos call "MAPL_FieldDestroy" so I should not have change the name to just "FieldDestroy", as I broke the public API which was causing the GEOSgcm build to fail.

OK - let's defer those kinds of changes until MAPL3.    There we may use things like `FieldDestroy` internally, and then have a compatibility layer than uses rename to the explicit name.    

I would hope destroying fields is very rare outside of MAPL though ...

PR #2210, opened 2023-06-26
Feature/wjiang/vertical indices

weiyuan-jiang:
## Description

This PR tries to improve the performance of vertical interpolation by saving the indexes and the linear weight.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist:

- [x] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [x] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #2251, opened 2023-07-25
fixed extension to the surface values in vertical interpolation

weiyuan-jiang:
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Added assert that may cause crash
-- Extended to the surface layer 
+- Fixed index bound and extended to the surface layer 
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed lower bound of plx ( non-flip) and fixed extension to the surface value for verticla interpolation 
+- Fixed lower bound of plx ( non-flip) and fixed extension to the surface value for vertical interpolation 
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Added assert that may cause crash
+- Extended to the surface layer 
+
 ### Removed
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed lower bound of plx ( non-flip) and fixed extension to the surface value for vertical interpolation 
+- Fixed extension to the surface in vertical interpolation 
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed index bound and extended to the surface layer 
+- Fixed lower bound of plx ( non-flip) and fixed extension to the surface value for verticla interpolation 
 
 ### Removed
 
--- a/base/MAPL_VerticalMethods.F90
+++ b/base/MAPL_VerticalMethods.F90
@@ -622,8 +622,9 @@ module MAPL_VerticalDataMod
       end where
     end if
 
-    deallocate(plx,_STAT)
-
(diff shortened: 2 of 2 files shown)


weiyuan-jiang:
@bena-nasa Please verify that this is a bug fix. I felt the original 'where' condition would never met. Also, if it is non-flip, the plx 's lower bound must be 1 in the implementation , otherwise it will be out of bound.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed index bound and extended to the surface layer
```
tclune, line 16: This comment is insufficient.    Bug fixes should give enough description that someone experiencing the bug might be able to notice the fix in the change log.   (After I review the remaining changes I may come back with a suggestion.)

base/MAPL_VerticalMethods.F90
```diff
@@ -585,48 +585,46 @@ subroutine VertInterp(v2,v3,pp,ple_,pl_,rc)
...
+          v2 = v3(:,:,km)
+      end where
     end if
 
-    if(flip) then
-       deallocate(plx,stat=status)
-       _VERIFY(STATUS)
-    end if
+    deallocate(plx,stat=status)
+    _VERIFY(STATUS)
```
tclune, line 627:
```suggestion
    deallocate(plx,_STAT)
```

CHANGELOG.md
tclune: Still not sure I understand what is actually fixed here.   But a better description would hint at something like "Fixed problem in treatment of vertical regridding ..."

weiyuan-jiang: if it is non-flip, the 3rd dimension of plx may start from 0 too.  The codes after that assume it starts from 1. That is the problem

weiyuan-jiang: There are 2 bugs here. 1) lbound of plx ;  2) "where" condition

weiyuan-jiang:
@bena-nasa For cell values vertical interpolation, there are no extension to the surface values ( no surface any way) . It becomes MAPL_UNDEF.  I guess it is what GEOS wants.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed lower bound of plx ( non-flip) and fixed extension to the surface value for vertical interpolation 
+
```
tclune, line 17: If it is 2 bugs then it should be 2 separate entries in the change log. 
```suggestion
- Fixed lower bound of plx ( non-flip) in vertical interpolation
- Fixed extension to the surface value for vertical interpolation 
```

tclune (approved):

weiyuan-jiang:
The lower bound of plx is probably not a bug. When it is passed in, the index is automatically starting from 1. @bena-nasa I would like your help to verify this PR.

mathomp4:
@weiyuan-jiang As you are presenting this as a hotfix to `main`, can you put an explanation in the main commit message as to what bug it is fixing, where it was encountered, etc. Or is there an issue you can point to?

Since you say this is non-zero-diff, we'll need a good explanation of why we are going non-zero-diff to convince @sdrabenh, @gmao-rreichle, etc.

bena-nasa:
I'll take a closer look at this, until I test this in detail, I'm not comfortable merging this since the vertical interpolation is used in 100% of every geos run so we cannot break it.

gmao-rreichle:
I can honestly say that I have no clue what this is about, but if there is a bug, I'm wondering if the omission of the `ppx<=ps*msn` condition is the right fix.  To me, this looks like it's intended to be a physics limit -- something like "pressure cannot be higher than surface pressure".  If it's an intentional limit that's implemented wrongly, the solution wouldn't be to omit the limit.  What I'm saying here may not make any sense, but I agree with @bena-nasa that we should proceed carefully.  I also agree with @mathomp4  and @tclune that a better Description of the bug in the PR's introductory comment box is needed.

weiyuan-jiang:
I run across the codes because of the other PR #2210 .  If I understand correctly, the conditions of the .and.  in the "where " block are contradictory.  So the codes there would never be executed.  If that is what GEOS wants ( after years of running), then this is not a bug any more.

tclune:
Is there a resolution to this issue?     I'd like to either merge or reject this PR unless there is still work going on to evaluate it.

gmao-rreichle:
> Is there a resolution to this issue? 

I have no idea. Has anyone checked with the modeling group to see if the `ppx<=ps*msn` condition at issue is an intentional physics check that inadvertently doesn't kick in?   I don't think the SI team should merge or reject this without confirmation from an atm model developer.

weiyuan-jiang:
This PR was rejected and would be closed. I had discussion with Ben and was convinced that we should not extend the values to the surface that has no values.

gmao-rreichle:
@weiyuan-jiang:  So did this get run by an atm model developer?  If not, I don't think the PR should be closed at this time.

PR #2252, opened 2023-07-26
Feature/bmaure/extdata on tiles

bena-nasa:
This PR provides a capability for extdata to fill variables on MAPL "tile" grids, aka our locstream used by the surface components that has been desired for ages. Just getting this PR in so @tclune and others can comment.

Note this is very different from griddedIO, there's no regridding, there's just not. This is a matter of reading the variable and moving the data where it needs to be as it, according to the locstream distribution used by the tiled components. This simply reads the variables (using the I-server just like the gridded variables) and scatters it according to the distribution the tiled grid has decided on.
I've created a simple tile io class that does what it needs to for this PR and nothing more and if the data/variable is on tile ExtData goes through this class. Also note that in there's a spot where normally when extdata runs a grid for the input file would be made. Now if the input file is on tiles, I just skip that, there's just no grid to be made, period in the current framework and no use to make one.

Finally I made an executive decision that for now, the "tile" files will have a single, non-time, dimension I have chosen the name of "tile_index" for as that is what the dimension represents, nothing more, nothing less.

This is the last piece needed so that work can proceed to retire MAPL_ReadForcing since ReadForcing is used to read a host of tile based climatologies in the surface components.

I have confirmed that I can add an import to a surface component (I chose catchment), and that the import is seen by ExtData on the tile grid and the results are very similar to ReadForcing, not the same though, small differences, just like I saw when doing the same thing with the gridded forcing like the SSTs.

With this there will be no excuse not to retire that code and update any components in GEOS that use it to use ExtData, just like the ocean can do now (although that is not on now by default).

## Types of changes

## Checklist:

### DIFF
--- a/griddedio/TileIO.F90
+++ b/griddedio/TileIO.F90
@@ -13,6 +13,7 @@ module MAPL_TileIOMod
    private
 
    type, public :: MAPL_TileIO
+     private
       type(ESMF_FieldBundle) :: bundle
       integer :: read_collection_id
       type(tile_buffer), allocatable :: tile_buffer(:)
--- a/griddedio/TileIO.F90
+++ b/griddedio/TileIO.F90
@@ -1,0 +1,120 @@
+#include "MAPL_Generic.h"
+
+module MAPL_TileIOMod
+   use ESMF
+   use pFIO
+   use MAPL_BaseMod
+   use MAPL_ExceptionHandling
+   use MAPL_CommsMod
+   use FIleIOSharedMod, only: MAPL_TileMaskGet
+
+   implicit none
+
+   private
+
+   type, public :: MAPL_TileIO
+      type(ESMF_FieldBundle) :: bundle
+      integer :: read_collection_id
+      type(tile_buffer), allocatable :: tile_buffer(:)
+      contains
+         procedure :: request_data_from_file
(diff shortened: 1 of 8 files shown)


griddedio/TileIO.F90
```diff
@@ -0,0 +1,120 @@
...
+   private
+
+   type, public :: MAPL_TileIO
+      type(ESMF_FieldBundle) :: bundle
+      integer :: read_collection_id
+      type(tile_buffer), allocatable :: tile_buffer(:)
+      contains
+         procedure :: request_data_from_file
+         procedure :: process_data_from_file
+   end type MAPL_TileIO
```
tclune, line 22: Can the data be private?
```suggestion
   type, public :: MAPL_TileIO
     private
      type(ESMF_FieldBundle) :: bundle
      integer :: read_collection_id
      type(tile_buffer), allocatable :: tile_buffer(:)
      contains
         procedure :: request_data_from_file
         procedure :: process_data_from_file
   end type MAPL_TileIO
```

tclune (dismissed):

tclune (approved):

bena-nasa:
@tclune, I accepted your change to make that private

PR #2296, opened 2023-08-03
Feature/hack for ExtData for using G5NR files with sample inputs for AIST project

bena-nasa:
I've updated the update_offset in the sampling definition in ExtData2G so that it can be negative. Since the ISO does not allow negative durations from what I can tell I just did this. Before passing the string to be converted to an ESMF time interval I check if it has a "-", if so, I strip that, then just negate the returned ESMF Time interval.

This will allow @metdyn to run ExtDataDriver.x in 2020 when the IODA fils are for, but set a -15 year offset in the sampling definition so he can use the G5NR files from 2005 to 2007

I implemented this by simply creating a new clock that is referenced in the component, by default it simply is set to the same time as whatever clock is passed in and the offset applied and offset default is 0.

## Types of changes

## Checklist:


mathomp4:
Is GEOSgcm okay with this change? I assume so since none of our extdata yaml files have the offset, but I just want to be sure.

bena-nasa:
> Is GEOSgcm okay with this change? I assume so since none of our extdata yaml files have the offset, but I just want to be sure.

I modified the strategy, but yes, this should be, since that update_offset in the sampling key words is not used as far as I know, the only thing that may use it was the CTM and all I did was allow it to represent a negative duration.

tclune (approved):
@bena-nasa   Not needed in this pass, but take a mental note that we may want to issue a pflogger warning when this option is used.  It is potentially dangerous.

PR #2309, opened 2023-08-16
Feature/tclune/mapl3 latlongeomfactory

tclune:
## Types of changes

## Checklist:


geom_mgr/latlon/HConfigUtils.F90
bena-nasa, line 1: Doesn't this belong in a different place? Seems very generic and widely useful

tclune: Definitely.   But I needed something "now".    I've assigned @darianboggs to create a full layer.  Collectively we need to decide on interfaces and where to put them.

bena-nasa:
@tclune, I don't see a hook in the new GeomFactory for the equivalent of append_var_metadata that was in the old factory

bena-nasa (commented):
I see a lot of the helper procedures needed to specify the bounds for the output file are no longer here. Append_variable_metdata caught my eye first, that probably needs to be returned. The rest generate_file_***_bounds, generate_file_reference_****, the question can the logic that those were implement be done in a way that it need not be attached to the specific grid for the IO layer.

bena-nasa (changes requested):
Seems zero diff to the old lat-lon grid factory. 
I do think that the append_variable_metdata helper procedure needs restoration. That is the only helper procedure from the old factory I see that I could not work around or that could be derived from the metadata and keep our existing IO system working with this new layer.

tclune:
> The one I would insist on being restored before I approve is append_variable_metadata.

Why does the metadata object need to be 1st created somewhere else?   The create metadata routine _is_ the former "append" plus the assumption that it will be first in the pipeline.    I can't immediately recall how I convinced myself of this, but will certainly amend the interface as the use cases come in.  Trying to keep it simple for the moment - just the stuff I can immediately see how it fits in the process.

bena-nasa:
> > The one I would insist on being restored before I approve is append_variable_metadata.
> 
> Why does the metadata object need to be 1st created somewhere else? The create metadata routine _is_ the former "append" plus the assumption that it will be first in the pipeline. I can't immediately recall how I convinced myself of this, but will certainly amend the interface as the use cases come in. Trying to keep it simple for the moment - just the stuff I can immediately see how it fits in the process.

Please go look at what the old append_variable_metdata is doing, I don't think you are understanding what it is for. This is NOT for appending the overall file metadata.
We add grid specific meta to the file variables, at least in the case of the cube, go look in the old factory. We add specific metadata to the variables so that panoply knows how to plot them.  The append_variable_metadata was a way to know for a particular grid, if any extra metadata needs to be attached to the file variables we will be writing.  Someone needs to know, if I'm a cube, every variable in the file needs these extra attributes. 
In reality every factory that is not a basic lat-lon should be adding "coordinate" attribute to the variables according to CF. So there really does need to be a place in general if I'm grid foo, there may need to be some extra attributes that need to be added to the variables.

tclune:
OK - I understand.   But still, it can wait for now.    It will be straightforward enough to bring in when the time comes.  Yes - we have to implement it for each factory subclass, but it's the same work now as it will be then, and my short term need is to use the factory to produce distinct grids so that generic can induce a regrid operation.

PR #2313, opened 2023-08-22
Read string attribute from netCDF variable

metdyn:
## Description

Previously  pfio only supports global string attributes.  This PR enables reading string attribute from netCDF variable in any VARID. 

## Related Issue

To read a IODA file with the string attribute, we need access to pfio/pfio_get_att_string.c.  The current setting allows for  varied = NC_GLOBAL only.  

Turned out:
C starts the varid from zero while Fortran from 1.   A constant shift is needed when using netCDF Fortran varid to access  netCDF-C function such as nc_get_att_string.   

Input: 
```
group: MetaData 
  variables:
             int64 dateTime(Location) ;
                 string dateTime:units = "seconds since 1970-01-01T00:00:00Z" ;
```
Output:  
```
ncid2, ncid, varid (Fortran)=       65536       65546           2
 xtype=12, len=          1
varid(C) = 1
 timeunits=seconds since 1970-01-01T00:00:00Z
```

## How Has This Been Tested?
gcc compiler

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Read string attribute from netCDF variable
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Read string attribute from netCDF variable
 
 ### Added
```
tclune, line 11:
```suggestion
### Added
- Added the ability to read string attributes of variables.   This is as opposed to "character" attributes - a distinction made by NetCDF.   Previously a small kludge had been used to allow reading string attributes, but was limited to attributes on the global var.

```

tclune (approved):

PR #2341, opened 2023-09-06
Trajectory sampler : regrid to locations in IODA files

metdyn:
## Description
Improved the trajectory sampler module in the History grid component to regrid model data to IODA lat-lon locations via locstream.  Additions include:

- code can read in file template for IODA files
- read netCDF string attribute from group
- add debug information

## How Has This Been Tested?
Tested with G5NR and IODA data on discover with 368 cores

## Types of changes

## Checklist:

### DIFF
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -3436,7 +3436,6 @@ ENDDO PARSER
          lgr => logging%get_logger('HISTORY.sampler')
          if (list(n)%timeseries_output) then
             if( ESMF_AlarmIsRinging ( list(n)%trajectory%alarm ) ) then
-               call list(n)%trajectory%close_file_handle(_RC)
                call list(n)%trajectory%create_file_handle(filename(n),_RC)
                list(n)%currentFile = filename(n)
                list(n)%unit = -1
@@ -3596,8 +3595,7 @@ ENDDO PARSER
          call list(n)%trajectory%regrid_accumulate(_RC)
          if( ESMF_AlarmIsRinging ( list(n)%trajectory%alarm ) ) then
             call list(n)%trajectory%append_file(current_time,_RC)
-            !!bug
-            !!the barrier did not work
+            call list(n)%trajectory%close_file_handle(_RC)
             call ESMF_GridCompGet(gc, vm=vm, _RC)
             call ESMF_VMGetCurrent(vm, _RC)
             call ESMF_VMbarrier(vm, _RC)
@@ -3866,9 +3864,9 @@ ENDDO PARSER
             call MAPL_GridGet(pgrid,globalCellCountPerDim=dims,_RC)
             IM = dims(1)
             JM = dims(2)
-            DLON   =  360._REAL64/float(IM)
+            DLON   =  360._REAL64/IM
             if (JM /= 1) then
-               DLAT   =  180._REAL64/float(JM-1)
+               DLAT   =  180._REAL64/(JM-1)
             else
                DLAT   =  1.0
             end if
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -59,6 +59,7 @@
   use gFTL_StringStringMap
   !use ESMF_CFIOMOD
   use pflogger, only: Logger, logging
+  use mpi
 
   implicit none
   private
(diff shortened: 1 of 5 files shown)


gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3861,9 +3866,9 @@ subroutine MAPL_GradsCtlWrite ( clock, state,list,fname,expid,expdsc,output_grid
             call MAPL_GridGet(pgrid,globalCellCountPerDim=dims,_RC)
             IM = dims(1)
             JM = dims(2)
-            DLON   =  360._REAL64/real(IM)
+            DLON   =  360._REAL64/float(IM)
```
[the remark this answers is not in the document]
tclune: Also completely unnecessary.   "IM" will be promoted to 64 bit precision when the division is processed.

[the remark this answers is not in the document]
tclune:
```suggestion
            DLON   =  360._REAL64/IM
```

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3861,9 +3866,9 @@ subroutine MAPL_GradsCtlWrite ( clock, state,list,fname,expid,expdsc,output_grid
...
             if (JM /= 1) then
-               DLAT   =  180._REAL64/real(JM-1)
+               DLAT   =  180._REAL64/float(JM-1)
```
tclune, line 3871:
```suggestion
               DLAT   =  180._REAL64/(JM-1)
```

tclune (changes requested):
Small but important changes needed.

gridcomps/History/MAPL_HistoryGridComp.F90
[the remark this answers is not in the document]
metdyn: Thanks, @tclune.  I had been staring at this change before I created the PR; I could not figure out why it happened.  I had thought NAG asked me to change it, but I donot remember at all.  

Now I figured out, it is because I did not merge this file from my working repository to the develop branch, instead I copied the MAPL_HistoryGridComp.F90 from my directory to here. So I missed the modification introduced by Matt (Aug 31 16:20:40 2023).  I will use a merge tool to do this. 

I agree  Fortran default promotion.  360._REAL64/IM  will do the work.   Other modifications 360._REAL64/ real(IM, kind=REAL64) is redundant.  I will make the changes.

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3596,6 +3596,11 @@ subroutine Run ( gc, import, export, clock, rc )
          call list(n)%trajectory%regrid_accumulate(_RC)
          if( ESMF_AlarmIsRinging ( list(n)%trajectory%alarm ) ) then
             call list(n)%trajectory%append_file(current_time,_RC)
+            !!bug
+            !!the barrier did not work
+            call ESMF_GridCompGet(gc, vm=vm, _RC)
+            call ESMF_VMGetCurrent(vm, _RC)
+            call ESMF_VMbarrier(vm, _RC)
```
metdyn, line 3603: I realized the usage of the ESMF_VMbarrier is right. I was complaining about this, because after this barrier step, the generated netCDF files by call append_file shows 2.1 MB and all the fields are there, except the 2d and 3d fields are zero.  I was expecting a full content in netCDF, but it does not happen.  So it is a separate problem.  

Maybe after
`
if (mapl_am_i_root()) then
 call this%file_handle%put_var(trim(item%xname),p_acc_rt_3d(:,:), start=[is,1],count=[nx,size(p_acc_rt_3d,2)])
endif
`
I should close the NC file on root.  Then I would get a finished NC file at the barrier point.
I will remove the comment. (So it is a bug)

tclune: Yes - I don't think NetCDF guarantees anything until the file is closed.  But I'm not an expert.

metdyn: @tclune  Yes! you are right ---  the  netCDF put_var does not mean it is writing to disk.  
I tested 
``` 
            call list(n)%trajectory%append_file(current_time,_RC)
            call list(n)%trajectory%close_file_handle(_RC)
            _FAIL('test close file')
```  
Now I immediately get the correct content 2d and 3d in NC files.  So that is the final for netCDF write to disk.

metdyn: I still see some netCDF latency phenomena on Mac that I donot understand as I am not an expert on netCDF.  But it does not affect the results and performance of the code, so I am fine with it. I will update the PR. Thanks.

metdyn:
This PR is out dated, and I will replace it with a new PR.

PR #2361, opened 2023-09-19
Feature/tclune/io bench

tclune:
## Description

Added structure to support a MAPL benchmark suite.

A few benchmarks introduced.

## How Has This Been Tested?

New benchmarks compiled and run on linux (intel) and laptos (OS X NAG).

## Types of changes

## Checklist:


CMakeLists.txt
```diff
@@ -224,7 +224,9 @@ add_subdirectory (gridcomps)
 add_subdirectory (griddedio)
 if (BUILD_WITH_FARGPARSE)
    add_subdirectory (tutorial)
-endif()
+ endif()
+ add_subdirectory (benchmarks)
+
```
mathomp4, line 229:
```suggestion
   add_subdirectory (benchmarks)
endif()
```

It looks like the benchmarks use fArgParse, so we want to make sure they are only built if we build with fArgParse.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added new benchmark suite.  Initial benchmarks are:
+	1. measuring raw bandwdith of a filesystem with multiple independent streams
```
mathomp4, line 13:
```suggestion
	1. measuring raw bandwidth of a filesystem with multiple independent streams 
```

PR #2374, opened 2023-09-24
Fixes #2373 - improved unused dummy macro

### LINKED ISSUE #2373: Improve macro for `_UNUSED_DUMMY`
The existing macro uses a PRINT statement and thus cannot go in PURE procedures.  A suggestion by @everythingfunctional is that an ASSOCIATE block can be used instead.    (Of course some care must then go to choosing a selector that does not conflict with the dummy argument.)

tclune:
## Motivation and Context

Want to be able to use macro in PURE procedures

## How Has This Been Tested?

Verified MAPL compiles with NAG.  This macro is not executable code, so should never change answers.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Modified fpp macro `_UNUSED_DUMMY(x) to use ASSOCIATE instead of PRINT.   With this change it can be used in PURE procedures.
+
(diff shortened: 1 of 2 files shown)


mathomp4 (approved):
Clever. I also like `q____` as a variable name. 😄

tclune:
I tested this PR with NAG and it worked.   I'll test again.

tclune:
Ah  - this is run time.  I did not check that.

The fix is to prepend with `if (.false.) then;  ...  ; endif`

PR #2381, opened 2023-09-26
Workarounds for MPICH

mathomp4:
## Description

This PR tracks various workarounds needed for MPICH as it does not support an interface in `use mpi` (that it should, see https://github.com/pmodels/mpich/issues/6691) and a bug in how it handles `ierror` (fixed in repo, see https://github.com/pmodels/mpich/issues/6693 and https://github.com/pmodels/mpich/pull/6694). The latter will probably be in MPICH 4.1.3, but that might be a while off.

Note that this also requires updates to pFlogger which has the same non-support (https://github.com/Goddard-Fortran-Ecosystem/pFlogger/pull/100) which is in develop there and will be released soon.

These were all pointed out to me by @climbfuji. I'm efforting to see if I can build a working MPICH stack on discover. It's been...a challenge.

## Motivation and Context

Let's MAPL (and hopefully GEOS) build with MPICH. And makes the life of @climbfuji better. (Though we might need to backport this stuff to a previous release cycle if it all works)

## How Has This Been Tested?

Ran with Intel Fortran/Intel MPI on discover and zero-diff. 🎉 

## Types of changes

## Checklist:


mathomp4:
Well, this seems to be safe. Zero-diff in testing.

mathomp4:
@climbfuji When you all use MAPL + MPICH, are you using our Oserver? If not, this might be enough for you (assuming you don't use pflogger as well).

mathomp4:
Note to @GEOS-ESM/mapl-team: I tested this on discover and zero-diff.

climbfuji:
@mathomp4 I was successful with backporting the workaround to 2.35.2 and 2.40.3. There are other issues with mpich4 with the ufs-weather-model that prevent us from testing if those backports are good or not. But I can share them with you if you wish.

mathomp4:
> No, they are patches that we run on those two versions in spack (our fork). I can create the PRs for that so that you can look at it).

Oh...okay. Well, it should be a pretty easy release if needed. But if this one is good, then we can work on getting out MAPL 2.42 for future work.

tclune (approved):
More checks - makes cmake initial step slower. :-(

PR #2390, opened 2023-09-27
Swath sampler step-1:  allow for destroying grid and regridder, and allocatable metadata in GriddedIO (donot merge)

metdyn:
## Description
For swath sampler to change grid at each epoch time, we need to destroy the outdated grid, route handle and metadata.  This is mostly worked out by @bena-nasa and modifications to GriddedIO are introduced to allow the time being written as a 2D variable as function of (x, y).  I put in this infrastructure change first to test the general framework.  The code is tested on Mac. 

## Types of changes

## Checklist:


griddedio/GriddedIO.F90
```diff
@@ -128,6 +129,11 @@ subroutine CreateFileMetaData(this,items,bundle,timeInfo,vdata,ogrid,global_attr
...
         integer :: status
 
+        if ( allocated (this%metadata) ) deallocate(this%metadata)
+        allocate(this%metadata)
+
+        call MAPL_FieldBundleDestroy(this%output_bundle, _RC)
+
         this%items = items
         this%input_bundle = bundle
         this%output_bundle = ESMF_FieldBundleCreate(rc=status)
```
metdyn, line 139: @bena-nasa: I was concerned about the added L135, until I see the L139.  So that is safe (first destroy bundle, then create bundle). It is easier to see the change in context on github than from the terminal. 
Meanwhile a question is what if this%output_bundle does not exist in the first hand, 
will  call MAPL_FieldBundleDestroy(this%output_bundle, _RC)  lead to an error?   
Maybe a safer way is  if (present (output_bundle)) destroy output_bundle.  Does ESMF handle this gracefully?

tclune: `output_bundle` is not an optional argument, so we cannot use `present(output_bundle)`.

But, yes, calling destroy on an uninitialized object should lead to an error condition.   I'm not 100% confident that ESMF correctly does this.  It would assume they have a derived type component with default initialization that they can use to check things.  (As oppossed to just assuming untouched memory has the value they expect.)

metdyn:
This PR for swath grid is out dated and will be replaced by a new one.

PR #2394, opened 2023-09-29
New SPLIT_CHECKPOINT option to replace read/write by face

bena-nasa:
This implements a new SPLIT_CHECKPOINT option that replaces the more limited read/write by face. This is because we know that on some file systems getting more files writing gives the best performance and this number is more than 6.

The general idea is that if you say "SPLIT_CHECKPOINT: YES" this triggers the checkpoints to be split. It still uses NUM_WRITERS to determine how many files to make.

The supplied checkpoint name, i.e. fvcore_internal_checkpoint for example now becomes a single yaml file if this is requested.
This YAML file for now just has the resolution and the number of files that were written.

The individual files would then be
fvcore_internal_checkpoint_0
fvcore_internal_checkpoint_1
fvcore_internal_checkpoint_n 

where n = NUM_WRITERS-1

When it comes time to read these back as restarts, I first see, can I make an HCONFIG out of the supplied restart name and if so does it have the keys I'm looking for that would denote a set of split restarts.

If so that triggers the SPLIT_RESTART option which then reads the N files backs.

If the the number of files does not divide NY an error is thrown. I am working python scripts that can be used to "resplit" the files.

Note this required moving some code around. The big thing was that since if it is a split restart, I need to essential make the number of restarts the "NUM_READERS", but in the old code, the communicators for all would have already been created. I needed to defer this so I moved where the communicators are created when reading. Now of course if the user has a single file input, they can still set NUM_READERS and go through the normal parallel NETCDF path. 

However, if I've detected this is a split file, I can override the num_readers for that particular file and the communicators only get created after this point.

I've done a start/stop regression test. I.E. the a control code for 24 hours, then with the new code take the same unspilt restarts, run 12 hours, output split restarts, read those split restarts, and output the normal single file restarts. This was 0-diff to the 24 hours control run

## Types of changes

## Checklist:


base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+       integer, optional, intent(out) :: rc
+
+       integer :: status, nx, ny, color, ny_by_writers, myid, j
+
+       nx = size(arrdes%i1)
+       ny = size(arrdes%j1)
+       _ASSERT(num_writers < ny,'num writers must be less than NY')
+       _ASSERT(mod(ny,num_writers)==0,'num writerss must evenly divide NY')
+       call mpi_comm_rank(full_comm,myid, status)
+       _VERIFY(status)
```
tclune, line 633:
```suggestion
       call mpi_comm_rank(full_comm,myid, _IERROR)
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Ycomm, status)
+       color = arrdes%NY0
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Xcomm, status) 
+       ny_by_writers = ny/num_writers
+       if (mod(myid,nx*ny/num_writers) == 0) then
+          color = 0
+       else
+          color = MPI_UNDEFINED
+       endif
+       call MPI_COMM_SPLIT(full_comm, color, myid, arrdes%writers_comm, status)
```
tclune, line 644: Unchecked status?

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+
+       integer :: status, nx, ny, color, ny_by_readers, myid, j
+
+       nx = size(arrdes%i1)
+       ny = size(arrdes%j1)
+       _ASSERT(num_readers < ny,'num readers must be less than NY')
+       _ASSERT(mod(ny,num_readers)==0,'num readers must evenly divide NY')
+    
+       call mpi_comm_rank(full_comm,myid, status)
+       _VERIFY(status)
```
tclune, line 671:
```suggestion
       call mpi_comm_rank(full_comm,myid, _IERROR)
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+       integer :: status, nx, ny, color, ny_by_writers, myid, j
+
+       nx = size(arrdes%i1)
+       ny = size(arrdes%j1)
+       _ASSERT(num_writers < ny,'num writers must be less than NY')
+       _ASSERT(mod(ny,num_writers)==0,'num writerss must evenly divide NY')
+       call mpi_comm_rank(full_comm,myid, status)
+       _VERIFY(status)
+       color =  arrdes%NX0
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Ycomm, status)
```
tclune, line 635:
```suggestion
       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Ycomm, _IERROR)
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+       integer, intent(in) :: num_writers
+       integer, optional, intent(out) :: rc
+
+       integer :: status, nx, ny, color, ny_by_writers, myid, j
+
+       nx = size(arrdes%i1)
+       ny = size(arrdes%j1)
+       _ASSERT(num_writers < ny,'num writers must be less than NY')
+       _ASSERT(mod(ny,num_writers)==0,'num writerss must evenly divide NY')
+       call mpi_comm_rank(full_comm,myid, status)
```
tclune, line 632:
```suggestion
       call mpi_comm_rank(full_comm,myid, _IERROR)
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+          color = 0
+       else
+          color = MPI_UNDEFINED
+       endif
+       call MPI_COMM_SPLIT(full_comm, color, myid, arrdes%writers_comm, status)
+       if (num_writers==ny) then
+          arrdes%IOgathercomm = arrdes%Xcomm
+       else
+            j = arrdes%NY0 - mod(arrdes%NY0-1,ny_by_writers)
+          call MPI_COMM_SPLIT(full_comm, j, myid, arrdes%IOgathercomm, status)
```
tclune, line 649:
```suggestion
          call MPI_COMM_SPLIT(full_comm, j, myid, arrdes%IOgathercomm, _IERROR)
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Ycomm, status)
+       color = arrdes%NY0
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Xcomm, status) 
+       ny_by_readers = ny/num_readers
+       if (mod(myid,nx*ny/num_readers) == 0) then
+          color = 0
+       else
+          color = MPI_UNDEFINED
+       endif
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%readers_comm, status)
```
tclune, line 682:
```suggestion
       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%readers_comm, _IERROR)
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+
+       nx = size(arrdes%i1)
+       ny = size(arrdes%j1)
+       _ASSERT(num_readers < ny,'num readers must be less than NY')
+       _ASSERT(mod(ny,num_readers)==0,'num readers must evenly divide NY')
+    
+       call mpi_comm_rank(full_comm,myid, status)
+       _VERIFY(status)
+       color =  arrdes%NX0
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Ycomm, status)
```
tclune, line 673:
```suggestion
       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Ycomm, _IERROR)
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+       color = arrdes%NY0
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Xcomm, status)
```
tclune, line 675:
```suggestion
       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%Xcomm, _IERROR) 
```

base/FileIOShared.F90
```diff
@@ -619,4 +617,78 @@ subroutine ArrDescrSet(ArrDes, offset, &
...
+          color = 0
+       else
+          color = MPI_UNDEFINED
+       endif
+       call MPI_COMM_SPLIT(full_comm, color, MYID, arrdes%readers_comm, status)
+       if (num_readers==ny) then
+          arrdes%IOscattercomm = arrdes%Xcomm
+       else
+          j = arrdes%NY0 - mod(arrdes%NY0-1,ny_by_readers)
+          call MPI_COMM_SPLIT(full_comm, j, MYID, arrdes%IOscattercomm, status)
```
tclune, line 687:
```suggestion
          call MPI_COMM_SPLIT(full_comm, j, MYID, arrdes%IOscattercomm, _IERROR)
```

tclune (changes requested):
Mostly ok.  But worth fixing up the error trapping before we merge.   I may have missed some checks, so please do another sweep.

base/NCIO.F90
```diff
@@ -3975,6 +3983,26 @@ subroutine add_fvar(cf,vname,vtype,dims,units,long_name,rc)
...
+          integer, intent(in) :: jm_world
+          integer, intent(in) :: num_writers
+          integer, intent(out), optional :: rc
+          integer :: status
+          type(ESMF_HConfig) :: hconfig
+          character(len=4) :: resolution
+          character(len=3) :: writers
+          character(len=128) :: yaml_content
+
+          write(resolution,'(I4)')jm_world
```
tclune, line 3997:
```suggestion
         _ASSERT(jm_world < 10**5, 'Format not wide enough')
          write(resolution,'(I5)')jm_world
```
We're just not that far from doubling C5760.

base/NCIO.F90
```diff
@@ -3975,6 +3983,26 @@ subroutine add_fvar(cf,vname,vtype,dims,units,long_name,rc)
...
+          write(writers,'(I3)')num_writers
```
tclune, line 3998:
```suggestion
          _ASSERT(num_writers < 10**3, 'Format not wide enough')
          write(writers,'(I3)')num_writers
```

base/NCIO.F90
```diff
@@ -3975,6 +3983,26 @@ subroutine add_fvar(cf,vname,vtype,dims,units,long_name,rc)
...
+       subroutine create_control_file(filename,jm_world,num_writers,rc)
+          character(len=*), intent(in) :: filename
+          integer, intent(in) :: jm_world
+          integer, intent(in) :: num_writers
+          integer, intent(out), optional :: rc
+          integer :: status
+          type(ESMF_HConfig) :: hconfig
+          character(len=4) :: resolution
+          character(len=3) :: writers
+          character(len=128) :: yaml_content
```
tclune, line 3995:
```suggestion
          character(len=:), allocatable :: yaml_content
```

tclune (changes requested):
Requested some minor changes for code safety.  (Fresh off an admonition about testing from last weeks workshop.)

bena-nasa:
@tclune @atrayano 
I finally after much struggle, a difficult to diagnose bug in the o-server that required tracking and fixing and probably making the code too ugly, have the generic split checkpoint writing with the o-server now too. As far as I'm concerned this seem to be working. I can start from the split checkpoints or a single checkpoint and I get to the same place at the end. It works both with no o-server or the o-server with both the mutligroup and "simple". I've also added a couple python utilities to split/recombine the checkpoints. Don't know if we want those, but I had them so figured might as well capture rather than just accidentally misplace the script.

So I guess we should merge (assuming CI tests pass) or not merge if I've made things too ugly and try to cleanup this module to an acceptable state

Probably then need to cleanup any script that was aware of the by_face logic and replace with new logic but that's not in this repo.

tclune:
If it works and you (@bena-nasa ) don't see any simple way to improve code quality, I say merge.   Safe to say that after MAPL3 settles down, a re-engineering effort for the o-server is (unfortunately) in order.  It is simply too important to leave in such a confusing state.

bena-nasa:
> If it works and you (@bena-nasa ) don't see any simple way to improve code quality, I say merge. Safe to say that after MAPL3 settles down, a re-engineering effort for the o-server is (unfortunately) in order. It is simply too important to leave in such a confusing state.

I don't think there's a "simple" or quick way to improve code quality, it's a question of how much time effort do we want to put in. It could be cleaned up, looking at the interfaces, seeing if anything can be simplified for all the write overloads, breaking/reorganizing this up into more logical programmatic structures where possible,  the o-server is just one piece of this code.

In some ways this is more complicated than history, there we only support output of a rather limited subset of the possible ESMF fields we have in the model. This has output every conceivable ESMF field. All with the unique gather/scatter algorithm used by nothing else in the code.

I looked like there are conflicts I need to resolve because develop changed but then I agree merge. And figure out where cleaning all this up lies in our priorities.

tclune (approved):

PR #2399, opened 2023-10-04
Fixes #2398

### LINKED ISSUE #2398: Missing return code checking when reading import restart
Just making an issue to reference in the PR. Turns out that for some reason, don't know if it was always there or a result of refactoring I assume Tom did ages ago. In any case, for the both the internal and import restart read in MAPL_GenericInitialize, there is some logic that says, if the call (the ESFMStateReadFromFile) fails print the state using our state print routine. But crucially in the internal state routine, in this block of code after the print it does a `_RETURN(ESMF_FAILURE)`. The similar procedure for the import state did not have this.

For various reasons this just lurked under the radar but recently when running a 3D-var DAS it was found that the job went wrong because reading the AGMC_Import_rst failed (so the increments don't get read) but the model kept running. Now the whole reason that call failed is because the 3 imports needed for GAAS going to ExtData were not marked with restart skip so the fact those aren't there triggered a failure becasue the DAS does NOT set MAPL_ENABLE_BOOTSTRAP: YES. So the call failed because the agcm_import_rst did not have those fields (nor should it!), but the model kept running along. Now we got the das back working by simply setting the restart skip in the AGC input file in GAAS, but the questioned remained, why did the model keep going.

Well the answer was blindingly obvious, if you don't return a failure the caller can't trap. I will submit a PR to fix, just a one liner...

Note this does not affect most das runs because they are all 4D var and the 4D var experiments does not rely on the restart reading in MAPL_Generic to ingest the increments. We checked and the only operational 3D var system is GEOS-IT and that was before the changes to GAAS came in, so no problem there.

bena-nasa:
Fixes #2398 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Adding missing check on the return status when reading import checkpoint in MAPL\_GenericInitialize
-
 ### Removed
 
 ### Deprecated
+
+## [2.41.1] - 2023-10-04
+
+### Fixed
+
+- Adding missing check on the return status when reading import checkpoint in MAPL\_GenericInitialize
 
 ## [2.41.0] - 2023-09-22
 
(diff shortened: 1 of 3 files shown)


atrayano (approved):

mathomp4 (approved):
Seems to run for me in a c24 run.

PR #2408, opened 2023-10-16
Feature/mapl3 geom mgr testing

tclune:
I only tested this with NAG on laptop as gfortran is currently broken in that environment.    Did not do anything especially complicated though, so ... maybe will pass CI without extra work on my part.


mathomp4 (approved):
Just the usual ADAS error.

PR #2419, opened 2023-10-26
Feature/bmauer/fixes #2418

### LINKED ISSUE #2418: Logic for when to use oserver in NCIO needs to be updated so replay doesn't crash when doing multi analysis replay
Just making an issue here for the PR

Some variations of replay work by writing the increments to file, in particular this happens when there are multiple analyses per predictor period. These files are then read in by agcm to so the increment can be applied. These files are written with MAPL_ESMFStateWriteToFile calls in GCM, the same routine used in MAPL_Generic. The problem is that in GCM no matter what is set, you do not want to write with the oserver as it is not appropriate since this is not a case where anything can be overlapped by computation, the corrector depends on these files.
However NCIO (which is ultimately what writes these) has logic that says, if write_by_oserver is set in the arrdes, then well write by oserver, of course whenever if is in a logical block if(write_by_oserver), the first this it does it asset that the client was passed, so it 's a bit redundant.

This is the problem with replay, gcm is not passing the oclient so the call fails if write_by_oserver is set.

 It seems like the logic should be, if the oclient is pass, use, if not, don't. It's up to the caller to pass it.
So I need to a bit of rearranging. 

This all came to a head because @mathomp4 found that for openMPI write_by_oserver is actually faster (even if you don't have separate oserver nodes), and since openMPI is what was are using on SCU17 for now

bena-nasa:
Fixes #2418 
This was needed to fix an issue @sdrabenh found with replay and the write_by_oserver, which is the default with openmpi

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Updated write_by_oserver logic so that the decision to write by the oserver is based on whether the output server client is passed in
 - Updated CI GEOSadas build to use special branch (as stock ADAS at the moment is too far behind GEOSgcm main)
 - Fix incorrect History print during runtime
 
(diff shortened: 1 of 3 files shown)


tclune (approved):
All those `if (have_oclients)` screams for a subclass.   But ... not now.

PR #2420, opened 2023-10-26
Automatic Code Generator Sample Code (0 diff)

JulesKouatchou:
## Description

- We provide a sample code that shows how to use the Automatic Code Generator tool. 
- The code reflects the example presented in the documentation.
- The code comes with a `spec` file and a `CMakeLists.txt` file needed for the tool to work properly.

## Motivation and Context

The sample code is meant to provide an example on the necessary changes users need to introduce in their application if they want to use the automatic code generator tool. 

## How Has This Been Tested?

Our main task was to only make sure that the code could compile by generating the necessary `__.h` files (four of them). The code is not used to create any executable.

## Types of changes

## Checklist:


docs/tutorial/grid_comps/code_generator/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+esma_set_this (OVERRIDE MAPL.mycomponent)
+
+set (srcs
+    MyComponent_GridComp.F90
+    )
+
```
tclune, line 6: The directory is not the `code_generator` so maybe come up with a good directory name and make the component agree with that?

"MyComponent" is minimally acceptable but maybe instead ACG_GridComp or some such?

tclune (changes requested):
Minor change requested.

mathomp4:
@JulesKouatchou Can you also include a CHANGELOG entry about this as well (along with @tclune 's suggestion?)

docs/tutorial/grid_comps/code_generator/CMakeLists.txt
JulesKouatchou: Because it is part of the tutorial sample codes, I propose as directory `sample_code_generation`. The name of the grid component will then be `SCG_GridComp`. I want to avoid `ACG` that is already the acronym of the tool. Any thought?

darianboggs: Given that `ACG` is a common abbreviation for the automatic code generator, I vote for `automatic_code_generator` for the directory, `ACG_GridComp.F90` for the GC, and `ACG_StateSpecs.rc` for the specs file.

tclune:
Not strongly opposed, but ... the directory is not the acg.  It is stuff that uses the acg.     Maybe obvious(ish) from context, but I'd prefer something that indicated that it does not in fact contain the ACG.

tclune:
Our usual convention is that gridcomps have the same name as the directory that they are in.

Long story, but the convention got botched during early deployment so we have dirs like:  `GEOSgcm_GridComp` that are supposed to be `GEOS_GcmGridComp`.   Going to Git was supposed to let us do that, but we've been afraid to pull the trigger.

docs/tutorial/grid_comps/CMakeLists.txt
```diff
@@ -4,4 +4,4 @@ add_subdirectory (leaf_comp_a)
 add_subdirectory (leaf_comp_b)
 add_subdirectory (parent_with_one_child)
 add_subdirectory (parent_with_two_children)
-add_subdirectory (code_generator)
+add_subdirectory (automatic_code_generator)
```
tclune, line 7:
```suggestion
add_subdirectory (automatic_code_generator_example)
```

JulesKouatchou:
- Changed the directory name (`automatic_code_generator`) and file names (`ACG_GridComp.F90` and `ACG_StateSpecs.rc`).
- Edited the `CHANGELOG.md` file
- Verified that the new sample code still compiles.

docs/tutorial/grid_comps/CMakeLists.txt
JulesKouatchou: @tclune I assume that you change the name from `automatic_code_generator` to `automatic_code_generator_example`. The documentation file also needs to be edited to reflect the name of the directory.

tclune: I did not change the name.   I made a suggestion in GitHub for the change.   You have apparently not committed that suggestion.  (Possibly you manually did something for that instead.)   But in either event, yes, I expected you to ensure it is all consistent.

JulesKouatchou: Changed the directory name and pushed the changes to the branch.

tclune (approved):

tclune (dismissed):

mathomp4:
I pushed a conflict fix up. I couldn't do it on the web.

tclune (approved):

PR #2434, opened 2023-10-30
Fixes #2433 (port to gfortran-13.2)

### LINKED ISSUE #2433: Internal compiler error with gfortran 13.2
Needs a workaround.   Compiler does not like array operation on an ASSOCIATE entity.  

Compiler bug, not MAPL bug.

tclune:
Note that unit tests still fail, but the code compiles.  Test failures might not be serious.


mathomp4:
Wow. commented out a lot of that file.

tclune:
oops - did not finish my process. Changing this to draft.

PR #2436, opened 2023-10-31
avoid collective call inside subroutine MAPL_GetGlobalHorzIJIndex

weiyuan-jiang:
It is expensive, inconvenient and unnecessary to have a collective call inside the subroutine  MAPL_GetGlobalHorzIJIndex
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Change the verification of the grid in MAPL_GetGlobalHorzIJIndex to avoid collective call
+
 ### Fixed
 
 ### Removed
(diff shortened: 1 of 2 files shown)


tclune (commented):
Have you run this with bounds checking?     The MAPL procedure that you have replaced returns the all the corners for the cells in the local domain, where the ESMF routines only return the left/south corners.    (Corners that are shared by 2 or more domains are only "owned" by one domain in ESMF.)

weiyuan-jiang:
> Have you run this with bounds checking? The MAPL procedure that you have replaced returns the all the corners for the cells in the local domain, where the ESMF routines only return the left/south corners. (Corners that are shared by 2 or more domains are only "owned" by one domain in ESMF.)

Yes. I only compare one left/south corner.  The other comparisons are on edge without end points.

tclune (approved):

PR #2445, opened 2023-11-16
WIP: Convert sends to ssends

mathomp4:
## Description

This is an odd test. While at SC23 I was reminded about all the odd sends MPI provides. So I thought, let's try ssends (aka synchronous sends). I then ran some C360 runs before and after and saw one interesting thing:

Current `develop`:
```
3721:--Finalize                                                1   132.238  16.03     0.017   0.00
3721:--Finalize                                                1   127.424  15.27     0.009   0.00
3722:--Finalize                                                1   112.208  13.75     0.006   0.00
```

Using Ssends:
```
3721:--Finalize                                                1   102.804  12.50     0.006   0.00
3721:--Finalize                                                1    99.120  11.93     0.009   0.00
3722:--Finalize                                                1    88.122  10.40     0.007   0.00
```

So...yeah. I mean, nothing else really changed, but the Finalize timer was consistently faster.

I'm wondering if one of the testers of @bena-nasa might be able to explore this? If there is one that can do checkpointing, maybe we could scale up and see if this is just an oddity of when I ran or something real?

## Types of changes

## Checklist:

### DIFF
--- a/Apps/time_ave_util.F90
+++ b/Apps/time_ave_util.F90
@@ -1463,7 +1463,7 @@ contains
             deallocate(buf)
          enddo
       else
-         call mpi_send(q,im*jm,MPI_FLOAT,peid0,mypet,MPI_COMM_WORLD,status)
+         call MPI_Ssend(q,im*jm,MPI_FLOAT,peid0,mypet,MPI_COMM_WORLD,status)
          _VERIFY(status)
       end if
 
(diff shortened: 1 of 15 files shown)


mathomp4:
Welp, new tests at C720 seems to show this is actually worse and slower. Which is sort of what I'd have expected since we are synchronizing! Closing!

PR #2465, opened 2023-12-01
Add type obs_platform to Trajectory Sampler 

metdyn:
## Description

Add to trajectory sampler the DEFINE_OBS_PLATFORM keyword (in HISTORY.rc) for reading multiple IODA files in a better format suggested by Arlindo.   To do this, we add union_platform function for observations.  Code adjustment is made to utilize new subroutines introduced in the swath sampler code in MAPL_ObsUtil.F90

```
Input:
  COLLECTIONS:
  geovals
  ::

  GRID_LABELS:
  ::

  geovals.CollectionType: IODA
  geovals.ObsPlatforms:   aircraft  airs_aqua
  geovals.template:       '%y4%m2%d2_%h2%n2z.nc4',
  geovals.format:         'CFIO',
  geovals.obs_file_interval:  '000000 060000'    #  fixed string format:  yymmdd hhmmss
  geovals.Epoch:          030000  # 060000                 #  fixed integer format:  hhmmss
  geovals.regrid_method:  'BILINEAR' ,
::


DEFINE_OBS_PLATFORM::

PLATFORM.aircraft::
  IODA_SCHEMA::
    index:     Location
    longitude: MetaData/longitude
    latitude:  MetaData/latitude
    time:      MetaData/dateTime
    file_name_template:  JEDI/ioda/Y2019/M08/D01/H00_06/aircraft.%y4%m2%d2T%h2%n2%S2Z.nc4
  ::
  GEOVALS_SCHEMA::
    geovals_fields::
       var2 ,  "Root" ,
    ::
  ::
::


PLATFORM.airs_aqua::
  IODA_SCHEMA::
    index:     Location
    longitude: MetaData/longitude
    latitude:  MetaData/latitude
    time:      MetaData/dateTime
    file_name_template: JEDI/ioda/Y2019/M08/D01/H00_06/airs_aqua.%y4%m2%d2T%h2%n2%S2Z.nc4
  ::
  GEOVALS_SCHEMA::
    geovals_fields::
       var2 ,  "Root"  ,
       var3 ,  "Root"  ,
    ::
  ::
::
```

Output:
aircraft.geovals.20190801_0000z.nc4  
airs_aqua.geovals.20190801_0000z.nc4
showing regridded values on var2d and var3d. 

## How Has This Been Tested?
Tested on Mac with ExtDataDriver. 

## Types of changes

## Checklist:


base/Plain_netCDF_Time.F90
```diff
@@ -511,3 +514,239 @@ subroutine convert_twostring_2_esmfinterval(symd, shms, interval, rc)
...
+       read (iunps, '(a100)', iostat = ios, err = 300) line
+       !       if (matchbgn (line, substring) ) return
+       !!write(6,*) 'line ', trim(line)
+       !!write(6,*) 'substring ', trim(substring)
+       if (matchbgn (trim(line), trim(substring)) ) return
+    enddo
+    return
+300 call error_nonstop ('scan_begin', &
+         'No '//trim(substring)//' block', abs (ios) )
+  end subroutine scan_begin
```
tclune, line 555: Don't make the change, but you can instead have `read()` return an error and then use a conditional on the error rather than branching.  

Is the error trapping handled correctly?  Cannot tell as this does not use our `_RC` mechanism.  Very important that all I/O does robust error trapping.

base/Plain_netCDF_Time.F90
```diff
@@ -511,3 +514,239 @@ subroutine convert_twostring_2_esmfinterval(symd, shms, interval, rc)
   end subroutine convert_twostring_2_esmfinterval
 
 end module Plain_NetCDF_Time
+
+
+module Fortran_read_file
```
tclune, line 519: This name is too generic.   

Also needs standard namespace protection in module name.   `mapl_`

base/Plain_netCDF_Time.F90
metdyn: Right, I agree the _RC mechanism is better.   There is not an error trapping here, I was trying to just to get it work.

tclune: ok - will approve for now and trust you'll circle back and fix that bit up later.

tclune (approved):

base/Plain_netCDF_Time.F90
metdyn: I changed the module name to MAPL_scan_pattern_in_file.

base/Plain_netCDF_Time.F90
metdyn: Actually the call name is mis-leading, it is not an error when reading the file reaches the end line.   I donot track the erorr because it ACTUALLY is not an error, but a note that the file reaches the end.

base/MAPL_ObsUtil.F90
```diff
@@ -375,33 +409,34 @@ function get_filename_from_template_use_index (obsfile_start_time, obsfile_inter
...
-          close(u)
-       else
-          ! exact file name
-          call fill_grads_template ( filename, file_template, &
-               experiment_id='', nymd=nymd, nhms=nhms, _RC )
-       end if
-
+    ! test on bash
+    !
+    cmd="bash -c 'ls "//trim(filename)//"' &> zzz_MAPL"
```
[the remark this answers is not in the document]
tclune: Probably wrap in a helper function as well.   Think about the functionality - then we can at least have a test and think about wrapping C instead.

base/Plain_netCDF_Time.F90
tclune: In that case, there are better options.   IOSTAT can be used instead of err, and you can check IOSTAT for EOF.   And stop/return if EOF but raise exception if somethinng else.

base/MAPL_ObsUtil.F90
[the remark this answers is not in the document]
tclune: @bena-nasa Could you look at this bit of code?   Is there already something in the ExtData utilities that might accomplish this?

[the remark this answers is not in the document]
metdyn: Thank you, @mathomp4  and @tclune, for getting to this point.  I had some discussion with @bena-nasa about this the week before and he can give more information.

The code tries to get file names from  MOD04_L2.A2017090.0100.*h5  to  MOD04_L2.A2017090.0155.*h5 at every 5 min interval, given that some files are missing:

bash -c "ls MOD04_L2.A2017090.0145.*h5"
ls: MOD04_L2.A2017090.0145.*h5: No such file or directory

bash -c "ls MOD04_L2.A2017090.0150.*h5"
MOD04_L2.A2017090.0150.051.NRT.h5

The * is needed because the filename itself sometimes contains file-processing time, which cannot be represented by time-template. 

In the source code, it is used in two places wrapped with MPI  irank=0 (mapl_am_I_root()), because otherwise this code crashes as multiple processes will write to the same file. 

- https://github.com/GEOS-ESM/MAPL/blob/feature/ygyu/mapl_obsmodule/base/MAPL_SwathGridFactory.F90#L538-L557

- https://github.com/GEOS-ESM/MAPL/blob/feature/ygyu/mapl_obsmodule/gridcomps/History/MAPL_HistoryTrajectoryMod_smod.F90#L546-L553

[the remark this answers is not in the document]
bena-nasa: @tclune @metdyn, no, I mean you are asking if there some sort of utility that simulations an "ls" with a wildcard... no, how would one even do that, you can only inquire/open given a file path, if you don't know the path, that's just not something Fortarn can do, you are asking an impossibility without file system commands.

I would say write a script that makes symlink with the untemplatable part stripped. So then no need to do a system call. That's what I would have done, given untemplable file paths for Fortran is unacceptable in my mind, that's what symlinks and a workflow if for before you run.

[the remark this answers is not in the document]
metdyn: OK, thank you, @bena-nasa.  I will delete this `ls ` system call in Fortran and  prevent a potential flaw in the code.  Thank you for the discussion.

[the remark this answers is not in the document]
bena-nasa: The more I think about, this "ls" needs go, maybe fine for development, but the fact is depending what's in the working directory when you do an "ls", it can be very slow. If you ever tried doing an ls in the temporary scratch that exists during DAS run you know what I mean. I'm just worried that this could be very bad for performance too depending on the circumstances.

So now I'm even more convinced this needs to be something addressed before execution as part of the workflow (i.e. make symlinks that have stripped the arbitrary part of the file names)

PR #2468, opened 2023-12-04
Feature/mapl3 port to ifort 2021.10.0

tclune:
Tested unit tests with 

- gfortran 12.3
- ifort 2021.10.0
- nag 7.1.40

on OS X.

Note that the pfio tests do not reliably pass on os x with any of these.    Not entirely reproducible from what I have seen.  Also some evidence that at least one ExtData test is not always reproducible.   These issues must be investigated as unreliable unit tests are worse than useless.


mathomp4:
I'm going to comment here to see if @tclune wants to keep this around. The conflict list is impressive enough I don't think I could solve it.

tclune:
I'm closing it for now.   Can always look for it again when I'm ready.

PR #2479, opened 2023-12-06
Hotfix/tclune/#2478 mapl abort

tclune:
Partially tested with laptop.   

(Very hard to do a unit test for this sort of thing.)


atrayano (approved):
Definitely a bug

PR #2490, opened 2023-12-11
fixes #2487

### LINKED ISSUE #2487: History Append Output is Broken on Develop
The nightly SCM tests have been going a bit wonky and I looked with @bena-nasa and it looks like the history append functionality is broken in `develop`. That is, if you run with `duration: 000000` it no longer appends, but writes every frequency.

Something most likely happened in History GridComp:

https://github.com/GEOS-ESM/MAPL/compare/main...develop#diff-0c9f15d29823f78a5dbd243a38a4d15479631b4143b114eba4bcbc095255b353

and most changes from @metdyn are protected by if blocks. But one:
```
         list(n)%currentFile = filename(n)
```
at line 3609 is not and maybe that is causing the issue?

I'll assign to @bena-nasa so he can take a look

bena-nasa:
Fixes #2487 
I think @metdyn was doing some work and committed a line he did not mean to which was resetting the current file every timestep which is should not.
This fixes it. I did confirm that the multi-step History as exercised by the SCM works once this is removed.

I didn't make this a hot fix because the offending code is not yet on main, so would not make sense to do a PR to main...

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed bug broken multi-step file output in History under certain template conditions
 - [#2433] Implemented workarounds for gfortran-13
 - Missing TARGET in GriddedIO - exposed runtime error when using NAG + debug.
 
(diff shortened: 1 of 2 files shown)


tclune (approved):

tclune:
> > This is a hotfix. Might be better to apply to main?
> 
> Like Ben said, the erroneous line does not exist in the main branch. This PR solves some of my puzzling on the output file names.

Ah - yes not hotfix then.

PR #2495, opened 2023-12-17
Feature/mapl3 dev/dec23

tclune:
Tested MAPL3 on OSX nag only.  But should be small cleanup changes - nothing risky.

### DIFF
--- a/generic3g/OuterMetaComponent.F90
+++ b/generic3g/OuterMetaComponent.F90
@@ -57,11 +57,11 @@ module mapl3g_OuterMetaComponent
       ! Hierarchy
       type(ChildComponentMap)                     :: children
       type(HierarchicalRegistry) :: registry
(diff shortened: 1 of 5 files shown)


mathomp4:
Intel seems to be throwing a build error:
```
/__w/MAPL/MAPL/geom_mgr/latlon/LatLonGeomFactory_smod.F90(279): error #6405: The same named entity from different modules and/or program units cannot be referenced.   [KEYWORDENFORCER]
------------^
/__w/MAPL/MAPL/geom_mgr/latlon/LatLonGeomFactory_smod.F90(276): error #6404: This name does not have a type, and must have an explicit type.   [UNUSABLE]
   module function make_file_metadata(this, geom_spec, unusable, chunksizes, rc) result(file_metadata)
-------------------------------------------------------^

```

But oddly GNU was fine. That is...interesting.

PR #2507, opened 2023-12-21
Memory Utility function

bena-nasa:
Santha mentioning that we have leaks, reminded me we made a little utility procedure that can be used anywhere that is linking the MAPL library. This was sitting a branch, but thought we should get this in cause seems like we have memory issues...

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_MemUtils.F90
+++ b/base/MAPL_MemUtils.F90
@@ -59,6 +59,7 @@ module MAPL_MemUtilsMod
   public MAPL_MemUtilsFree
   public MAPL_MemCommited
   public MAPL_MemUsed
+  public MAPL_MemReport
 
 #ifdef _CRAY
   public :: hplen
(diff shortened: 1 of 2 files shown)


base/MAPL_MemUtils.F90
```diff
@@ -767,4 +768,32 @@ subroutine get_unit ( iunit )
...
+    if (present(decorator)) then
+       extra_message = decorator
+    else
+       extra_message = ""
+    end if
+    call MAPL_MemUsed(mem_total,mem_used,percent_used)
+    call MAPL_MemCommited(committed_total,committed,percent_committed)
+    call MPI_Comm_Rank(comm,rank,status)
+    if (rank == 0) write(*,'("Mem report ",A20," ",A30," ",i7," ",f5.1,"% : ",f5.1,"% Mem Comm:Used")')trim(extra_message),file_name,line,percent_committed,percent_used
+
```
tclune, line 796: I'm going to open a separate issue to update this to use pFLogger.  This will allow us to optionally track on multiple processes and/or send the report elsewhere.

tclune (approved):

PR #2514, opened 2023-12-24
Feature/tclune/mapl3 dec23

tclune:
Major refactoring of MAPL3.

Primary accomplishment was the merger of classes UserComponent and ComponentHandler (previously ChildComponent).


tclune:
Have an updated version.   Maybe it accidentally fixes the CI issue that does not show up on my laptop.

PR #2516, opened 2024-01-03
MAPL3 Interface between ESMF_Field and udunits2

darianboggs:
MAPL3 Interface between ESMF_Field and udunits2

## Description

MAPL3 needs to convert values in compatible ESMF_Field instances. udunits2 will be used to convert values between ESMF_Field instances with compatible units, like meters and kilometers

## Related Issue

None

## Motivation and Context

See Description.

## How Has This Been Tested?

Since this is new functionality that nothing else depends on yet, I ran new tests specific to this feature and existing tests. Success.

## Types of changes

## Checklist:


mathomp4:
Huh. This is interesting. I mean udunits2 should be in the docker image! I'll take a look...

mathomp4:
Aww nuts. I see the issue. The Docker images need `texinfo` the build has been "silently" failing and we haven't seen the results of that until now.

A fix will need to wait until tomorrow when I can get on AWS from Goddard.

cmake/Findudunits.cmake
```diff
@@ -0,0 +1,57 @@
+# (C) Copyright 2022- UCAR.
+#
+# Try to find the udunits headers and library
+#
+# This module defines:
+#
+#   - udunits::udunits       - The udunits shared library and include directory, all in a single target.
```
tclune, line 7: This file should maybe go into ESMA_CMake.  Need to discuss with @mathomp4

field_utils/CMakeLists.txt
```diff
@@ -7,6 +7,11 @@ set(srcs
   FieldUtilities.F90
   FieldUnaryFunctions.F90
   FieldBinaryOperations.F90
+  FieldUnits.F90
+  udunits2.F90
+  udunits2interfaces.F90
+  udunits2encoding.F90
+  udunits2status.F90
```
tclune, line 14: I am not really sure where I want these to live.   But we can move after we get this change merged.

field_utils/MockField.F90
```diff
@@ -0,0 +1,170 @@
+module MockField_mod
+
+   implicit none
```
tclune, line 3: Is this file for testing?  If so, shouldn't it be in a test directory.

field_utils/MockField.F90
```diff
@@ -0,0 +1,170 @@
...
+   private
+
+   integer, parameter :: MAXLEN = 80
+   integer, parameter :: SUCCESS = 0
+   integer, parameter :: ERROR = SUCCESS - 1
+
+   ! Mock for ESMF_Field
+   type :: MockField
+      private
+      real(R64), allocatable :: f_(:, :)
```
tclune, line 16: I do not see where R64 comes from.  How can this even compile?

field_utils/udunits2encoding.F90
```diff
@@ -0,0 +1,14 @@
+module udunits2encoding
```
tclune, line 1: If these files belong in MAPL they need MAPL namespace guards.  `mapl_`

tclune (commented):
I need to delve deeper.  This is all I could do during another meeting.

field_utils/FieldUnits.F90
```diff
@@ -0,0 +1,80 @@
...
+! InitializeFieldUnits must be called first, and FinalizeFieldUnits must be called last.
+! InitializeFieldUnits and FinalizeFieldUnits are called once, before and after,
+! respectively, all GetFieldUnitsConverter and conv % convert calls.
+
+! For a given FieldUnitsConverter, GetFieldUnitsConverter and conv % convert
+! cannot be called before InitializeFieldUnits or after FinalizeFieldUnits
+! and conv % convert cannot be called before calling GetFieldUnitsConverter for conv.
+
+#include "MAPL_Generic.h"
+module FieldUnits
```
tclune, line 29:
```suggestion
module mapl_FieldUnits
```

darianboggs:
I added a C function (see `udunits2.c`) to disable error messages originating from udunits2 along with modifications to existing code (see `udunits2.F90`, `udunits2interfaces.F90`, and `CMakeLists.txt`) The C function still needs to be linked in. So, for now, the capability to disable udunits2 error messages is not compiled (via a preprocessor variable). Once the C function is linked in, the capability can be enabled and the preprocessor code eliminated.

cmake/Findudunits.cmake
darianboggs: Matt fixed.

field_utils/CMakeLists.txt
darianboggs: OK

field_utils/MockField.F90
darianboggs: fixed

darianboggs:
Pushed up changes incorporating conversations. Have not moved directory yet. Will consult @tclune and @mathomp4 later this morning to move.

darianboggs:
One remaining issue is that I had to add a C function to suppress excessive informative messages to STDERR from udunits. To make this work, I included an excerpt of `udunits2.h` in `field_utils`. It's not great, but it works. It might be possible to compile it to a small library.

field_utils/mapl_udunits2cfunc.c
```diff
@@ -0,0 +1,16 @@
...
+ * which disables error messages from udunits2
+ * udunits2 requires a ut_error_message_handler be passed
+ * into ut_set_error_message_handler to change the error handler,
+ * and ut_error_message_handler is a function with a variadic list
+ * of arguments, which is not possible in Fortran.
+*/
+ut_error_message_handler ut_set_ignore_error_message_handler()
+{
+    return ut_set_error_message_handler(ut_ignore);
+}
```
tclune, line 16: @darianboggs Not that there is a problem with including C code here, but ... why couldn't you just add an interoperable interface in `mapl_udunits2interfaces.F90` and then called it in the top level initialize?

tclune: Seems like what we would want to do for the external package.  It then allows a Fortran user to specify other error handling -not that we expect to do so.

field_utils/mapl_udunits2cfunc.c
```diff
@@ -0,0 +1,16 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include "udunits2.h"
+
+/* Helper function to augment udunits2 error handling
+ * Sets the udunits2 error handler to ut_ignore
+ * which disables error messages from udunits2
```
tclune, line 7: Q:   If this is just disabling error messages, do we instead want to pass a procedure that directs error messages to MAPL's error handling routine?

If it is also disabling "info" messages, then, yes, we want to shut it all down.

field_utils/mapl_udunits2cfunc.c
tclune: Oh  - I see now that `ut_ignore` is a func not an enum.   That might make the Fortran interface a bit difficult - outside of my ken ...   But probably worth exploring.

field_utils/mapl_udunits2status.F90
```diff
@@ -0,0 +1,28 @@
...
+
+   implicit none
+
+   enum, bind(c)
+       enumerator :: &
+       UT_SUCCESS = 0, & ! Success
+       UT_BAD_ARG, & ! An argument violates the function's contract
+       UT_EXISTS, & ! Unit, prefix, or identifier already exists
+       UT_NO_UNIT, & ! No such unit exists
+       UT_OS, & ! Operating-system error. See "errno".
```
tclune, line 13: Documenting here is of course quite useful for the user.

field_utils/mapl_udunits2status.F90
```diff
@@ -0,0 +1,28 @@
...
+! The values are the same as the udunits2 utStatus C enum
+module mapl_udunits2status
+
+   implicit none
+
+   enum, bind(c)
+       enumerator :: &
+       UT_SUCCESS = 0, & ! Success
+       UT_BAD_ARG, & ! An argument violates the function's contract
+       UT_EXISTS, & ! Unit, prefix, or identifier already exists
```
tclune, line 11: Q:   Are you certain that the C layer does not play any funny games with the values of the enums?   I always get nervous in this context because the "coincidence" in the ordering is fragile under change.

If the C values are defined with the preprocessor we could possible directly assign values.. E.g.,
``` 
UT_BAD_ARG = c_ut_bad_arg
```
Means that the Fortran name would need to be spelled differently (or at least capitalized differently.

tclune (approved):
I had a few questions I'd like addressed.  But will merge in the mean time.

mathomp4:
Also, @darianboggs, one thing that might be useful in a future PR (or wiki by @JulesKouatchou) is an example of how to use this functionality "outside" of MAPL, i.e., as a program. My guess is at some point, someone might like to write a program in something like GMAO_Shared which does unit conversion without full GEOS.

Or, I guess, if we move some of this into Baselibs at some point, then people can do it without MAPL. Udunits is so crazy to use that having a demo would be good!

field_utils/mapl_udunits2cfunc.c
darianboggs: It includes informational/warning messages. udunits sends messages to the effect of "Unit A is being used instead of Unit B" for several units to the error handler each time the units database, which currently is 1 time per run. There are also status codes reporting success, failure, or information. So it's not an error. To avoid the helper function, we could allow the messages since it only happens once per run. I don't like that from a design point of view, but practicality may win out here.

field_utils/mapl_udunits2cfunc.c
darianboggs: I agree that it would be better to use `mapl_udunits2interfaces.F90`.

Turning off the output requires 2 functions. `ut_set_error_message_handler` sets the error_message_handler, and `ut_ignore` is a `ut_error_message_handler` that is passed in. Unfortunately, `ut_error_message_handler` instances have variadic arguments, like `printf`. I don't think the Fortran interface to C allows functions with variadic arguments. The default `ut_error_message_handler` sends all messages to STDERR.

So while we want to use the `ut_ignore` handler, which ignores all messages and does nothing, I can't write an interface for `ut_ignore`. I thought about using a `c_funptr` in the interface for `ut_set_error_message_handler`, but I would still need to pass in `ut_ignore` as the `c_funptr`.

tclune: ok - thanks for the reminder.

darianboggs: abstract interface
    
    subroutine ut_set_error_message_handler(handler) bind(c, name="ut_set_error_message_handler")
      import :: c_funptr
      type(c_funptr), intent(in) :: handler
    end subroutine ut_set_error_message_handler

    subroutine ut_ignore(fmt, args) bind(c, name="ut_ignore")
      import :: ??? ! not sure what to use
      character(kind=c_char), intent(in) :: fmt(*)
      type(va_list), intent(in) :: args ! Can't do this
    end subroutine ut_ignore
    
    end interface

field_utils/mapl_udunits2interfaces.F90
```diff
@@ -0,0 +1,148 @@
...
+         type(c_ptr), value, intent(in) :: converter
+         real(c_float), intent(in) :: in_(*)
+         integer(c_int), value, intent(in) :: count_
+         real(c_float), intent(out) :: out_(count_)
+      end subroutine cv_convert_floats
+
+      ! Return type(c_ptr) to ut_unit
+      ! UT_SUCCESS indicates that the function ran successfully.
+      ! Other ut_status codes indicate cause of failure.
+      ! Use ut_get_status to check error condition.
```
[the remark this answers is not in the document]
darianboggs: True. Perhaps when we refine it.

field_utils/mapl_udunits2status.F90
darianboggs: Yes. In this case, I think that bringing over the comments is helpful.

field_utils/mapl_udunits2status.F90
darianboggs: I agree. I copied and pasted from `udunits2` header file(s?). From what I understand, Fortran enums are not really that useful. I thought about making these regular `integer(c_int)` constants.

The C values are in a C typedef enum `ut_status` in a udunits2 header file with `UT_STATUS = 0` and the rest following in the order I used.

With udunits2 in BASELIBS, we can't access the header files. The preprocessor could parse those header files for the `ut_status` values and generate constants, but we would need a copy of the header files included with MAPL, which means we would basically be back where we started, since we wouldn't receive updates for the values. Ugh!

field_utils/mapl_udunits2cfunc.c
tclune: No - I don't want any output to STDOUT if things are "good".

field_utils/mapl_udunits2status.F90
tclune: ok

field_utils/mapl_udunits2cfunc.c
darianboggs: Resolving for now to allow merge.

field_utils/mapl_udunits2cfunc.c
darianboggs: Resolving for now to allow merge.

darianboggs: I know this is merged.

For completeness: the error message handler writes 9 long lines to STDERR when it starts up. Each line reports an override like this:

> Definition of "au" in "/discover/swdev/gmao_SIteam/Baselibs/ESMA-Baselibs-7.17.0/x86_64-pc-linux-gnu/gfortran_12.1.0-openmpi_4.1.3-SLES12/Linux/share/udunits/udunits2-accepted.xml", line 123, overrides prefixed-unit "1.6605402e-45 kilogram"

So it is the normal course. We haven't made any changes to the units.

So udunits2 produces messages that are sent to the ut_error_message_handler, which defaults to STDERR, if the message is not an error. We can change the ut_error_message_handler to ut_ignore (or another handler), but it has to be done in C since the handler has variadic arguments which can't be interfaced from Fortran.

PR #2528, opened 2024-01-12
updates for geostitionary output part 1

bena-nasa:
@amdasilva  requested that History be able to output geostationary data. This can be described via rectilinear grid with some missing points. We can extend the XY grid factory for this.
For part1 this PR will assume the user has converted them to a form that the XY grid factory understands
I tested by obtaining the coordintes for GOES-16 here
https://www.star.nesdis.noaa.gov/atmospheric-composition-training/satellite_data_goes_imager_projection.php
and converting to the form needed by XY factory.

The new features are thus
1. The xy grid factory now tests if any points are missing, if so it adds a mask to the ESMF grid and masks at the appropriate points
2. A few new MAPL constants were defined to be the masking parameters
3. The MAPL wrapper to ESMF regridders was updated so that it checks if the destination grid has a mask. If so it adds the dstMaskValues argument to the restore. For now limited to linear regrid, could be extended to other methods

From here @metdyn can extend the grid factory to support the input format @amdasilva showed us that requires convert to raw input to lats and Lons on the sphere.

Finally here's a picture so show that this does work, the heights regridded to goes-16 view

Screenshot 2024-01-11 at 5 21 07 PM

## Types of changes

## Checklist:

### DIFF
--- a/base/MAPL_EsmfRegridder.F90
+++ b/base/MAPL_EsmfRegridder.F90
@@ -12,6 +12,7 @@ module MAPL_EsmfRegridderMod
    use MAPL_GridManagerMod
    use MAPL_BaseMod, only: MAPL_undef, MAPL_GridHasDE
    use MAPL_RegridderSpecRouteHandleMap
+   use MAPL_ConstantsMod
    implicit none
    private
 
@@ -138,7 +139,7 @@ contains
       if (HasDE) then
          call ESMF_FieldGet(dst_field,localDE=0,farrayPtr=p_dst,rc=status)
          _VERIFY(status)
-         p_dst = 0
+         p_dst = MAPL_UNDEF
       end if
 
       call this%do_regrid(src_field, dst_field, rc=status)
@@ -193,7 +194,7 @@ contains
       if (HasDE) then
          call ESMF_FieldGet(dst_field,localDE=0,farrayPtr=p_dst,rc=status)
          _VERIFY(status)
-         p_dst = 0
+         p_dst = MAPL_UNDEF
       end if
 
       call this%do_regrid(src_field, dst_field, rc=status)
@@ -245,7 +246,7 @@ contains
       if (HasDE) then
          call ESMF_FieldGet(dst_field,localDE=0,farrayPtr=p_dst,rc=status)
          _VERIFY(status)
-         p_dst = 0
+         p_dst = MAPL_UNDEF
       end if
 
       call this%do_regrid(src_field, dst_field, doTranspose=.true., rc=status)
@@ -328,7 +329,7 @@ contains
       if (HasDE) then
          call ESMF_FieldGet(dst_field,localDE=0,farrayPtr=p_dst,rc=status)
          _VERIFY(status)
-         p_dst = 0
+         p_dst = MAPL_UNDEF
       end if
 
       call this%do_regrid(src_field, dst_field, rc=status)
@@ -411,7 +412,7 @@ contains
       if (HasDE) then
          call ESMF_FieldGet(dst_field,localDE=0,farrayPtr=p_dst,rc=status)
          _VERIFY(status)
-         p_dst = 0
+         p_dst = MAPL_UNDEF
       end if
 
       call this%do_regrid(src_field, dst_field, rc=status)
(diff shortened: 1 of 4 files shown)


metdyn:
Thank you, @bena-nasa, for your work!   The code is working with the lambert_grid on the Mac and I am proceeding based on your PR.

base/MAPL_EsmfRegridder.F90
```diff
@@ -1493,16 +1496,25 @@ subroutine create_route_handle(this, kind, rc)
...
+                   & factorList=factorList, factorIndexList=factorIndexList, &
+                   & routehandle=route_handle, unmappedaction=unmappedaction, _RC)
+           else
+              call ESMF_FieldRegridStore(src_field, dst_field, &
+                   & regridmethod=ESMF_REGRIDMETHOD_BILINEAR, &
+                   & linetype=ESMF_LINETYPE_GREAT_CIRCLE, & ! closer to SJ Lin interpolation weights?
+                   & srcTermProcessing = srcTermProcessing, &
+                   & factorList=factorList, factorIndexList=factorIndexList, &
+                   & routehandle=route_handle, unmappedaction=unmappedaction, _RC)
+           end if
```
tclune, line 1514: I will not ask you to make the change, but there is a way to avoid the else/duplication here:
```suggestion
              integer, allocatable :: dstMaskValues(:)
              ...
              if (has_mask) dstMaskValues = [MAPL_MASK_OUT] ! otherwise unallocated
              
              call ESMF_FieldRegridStore(src_field, dst_field, &
                   & dstMaskValues = dstMaskValues, & ! unallocated treated as not PRESENT
                   & regridmethod=ESMF_REGRIDMETHOD_BILINEAR, &
                   & linetype=ESMF_LINETYPE_GREAT_CIRCLE, & ! closer to SJ Lin interpolation weights?
                   & srcTermProcessing = srcTermProcessing, &
                   & factorList=factorList, factorIndexList=factorIndexList, &
                   & routehandle=route_handle, unmappedaction=unmappedaction, _RC)
```

base/MAPL_XYGridFactory.F90
```diff
@@ -894,4 +896,36 @@ subroutine file_has_corners(this,rc)
...
+
+      integer(ESMF_KIND_I4), pointer :: mask(:,:)
+      real(ESMF_KIND_R8), pointer :: fptr(:,:)
+      integer :: i,j,status
+      type(ESMF_VM) :: vm
+      integer :: has_undef, local_has_undef
+
+      call ESMF_GridGetCoord(grid, coordDim=1, localDE=0, &
+         staggerloc=ESMF_STAGGERLOC_CENTER, &
+         farrayPtr=fptr, rc=status)
```
tclune, line 912: Untrapped return code.
```suggestion
         farrayPtr=fptr, _RC)
```

base/MAPL_XYGridFactory.F90
```diff
@@ -894,4 +896,36 @@ subroutine file_has_corners(this,rc)
...
+      local_has_undef = 0
+      if (any(fptr == MAPL_UNDEF)) local_has_undef = 1
+      call ESMF_VMGetCurrent(vm,_RC)
+      call ESMF_VMAllFullReduce(vm, [local_has_undef], has_undef, 1, ESMF_REDUCE_MAX, _RC) 
+      if (has_undef == 0) then
+         _RETURN(_SUCCESS)
+      end if
```
tclune, line 919: Common pattern so I made a macro for this.   There is also `_RETURN_UNLESS(cond)`
```suggestion
_RETURN_IF(has_undef == 0)    
```

tclune (changes requested):
Very nice implementation _and_ description.

A couple of small "suggested changes", and a bigger "something to think about".

tclune (approved):

base/MAPL_EsmfRegridder.F90
bena-nasa: @metdyn could clean this up on a subsequent PR. I'll just resolve for now.

tclune (approved):

base/MAPL_EsmfRegridder.F90
metdyn: Thanks, Toms' suggestion works instantly!   I have added this into the next PR.

PR #2548, opened 2024-01-19
Feature/tclune/mapl3 field spec

tclune:
Tested on laptop.

Builds with nag and gfortran 12.3
Unit tests run with nag, except pfio  and extdata ... sigh.


bena-nasa (approved):

tclune:
Forcing merge.  Passed before branch update so ...

PR #2552, opened 2024-01-23
Feature/tclune/mapl3 history

tclune:
Passes tests with NAG on OS X.


mathomp4 (approved):
Trust in @tclune and @darianboggs and the unit tests!

PR #2570, opened 2024-02-06
MAPL HConfig Get

darianboggs:
## Description

Draft implementation of MAPL_GetResource using hconfig and a new package for MAPL / HConfig interaction

## Motivation and Context

MAPL3 will use the ESMF_HConfig derived type for configurations instead of ESMF_HConfig.

## Types of changes

## Checklist:


darianboggs:
This feature is not complete, but it requires feedback on the final direction including interfacing with the MetaComp and pflogger. The private module tests pass. The public interface is a pass-through procedure. Additional MAPL_GetResource specific procedures need to be added for the MetaComp and for array values.

CMakeLists.txt
```diff
@@ -249,6 +249,7 @@ endif()
 
 add_subdirectory (geom_mgr)
 add_subdirectory (regridder_mgr)
+add_subdirectory (hconfig)
```
tclune, line 252: `hconfig_utils` might be better.   Not worth a lot of back and forth though.

generic3g/MAPL_Generic.F90
```diff
@@ -575,26 +598,85 @@ subroutine gridcomp_connect_all(gridcomp, src_comp, dst_comp, rc)
...
+      type(ESMF_GridComp), intent(inout) :: gridcomp
+      type(ESMF_HConfig), intent(out) :: hconfig
       integer, optional, intent(out) :: rc
 
       integer :: status
-      logical :: has_key
+      type(ESMF_Config) :: config
+      
+      call ESMF_GridCompGet(gridcomp, config=config, _RC)
+      call ESMF_ConfigGet(config, hconfig=hconfig, _RC)
```
tclune, line 610: This utility will force me (us) to resolve a problem in the existing implementation.   OuterMeta also has-a hconfig object.   I'm not yet confident that it is the same reference as one obtains with the calls you provided.    We'll discuss - no change for now.
```suggestion
      call ESMF_GridCompGet(gridcomp, config=config, _RC)
      call ESMF_ConfigGet(config, hconfig=hconfig, _RC)
```

generic3g/MAPL_Generic.F90
```diff
@@ -575,26 +598,85 @@ subroutine gridcomp_connect_all(gridcomp, src_comp, dst_comp, rc)
...
       
-      has_key = ESMF_HConfigIsDefined(hconfig, keystring=keystring, _RC)
-      if (has_key) then
-         value = ESMF_HConfigAsSTring(hconfig, keystring=keystring, _RC)
-         _RETURN(_SUCCESS)
+
+      _RETURN(_SUCCESS)
+   end subroutine gridcomp_get_hconfig
+
+   subroutine mapl_resource_get_scalar(hconfig, keystring, value, unusable, default, logger, is_default, found, rc)
```
tclune, line 616: I thought this procedure would go in the new directory you created.  Please explain your thinking.

hconfig/mapl3hconfig_get.F90
```diff
@@ -0,0 +1,39 @@
...
+      character(len=*), intent(in) :: keystring
+      class(*), intent(inout) :: value
+      character(len=:), allocatable, intent(inout) :: message
+      class(KeywordEnforcer), optional, intent(in) :: unusable
+      logical, optional, intent(out) :: value_is_set
+      integer, optional, intent(out) :: rc
+      logical :: found
+
+      _UNUSED_DUMMY(unusable)
+
```
tclune, line 31: Weiyuan suggests that these macros be moved to bottom of procedures.  I agree.

hconfig/mapl3hconfig_get.F90
```diff
@@ -0,0 +1,39 @@
...
+      call get_value(hconfig, value, found, message, keystring, _RC)
+      if(present(value_is_set)) value_is_set = found
```
tclune, line 33: If `get_value()` has `found` as an optional argument, then you could simplify the above as:
```suggestion
      call get_value(hconfig, value, found=value_is_set, message, keystring, _RC)
```
Which then begs the need for this particular layer of the code.  I'll look at `get_value()` next, so maybe this comment is premature.

hconfig/mapl3hconfig_get_private.F90
```diff
@@ -0,0 +1,107 @@
...
+      character(len=:), allocatable :: valuestring
+
+      integer :: status
+      integer :: ios
+      character(len=MAXSTRLEN) :: rawstring
+
+      found = ESMF_HConfigIsDefined(hconfig, keystring=keystring, _RC)
+      if(.not. found) then
+         _RETURN(_SUCCESS)
+      end if
```
tclune, line 39: This is SUCCESS?

hconfig/mapl3hconfig_get_private.F90
```diff
@@ -0,0 +1,107 @@
...
+
+      select type(value)
+      type is (integer(kind=ESMF_KIND_I4))
+         typestring = 'I4'
+         value = ESMF_HConfigAsI4(hconfig, keyString=keystring, _RC)
+         write(rawstring, fmt='(I12)', iostat=ios) value
```
tclune, line 45: You don't need to do this for rawstring.  You can just do a separate `HConfigAsString` for those.  Then you can ignore concerns about formatting.

hconfig/mapl3hconfig_get_private.F90
```diff
@@ -0,0 +1,107 @@
...
+         value = ESMF_HConfigAsR4(hconfig, keyString=keystring, _RC)
+         write(rawstring, fmt='(G17.8)', iostat=ios) value
+      type is (real(kind=ESMF_KIND_R8))
+         typestring = 'R8'
+         value = ESMF_HConfigAsR8(hconfig, keyString=keystring, _RC)
+         write(rawstring, fmt='(G24.16)', iostat=ios) value
+      type is (logical)
+         typestring = 'L'
+         value = ESMF_HConfigAsLogical(hconfig, keyString=keystring, _RC)
+         write(rawstring, fmt='(L1)', iostat=ios) value
```
tclune, line 61: This is the one case you might want to keep the rawstring as there are many ways to spell `true` in yaml.

hconfig/mapl3hconfig_get_private.F90
```diff
@@ -0,0 +1,107 @@
...
+      character(len=*), intent(in) :: keystring
+      character(len=*), intent(in) :: valuestring
+      integer, intent(in) :: valuerank
+      character(len=*), parameter :: J_ = ', '
+
+      if(valuerank > 0) then
+         message = typestring //J_// keystring //J_// valuestring //J_// rankstring(valuerank)
+      else
+         message = typestring //J_// keystring //J_// valuestring
+      end if
```
tclune, line 91:
```suggestion
      message = typestring //J_// keystring //J_// valuestring
      if(valuerank > 0) then
         message = message//J_// rankstring(valuerank)
      end if
```
Never nester.

hconfig/mapl3hconfig_get_private.F90
```diff
@@ -0,0 +1,107 @@
...
+   function rankstring(valuerank) result(string)
+      character(len=:), allocatable :: string
+      integer, intent(in) :: valuerank
+
+      if(valuerank > 0) then
+         string = '(:' // repeat(',:', valuerank-1) // ')'
+      else
+         string = ''
+      end if
+
```
tclune, line 104:
```suggestion
     string=''
      if(valuerank > 0) then
         string = '(:' // repeat(',:', valuerank-1) // ')'
      end if
```
Never nester.

tclune (changes requested):
Made my first pass.  A few things that should be discussed soon.

generic3g/MAPL_Generic.F90
darianboggs: My plan was to separate the `ESMF_HConfig` handling from the logger and default handling. `hconfig_utils` uses parts of `ESMF`. `mapl3hconfig_get` (public) uses `ESMF_HConfig`. `mapl3hconfig_get_private` uses `ESMF_HConfig`, some `ESMF_HConfig` procedures, and ESMF kinds. It passes up message strings up instead of directly accessing the logger. Since there is no logger associated with the hconfig, I decided not to pass it in, because the calling procedure would have to get the logger to pass in. Similarly, there is no default value for a keystring in the hconfig, so I decided to let the calling procedure handle the default. I thought about throwing an exception **if** a value for the keystring is not found **and** no default is found, but I decided that it wasn't a true error condition. So, I added a flag to signal that the value was not found, and let the calling procedure use that flag to handle that case.

geom_mgr/tests/Test_HConfigUtils.pf
```diff
@@ -0,0 +1,111 @@
+module Test_HConfigUtils
+   use funit
+   use ESMF
```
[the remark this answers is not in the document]
darianboggs: Agreed.

hconfig/mapl3hconfig_get.F90
darianboggs: OK. I have no opinion. That's fine.

hconfig/mapl3hconfig_get_private.F90
darianboggs: Yes. It is, in essence, checking a boolean condition, and if true, returning a value.   **Checking** a boolean condition has three outcomes: true, false, error ("bottom"). I wanted to distinguish between: (1) the value is false, and (2) there was an error checking the condition. Our standard exception produces strings, stacktraces, and integers. So, determining the error requires parsing the message string or checking the integer value of the return code. Parsing the message to drive program logic is rather dangerous, and beyond 0 is success, we don't have code that looks up error conditions based on the integer return code. (Let me know if there is such code.)

If `default` (currently in the calling procedure) were mandatory, it would be impossible to not set the value, unless there were some other error condition. In that case, the flag that indicates if the value returned equals the default, would be mandatory. That flag is optional right now, and it is used for logging since we want to indicate if the value equals the default. I wanted to push that logic higher.

At the higher level, there is an `_FAIL` condition if: (1) the value is not found, (2) there is no default value, **and** (3) there is no `found` flag. (It's an `_ASSERT` statement.)

I don't like a mandatory default value because that invites passing in a default that may not be useful, just to pass in a default. The optional arguments should make it possible to handle the condition that the keystring is not found.

We could be stricter and say that the default is mandatory, and that the user needs to be careful to check that the value returned is a sensible value. Because the HConfig values are run-time, it's hard to safeguard against a user failing to include a keystring a file, especially since some users don't want a runtime error in that case.

hconfig/mapl3hconfig_get_private.F90
darianboggs: I did that originally, but then I tried to parse the string to produce the log message format we wanted:

    TY KEY VALUE[(RANK)
    
where TY is "CH", "L", "R4", "R8", "I4", or "I8" (The format can be tweaked as needed.) Since the string is coming from ESMF, I didn't see a way to control the format. So, I used the Fortran `write` to control the format of the value. The message is a comma-separated string for now until I have the logger part in place.

hconfig/mapl3hconfig_get_private.F90
darianboggs: See the comment above.

hconfig/mapl3hconfig_get_private.F90
darianboggs: It weighed to (1) allocate once with an `if/else/end if` block or (2) (possibly) allocate twice with a simple `if` statement. I guess the allocate is not an issue. I suppose the compiler can make it efficient.

hconfig/mapl3hconfig_get_private.F90
darianboggs: I got rid of `if/else/end if` with an `_ASSERT`. There is never any valid reason to call the function `rankstring` with rank < 1.

hconfig/mapl3hconfig_get.F90
darianboggs: Done.

generic3g/MAPL_Generic.F90
```diff
@@ -143,10 +153,23 @@ module mapl3g_Generic
       procedure :: gridcomp_connect_all
    end interface MAPL_ConnectAll
 
-
+   ! MAPL_ResourceGet
+   !  This will have 4 specific procedures:
+   !     scalar value from hconfig
+   !     scalar value from metacomp
```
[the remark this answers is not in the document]
darianboggs: We may need 2 cases, only: scalar and array/sequence. Both would be for _gridcomp_. If we want to be able to call `MAPL_ResourceGet` for **both** _gridcomp_ **and** _metacomp_, it would double, obviously. If we want to be able call `MAPL_ResourceGet` for _hconfig_, as well, there would be 2 more. I don't think we really need the version for the _hconfig_ itself, since this is in `MAPL_Generic.F90` in `generic3g`. The _hconfig_ is not tied to anything else in MAPL (except exception handling). _hconfig_utils_ has a public procedure that can be used if we want to get an _hconfig_ value elsewhere. So my current plan is to fold the _hconfig_ version into the _gridcomp_ version. We should talk.

CMakeLists.txt
darianboggs: Done

generic3g/MAPL_Generic.F90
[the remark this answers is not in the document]
darianboggs: Now there are two specific procedures for `MAPL_ResourceGet`:

1. with gridcomp (calls 2.)
2. with hconfig

The hconfig is accessible with `MAPL_GridCompGet`, so an intermediate procedure is not required for the gridcomp. However, if there is a use case for a standalone metacomp procedure, we can add that. Of course this will double for arrays/sequences.

hconfig_utils/tests/Test_mapl3hconfig_get_private.pf
```diff
@@ -0,0 +1,248 @@
...
+      class(*), intent(in) :: actual, expected
+      character(len=*), optional, intent(in) :: prolog, bridge, epilog
+      character(len=:), allocatable :: actual_string, expected_string
+      character(len=:), allocatable :: prolog_, epilog_, bridge_
+
+      if(present(prolog)) then
+         prolog_ = trim(adjustl(prolog)) // SPACE
+      else
+         prolog_ = ''
+      end if
```
tclune, line 194:
```suggestion
      prolog_ = ''
      if (present(prolog)) then
         prolog_ = trim(adjustl(prolog)) // SPACE
      end if
```

tclune: And repeat.

hconfig_utils/hconfig_macros.h
```diff
@@ -0,0 +1,11 @@
+#if !defined MTYPE
+#define MTYPE VTYPE
+#endif
```
tclune, line 3: Please indent in CPP blocks.   CPP is confusing enough as it is.
```suggestion
#if !defined MTYPE
#   define MTYPE VTYPE
#endif
```

hconfig_utils/hconfig_r4.F90
```diff
@@ -0,0 +1,10 @@
+module hconfig_r4
```
tclune, line 1: I am thinking that we need a mapl prefix for even this semi-hidden modules.   But feel free to explain a rationale for otherwise.

hconfig_utils/mapl3hconfig_get.F90
```diff
@@ -0,0 +1,10 @@
+module mapl3hconfig_get
```
tclune, line 1: The prefix being used for mapl3 is `mapl3g_`
```suggestion
module mapl3g_hconfig_get
```

darianboggs:
All cases work (I4, I8, R4, R8, Logical, and String: scalar and array). except array of strings. Array of strings work differently with `ESMF_HConfigAsStringSeq`, which requires an allocatable return value **and** an extra `stringlen` parameter. (ESMF documentation gives an explantion.) I worked on a fix, and I am testing it now.

tclune:
Arrays of strings are inherently klunky.

1. Are we certain we have a use case for this?
2. Would we be better served with a StringVector?   (Manually unpack from ESMF one string at a time.)

hconfig_utils/tests/Test_mapl3hconfig_get_private.pf
darianboggs: will do

hconfig_utils/hconfig_r4.F90
darianboggs: will do

hconfig_utils/mapl3hconfig_get.F90
darianboggs: will do

darianboggs:
My fix for string arrays did not work. I am going to remove that case. I'll implement later as needed. (See above for details.)

geom_mgr/latlon/LatAxis_smod.F90
```diff
@@ -77,7 +77,7 @@ module function make_LatAxis_from_hconfig(hconfig, rc) result(axis)
       real(kind=R8), allocatable :: centers(:), corners(:)
       type(AxisRanges) :: ranges
 
-      call MAPL_GetResource(jm_world, hconfig, 'jm_world', _RC)
+      call MAPL_HConfigGet(hconfig, 'jm_world', jm_world, _RC)
```
tclune, line 80:
```suggestion
      call MAPL_HConfigGet(hconfig, 'jm_world', jm_world, found=found, _RC)
      _ASSERT(found, 'jm_world not found')
```

darianboggs:
Comments more explicit, variables renamed/reordered, tests more explicit.
handle_default_i4 only sets the value to default if compare_only is .FALSE. When it is called by set_value_i4, compare_only=this%found (value was found in HConfig). So if this%found, value is set already, and it can be safely compared, and the value is not changed; otherwise value is set to default.

darianboggs:
I have updated my local branch with release/MAPL-v3. So I will not update the repository version.

tclune:
You can just delete the branch here and then push again.  Not sure what that does to the actual PR though.

hconfig_utils/mapl3g_hconfig_get_private.F90
```diff
@@ -0,0 +1,226 @@
...
+      class(Logger_t), optional, target, intent(inout) :: logger
+      integer, optional, intent(out) :: rc
+      integer :: status = _FAILURE
+      type(HConfigGetter) :: getter
+      type(logger_t), pointer :: logger_
+      logical :: found = .FALSE.
+
+      if(present(valueset)) valueset = .FALSE.
+      if(.not. present(valueset)) status = _FAILURE
+      if(present(logger)) logger_ => logger
```
tclune, line 35:
```suggestion
      logger_ => null()
      if(present(logger)) logger_ => logger
```

hconfig_utils/mapl3g_hconfig_get_private.F90
```diff
@@ -0,0 +1,226 @@
...
+
+      found = ESMF_HConfigIsDefined(hconfig, keyString=label, _RC)
+      getter = HConfigGetter(hconfig, label, found)
+      if(.not. (found .or. present(default))) then
+         if(present(rc)) rc = status
+         return
+      end if
```
tclune, line 42:
```suggestion
     _RETURN_UNLESS(found .or. present(default)))
```

hconfig_utils/mapl3g_hconfig_get_private.F90
```diff
@@ -0,0 +1,226 @@
...
+
+      select type(value)
+      type is (integer(ESMF_KIND_I4))
+         call get_value(getter, value, default, _RC)
+      class default
+         _FAIL('Unsupported type provided for label <'//getter%label//'>')
```
tclune, line 48:
```suggestion
         _FAIL('Unsupported type requested for label <'//getter%label//'>')
```

hconfig_utils/mapl3g_hconfig_getter.F90
```diff
@@ -0,0 +1,235 @@
...
+   type(HConfigGetter) function construct(hconfig, label, found)
+      type(ESMF_HConfig), intent(in) :: hconfig
+      character(len=*), intent(in) :: label
+      logical, intent(in) :: found
+
+      construct%hconfig=hconfig
+      construct%label=label
+      construct%found=found
+      construct%typestring=EMPTY_STRING
+      construct%valuestring=EMPTY_STRING
```
tclune, line 43: please put space on either side of assignment `=`.  (But not when `=` is used for keywords in procedure references.)

hconfig_utils/mapl3g_hconfig_getter.F90
```diff
@@ -0,0 +1,235 @@
...
+      integer(kind=ESMF_KIND_I4), intent(out) :: value !macro VALTYPE
+      character(len=*), parameter :: fmt_ = DEFAULT_FORMAT_STRING !macro FMTSTR
+      integer(kind=ESMF_KIND_I4) :: default_ !macro VALTYPE
+      type(HConfigGetter), intent(inout) :: getter
+      class(*), optional, intent(in) :: default
+      integer, optional,intent(out) :: rc
+      integer :: status = 0
+      character(len=MAXSTRLEN) :: buffer
+
+      getter%typestring = 'TYPESTRING_'
```
tclune, line 57: I think there is a problem here.   Macros don't expand inside of quotes - at least on some compilers.   But I could be wrong.  Usually this is done by making the macro have the quotes:
```
#define TYPESTRING_ 'I4'
```

tclune: But this is not even parameterized code, so ... not sure what is supposed to be here.

hconfig_utils/tests/Test_mapl3g_hconfig_get_private.pf
```diff
@@ -0,0 +1,225 @@
+module Test_mapl3g_hconfig_get_private
```
tclune, line 1: No need to repeat the namespace protection here.
```suggestion
module Test_hconfig_get_private
```

hconfig_utils/tests/Test_mapl3g_hconfig_getter.pf
```diff
@@ -0,0 +1,110 @@
+module Test_mapl3g_hconfig_getter
```
tclune, line 1: Delete mapl3g here.

tclune (changes requested):
See inline comments.

hconfig_utils/mapl3g_hconfig_get_private.F90
darianboggs: Done

hconfig_utils/mapl3g_hconfig_get_private.F90
darianboggs: Done

hconfig_utils/mapl3g_hconfig_getter.F90
darianboggs: Fixed

hconfig_utils/tests/Test_mapl3g_hconfig_getter.pf
darianboggs: Done

hconfig_utils/mapl3g_hconfig_getter.F90
darianboggs: Done

hconfig_utils/tests/Test_mapl3g_hconfig_get_private.pf
darianboggs: Done

darianboggs:
I eliminated the separate specific subroutines for the `MAPL_ResourceGet` interface in `MAPL_Generic` (3g) for hconfig (without a grid component). Now `MAPL_ResourceGet` requires a grid comp. You can access `hconfig_utils` without a grid component by first creating a `HConfigParams` instance with an `ESMF_HConfig` instance and a label (keystring), and then calling `MAPL_HConfigGet`. A new specific subroutine using a `ESMF_HConfig` and label for each value type can be created later for `MAPL_HConfigGet` later as needed.

generic3g/MAPL_Generic.F90
```diff
@@ -40,15 +40,17 @@ module mapl3g_Generic
...
    use :: esmf, only: ESMF_SUCCESS
    use :: esmf, only: ESMF_Method_Flag
    use :: esmf, only: ESMF_STAGGERLOC_INVALID
    use :: esmf, only: ESMF_StateIntent_Flag
    use :: esmf, only: ESMF_STATEINTENT_IMPORT, ESMF_STATEINTENT_EXPORT, ESMF_STATEINTENT_INTERNAL
    use :: esmf, only: ESMF_TypeKind_Flag, ESMF_TYPEKIND_R4
+   use :: esmf, only: ESMF_KIND_I4, ESMF_KIND_I8, ESMF_KIND_R4, ESMF_KIND_R8
    use :: esmf, only: ESMF_StateItem_Flag, ESMF_STATEITEM_FIELD, ESMF_STATEITEM_FIELDBUNDLE
    use :: esmf, only: ESMF_STATEITEM_STATE, ESMF_STATEITEM_UNKNOWN
+!   use hconfig3g
```
tclune, line 52:
```suggestion
```

generic3g/MAPL_Generic.F90
```diff
@@ -40,15 +40,17 @@ module mapl3g_Generic
...
+   use mapl3g_hconfig_get
```
tclune, line 53: Why isn't this `mapl3g_hconfig_utils`?

generic3g/MAPL_Generic.F90
```diff
@@ -143,10 +145,20 @@ module mapl3g_Generic
...
+      module procedure :: mapl_resource_get_i8_gc
+      module procedure :: mapl_resource_get_r4_gc
+      module procedure :: mapl_resource_get_r8_gc
+      module procedure :: mapl_resource_get_logical_gc
+      module procedure :: mapl_resource_get_i4seq_gc
+      module procedure :: mapl_resource_get_i8seq_gc
+      module procedure :: mapl_resource_get_r4seq_gc
+      module procedure :: mapl_resource_get_r8seq_gc
+      module procedure :: mapl_resource_get_logical_seq_gc
+      module procedure :: mapl_resource_get_string_gc
```
tclune, line 159: Latest Fortran versions dropped the need for `module` in this context.   And ... we don't need  the `mapl_` prefix on procedure names except for the public interface.
```suggestion
      procedure :: resource_get_i4_gc
      procedure :: resource_get_i8_gc
      procedure :: resource_get_r4_gc
      procedure :: resource_get_r8_gc
      procedure :: resource_get_logical_gc
      procedure :: resource_get_i4seq_gc
      procedure :: resource_get_i8seq_gc
      procedure :: resource_get_r4seq_gc
      procedure :: resource_get_r8seq_gc
      procedure :: resource_get_logical_seq_gc
      procedure :: resource_get_string_gc
```

generic3g/MAPL_Generic.F90
```diff
@@ -574,26 +586,271 @@ subroutine gridcomp_connect_all(gridcomp, src_comp, dst_comp, rc)
...
-         value = ESMF_HConfigAsSTring(hconfig, keystring=keystring, _RC)
-         _RETURN(_SUCCESS)
-      end if
+      call ESMF_GridCompGet(gridcomp, config=config, _RC)
+      call ESMF_ConfigGet(config, hconfig=hconfig, _RC)
+
+      _RETURN(_SUCCESS)
+   end subroutine gridcomp_get_hconfig
+
+   subroutine mapl_resource_get_i4_gc(gc, keystring, value, unusable, default, value_set, rc)
```
tclune, line 603:
```suggestion
   subroutine resource_get_i4_gc(gc, keystring, value, unusable, default, value_set, rc)
```
I'll leave it to you to propagate to the other procedures.

generic3g/MAPL_Generic.F90
```diff
@@ -574,26 +586,271 @@ subroutine gridcomp_connect_all(gridcomp, src_comp, dst_comp, rc)
...
+      logical, optional, intent(out) :: value_set
+      integer, optional, intent(out) :: rc
+      class(Logger_t), pointer :: logger
+      type(ESMF_HConfig) :: hconfig
+      type(HConfigParams) :: params
+      integer :: status
+
+      call MAPL_GridCompGet(gc, hconfig=hconfig, logger=logger, _RC)
+      params = HConfigParams(hconfig, keystring, value_set, logger)
+      call MAPL_HConfigGet(params, value, default, _RC)
```
tclune, line 641: This is not what we discussed the other day.   HConfigGet should take an hconfig object.

geom_mgr/latlon/LatAxis_smod.F90
```diff
@@ -136,7 +141,8 @@ module function get_lat_range(hconfig, jm_world, rc) result(ranges)
       _ASSERT(has_range .neqv. has_pole, 'Exactly one of lon_range or pole must be defined in hconfig')
 
       if (has_range) then ! is_regional
-         call MAPL_GetResource(t_range, hconfig, 'lat_range', _RC)
+         t_range = ESMF_HConfigAsR4Seq(hconfig, keyString='lat_range', _RC)
+!         call MAPL_HConfigGet(hconfig, 'lat_range', t_range, _RC)
```
tclune, line 145:
```suggestion
```

geom_mgr/latlon/LatLonGeomSpec_smod.F90
```diff
@@ -3,7 +3,7 @@
 submodule (mapl3g_LatLonGeomSpec) LatLonGeomSpec_smod
    use mapl3g_CoordinateAxis
    use mapl3g_GeomSpec
-   use mapl3g_HConfigUtils
+!   use hconfig3g
```
tclune, line 6:
```suggestion
```

geom_mgr/latlon/LatLonGeomSpec_smod.F90
```diff
@@ -78,8 +78,10 @@ function make_decomposition(hconfig, dims, rc) result(decomp)
       _ASSERT(has_ims .eqv. has_jms, 'ims and jms must be both defined or both undefined')
 
       if (has_ims) then
-         call MAPL_GetResource(ims, hconfig, 'ims', _RC)
-         call MAPL_GetResource(jms, hconfig, 'jms', _RC)
+         ims = ESMF_HConfigAsI4Seq(hconfig, keyString='ims', _RC)
+         jms = ESMF_HConfigAsI4Seq(hconfig, keyString='jms', _RC)
+!         call MAPL_HConfigGet(hconfig, 'ims', ims,  _RC)
+!         call MAPL_HConfigGet(hconfig, 'jms', jms, _RC)
```
tclune, line 84:
```suggestion
```

geom_mgr/latlon/LatLonGeomSpec_smod.F90
```diff
@@ -89,8 +91,10 @@ function make_decomposition(hconfig, dims, rc) result(decomp)
       _ASSERT(has_nx .eqv. has_ny, 'nx and ny must be both defined or both undefined')
 
       if (has_nx) then
-         call MAPL_GetResource(nx, hconfig, 'nx', _RC)
-         call MAPL_GetResource(ny, hconfig, 'ny', _RC)
+         nx = ESMF_HConfigAsI4(hconfig, keyString='nx', _RC)
+         ny = ESMF_HConfigAsI4(hconfig, keyString='ny', _RC)
+!         call MAPL_HConfigGet(hconfig, 'nx', nx, _RC)
+!         call MAPL_HConfigGet(hconfig, 'ny', ny, _RC)
```
tclune, line 97:
```suggestion
```

geom_mgr/latlon/LonAxis_smod.F90
```diff
@@ -2,8 +2,9 @@
 
 submodule (mapl3g_LonAxis) LonAxis_smod
    use mapl_RangeMod
-   use mapl3g_HConfigUtils
    use mapl_ErrorHandling
+!   use hconfig3g
```
tclune, line 6:
```suggestion
```

geom_mgr/latlon/LonAxis_smod.F90
```diff
@@ -121,11 +127,14 @@ logical module function supports_hconfig(hconfig, rc) result(supports)
...
 
-      has_im_world = ESMF_HConfigIsDefined(hconfig, keystring='im_world', _RC)
+      has_im_world = ESMF_HConfigIsDefined(hconfig, keystring='lon_range', _RC)
+!      has_im_world = MAPL_HConfigKeystringFound(hconfig, keystring='im_world', _RC)
       _RETURN_UNLESS(has_im_world)
 
       has_lon_range = ESMF_HConfigIsDefined(hconfig, keystring='lon_range', _RC)
       has_dateline = ESMF_HConfigIsDefined(hconfig, keystring='dateline', _RC)
+!      has_lon_range = MAPL_HConfigKeystringFound(hconfig, keystring='lon_range', _RC)
+!      has_dateline = MAPL_HConfigKeystringFound(hconfig, keystring='dateline', _RC)
```
tclune, line 137:
```suggestion
```

hconfig_utils/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+esma_set_this (OVERRIDE MAPL.hconfig_utils)
+
+set(srcs
+  HConfig3G.F90
+  mapl3g_hconfig_get.F90
+  mapl3g_hconfig_params.F90
+  mapl3g_hconfig_get_private.F90
+  mapl3g_generalized_equality.F90
+  mapl3g_get_hconfig.F90
```
tclune, line 9: Filenames should not have `mapl_`.  Namespace protection is for module names.
```suggestion
  hconfig_get.F90
  hconfig_params.F90
  hconfig_get_private.F90
  generalized_equality.F90
  get_hconfig.F90
```

hconfig_utils/mapl3g_generalized_equality.F90
```diff
@@ -0,0 +1,108 @@
...
+
+   use :: esmf, only: ESMF_KIND_R4, ESMF_KIND_R8, ESMF_KIND_I4, ESMF_KIND_I8
+   implicit none
+   private
+
+   public :: are_equal
+
+   interface are_equal
+      module procedure :: equals_i4_scalar
+      module procedure :: equals_i8_scalar
```
tclune, line 11:
```suggestion
      procedure :: equals_i4_scalar
      procedure :: equals_i8_scalar
```
Propagate to the rest.

hconfig_utils/mapl3g_generalized_equality.F90
```diff
@@ -0,0 +1,108 @@
...
+      lval = (u == v)
+
+   end function equals_r8_scalar
+
+   logical function equals_l_scalar(u, v) result(lval)
+      logical, intent(in) :: u, v
+
+      lval = (u .eqv. v)
+
+   end function equals_l_scalar
```
tclune, line 58: Not a complaint, but wouldn't it have been much easier to just pass `.eqv.`, `==`, and `all(==)` as tokens to your template rather than implementing such trivial procedures?

hconfig_utils/mapl3g_get_hconfig.F90
```diff
@@ -0,0 +1,156 @@
...
+   use :: esmf, only: ESMF_HConfigAsR8, ESMF_HConfigAsR8Seq
+   use mapl_ErrorHandling
+
+   implicit none
+   private
+
+   public :: get_hconfig
+
+   interface get_hconfig
+      module procedure :: get_hconfig_as_i4
```
tclune, line 20: Remove `module`

hconfig_utils/mapl3g_hconfig_get_private.F90
```diff
@@ -0,0 +1,149 @@
...
+   use :: esmf, only: ESMF_KIND_R4, ESMF_KIND_R8, ESMF_KIND_I4, ESMF_KIND_I8
+   use :: esmf, only: ESMF_HConfig, ESMF_HConfigIsDefined
+   use mapl_ErrorHandling
+
+   implicit none
+   private
+   public :: get_value, HConfigParams
+
+   interface get_value
+      module procedure :: get_value_i4
```
tclune, line 16:
```suggestion
      procedure :: get_value_i4
```

hconfig_utils/mapl3g_hconfig_get_private.F90
```diff
@@ -0,0 +1,149 @@
...
+#include "mapl3g_hconfig_get_value_template.h"
+   end subroutine get_value_i4seq
+
+   subroutine get_value_i8seq(params, value, default, valuestring, rc)
+      integer(kind=ESMF_KIND_I8), dimension(:), allocatable, intent(inout) :: value
+      integer(kind=ESMF_KIND_I8), dimension(:), optional, intent(in) :: default
+      character(len=*), parameter :: typestring = 'I8'
+      character(len=*), parameter :: edit_descriptor = EDIT_DESC_I8
+#include "mapl3g_hconfig_get_value_template.h"
+   end subroutine get_value_i8seq
```
tclune, line 115: This approach is generally harder to understand than one where the template also includes the procedure interface.   It just means you have to also define the proc name as a token.

tclune (changes requested):
Mostly a bunch of trivial things to change, but there are a few substantive questions mixed in.

generic3g/MAPL_Generic.F90
```diff
@@ -1,18 +1,18 @@
 #include "MAPL_ErrLog.h"
 
 !---------------------------------------------------------------------
-!
+
```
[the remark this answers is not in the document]
darianboggs: I don't know what happened here. I restored the '!'.

generic3g/MAPL_Generic.F90
darianboggs: Done

hconfig_utils/mapl3g_hconfig_get_private.F90
darianboggs: True. The value and default arguments cannot appear in the template. Actually, the types, dimensions, and allocatable status cannot appear in the template. It's possible to replace these with macros:

- type of value and default
- kind of value and default
- len for value (character)
- len for default (character)
- dimension(:) for arrays
- allocatable for arrays

The downside is that the macros change for each subroutine. It seemed a bad idea to redefine the macros multiple times in the same source file. I noticed that the preprocessor reports each time they are redefined. By carefully grouping the subroutines, some of the redefinitions can be avoided, but there are still multiple redefinitions.

typestring and edit_descriptor can be set by macros that change for each subroutine

generic3g/MAPL_Generic.F90
darianboggs: I saw some places where the **package** module had the mapl3g prefix and some where it did not have the prefix. So this could be `use mapl3g_hconfig_utils` or `use hconfig_utils`. Will fix in morning.

generic3g/MAPL_Generic.F90
darianboggs: Will change in morning.

hconfig_utils/CMakeLists.txt
darianboggs: Will change CMakeLists.txt and actual filenames in morning

hconfig_utils/mapl3g_generalized_equality.F90
darianboggs: Will change in morning

hconfig_utils/mapl3g_generalized_equality.F90
darianboggs: I originally had a number of macros like EQUALS_ that were redefined for each subroutine. In order to reduce the number of macros that were redefined for each subroutine, I switched to this. I'll look at this in the morning, but I think that this could be simplified with IS_ARRAY and IS_LOGICAL macros change the relation type.

hconfig_utils/mapl3g_get_hconfig.F90
darianboggs: Will change

hconfig_utils/mapl3g_generalized_equality.F90
tclune: Note that it is also possible that you can use `all()` even in the scalar case.  Might be confusing, but if it simplifies the other logic...   I think it will become a common pattern when Fortran has proper templates.

hconfig_utils/CMakeLists.txt
darianboggs: Done

hconfig_utils/mapl3g_generalized_equality.F90
darianboggs: Done

hconfig_utils/mapl3g_get_hconfig.F90
darianboggs: Done

hconfig_utils/mapl3g_hconfig_get_private.F90
darianboggs: I'll leave this for the refactor.

hconfig_utils/mapl3g_generalized_equality.F90
darianboggs: I'll leave this to the refactor. I wasn't able to use `ALL()` with a scalar, but there is probably a way to do that. In refactor, I'll probably just use IS_ARRAY and IS_LOGICAL to set the relational operation.

hconfig_utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+                WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+                MAX_PES 4
+                )
+set_target_properties(MAPL.hconfig_utils.tests PROPERTIES Fortran_MODULE_DIRECTORY ${MODULE_DIRECTORY})
+
+if (APPLE)
+  set(LD_PATH "DYLD_LIBRARY_PATH")
+else()
+  set(LD_PATH "LD_LIBRARY_PATH")
+endif ()
```
tclune, line 21: Note to @mathomp4  Would it make sense to set this globally for GEOS.   (Possibly with longer name.)

tclune (approved):
Whew!

generic3g/MAPL_Generic.F90
darianboggs: Working on implementing change.

PR #2575, opened 2024-02-06
Fixes #2569. Convert MAPL to use ESMF::ESMF target

### LINKED ISSUE #2569: Use ESMF::ESMF CMake target
ESMF's FindESMF.cmake [now has a CMake target called `ESMF::ESMF`](https://github.com/esmf-org/esmf/commit/da8f4108c65605a4762b2383f025be3ba5fd6543) which has the benefit of avoiding ambiguity between library name and target name when invoking `target_link_libraries()`. If MAPL can move to using this target name, then when UFS Weather Model/GOCART imports MAPL, we can avoid the spurious `-lESMF` flag we've been seeing because of a mismatch in the case of the target name.

I'm happy to help however I can with making this happen, but I don't know MAPL or its build system very well.

mathomp4:
## Description

This PR moves MAPL to use `ESMF::ESMF` target instead of the `esmf` or `ESMF` target. This was causing issues with UFS builds (see #2569). 

This requires the use of ESMA_cmake v3.40.0 so that non-Baselibs builds of MAPL are provided this target. GEOSgcm `main` now uses that version as of https://github.com/GEOS-ESM/GEOSgcm/pull/746

## Related Issue

Fixes #2569 

## Motivation and Context

Hopefully fixes build issues by our UFS colleagues.

## How Has This Been Tested?

Builds of MAPL and GEOSgcm. Frankly, if it builds, that's all we need as it is a CMake change only. But testing shows it is zero-diff.

## Types of changes

## Checklist:


mathomp4:
Well, crap. I forgot about GEOSldas. Let me go update them.

tclune (approved):

PR #2599, opened 2024-02-12
Fix non-Baselibs build

mathomp4:
# NOTE NOTE NOTE

Note to @mathomp4: Because this needs an update to ESMF 8.6.1 to work, this will *not* be a hotfix and all this will need to go against develop. But I'm leaving the PR here so I don't lose anything until ready.

---

## Description

As found by @AlexanderRichert-NOAA in https://github.com/spack/spack/pull/42619 I was a bit too heavy-handed with the sed and changed too many things to `ESMF::ESMF`. As such, I broke the Spack build of MAPL 2.44

This should be the right fix, though testing is underway. I can say I can build MAPL using spack libraries. I'm trying `spack install mapl` now.

Related: https://github.com/GEOS-ESM/ESMA_cmake/pull/354

I'm still trying to figure out the right stuff in ESMA_cmake to get this to work. I think I broke...something.

## Motivation and Context

Fixes the spack build of MAPL!

## How Has This Been Tested?

Testing now. As this is a pure CMake bug, if it builds, it works.

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,8 @@
 CMakeUserPresets.json
 
 # Ignore possible symlinked build and install directories
-build-*
-install-*
+build*/
+install*/
 
 *.swp
 *.swo
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,12 @@
 CMakeUserPresets.json
 
 # Ignore possible symlinked build and install directories
-build*/
-install*/
+# Note: we can't use a / at the end because they might be
+# symlink to a directory
+build*
+install*
+
+# If you build with spack libraries, you can get spack log files
 spack*.log
 
 *.swp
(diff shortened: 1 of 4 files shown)


AlexanderRichert-NOAA:
As part of the testing, please check that the name of the ESMF target in MAPL-targets.cmake is `ESMF::ESMF`-- when I hacked my way around the 'target_link_libraries can not be used on an ALIAS target' issue, the target name in MAPL-targets.cmake was still `ESMF`, and as far as I can tell that's the thing that we want to change (so that when we import MAPL we won't run into conflicting or nonexistent names for the ESMF target).

mathomp4:
> As part of the testing, please check that the name of the ESMF target in MAPL-targets.cmake is `ESMF::ESMF`-- when I hacked my way around the 'target_link_libraries can not be used on an ALIAS target' issue, the target name in MAPL-targets.cmake was still `ESMF`, and as far as I can tell that's the thing that we want to change (so that when we import MAPL we won't run into conflicting or nonexistent names for the ESMF target).

Ooh. Okay. Let me look at that too once I can build.

PR #2600, opened 2024-02-13
Mask sampler for geostationary satellite

metdyn:
## Description

Mask array is used when observation resolution is higher than the MODEL. 
When obs locations are at 500 m resolution in ABI and AHI, we want to output loc stream points from the cubed sphere grid, corresponding to the masked positions and their nearest neighbor halo points on CS.  

Other added functions:
- Add mask sampler for geostationary satellite (GEOS-R series)
- Add geostation name into NC for station sampler
- Add mapping between the IODA loc_index and trajectory NC output loc_index
- Corrected a _RETURN in  subroutine MAPL_GetGlobalHorzIJIndex (@weiyuan-jiang), when I realized my run crashes otherwise.   Only after the ESMF_functionalities are called using global PE, then partial PE can return with success. 

Screenshot 2024-02-13 at 2 29 51 PM

## Types of changes

## Checklist:


base/Base/Base_Base_implementation.F90
```diff
@@ -2821,8 +2817,12 @@ module subroutine MAPL_GetGlobalHorzIJIndex(npts,II,JJ,lon,lat,lonR8,latR8,Grid,
     II = -1
     JJ = -1
 
-    ! The edge points are assigned in the order of face 1,2,3,4,5,6
+    ! when npts=0 on localDE, return here
+    if (npts == 0 ) then
+      _RETURN(_SUCCESS)
+    endif
```
tclune, line 2823:
```suggestion
    ! Return if no local points
   _RETURN_IF(npts == 0)
```

base/MAPL_ObsUtil.F90
```diff
@@ -166,6 +172,71 @@ subroutine time_real_to_ESMF (times_R8_1d, times_esmf_1d, datetime_units, rc)
...
+  subroutine create_timeunit (time, datetime_units, input_unit, rc)
+    type(ESMF_Time), intent(in) :: time
+    character(len=*), intent(out) :: datetime_units
+    character(len=*), optional, intent(in) :: input_unit
+
+    integer, optional, intent(out) :: rc
+
+    integer :: i, len
+    integer :: status
+    character(len=20) :: string
```
tclune, line 226: Is 20 characters guranteed to be sufficient by ESMF?

base/MAPL_ObsUtil.F90
```diff
@@ -166,6 +172,71 @@ subroutine time_real_to_ESMF (times_R8_1d, times_esmf_1d, datetime_units, rc)
...
+
+    call ESMF_timeget (time, timestring=string, _RC)
+    if (present(input_unit)) then
+       datetime_units = trim(input_unit)//' since '//trim(string)
+    else
+       datetime_units = 'seconds since '//trim(string)
+    end if
+    !!print*, 'datetime_units:', trim(datetime_units)
```
tclune, line 234:
```suggestion
    datetime_units = 'seconds'
    if (present(input_unit)) datetime_units = trim(input_unit)
    datetime_units = trim(datetime_units )// ' since ' //t rim(string)
```
(Avoid "else" clauses.)

base/MAPL_ObsUtil.F90
```diff
@@ -166,6 +172,71 @@ subroutine time_real_to_ESMF (times_R8_1d, times_esmf_1d, datetime_units, rc)
...
+       end select
+    enddo
+
+    _RETURN(_SUCCESS)
+  end subroutine time_ESMF_to_real
+
+
+  subroutine create_timeunit (time, datetime_units, input_unit, rc)
+    type(ESMF_Time), intent(in) :: time
+    character(len=*), intent(out) :: datetime_units
```
tclune, line 219: This interface would (possibly) be simpler if the string were allocatable.   Otherwise we have to `trim()` in various spots.  (See change below.)

base/MAPL_ObsUtil.F90
```diff
@@ -250,12 +319,14 @@ subroutine Find_M_files_for_currTime (currTime, &
...
-       obsfile_Te_index = n2
-    end if
+    obsfile_Ts_index = n1 - 1   ! downshift by 1
+    obsfile_Te_index = n2
+!    if ( dT2_s - n2*dT0_s < 1 ) then
+!       obsfile_Te_index = n2 - 1
+!    else
+!       obsfile_Te_index = n2
+!    end if
+
```
tclune, line 329: Eliminate commented code?
```suggestion
```

base/MAPL_ObsUtil.F90
```diff
@@ -668,6 +736,68 @@ subroutine sort_four_arrays_by_time(U,V,T,ID,rc)
...
+    integer :: i, len
+    real(ESMF_KIND_R8), allocatable :: XX(:)
+
+    _ASSERT (size(X)==size(IA), 'X and IA (its index) differ in dimension')
+    len = size (X)
+    allocate (XX(len))
+    XX(:) = X(:)
+    do i=1, len
+       X(i) = XX(IA(i))
+    enddo
```
tclune, line 776: Check me on this, but I think vanilla Fortran vector indices does this operation.   If it does, please eliminate the unused local variables.   And then consider whether it is clearer to have this procedure or just a comment with the inlined expression below ...
```suggestion
 x = x(IA)
```

base/Plain_netCDF_Time.F90
```diff
@@ -287,12 +287,12 @@ subroutine get_att_real_netcdf(filename, varname, att_name, att_value, group_nam
     character(len=*), intent(in) :: varname
     character(len=*), intent(in) :: att_name
     real(REAL64), intent(out) :: att_value
-    character(len=*), optional, intent(out) :: group_name    
+    character(len=*), optional, intent(out) :: group_name
```
tclune, line 290: I've not looked closely, but generally an allocatable string is preferable when `intent(out)` (or result of function.  But don't change if it requires changing lots of other declarations.

base/Plain_netCDF_Time.F90
```diff
@@ -396,12 +396,12 @@ subroutine time_esmf_2_nc_int(time, tunit, n, rc)
...
-!    call ESMF_TimeGet(time, timestring=STR1, _RC)
-!    write(6, '(2x,a,2x,a)') 'time=', trim(STR1)
-!    call ESMF_TimeGet(time0, timestring=STR1, _RC)
-!    write(6, '(2x,a,2x,a)') 'time0=', trim(STR1)
+    !! test
+    !!write(6, '(2x,a,2x,a)') 'tunit=', trim(tunit)
+    !!call ESMF_TimeGet(time, timestring=STR1, _RC)
+    !!write(6, '(2x,a,2x,a)') 'time=', trim(STR1)
+    !!call ESMF_TimeGet(time0, timestring=STR1, _RC)
+    !!write(6, '(2x,a,2x,a)') 'time0=', trim(STR1)
```
tclune, line 404: Eliminate commented code?
```suggestion
```

base/Plain_netCDF_Time.F90
```diff
@@ -441,12 +440,17 @@ subroutine parse_timeunit_i4(tunit, n, t0, dt, rc)
...
-    isec=n
+    if (trim(s_unit) == 'seconds') then
+       isec=n
+    elseif (trim(s_unit) == 'minutes') then
+       isec=n * 60
+    elseif (trim(s_unit) == 'hours') then
+       isec=n * 3600
+    else
+       _FAIL ('time_unit not implemented')
+    end if
```
tclune, line 451: This logic is duplicated below for the I8 interface.   Couldn't the I4 interface just call the I8 interface, e.g.:
```f90
integer(kind=ESMF_KIND_I8) :: n8
n8 = n
call parse_timeunit(tunit, n8, t0, dt, _RC)
```

base/Plain_netCDF_Time.F90
```diff
@@ -482,18 +485,70 @@ subroutine parse_timeunit_i8(tunit, n, t0, dt, rc)
...
-         calendar=gregorianCalendar, _RC)
+    call ESMF_timeSet(t0, yy=y,mm=m,dd=d,h=hour,m=min,s=sec, _RC)
     call ESMF_timeintervalSet(dt, d=0, h=0, m=0, s_i8=isec, _RC)
 
     _RETURN(_SUCCESS)
 
   end subroutine parse_timeunit_i8
 
+  subroutine diff_two_timeunits (tunit1, tunit2, x, rc)
+    character(len=*), intent(in) :: tunit1
```
tclune, line 506: Wouldn't this be easier to use and understand if it was a function that returns x rather than having an `INTENT(OUT)` argument ?

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -5395,15 +5416,12 @@ subroutine regen_rcx_for_obs_platform (config, nlist, list, rc)
        nseg_ub=0
        do while (ios == 0)
           read (unitr, '(A)' ) line
```
tclune, line 5418: I know it is not part of this pull request, but shouldn't there be an IOSTAT argument for the READ statement?   Seems robust, but I'm worried about trapping errors in running code.  I/O is the most common place for unexpected things to happen.

gridcomps/History/Sampler/MAPL_GeosatMaskMod.F90
```diff
@@ -0,0 +1,167 @@
+module MaskSamplerGeosatMod
```
tclune, line 1: Please change all "GeoSat" to "GeoStat".   (This includes renaming the file itself.)

gridcomps/History/Sampler/MAPL_StationSamplerMod.F90
```diff
@@ -562,11 +571,13 @@ subroutine get_file_start_time(this,start_time,time_units,rc)
...
 
   ! TODO: delete and use system utilities when available
-  Subroutine count_substring (str, t, ncount)
+  Subroutine count_substring (str, t, ncount, rc)
     character (len=*), intent(in) :: str
     character (len=*), intent(in) :: t
     integer, intent(out) :: ncount
+    integer, optional, intent(out) :: rc
     integer :: i, k, lt
+    integer :: status
```
tclune, line 580: This local variable is unused, so why add it?
```suggestion
```

gridcomps/History/Sampler/MAPL_StationSamplerMod.F90
```diff
@@ -562,11 +571,13 @@ subroutine get_file_start_time(this,start_time,time_units,rc)
   end subroutine get_file_start_time
 
   ! TODO: delete and use system utilities when available
-  Subroutine count_substring (str, t, ncount)
+  Subroutine count_substring (str, t, ncount, rc)
     character (len=*), intent(in) :: str
     character (len=*), intent(in) :: t
     integer, intent(out) :: ncount
+    integer, optional, intent(out) :: rc
```
tclune, line 578: There is no failure mode in this procedure, so maybe you did not need the change?

gridcomps/History/Sampler/MAPL_StationSamplerMod.F90
```diff
@@ -576,6 +587,45 @@ Subroutine count_substring (str, t, ncount)
...
+    integer :: i, j, k
+    integer :: status
+
+    i=index(line, ',')
+    j=index(line(i+1:), ',')
+    _ASSERT (i>0, 'not CSV format')
+    _ASSERT (j>0, 'CSV format: find only 1 comma, should be > 1')
+    j=i+j
+
+    read(line(1:i-1), '(a100)')  name
```
tclune, line 609: Definitely what to check IOSTAT in reads for this routine.

gridcomps/History/Sampler/MAPL_StationSamplerMod.F90
```diff
@@ -576,6 +587,45 @@ Subroutine count_substring (str, t, ncount)
...
+    k=index(line(i+1:j-1), '.')
+    if (k > 0) then
+       read(line(i+1:j-1), *) lon
+    else
+       read(line(i+1:j-1), *) i
+       lon = i
+    endif
```
tclune, line 616: `k` is not used in this block.  I think the line should be deleted.

base/MAPL_ObsUtil.F90
metdyn: Great!  This works!  Thanks!

base/MAPL_ObsUtil.F90
```diff
@@ -668,6 +736,68 @@ subroutine sort_four_arrays_by_time(U,V,T,ID,rc)
...
+    integer :: i, len
+    integer, allocatable :: XX(:)
+
+    _ASSERT (size(X)==size(IA), 'X and IA (its index) differ in dimension')
+    len = size (X)
+    allocate (XX(len))
+    XX(:) = X(:)
+    do i=1, len
+       X(i) = XX(IA(i))
+    enddo
```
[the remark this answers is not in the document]
metdyn: This approach is so convenient and I find we donot even need these two subroutines (they are deleted now).

base/Plain_netCDF_Time.F90
metdyn: I realized there a need to add an optional output dt_ESMF to the subroutine.

gridcomps/History/MAPL_HistoryGridComp.F90
metdyn: Nice, that is a good point!  I see an example in the MAPL code: `Read(fIDGCHP,'(a)',IOStat=RC_) currLine`. I changed these similar lines in the code to follow the style to catch unexpected situations.

base/Plain_netCDF_Time.F90
metdyn: Thanks for pointing out this approach, which is much better!

base/Plain_netCDF_Time.F90
metdyn: I will think about the allocatable character in the coding later on. Here there are many hard coded string with fixed length, requiring trim. I will do a clean up in another PR.

tclune (dismissed):

metdyn:
Fixed a few bugs in the sampler code.
- Add allocate(X, _STAT) to sampler codes
- Skip destroy_regen_grid when list(n)%end_alarm is active (the last time step in sampler)
- Add extract_unquoted_item(STR1) to fix a string bug in geoval_xname(mx_ngeoval) in trajectory sampler

tclune (approved):

tclune (approved):

PR #2626, opened 2024-02-28
Add remove attribute procedure to FileMetadata

bena-nasa:
In the interp_restart code, to produce the output file I "cheat". Basically load the input file and produce a fileMetadata object from it. I then copy this to a new fileMetadata object and change the dimensions in object. Then I can use this to produce the new file.

The problem is that if the input is stretched but the output is not, @wmputman found that the stretching parameters were getting into the output file in his new updated inpterp_restarts.x version that can handle both and input and/or output stretched grid. So I needed a procedure to remove them if needed before writing.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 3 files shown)


atrayano (approved):

tclune:
Longer term is there an option to simply pass a correct metadata object when creating the new file?   This seems kludgy.

PR #2631, opened 2024-03-06
Store regrid weights

bena-nasa:
This PR provide several things.
1. The ability for MAPL_ESMFRegridder to retrieve and store the route handle weights from/to a file. Off by default. If turned on it will take the source grid dimensions, source layout, destination dimensions, destination layout, the regrids method and generate a unique file name. If the file exists, it will read rather than generate weights. If the file does not exists will generate weight as normal and then write them to a file.
2. An expanded use the "hints" option to enable item
3. Disable the computation of the transpose by default
4. Another hint to enable the transpose in make_regridder
5. Options in the History.rc and extdata.yaml files to turn on the the write/read weights option

These weights are of course specific to a particular layout so only for a longer run where multiple executions of the model will be done and the layout doesn't change.

## Types of changes

## Checklist:


base/MaplGrid.F90
```diff
@@ -325,6 +329,21 @@ subroutine MAPL_GridGet(GRID, globalCellCountPerDim, localCellCountPerDim, RC)
...
+         call MAPL_DistGridGet(distgrid, &
+            minIndex=minindex, &
+            maxIndex=maxindex, _RC)
+         call MAPL_GetImsJms(Imins=minindex(1,:),Imaxs=maxindex(1,:),&
+           Jmins=minindex(2,:),Jmaxs=maxindex(2,:),Ims=ims,Jms=jms,_RC)
+
+         layout(1) = size(ims)
+         layout(2) = size(jms)
+      end if
+
```
atrayano, line 346: I think we should deallocate ims and jms at this point

base/MaplGrid.F90
```diff
@@ -325,6 +329,27 @@ subroutine MAPL_GridGet(GRID, globalCellCountPerDim, localCellCountPerDim, RC)
...
+         call MAPL_DistGridGet(distgrid, &
+            minIndex=minindex, &
+            maxIndex=maxindex, _RC)
+         if (associated(ims)) then
+            nullify(ims)
+         end if
+         if (associated(jms)) then
+            nullify(jms)
+         end if
```
tclune, line 345: No need to be so verbose here.  (I had to check/learn that nullify accepts multiple items.)
```suggestion
         nullify(ims, jms)
```

tclune: And as per Atanas' comment below - is this a memory leak?

tclune (approved):
It is unfortunate that we needed such substantial changes in ESMFRegridder, as the MAPL3G version is much cleaner.  OTOH, this probably helps us to "get it right" in 3g rather than rushing to get something working in there.

base/MaplGrid.F90
bena-nasa: @tclune  I don't know what is going on. That seems to be necessary to make gfortran happy since it complains IMS is associated going in the call. If I try deallocating it rather than nullifying it it says it can't deallocate something not allocated.
Really there's no reason that ims/jms passed needs to be a pointer at all from what I can see. Seems like it could be turned into an allocatable. Unfortunately that call is used elsewhere so I would have to fix every there it is called to make it an allocatable rather than a pointer?
Maybe I should?

tclune: 2 separate issues.   **If** the pointers do not "own"  the memory, then my suggested change is just simpler.  Than your original.  The pointers end up null regardless of their state.

But **if** the pointers do "own" the memory, then we should/must deallocate.  Then the code should be:
```
if (associated(ims)) then
   deallocate(ims)
endif
```
Then `ims` ends up in null() in any event and any associated memory is freed.  I'll try to refresh my memory to see which case should apply.

tclune:
> > It is unfortunate that we needed such substantial changes in ESMFRegridder, as the MAPL3G version is much cleaner. OTOH, this probably helps us to "get it right" in 3g rather than rushing to get something working in there.
> 
> @tclune Yeah, but unfortunately I can't think of significantly cleaner way to do this. 

I was not critiquing your change - just came out wrong.   I meant I'd rather have treated this layer as frozen as it has been superseded.  But ... as we cannot yet use MAPL3 for the full model, and  we need the capability "now" ... we must implement it twice.

base/MaplGrid.F90
bena-nasa: @atrayano yep, made that change

atrayano (approved):

PR #2446, opened 2023-11-16
change implementation of MAPL_getHorzIJIndex

weiyuan-jiang:
This PR is replacing MAPL_GetHorzIJIndex with MAPL_GetGlobalHorzIJIndex. The indexes are converted back to the local index so the results are not changed. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Trivial change (affects only documentation or cleanup)

## Checklist:

- [ ] I have tested this change with a run of GEOSgcm (if non-trivial)
- [x] I have added one of the required labels (0 diff, 0 diff trivial, 0 diff structural, non 0-diff)
- [x] I have updated the CHANGELOG.md accordingly following the style of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how)

PR #2636, opened 2024-03-07
unify HorzIJIndex search

weiyuan-jiang:
unify HorzIJIndex search. This PR replace this [PR
](https://github.com/GEOS-ESM/MAPL/pull/2446)
## Types of changes

## Checklist:

### DIFF
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -2592,13 +2592,16 @@ contains
     real(ESMF_KIND_R8), pointer     :: lats(:,:)
     real(ESMF_KIND_R8), allocatable :: elons(:)
     real(ESMF_KIND_R8), allocatable :: elats(:)
-    integer :: i,iiloc,jjloc
+    integer :: i,iiloc,jjloc, i1, i2, j1, j2
     real(ESMF_KIND_R4) :: lonloc,latloc
     logical                 :: localSearch
     real(ESMF_KIND_R8), allocatable :: target_lons(:),target_lats(:)
-    real(ESMF_KIND_R8), allocatable :: corner_lons(:,:),corner_lats(:,:),center_lats(:,:),center_lons(:,:)
     type(ESMF_CoordSys_Flag) :: coordSys
     character(len=ESMF_MAXSTR) :: grid_type
+
+    if (npts == 0 ) then
+      _RETURN(_SUCCESS)
+    endif
 
     ! if the grid is present then we can just get the prestored edges and the dimensions of the grid
     ! this also means we are running on a distributed grid
@@ -2615,6 +2618,7 @@ contains
     else
        localSearch = .false.
     end if
+
     allocate(target_lons(npts),target_lats(npts))
     if (present(lon) .and. present(lat)) then
        target_lons = lon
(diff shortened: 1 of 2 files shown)


weiyuan-jiang:
@metdyn Would you please help me test the branch?  You reported error last time.

metdyn:
Sure. I will run the test for sampler code.

mathomp4:
The GEOSadas failure looks like the usual race condition.

atrayano (commented):
I have several comments:
1) I am not sure if the "accurate" calculation works at all with stretched CS grids (i.e. with Schmidt transform). Even if it does, when the grid is stretched, the "shift" is 0. We need some protection in place
2) The calls to get the corners from ESMF:
a) at the very least, we should verify the return status
b) in practice, the corodinates are stored as R8 precision. But since this is not required by ESMF, the get call might fail when R8 pointer is passed
c) I prefer to use MAPL_GetCorner routine: the shape of the corners form ESMF is (IM, JM) instead of (IM+1, JM+1). We have the proper logic ( and halo) to get all carners in the MAPL routine
3) if we set a tolerance, why are we comparing against 5*tol?

metdyn:
@weiyuan-jiang,  no complaints were found in the sampler code (i.e., ./MAPL_GeosatMaskMod_smod.F90:       
call MAPL_GetHorzIJIndex(nx,II,JJ,lonR8=obs_lons,latR8=obs_lats,grid=grid,_RC)

tclune:
> 1. I am not sure if the "accurate" calculation works at all with stretched CS grids (i.e. with Schmidt transform). Even if it does, when the grid is stretched, the "shift" is 0. We need some protection in place

Yes - but this will need to wait.   The general case is much harder and will only get worse when we start reading coords from a file and/or use stretched grids!


>    b) in practice, the corodinates are stored as R8 precision. But since this is not required by ESMF, the get call might fail when R8 pointer is passed

I would think that just trapping on the ESMF return code would be sufficient here.  We don't need to support R4 coordinates just because ESMF does.  This is very specific to CS after all.

>    c) I prefer to use MAPL_GetCorner routine: the shape of the corners from ESMF is (IM, JM) instead of (IM+1, JM+1). We have the proper logic ( and halo) to get all corners in the MAPL routine

Agreed, but only if the algorithm needs them to do its logic.  Did not look closely.

> 3. if we set a tolerance, why are we comparing against 5*tol?

I presume you mean the lines should be fused:
` real, parameter :: TOLERANCE = 0.5 * espilon(1.)`
?

And a little thought should go into whether that should be R8, and if so `epsilon(1.d0)`

weiyuan-jiang:
> I have several comments:
> 
> 1. I am not sure if the "accurate" calculation works at all with stretched CS grids (i.e. with Schmidt transform). Even if it does, when the grid is stretched, the "shift" is 0. We need some protection in place

   Is the original subroutine correct? Is there any hint that it is shift or stretched ? We need the information make it right.

> 2. The calls to get the corners from ESMF:
>    a) at the very least, we should verify the return status
>    b) in practice, the corodinates are stored as R8 precision. But since this is not required by ESMF, the get call might fail when R8 pointer is passed
>    c) I prefer to use MAPL_GetCorner routine: the shape of the corners from ESMF is (IM, JM) instead of (IM+1, JM+1). We have the proper logic ( and halo) to get all corners in the MAPL routine

      c) MAPL_GetCorner  is a collective call that I want to avoid. It only needs several points to verify that is is the "correct" grid

> 3. if we set a tolerance, why are we comparing against 5*tol?

   I relax the criteria because I am not sure how the "accurate" one is calculated.  The relaxed standard should be good enough for very high resolution.

atrayano:
The routine MAPL_GetCorner internally does the collective communication only once. Then it stores the corners as 1-d attribute, and next time we call the routine, it only retrieves a local copy

base/Base/Base_Base_implementation.F90
```diff
@@ -2592,14 +2592,17 @@ module subroutine MAPL_GetHorzIJIndex(npts,II,JJ,lon,lat,lonR8,latR8,Grid, rc)
...
     logical                 :: localSearch
     real(ESMF_KIND_R8), allocatable :: target_lons(:),target_lats(:)
-    real(ESMF_KIND_R8), allocatable :: corner_lons(:,:),corner_lats(:,:),center_lats(:,:),center_lons(:,:)
     type(ESMF_CoordSys_Flag) :: coordSys
     character(len=ESMF_MAXSTR) :: grid_type
 
+    if (npts == 0 ) then
+      _RETURN(_SUCCESS)
+    endif
+
```
tclune, line 2605:
```suggestion
    _RETURN_IF(npts == 0 ) 
```

tclune (approved):

tclune:
@weiyuan-jiang I'm going to add a separate ticket that something needs to check that this is not a stretched grid.   
But will let this PR go through as that is not currently a concern.
Note there are conflicts to resolve.

weiyuan-jiang:
> The GEOSadas failure looks like the usual race condition.

Could you post the failed message? If it fails because of the race condition, we can either add a barrier or change the code elsewhere.  I think adding a barrier here is a bad idea.

mathomp4:
> > The GEOSadas failure looks like the usual race condition.
> 
> Could you post the failed message? If it fails because of the race condition, we can either add a barrier or change the code elsewhere. I think adding a barrier here is a bad idea.

That's a rabbit hole you don't want to go down. The GSI is not fun to deal with and it would be a nigh rewrite to get it "good". Since GSI is going away eventually I figure we deal with odd failures here and there.

tclune (approved):

mathomp4 (approved):

mathomp4:
@weiyuan-jiang This has been merged, though note the concerns from @tclune here:

https://github.com/GEOS-ESM/MAPL/issues/2643

PR #2645, opened 2024-03-12
Minor improvement on swath sampler

metdyn:
## Description

Minor changes to swath grid
- Give informative error message when swath grid Epoch does not equal swath sampler frequency
- Duplicated/Adapted subroutine RegridVector from GriddedIO.F90 to MAPL_EpochSwathMod.F90 (changing class name for ::this) because of the need to regrid and accumulate for swath sampler

## Types of changes

## Checklist:


tclune (commented):
Do these changes actually do the tangent --> cartesian --> tangent needed for vectors?  Or just some of the initial work to populate the bundle?   If the answer is yes to the former, then please point me to that because I did not see it.

metdyn:
It took me a while to find the overload in AbstractRegridder.   
Indeed, the tangent --> cartesian --> tangent is used as in 
https://github.com/GEOS-ESM/MAPL/blob/feature/ygyu/sampler_aist_test/base/MAPL_EsmfRegridder.F90#L562-L595

The code goes through MAPL_EsmfRegridder module via
https://github.com/GEOS-ESM/MAPL/blob/feature/ygyu/sampler_aist_test/base/NewRegridderManager.F90#L163-L164

I print out an output that shows
 TestDriver Date: 2019/08/01  Time: 00:40:00    0.0%Memory Committed
 regrid_vector_3d_real32


My statement the PR description is truly misleading and I have corrected it.  I did not add any regridding subroutines, all the functions are already in the MAPL, i.e., The griddedio.F90 code is here 
https://github.com/GEOS-ESM/MAPL/blob/feature/ygyu/sampler_aist_test/griddedio/GriddedIO.F90#L670-L835. 

I adapted the regridvector code for the swath sampler, because of the need to regrid and accumulate the interpolated values along the swath axis (only changed one line for the class name of this)

metdyn:
I have built and run GEOS with this feature branch with C180 resolution

gridcomps/History/Sampler/MAPL_EpochSwathMod.F90
```diff
@@ -193,6 +193,39 @@ subroutine config_accumulate (this, key, cf, rc)
...
+    local_swath_epoch_sec = MAPL_nsecf( time_integer )
+
+    con = (hq_epoch_sec == local_swath_epoch_sec) .AND. (hq_epoch_sec == freq_sec)
+    if (mapl_am_i_root()) then
+       if (.not. con) then
+          write(6, '(2x,a,2x,i10)') 'hq_epoch_sec', hq_epoch_sec
+          write(6, '(2x,a,2x,i10)') 'local_swath_epoch_sec', local_swath_epoch_sec
+          write(6, '(2x,a,2x,i10)') 'freq_sec', freq_sec
+       end if
+    end if
```
tclune, line 223: Not in this pass, but please follow up with a PR to replace `write` with pFlogger calls.

tclune (approved):

PR #2658, opened 2024-03-18
fixes #2657

### LINKED ISSUE #2657: Bit shaving and binary file output in History
The LDAS folks were finding that turning collections to instantaneous from time-averaged in History was changing the GEOSldas.x checkpoint. I looked into it and it was indeed the case. The issue though is the bit shaving in the binary code path.

Looks like since v2.8.6 of MAPL binary output has had a bit-shaving option. Turns out there's a possible bug. If bit shaving basically the "output state" of the collection is passed to an optional routine to bit shave based on user input.

Looks like if instantaneous output, we are bit-shaving the actual fields from the component. Because of the friendliness business in MAPL, many exports are more than just diagnostic and this can change the model state. I confirmed that turning off big shaving got rid of the non-zero diff.

If it is time-averaged, we've made a new buffer for the accumulation and this is what is bit shaved. When the accumulation is done, we shave that buffer, it has no effect on the component fields.

So if we are bit shaving instataneous output, we need to make a copy.

bena-nasa:
## Related Issue

As the title says this fixes #2657 
If the the binary output is bit shaved, I copy to a new state before writing. 

In reality this is only necessary for instantaneous output  but I was trying to avoid a whole slew of conditions and it doesn't matter in the end if we make a few extra copies

I've marked this as zero-diff because is is zero-diff for the GEOSgcm.x although it is not for the LDAS.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix bug where bit-shaved, instantaneous binary output in History was modifying the original export state passed
+
 ### Removed
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix bug where bit-shaved, instantaneous binary output in History was modifying the original export state passed
-
 ### Removed
 
 ### Deprecated
+
+## [2.44.1] - 2024-03-19
+
+### Fixed
+
+- Fix bug where bit-shaved, instantaneous binary output in History was modifying the original export state passed
 
 ## [2.44.0] - 2024-02-08
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ endif ()
 
 project (
   MAPL
-  VERSION 2.44.0
+  VERSION 2.44.1
   LANGUAGES Fortran CXX C)  # Note - CXX is required for ESMF
 
 # Set the possible values of build type for cmake-gui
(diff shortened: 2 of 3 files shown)


atrayano (approved):

weiyuan-jiang (approved):

tclune (dismissed):
@bena-nasa    I think this needs to be changed to **nonzero** diff.   If I understood correctly it really can change answers.

Also ...  does this possibly impact the results that the modeling team reported yesterday?  Were we shaving one of their fields or only a diagnostic?

mathomp4:
Converting to draft. Per @biljanaorescanin the debug runs of the LDAS are throwing a failure:

https://github.com/GEOS-ESM/GEOSldas/pull/734#issuecomment-2007063651

Intel:
```
forrtl: severe (408): fort: (3): Subscript https://github.com/GEOS-ESM/GEOSldas/issues/1 of the array DOIT has value -1 which is less than the lower bound of 1

Image PC Routine Line Source
libMAPL.base.so 00002B36FD528A78 biniomod_mp_mapl_ 1988 BinIO.F90
libMAPL.history.s 00002B36FAC560FA mapl_historygridc 3641 MAPL_HistoryGridComp.F90
```
GNU:
```
At line 1988 of file /discover/nobackup/borescan/SystemTests/builds/LDAS_GNUCONUS/CURRENT/GEOSldas/src/Shared/@MAPL/base/BinIO.F90
Fortran runtime error: Index '-1' of dimension 1 of array 'doit' below lower bound of 1
```

I suppose the "good" thing is it's the same error it seems.

mathomp4:
> @bena-nasa I think this needs to be changed to **nonzero** diff. If I understood correctly it really can change answers.
> 
> Also ... does this possibly impact the results that the modeling team reported yesterday? Were we shaving one of their fields or only a diagnostic?

@tclune I don't think anyone in the AGCM world operates in such a way to trigger this. But I'll work with @bena-nasa to get specifics on when this could be triggered so we can let people know in the release notes and changelog.

bena-nasa:
@biljanaorescanin 
Matt showed me the error you were seeing in the debug run. I will look into this.  Hang tight

mathomp4:
The CI failure looks to be some weird issue on their side. I'm re-launching

biljanaorescanin:
With last commit all my land tests passed.

bena-nasa:
I did some spot checking of the output files by using that converter, made sure that the shaved vs non-shaved output was right in that each variable was "right" in that they were basically the same modulo the bit shaving. I.E. I didn't do something like somehow write the variables in the wrong order in the binary file in say the shaved path.

PR #2664, opened 2024-03-21
Feature/tclune#2663 port to nag 7.201

tclune:
Compiles with NAG 7.201


mathomp4 (approved):

tclune:
@mathomp4 - Apparently this fixes 7.2.01 but then breaks 7.1.40 in precisely the same way.    I think we can probably abandon 7.1 until NAG provides a fix for 7.2.  (Sort of surprised we've not seen one yet.)

PR #2670, opened 2024-03-22
Feature/bmauer/fixes #2440

### LINKED ISSUE #2440: In HISTORY, allow arithmetic involving a bundle field
HISTORY supports the output of bundle fields, it also supports arithmetic on fields, but it does not allow these two in combination,  i.e.  it does not allow arithmetic involving bundle fields.  Could this feature be added?  An example "use case" would be the units conversion of Increment bundle fields.  Thanks for considering this.

bena-nasa:
Fixes issue #2440 
This allows one to do arithmetic expressions with bundles.
How, still no arithmetic allowed with any collection that has the split field option.
I also added some new features to ExtDataDriver.x to add a bundle to the export state so I could debug this

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Allow arithmetic operations to be performed on fields from bundles in History
 - Adapted subroutine RegridVector from GriddedIO.F90 to MAPL_EpochSwathMod.F90 (changing class name for this)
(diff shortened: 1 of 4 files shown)


atrayano (dismissed):

mathomp4:
Looks like exactly *one* ExtData case is failing? Interesting.

bena-nasa:
> Looks like exactly _one_ ExtData case is failing? Interesting.

Yep, I just reproduced, case18, that's weird. I'l investigate.

mathomp4 (approved):

PR #2684, opened 2024-03-27
fixes #2683

### LINKED ISSUE #2683: Multi-dimension coordinate variables are not picking up the deflate level of a collection in History
A user was asking why the cubed-sphere files produced by History were so big compared to a lat-lon output file given a similar number of variables and comparable number of total grid points.  The answer is that the cubed-sphere files have both the centers and corners in the file, in double precision. 

But one thing we did notice, and the point of the issue, is that if the collection set a non-zero deflate level, while the variables were compressed in the output file, the center and corner coordinates were not being compressed.

So this was clearly an oversight on my part and should be fixed. There seems no reason that if the variables are being compressed, that the coordinates should not either for consistency. I've already tested this so just making an issue for the PR.

bena-nasa:
This fixes #2683
It was a silly oversight on my part. If the variables in a collection are compressed, compress the coordinate variables too, just silly not too...

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix inconsistency in History output so that multi-dimensional coordinate variables are also compressed if requested in the
-  collection
+- Fix inconsistency in History output so that multi-dimensional coordinate variables are also compressed if requested in the collection
 - Minor workaround to enable NAG 7.2.01 to compile.  (Reproducer submitted to NAG.)
 - Fixed bug with split restart files
 - Removed unnecessary memory allocation for tile reads. This is critical for high res runs on SCU17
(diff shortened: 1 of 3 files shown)


tclune (changes requested):
I think we need to do this as a hotfix on whichever release of MAPL R21C is using.

Please check.

mathomp4:
> I think we need to do this as a hotfix on whichever release of MAPL R21C is using.
> 
> Please check.

@tclune R21C has a branch here called `R21C`. It's essentially v2.35.3+more. This is the branch naming scheme preferred by the R21C folks (you'll see the branch in many places).

bena-nasa:
> I think we need to do this as a hotfix on whichever release of MAPL R21C is using.
> 
> Please check.

@tclune 

Don't we also want this in develop so it gets into a future release? That's why I made this here.

I know we also will need to do this as a patch for the MAPL R21C version if that is what Amal wants. I will check on that of course.

bena-nasa:
I was just testing this in ExtDataDriver.x, thought this was a harmless PR, but just tried in model and things crashed. I've put a DNA on for now while I debug

mathomp4:
> I think we need to do this as a hotfix on whichever release of MAPL R21C is using.
> 
> Please check.

Superseded by #2689

PR #2695, opened 2024-03-29
Mapl3/tclune/per component clock

tclune:
Tested with NAG 7.2.01 and gfortran 13.2 on OSX.  Unit tests only.

### DIFF
--- a/generic3g/registry/HierarchicalRegistry.F90
+++ b/generic3g/registry/HierarchicalRegistry.F90
@@ -454,6 +454,8 @@ contains
 
    ! "this" is _source_ registry
    subroutine add_state_extension(this, extension_pt, src_spec, extension, source_coupler, rc)
+      use mapl3g_ESMF_Subset, only: ESMF_Clock
+      use mapl3g_MultiState
       class(HierarchicalRegistry), target, intent(inout) :: this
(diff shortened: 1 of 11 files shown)


generic3g/registry/HierarchicalRegistry.F90
```diff
@@ -466,14 +468,17 @@ subroutine add_state_extension(this, extension_pt, src_spec, extension, source_c
...
+
       action = src_spec%make_action(extension, _RC)
       new_coupler = make_coupler(action, source_coupler, _RC)
       ! Need to ensure the stored copy of driver is kept and others are just pointers.
       allocate(new_driver)
       call this%export_couplers%insert(extension_pt, new_driver)
       deallocate(new_driver)
       new_driver => this%export_couplers%of(extension_pt)
-      new_driver = GriddedComponentDriver(new_coupler)
+      ! TODO: need to cretae clock and multi-state.  But this is the wrong layer for such a thing.
```
atrayano, line 480:
```suggestion
      ! TODO: need to create clock and multi-state.  But this is the wrong layer for such a thing.
```

atrayano (approved):

atrayano (approved):

PR #2712, opened 2024-04-05
Updates toward GCM v12-rc1

sdrabenh:
## Description
These are updates from @wputman towards the release of GCM v12. Three files required hand-resolving of merge conflicts:
   1. base/FileIOShared.F90
   2. base/MAPL_EsmfRegridder.F90
   3. gridcomps/ExtData2G/ExtDataGridCompNG.F90
 
I would request a careful review of these changes from those more familiar with the concurrent bug fixes, namely @bena-nasa and @mathomp4 

## Types of changes

## Checklist:


base/FileIOShared.F90
```diff
@@ -599,11 +599,11 @@ subroutine ArrDescrCreateWriterComm(arrdes, full_comm, num_writers, rc)
        integer, intent(in) :: num_writers
        integer, optional, intent(out) :: rc
 
-       integer :: status, nx, ny, color, ny_by_writers, myid, j, writer_rank
+       integer :: status, nx, ny, color, ny_by_writers, myid, j
 
        nx = size(arrdes%i1)
        ny = size(arrdes%j1)
-       _ASSERT(num_writers <= ny,'num writers must be less or equal to than NY')
+       _ASSERT(num_writers < ny,'num writers must be less than NY')
```
mathomp4, line 606: I'm pretty sure this has to be `<=` here and not just `<`. I believe the SCM will fall apart otherwise.

base/MAPL_EsmfRegridder.F90
```diff
@@ -1468,6 +1468,7 @@ subroutine create_route_handle(this, kind, rc)
            rh_file_exists = .false.
         end if
         if (rh_file_exists) then
+           if (mapl_am_I_root()) write(*,*)"bmaa reading "//trim(rh_file)
```
mathomp4, line 1471: If we want this print here, it should be a logger call...

base/NCIO.F90
```diff
@@ -1125,12 +1127,18 @@ subroutine MAPL_VarWriteNCpar_R4_2d(formatter, name, A, ARRDES, lev, offset2, oC
...
        call mpi_gatherv( a, size(a), MPI_REAL, recvbuf, recvcounts, displs, MPI_REAL, &
                       0, arrdes%iogathercomm, status )
        _VERIFY(STATUS)
+       t1 = mpi_wtime()
+       call mpi_allreduce(t1-t0,maxtime,1, MPI_REAL, MPI_MAX, arrdes%iogathercomm, status )
+       _VERIFY(STATUS)
 
        if(myiorank==0) then
 
+         !write(*,'(A,E11.4)') 'Max time spent in R4_2d mpi_gatherv (seconds) ', maxtime
```
mathomp4, line 1140: This should be deleted or turned into a logger call.

base/NCIO.F90
```diff
@@ -2701,12 +2711,18 @@ subroutine MAPL_VarWriteNCpar_R8_2d(formatter, name, A, ARRDES, lev, offset2, oC
...
        call mpi_gatherv( a, size(a), MPI_DOUBLE_PRECISION, recvbuf, recvcounts, displs, &
                          MPI_DOUBLE_PRECISION, 0, arrdes%iogathercomm, status )
        _VERIFY(STATUS)
+       t1 = mpi_wtime()
+       call mpi_allreduce(t1-t0,maxtime,1, MPI_REAL, MPI_MAX, arrdes%iogathercomm, status )
+       _VERIFY(STATUS)
 
        if(myiorank==0) then
 
+         !write(*,'(A,E11.4)') 'Max time spent in R8_2d mpi_gatherv (seconds) ', maxtime
```
mathomp4, line 2724: This should be deleted or turned into a logger call.

base/MAPL_EsmfRegridder.F90
bena-nasa: We don't have access to the logger here.

base/NCIO.F90
bena-nasa: I deleted it

base/FileIOShared.F90
bena-nasa: yep I fixed this

base/NCIO.F90
bena-nasa: removed

base/MAPL_EsmfRegridder.F90
mathomp4: Hmm. Well, if we don't have access to logger, then if nothing else the print should be "generalized". No initials, etc. but make it look like a "MAPL" print.

bena-nasa: I changed it so it's just the filename and the message

mathomp4: Ah. Okay. I guess I thought github would update this comment thing. I guess not!

mathomp4:
Well, as far as I can see, this PR is fine for MAPL and GEOS. I tested it and it is zero-diff. Obviously the other branches have issues, but the changes here are safe. We can pull in whenever.

base/MAPL_EsmfRegridder.F90
tclune: And certainly without anyone's initials.

tclune (commented):
I found nothing problematic beyond the items already highlighted by @mathomp4.

base/MAPL_EsmfRegridder.F90
```diff
@@ -1457,7 +1458,7 @@ subroutine create_route_handle(this, kind, rc)
      spec = this%get_spec()
 
      if (route_handles%count(spec) == 0) then  ! new route_handle
-        file_weights = IAND(spec%hints,REGRID_HINT_FILE_WEIGHTS) /= 0
+        file_weights = .true. !IAND(spec%hints,REGRID_HINT_FILE_WEIGHTS) /= 0
```
mathomp4, line 1461: @bena-nasa Do you know what this would do?

gridcomps/ExtData2G/ExtDataGridCompNG.F90
```diff
@@ -108,7 +109,7 @@ MODULE MAPL_ExtDataGridComp2G
                                           !! are not actually required
      type(ESMF_Config)    :: CF
      logical              :: active = .true.
-     logical              :: file_weights = .false.
+     logical              :: file_weights = .true.
```
mathomp4, line 112: Do we want this on all the time?

mathomp4:
Note: The CI failure is real. I am honestly flabbergasted that GNU Debug run worked. I guess it says something for the robustness of MAPL or ESMF or something that the internal state could be made.

The issue is a very real bug in the Solar GridComp. I can maybe see ways around it, but I'd probably need @dr0cloud help to figure out the "best" way as he knows RRTMG internals better than I do (note, not sure there is a "right" way, but maybe there is a way).

base/MAPL_EsmfRegridder.F90
bena-nasa: We 100% do not want this @mathomp4

gridcomps/ExtData2G/ExtDataGridCompNG.F90
bena-nasa: @mathomp4 this is at least debatable, although I'm surprised Bill had this given the issues with this I still have not been able to reproduce with the reading/writing of the weights at "Bill" scale

PR #2730, opened 2024-04-11
Mapl3/tclune/#2720 set geom

tclune:
## Types of change(s)
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist
- [ ] Tested this change with a run of GEOSgcm
- [x] Ran the Unit Tests (`make tests`)

## Description
MAPL_SetGeom() should now work from user gridcomps.

PR #2731, opened 2024-04-11
Baseline commit for beginning of History development

bena-nasa:
@tclune
@darianboggs 

This baseline PR implemented the following:

1. Updates to Cap3g I found were needed
2. Updates to History3G to get it working to the point that you can add a collection gridcomp and set it's geometry
3. Add a generic gridcomp directory underneath grid comps
4. Make sure I can run Cap and History via GEOS.x
5. Add a non pf unit test that just makes sure this can run in cap3g. This tests time steps cap, and in History for now just adds a geometry to the History collection.

I did confirm that the GEOS.x works, I cap pass the cap.yaml to it and it seems to work.
I added an option to CapGridComp to disable ExtData for now since it's vaporware

Issues:
There were several.
We need to figure out how to properly initialize the logger so it can be used in History etc...
Since ESMF does not support ISO strings for TImeIntervals yet I borrowed some code to parse and set the time interval in Cap.F90, had to do something here...

I've probably chosen wrong names for things (like the new directory I added under gridcomps).
I just borrowed how to run the test in cap from the extdata tests because that's where I knew to look. Probably not what we want to do but that's what I knew how to do.

## Types of change(s)

## Checklist


gridcomps/History3G/HistoryCollectionGridComp_private.F90
```diff
@@ -31,4 +33,31 @@ function make_geom(hconfig, rc) result(geom)
       _RETURN(_SUCCESS)
    end function make_geom
 
+   !subroutine make_import_state(gridcomp, hconfig, rc)
```
tclune, line 36: This procedure will need a better name.  (I know it is commented out for now.)   It is not making the ESMF Import State per se.  Rather it is registering import items.  So maybe `register_imports()`?

gridcomps/History3G/HistoryGridComp.F90
```diff
@@ -49,12 +49,11 @@ subroutine setServices(gridcomp, rc)
...
       do while (ESMF_HConfigIterLoop(iter, iter_begin, iter_end, rc=status))
          _VERIFY(status)
-
-         collection_name = ESMF_HConfigAsStringMapKey(iter, _RC)
+         collection_name = ESMF_HConfigAsString(iter, _RC)
          child_hconfig = make_child_hconfig(hconfig, collection_name)
          child_name = make_child_name(collection_name, _RC)
          call MAPL_AddChild(gridcomp, child_name, user_setservices(collection_setServices), child_hconfig, _RC)
-         call ESMF_HConfigDestroy(child_hconfig, _RC)
+         !call ESMF_HConfigDestroy(child_hconfig, _RC)
```
tclune, line 56: Still debating myself.      It is a general rule that objects should be destroyed in the same scope (or at the same level) where they are created.   So I'm strongly tempted to have the child create a deep copy for using.

gridcomps/cap3g/Cap.F90
```diff
@@ -17,6 +17,7 @@ module mapl3g_Cap
 
 
    subroutine MAPL_run_driver(hconfig, unusable, rc)
+      USE MAPL_ApplicationSupport
```
tclune, line 20: Definitely want to find a better name for this module ...

gridcomps/cap3g/Cap.F90
```diff
@@ -62,16 +66,22 @@ function create_clock(hconfig, rc) result(clock)
       type(ESMF_Time) :: startTime, stopTime, end_of_segment
       type(ESMF_TimeInterval) :: timeStep, segment_duration
       type(ESMF_HConfig) :: clock_config
+      type(ESMF_Calendar) :: calendar
```
tclune, line 69:
```suggestion
```
Not needed.

gridcomps/cap3g/Cap.F90
```diff
@@ -62,16 +66,22 @@ function create_clock(hconfig, rc) result(clock)
...
       
       clock_config = ESMF_HConfigCreateAt(hconfig, keystring='clock', _RC)
 
+      calendar = ESMF_CalendarCreate(ESMF_CALKIND_GREGORIAN, name='CapCal', _RC)
```
tclune, line 73:
```suggestion
```
Not needed.

gridcomps/cap3g/Cap.F90
```diff
@@ -62,16 +66,22 @@ function create_clock(hconfig, rc) result(clock)
...
+      call ESMF_CalendarSetDefault(ESMF_CALKIND_GREGORIAN,_RC)
       call set_time(startTime, 'start', clock_config, _RC)
+      call ESMF_TimePrint(startTime, options='string', prestring='start time set: ' ,_RC)
```
tclune, line 76: Not needed for this sweep, but this should go through logger.

gridcomps/History3G/HistoryGridComp.F90
bena-nasa: @tclune I agree, if you can add that we can restore the destroy.

gridcomps/History3G/HistoryCollectionGridComp_private.F90
bena-nasa: @tclune, ok , nothing has been implemented yet so we can call it whatever we want.

gridcomps/cap3g/Cap.F90
```diff
@@ -84,11 +94,73 @@ subroutine set_time_interval(interval, key, hconfig, rc)
...
+            read(time_string(lpos+1:cpos-1),*)min
+            lpos = cpos
+         end if
+         cpos = index(time_string,'S')
+         if (cpos /= 0) then
+            read(time_string(lpos+1:cpos-1),*)sec
+            lpos = cpos
+         end if
+      end if
+      call ESMF_TimeIntervalSet(interval, yy=year, mm=month, d=day, h=hour, m=min, s=sec,_RC)
```
tclune, line 163: @darianboggs should have a procedure for this if ESMF does not provide it directly.  (I think we opened a ticket with them.)

gridcomps/cap3g/CapGridComp.F90
```diff
@@ -51,15 +44,23 @@ subroutine setServices(gridcomp, rc)
...
+      call MAPL_ResourceGet(gridcomp, keystring='root_name', value=cap%root_name, _RC)
+      call MAPL_ResourceGet(gridcomp, keystring='history_name', value=cap%history_name, default='HIST', _RC)
+
+      if (cap%run_extdata) then 
+         call MAPL_ConnectAll(gridcomp, src_comp=cap%extdata_name, dst_comp=cap%root_name, _RC)
+      end if
+      if (cap%run_history) then
+         !call MAPL_ConnectAll(gridcomp, src_comp=cap%root_name, dst_comp=cap%history_name, _RC)
+         outer_meta => get_outer_meta_from_inner_gc(gridcomp,_RC)
+         call outer_meta%connect_all(cap%root_name, cap%history_name, _RC)
```
tclune, line 62:
```suggestion
         call MAPL_ConnectAll(gridcomp, src_comp=cap%root_name, dst_comp=cap%history_name, _RC)
```
Test again this way to see if it now works.

gridcomps/generic_gridcomps/SimpleLeafGridComp.F90
```diff
@@ -0,0 +1,85 @@
+#include "MAPL_Generic.h"
+
+module mapl3g_SimpleLeafGridComp
```
tclune, line 3: My intent here was for you to copy over SimpleParentGridComp and rename to ConfigurableGridComp.

And there will be only one, so the directory itself should be `gridcomps/configurable`

tclune (changes requested):
See inline comments.

gridcomps/cap3g/CapGridComp.F90
bena-nasa: @tclune, are you changed this. I still get that error?

gridcomps/generic_gridcomps/SimpleLeafGridComp.F90
bena-nasa: @tclune will the name

gridcomps/cap3g/Cap.F90
bena-nasa: @tclune.  ESMF has not implemented this yet. If @darianboggs has a procedure then let's use that. I didn't know he did

darianboggs: ESMF_TimeIntervalSet does support a ISO8601 datetime string as input. The argument is `timeIntervalString`.

gridcomps/cap3g/CapGridComp.F90
tclune: ok - I should be able to troubleshoot once this is merged.

gridcomps/generic_gridcomps/SimpleLeafGridComp.F90
bena-nasa: @tclune, named changed

gridcomps/cap3g/Cap.F90
bena-nasa: @tclune. maybe we should just keep the code I have, I know it works, the ESMF functionality is coming, just not in a release yet.
I can't seem to use anything from MAPL_BaseMod in cap3g (want to figure this too, I don't understand), but @darianboggs thought the ISO8601 that was implemented maybe not have been tested throughly. I know the code I put there works and this is just for development and can be stripped out as when ESMF has this.

mathomp4:
An interesting warning message from CI:
```
PMIx was unable to find a usable compression library
on the system. We will therefore be unable to compress
large data streams. This may result in longer-than-normal
startup times and larger memory footprints. We will
continue, but strongly recommend installing zlib or
a comparable compression library for better user experience.

You can suppress this warning by adding "pcompress_base_silence_warning=1"
to your PMIx MCA default parameter file, or by adding
"PMIX_MCA_pcompress_base_silence_warning=1" to your environment.
```
I'm not sure I've ever seen that. 

However, Intel MPI also threw the same failure, so I don't think this is the issue.

tclune:
The errors here are similar to the errors with my other PR (#2730).    I think the CI is not finding the UDUnits database.

gridcomps/cap3g/Cap.F90
tclune: That's fine.   But open another ticket to repair this in the future.

tclune (approved):

PR #2746, opened 2024-04-15
Use glob function to allow wild character in template

metdyn:
## Types of change(s)

## Checklist

## Description

In the current sampler code (interpolation for each time interval along model run),  reading a filename with template sometimes requires us to rename the observation file name to concord with the template.  For example, 
```
template = MYD04_L2.A%y4%D3.%h2%n2.hdf
obs file = MYD04_L2.A2019262.2305.061.2019263152708.hdf
requiring user to either link the file (or rename it)
MOD04_L2.A2019212.2330.hdf -> MOD04_L2.A2019212.2330.061.2019213072220.hdf
```

This is simplified, if wild character is allowed
```
template = MYD04_L2.A%y4%D3.%h2%n2*.hdf
```
@amdasilva suggested using glob function to enable * character in template.  
This PR is a code implementation.

Example input:
```
PLATFORM.airs_aqua::
    file_name_template:  airs_aqua.%y4%m2%d2T%h2%n2*.nc4
PLATFORM.amsr2_gcom-w1::
    file_name_template: amsr2_gcom-w1.%y4%m2%d2T%h2%n2*.nc4
```

Sampler code output:
```
        sampler: DEBUG: non-exist: filename fid j  :          0
        sampler: DEBUG: non-exist: missing filename: /Users/yyu11/ModelData/JEDI/ioda/Y2019/M08/D01/H00_06/airs_aqua.20190731T2100*.nc4
        sampler: DEBUG: non-exist: filename fid j  :          1
        sampler: DEBUG: non-exist: missing filename: /Users/yyu11/ModelData/JEDI/ioda/Y2019/M08/D01/H00_06/airs_aqua.20190801T0300*.nc4
        sampler: DEBUG: exist: filename fid j      :          0
        sampler: DEBUG: exist: true filename       : /Users/yyu11/ModelData/JEDI/ioda/Y2019/M08/D01/H00_06/amsr2_gcom-w1.20190731T210000Z.nc4
        sampler: DEBUG: exist: filename fid j      :          1
        sampler: DEBUG: exist: true filename       : /Users/yyu11/ModelData/JEDI/ioda/Y2019/M08/D01/H00_06/amsr2_gcom-w1.20190801T030000Z.nc4
        sampler: DEBUG: nobs from input file:        12584
```

### DIFF
--- a/base/MAPL_ObsUtil.F90
+++ b/base/MAPL_ObsUtil.F90
@@ -948,7 +948,6 @@ contains
     filename=""
     if (slen>0) filename(1:slen)=c_filename(1:slen)
 
-    deallocate(c_search_name)
     return
   end subroutine fglob
 
--- a/base/MAPL_ObsUtil.F90
+++ b/base/MAPL_ObsUtil.F90
@@ -10,6 +10,7 @@ module MAPL_ObsUtilMod
   use pFIO_FileMetadataMod, only : FileMetadata
   use pFIO_NetCDF4_FileFormatterMod, only : NetCDF4_FileFormatter
   use, intrinsic :: iso_fortran_env, only: REAL32, REAL64
+  use, intrinsic :: iso_c_binding
   implicit none
   integer, parameter :: mx_ngeoval = 60
   ! GRS80 by Moritz
@@ -58,6 +59,18 @@ module MAPL_ObsUtilMod
      module procedure sort_three_arrays_by_time
      module procedure sort_four_arrays_by_time
   end interface sort_multi_arrays_by_time
+
+  interface
+     function f_call_c_glob(search_name, filename, slen) &
+           & result(stat)    bind(C, name="glob_C")
+       use, intrinsic :: iso_c_binding
+       implicit none
+       integer :: stat
+       character (kind=c_char), intent(in) :: search_name(*)
+       character (kind=c_char), intent(out) :: filename(*)
+       integer, intent(inout) :: slen
+     end function f_call_c_glob
+  end interface
 
 contains
 
@@ -546,12 +559,14 @@ contains
     type(ESMF_TimeInterval) :: dT
     type(ESMF_Time) :: time
     integer :: i, j, u
+    logical :: allow_wild_char
 
     character(len=ESMF_MAXSTR) :: file_template_left
     character(len=ESMF_MAXSTR) :: file_template_right
     character(len=ESMF_MAXSTR) :: filename_left
     character(len=ESMF_MAXSTR) :: filename_full
     character(len=ESMF_MAXSTR) :: filename2
+    character(len=ESMF_MAXSTR) :: filename3
     character(len=ESMF_MAXSTR) :: cmd
 
     call ESMF_TimeIntervalGet(obsfile_interval, s_r8=dT0_s, rc=status)
@@ -565,9 +580,34 @@ contains
 
     ! parse time info
     !
-    call fill_grads_template ( filename, file_template, &
-         experiment_id='', nymd=nymd, nhms=nhms, _RC )
-    inquire(file= trim(filename), EXIST = exist)
+    !
+    allow_wild_char=.true.
+    !
+    j= index(file_template, '*')
+    if (j>0) then
+       ! wild char exist
+       !!print*, 'pos of * in template =', j
+       file_template_left = file_template(1:j-1)
+       call fill_grads_template ( filename_left, file_template_left, &
+            experiment_id='', nymd=nymd, nhms=nhms, _RC )
+       filename2= trim(filename_left)//trim(file_template(j:))
+       call fglob(filename2, filename3, rc=status)
+       if (status==0) then
+          !  the *-file is found
+          exist=.true.
+          filename=trim(filename3)
+       else
+          !  the *-file is not found
+          exist=.false.
+          filename=filename2
+       end if
+       !
+    else
+       ! exact file name
+       call fill_grads_template ( filename, file_template, &
+            experiment_id='', nymd=nymd, nhms=nhms, _RC )
+       inquire(file= trim(filename), EXIST = exist)
+    end if
 
     _RETURN(_SUCCESS)
 
@@ -912,5 +952,28 @@ contains
 121   format (2x, a,10(2x,i8))
 
   end subroutine test_conversion
+
+
+  subroutine fglob(search_name, filename, rc)     ! give the last name
+    character(len=*), intent(in) ::  search_name
+    character(len=*), intent(INOUT) :: filename
+    integer, optional, intent(out)  :: rc
+
+    character(kind=C_CHAR, len=:), allocatable :: c_search_name
+    character(kind=C_CHAR, len=512) :: c_filename
+    integer n, status, slen
+
+    n=len(trim(search_name))
+    allocate(character(kind=C_CHAR,len=n+1) :: c_search_name)
+    c_search_name(1:n)=search_name(1:n)
+    c_search_name(n+1:n+1)=c_null_char
+
+    rc = f_call_c_glob(c_search_name, c_filename, slen)
+    filename=""
+    if (slen>0) filename(1:slen)=c_filename(1:slen)
+
+    deallocate(c_search_name)
+    return
+  end subroutine fglob
 
 end module MAPL_ObsUtilMod
--- a/base/MAPL_ObsUtil.F90
+++ b/base/MAPL_ObsUtil.F90
@@ -574,12 +574,9 @@ contains
 
     ! parse time info
     !
-    !
     allow_wild_char=.true.
     j= index(file_template, '*')
-    if (.NOT. allow_wild_char .AND. j>0) then
-       _FAIL("* is not allowed in template")
-    end if
+    _ASSERT ( j==0 .OR. allow_wild_char, "* is not allowed in template")
     call fill_grads_template ( filename, file_template, &
          experiment_id='', nymd=nymd, nhms=nhms, _RC )
     if (j==0) then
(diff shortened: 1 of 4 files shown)


mathomp4:
Would it be possible to make unit test(s) for this? Only because I'd want to make sure that Intel and GNU both work with this. (Only because Fortran + Strings = 🤷🏼 sometimes as @tclune and @darianboggs can attest to!)

mathomp4:
> > Would it be possible to make unit test(s) for this? Only because I'd want to make sure that Intel and GNU both work with this. (Since Fortran + Strings = 🤷🏼 sometimes as @tclune and @darianboggs can attest to!)
> > Thanks for reminding, I have
> > `  1) gcc-gfortran/12.3.0   2) openmpi/5.0.0rc12   3) Baselibs/7.17.1` on mac,
> > Then under build dir, I run ctest and I get the following, which does not seem to be related to this PR.   Do you get the same output for ctest?
> 
> 97% tests passed, 2 tests failed out of 66
> 
> Label Time Summary: ESSENTIAL = 118.08 sec_proc (56 tests) EXTDATA1G_BIG_TESTS = 1.69 sec_proc (1 test) EXTDATA2G_BIG_TESTS = 1.70 sec_proc (1 test) PERFORMANCE = 2.95 sec_proc (6 tests)
> 
> Total Test time (real) = 125.43 sec
> 
> The following tests FAILED: 25 - ExtData1G_case12 (Failed) 49 - ExtData2G_case12 (Failed)

Ah. You'd want to run:

```
ctest -L 'ESSENTIAL'
```

which just runs the essential cases. This is what happens if you run `make tests` we just run those.

The reason Case 12 fails is that Case 12 requires 216 cores so...big! (I wonder if there's a way to have ctest avoid that if there aren't enough cores available...)

metdyn:
@mathomp4 , that is really helpful!
ctest -L 'ESSENTIAL'
100% tests passed, 0 tests failed out of 56.

I have checked the box in the CheckList   Run the unit test (make tests)

mathomp4:
Did you add a unit test for your code? Or is that sort of...out of reach? Maybe the model is the unit test!

base/MAPL_ObsUtil.F90
```diff
@@ -546,12 +559,14 @@ function get_filename_from_template_use_index (obsfile_start_time, obsfile_inter
     type(ESMF_TimeInterval) :: dT
     type(ESMF_Time) :: time
     integer :: i, j, u
+    logical :: allow_wild_char
```
tclune, line 562: This variable appears to be unused.
```suggestion
```

base/MAPL_ObsUtil.F90
```diff
@@ -565,9 +580,34 @@ function get_filename_from_template_use_index (obsfile_start_time, obsfile_inter
 
     ! parse time info
     !
-    call fill_grads_template ( filename, file_template, &
-         experiment_id='', nymd=nymd, nhms=nhms, _RC )
-    inquire(file= trim(filename), EXIST = exist)
+    !
+    allow_wild_char=.true.
```
tclune, line 584:
```suggestion
```

base/MAPL_ObsUtil.F90
```diff
@@ -565,9 +580,34 @@ function get_filename_from_template_use_index (obsfile_start_time, obsfile_inter
...
+    !
+    j= index(file_template, '*')
+    if (j>0) then
+       ! wild char exist
+       !!print*, 'pos of * in template =', j
+       file_template_left = file_template(1:j-1)
+       call fill_grads_template ( filename_left, file_template_left, &
+            experiment_id='', nymd=nymd, nhms=nhms, _RC )
```
tclune, line 592: Won't `fill_grads_template()` conveniently ignore the `*` at this step?   I.e., can't we fill the template before the `if (j > 0)`?

tclune: I ask because the logic can be greatly simplified if so.

base/MAPL_ObsUtil.F90
```diff
@@ -913,4 +953,27 @@ subroutine test_conversion
...
 
+
+  subroutine fglob(search_name, filename, rc)     ! give the last name
+    character(len=*), intent(in) ::  search_name
+    character(len=*), intent(INOUT) :: filename
+    integer, optional, intent(out)  :: rc
+
+    character(kind=C_CHAR, len=:), allocatable :: c_search_name
+    character(kind=C_CHAR, len=512) :: c_filename
+    integer n, status, slen
```
tclune, line 964:
```suggestion
    integer status, slen
```

base/MAPL_ObsUtil.F90
```diff
@@ -913,4 +953,27 @@ subroutine test_conversion
...
+
+    n=len(trim(search_name))
+    allocate(character(kind=C_CHAR,len=n+1) :: c_search_name)
+    c_search_name(1:n)=search_name(1:n)
+    c_search_name(n+1:n+1)=c_null_char
```
tclune, line 969:
```suggestion
    c_search_name = trim(search_name) // C_NULL_CHAR
```

base/MAPL_ObsUtil.c
```diff
@@ -0,0 +1,31 @@
...
+    return (error);
+
+  int i = 0;
+  for (; globlist.gl_pathv[i] ; ++i)
+    // printf("f = %s\n", globlist.gl_pathv[i]);
+    ;
+  s = globlist.gl_pathv[--i];
+  for (i=0; *(s+i) != '\0'; i++)
+    *(filename+i) = *(s+i);
+  if (i>512) return error;
```
tclune, line 27: Magic number.    Am guessing it is something like max len of file name allowed by the OS?     Please make it a named parameter (or whatever the C equivalent to that is.)

tclune:
> Did you add a unit test for your code? Or is that sort of...out of reach? Maybe the model is the unit test!

Ideally unit tests won't access the filesystem, so a bit tricky here.    Let's not let this get in the way of approval.   But the follow up will be to work with @darianboggs to add some sort of basic sanity check.   I guess create a directory with some empty files with various names and then see what glob returns?

base/MAPL_ObsUtil.F90
metdyn: It is a hard coded logical.

base/MAPL_ObsUtil.F90
metdyn: Oh, I see you are right. Yes, we can fill the template while ignoring the `*`. 
I had  a concern (or a misconception) that `fill_grads_template` can not handle `*`, which leads me to chop the string into two pieces then bind them together.  
I have changed the code to make the logic clean.  Thank you!

base/MAPL_ObsUtil.F90
metdyn: Thanks! This helps!  I have changed the code.

base/MAPL_ObsUtil.c
metdyn: I added `int MAXLEN = 512;` as an upper bound

base/MAPL_ObsUtil.F90
metdyn: the unused variables deleted

base/MAPL_ObsUtil.F90
```diff
@@ -582,31 +576,24 @@ function get_filename_from_template_use_index (obsfile_start_time, obsfile_inter
...
+       _FAIL("* is not allowed in template")
+    end if
+    call fill_grads_template ( filename, file_template, &
+         experiment_id='', nymd=nymd, nhms=nhms, _RC )
+    if (j==0) then
+       ! exact file name
+       inquire(file= trim(filename), EXIST = exist)
+    else
+       ! now filename is:  file*.nc
+       call fglob(filename, filename2, rc=status)
```
tclune, line 590: GitHub won't let me propose the relevant code here because some of the lines are pre-existing.   The bit just below this can be replaced with:
```
   exist = .false.
   if (status == 0) then
      exist = .true.
     filename = trim(filename2)
   end if
```

I.e., eliminate the else block.

I'd like to avoid the nested conditional but not liking the options at first glance, so we'll settle just for this.

base/MAPL_ObsUtil.F90
```diff
@@ -582,31 +576,24 @@ function get_filename_from_template_use_index (obsfile_start_time, obsfile_inter
...
-       ! wild char exist
-       !!print*, 'pos of * in template =', j
-       file_template_left = file_template(1:j-1)
-       call fill_grads_template ( filename_left, file_template_left, &
-            experiment_id='', nymd=nymd, nhms=nhms, _RC )
-       filename2= trim(filename_left)//trim(file_template(j:))
-       call fglob(filename2, filename3, rc=status)
+    if (.NOT. allow_wild_char .AND. j>0) then
+       _FAIL("* is not allowed in template")
+    end if
```
tclune, line 582:
```suggestion
   _ASSERT(j == 0 .or. allow_wild_char, "* is not allowed in template")
```

metdyn: Thank you for the suggestion! Changed.

metdyn:
@mathomp4, thanks for the suggestion.  The full sampler code running with all observation locations will be the test (I have executed a part of it).

base/MAPL_ObsUtil.F90
```diff
@@ -913,4 +933,23 @@ subroutine test_conversion
...
+    character(kind=C_CHAR, len=:), allocatable :: c_search_name
+    character(kind=C_CHAR, len=512) :: c_filename
+    integer slen
+
+    c_search_name = trim(search_name)//C_NULL_CHAR
+    rc = f_call_c_glob(c_search_name, c_filename, slen)
+    filename=""
+    if (slen>0) filename(1:slen)=c_filename(1:slen)
+
+    deallocate(c_search_name)
```
tclune, line 951:
```suggestion
```

tclune (approved):

PR #2751, opened 2024-04-17
Register imports for History3G

darianboggs:
## Types of change(s)

## Checklist

## Description
Registers imports for History3G based on collections defined in History yaml file.

### DIFF
--- a/gridcomps/History3G/HistoryCollectionGridComp_private.F90
+++ b/gridcomps/History3G/HistoryCollectionGridComp_private.F90
@@ -6,17 +6,14 @@ module mapl3g_HistoryCollectionGridComp_private
    use esmf
    use Mapl_ErrorHandling
    use mapl3g_geom_mgr
-   use gftl2_StringStringMap
 
    implicit none
    private
 
    public :: make_geom, register_imports
 
-   interface get_short_names
-      module procedure :: get_short_names_array
-!      module procedure :: get_short_names_vector
-   end interface get_short_names
+   character(len=*), parameter :: VARIABLE_DELIMITER = '.'
+   character(len=*), parameter :: DELIMITER_REPLACEMENT = '/'
 
 contains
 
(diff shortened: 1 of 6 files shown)


gridcomps/History3G/HistoryCollectionGridComp_private.F90
```diff
@@ -33,31 +36,71 @@ function make_geom(hconfig, rc) result(geom)
...
-   !subroutine make_import_state(gridcomp, hconfig, rc)
-      !type(ESMF_GridComp), intent(inout) :: gridcomp
-      !type(ESMF_HConfig), intent(in) :: hconfig
-      !integer, optional, intent(out) :: rc
+   subroutine register_imports(gridcomp, hconfig, rc)
+      type(ESMF_GridComp), intent(inout) :: gridcomp
+      type(ESMF_HConfig), intent(in) :: hconfig
+      integer, optional, intent(out) :: rc
+      character(len=*), parameter :: VAR_LIST_KEY = 'var_list'
+      type(ESMF_HConfigIter) :: iter, iter_begin, iter_end
```
tclune, line 44:
```suggestion
      type(ESMF_HConfigIter) :: iter
```

gridcomps/History3G/HistoryCollectionGridComp_private.F90
```diff
@@ -33,31 +36,71 @@ function make_geom(hconfig, rc) result(geom)
...
-      !type(VariableSpec) :: varspec
-      !integer :: status
+      isScalar = ESMF_HConfigIsScalarMapKey(item, _RC)
+      _ASSERT(isScalar, 'Variable list item does not have a scalar name.')
 
-      !var_list = ESMF_HConfigCreateAt(hconfig, keystring='var_list', _RC)
-      !iter_begin = ESMF_HConfigIterBegin(var_list,_RC)
-      !iter_end = ESMF_HConfigIterEnd(var_list,_RC)
-      !iter = iter_begin
+      isMap = ESMF_HConfigIsMapMapVal(item, _RC)
```
tclune, line 80: I'm a bit confused.  How can you assert that it is both a scalar and a map?

tclune (commented):
Looks good except the inline question that I'd like answered before I opprove.

gridcomps/History3G/HistoryCollectionGridComp_private.F90
darianboggs: The map key is Scalar (String). The map value is a Map. `item` is a Map (HConfig) [Map 1].  It maps the History Collection item name (alias) to a Map [Map 2]. Map 2 maps 'expr' to a String expression, which in the simplest case is a single import. For the simplest case, treating the Map 1 value as a String is simple, and it could be parsed to get the String following 'expr:'. In the future, an item in the collection could have attributes in addition to 'expr'. It was simpler to treat the Map 1 value as a Map, which avoids parsing and allows the additional attributes in the future.

tclune: Ah.  Sorry.

tclune (approved):

PR #2732, opened 2024-04-11
work around #2703 and release pfio memory as early as possible

weiyuan-jiang:
## Types of change(s)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Trivial change (affects only documentation or cleanup)

## Checklist
- [x] Tested this change with a run of GEOSgcm
- [x] Ran the Unit Tests (`make tests`)

## Description
Save data to a integer array pointer to work around #2703 . This PR also releases the pfio memory as early as possible

PR #2752, opened 2024-04-17
use int array to avoid cast

weiyuan-jiang:
## Types of change(s)

## Checklist

## Description
Just  a temporary  branch for comparison


weiyuan-jiang:
@tclune This is my attempt to work around intel bug. It avoids the cast from calss(*) to integer. But somehow it still fails at the same point as before

weiyuan-jiang:
it is merged to #2732

PR #2754, opened 2024-04-17
fixes #2753

### LINKED ISSUE #2753: Complete make_info in FieldSpec.F90
Work was started on setting the ESMF_Info object when the actual ESMF field is created in FieldSpec.F90, this just needs to be completed. Making an issue here to reference.

bena-nasa:
Fixes #2753 

This adds all available information to the ESMF_Info object attached to fields when they are created.
I updated the units tests.
## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/specs/FieldSpec.F90
+++ b/generic3g/specs/FieldSpec.F90
@@ -705,6 +705,7 @@ contains
       integer :: status
       type(ESMF_Info) :: ungridded_dims_info
       type(ESMF_Info) :: vertical_dim_info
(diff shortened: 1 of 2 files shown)


tclune (approved):
We should probably start splitting the test apart into multiple tests to make things a bit more orthogonal when things go wrong.

mathomp4:
You can add "Changelog Skip" to skip the changelog enforcer. Though maybe I should make the enforcer by default skip the changelog if the MAPL3 label is set?

PR #2769, opened 2024-04-19
Fix an error in filename when wild character is introduced into template

metdyn:
## Types of change(s)

## Checklist

## Description
A bug in filename output is found then fixed when reading in 
template=`aist-nr/data/ioda_reshuffle/%y4%m2%d2/geos_atmosphere/omi_aura.%y4%m2%d2T%h2%n2%S2Z.nc4`


tclune (approved):
Approving - but I think it would be interesting to eliminate the nested conditionals that were added here.    But ... other priorities.

PR #2775, opened 2024-04-22
Fixed missing error condition.

tclune:
## Types of change(s)

## Checklist

## Description

Previously, MatchConnection would not fail if any of the destination
points lacked a matching source point.     Essentially a loop would
become degenerate and skip the destination.   Failure would still
happen much later, but in a far less clear manner.

With the change an error of the form below is generated:

   <No matching source found for connection dest:  Virtual{intent: <import>, name: <AGCM/E_1>}>

(A proper unit test is called for.)

### DIFF
--- a/generic3g/connection/MatchConnection.F90
+++ b/generic3g/connection/MatchConnection.F90
@@ -87,7 +87,6 @@ contains
       dst_v_pts = dst_registry%filter(dst_pt%v_pt)
 
       do i = 1, dst_v_pts%size()
-         _HERE, i
          dst_pattern => dst_v_pts%of(i)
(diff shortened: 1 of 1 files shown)


generic3g/connection/MatchConnection.F90
```diff
@@ -86,6 +87,7 @@ recursive subroutine connect(this, registry, rc)
       dst_v_pts = dst_registry%filter(dst_pt%v_pt)
 
       do i = 1, dst_v_pts%size()
+         _HERE, i
```
mathomp4, line 90: Do you want this `_HERE` here?

mathomp4, line 90:
```suggestion
```

bena-nasa (approved):

Issue #2792, opened 2024-04-25
Using PFIO server in new History3G design (stage, done, wait)


bena-nasa:
Current the PFIO server of any variety uses the the `collective_stage_data` method to stage data to be sent to the output server. Then after all staging has been done the client must call 
```
      call o_Clients%done_collective_stage(_RC)
      call o_Clients%post_wait()
```
This presents a problem for the proposed History3G design. The individual collection components would call the `collective_stage_data`, however, the logical place for the `done` and `wait` is in the main History component, not the collection components after all the children have run.

However, we also have no need to call the done and wait if the no component actually wrote anything. But the parent has no way to know this information, that the providence of the child...

Just putting here so we don't forget about this

tclune:
Interesting.  Any idea if we lose any performance in practice by always posting done/wait?     

Presumably old History has some check to see if any collections are active in a time step.   It would not be unreasonable to have such logic in 3G.   The gridcomp is awfully thin as is.  Almost begging for a bit more responsibility.

bena-nasa:
@tclune 
Right in the old History we only call the done/wait if SOME collection wrote that timestep, other it is not called. I would think we would want this in the new History, it's just the the logical place in HistoryGridComp, tNOT he HistoryCollectionGridComp, how would HistoryGridComp know if any of its children actually wrote. I could parse the same file and make the same alarms, but that seems like its just duplicating the same logic...

As for performance, the question is, if we call the done/wait all the time in HistoryGridcomp, I guess what if somehow the front end of the server were still transmitting data to the backend, would that cause the front to be unresponsive and block  on the client side?

@weiyuan-jiang do you know the answer to that offhand?

tclune:
> I could parse the same file and make the same alarms, but that seems like its just duplicating the same logic...

Ideally we will find something a bit more clever, but it may well come down to something like this.

tclune:
> As for performance, the question is, if we call the done/wait all the time in HistoryGridcomp, I guess what if somehow the front end of the server were still transmitting data to the backend, would that cause the front to be unresponsive and block on the client side?

Yes - probably cannot take that risk.

tclune:
Maybe the client itself can decide that it has nothing in that iteration?     Not good probably, but just trying to get more ideas on the table.

bena-nasa:
> Maybe the client itself can decide that it has nothing in that iteration? Not good probably, but just trying to get more ideas on the table.

[...]

Issue #2736, opened 2024-04-12
Cleanup time interval creation in Cap.F90 in cap3g in future


bena-nasa:
Just a reminder to 
The Cap.F90 in MAPL3 needs to create ESMF time intervals. I put some stopgap simple code to parse the ISO time interval. This should be replaced in the future. 
ESMF has implemented the ability to pass an ISO duration but it is not in a release yet. When this is released we should update this.

The other possibility is the ISO8601 code in MAPL, but that has issues that need resolving as documented in issue #2735

tclune:
done

PR #2800, opened 2024-04-30
Fixes #2793 basic time spec

### LINKED ISSUE #2793: Implement time schema for History3G where possible
Issue for task for History3G, basic time schema, needs to parsed and some actions taken
```
time_spec:
  six_hourly: &six_hourly​
    mode: instantaneous​
    frequency: <ISO duration>  # e.g., P6H​
    ref_time: <ISO time>       # e.g., T21H (no date)​
    start: <ISO time>  # e.g., 2020-01-01​
    end:   <ISO time>  # default is end of universe
```
Each collection will reference a time spec. For now the mode is irrelevant until @tclune or someone else implements the ability of the coupling layer to time-average and a way to tell the coupling to do this.

The other items are straightforward:

Given the frequency and ref_time (if no ref time default to 0z) create an alarm we can use to decide when to write in the run method. The alarm should be stored in the collection components private state.

The other two the start and end are simply a time that if you are before or after the time don't write anything.
The default for start is simply when the application initialized, i.e. whatever the initial time of the clock is. The end default it the end of the universe or the closest representable thing in an ESMF time.
Likewise this information will have to go in the collection components private state.

bena-nasa:
This adds some basic timespec stuff to History3G as detailed in #2793 

Adds a basic frequency alarm and start/stop ability for the collection.

For the collection start/stop, if not specified assume it is the begin/end of the clock.

Note the alarm business is a bit mess, borrowed from old History all this to created an alarm that given a first time will ring periodically after that. This is due to the issues with ESMF alarms and is needed unfortunately.

For some reason in the ESMF_Initialized used by the PF units tests I had to set a default calendar, it I did not it would not let me create times that had a year in them.

Finally removed a little routine I added to Cap.F90 to parse an ISO duration to get an ESMF_TimeInterval because I remembered I had this in base and it's now needed two places. When ESMF supports TimeIntervals from strings, this can go away.

## Types of change(s)

## Checklist

### DIFF
--- a/base/TimeStringConversion.F90
+++ b/base/TimeStringConversion.F90
@@ -11,6 +11,7 @@ module MAPL_TimeStringConversion
    public :: string_to_integer_date
    public :: string_to_esmf_time
    public :: string_to_esmf_timeinterval
+   public :: hconfig_to_esmf_timeinterval
 
 contains
 
@@ -238,5 +239,23 @@ contains
       _RETURN(_SUCCESS)
 
    end function string_to_esmf_timeinterval
+
+   function hconfig_to_esmf_timeinterval(hconfig, key, unusable, rc) result(time_interval)
+      type(ESMF_TimeInterval) :: time_interval
+      type(ESMF_HConfig), intent(in) :: hconfig
+      character(len=*), intent(in) :: key
+      class(KeywordEnforcer), optional, intent(in) :: unusable
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+      character(len=:), allocatable :: iso_duration
+
+      _UNUSED_DUMMY(unusable)
+
+      iso_duration = ESMF_HConfigAsString(hconfig, keystring=key, _RC)
+      time_interval = string_to_esmf_timeinterval(iso_duration, _RC)
+
+      _RETURN(_SUCCESS)
+   end function hconfig_to_esmf_timeinterval
 
 end module MAPL_TimeStringConversion
--- a/gridcomps/History3G/HistoryCollectionGridComp.F90
+++ b/gridcomps/History3G/HistoryCollectionGridComp.F90
@@ -16,7 +16,6 @@ module mapl3g_HistoryCollectionGridComp
    type :: HistoryCollectionGridComp
 !#      class(Client), pointer :: client
       type(ESMF_FieldBundle) :: output_bundle
-      type(ESMF_Alarm) :: write_alarm
       type(ESMF_Time) :: start_stop_times(2)
    end type HistoryCollectionGridComp
 
@@ -65,17 +64,19 @@ contains
       type(ESMF_HConfig) :: hconfig
       type(ESMF_Geom) :: geom
       type(ESMF_Alarm) :: alarm
+      character(len=ESMF_MAXSTR) :: name
 
       ! To Do:
       ! - determine run frequencey and offset (save as alarm)
       call MAPL_GridCompGet(gridcomp, hconfig=hconfig, _RC)
+      call ESMF_GridCompGet(gridcomp, name=name, _RC)
 
       _GET_NAMED_PRIVATE_STATE(gridcomp, HistoryCollectionGridComp, PRIVATE_STATE, collection_gridcomp)
       collection_gridcomp%output_bundle = create_output_bundle(hconfig, importState, _RC)
 
       call MAPL_GridCompGet(gridcomp, geom=geom, _RC)
 
-      collection_gridcomp%write_alarm = create_output_alarm(clock, hconfig, _RC)
+      call create_output_alarm(clock, hconfig, trim(name), _RC)
       collection_gridcomp%start_stop_times = set_start_stop_time(clock, hconfig, _RC)
 
       _RETURN(_SUCCESS)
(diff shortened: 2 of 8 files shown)


base/TimeStringConversion.F90
```diff
@@ -239,4 +240,22 @@ function string_to_esmf_timeinterval(time_interval_string,unusable,rc) result(ti
 
    end function string_to_esmf_timeinterval
 
+   function hconfig_to_esmf_timeinterval(hconfig, key, unusable, rc) result(time_interval)
```
tclune, line 243: Hopefully some day, this will be a single call to ESMF, which then can just be inserted where needed?

gridcomps/History3G/HistoryCollectionGridComp_private.F90
```diff
@@ -137,4 +143,75 @@ function create_output_bundle(hconfig, import_state, rc) result(bundle)
...
+
+      call ESMF_ClockGet(clock, currTime=currTime, timeStep=time_interval, startTime = startTime, _RC)
+      int_time = 0 
+
+      time_hconfig = ESMF_HConfigCreateAt(hconfig, keyString='time_spec', _RC)
+
+      has_frequency = ESMF_HConfigIsDefined(time_hconfig, keyString='frequency', _RC)
+      if (has_frequency) then
+         time_interval = hconfig_to_esmf_timeinterval(time_hconfig, 'frequency', _RC)
+      end if
```
tclune, line 168: Seems like we need a `time_interval` even if there is not a 'frequency'.  Should default to heartbeat.

base/TimeStringConversion.F90
bena-nasa: @tclune, ESMF has the ability to take an an ISO duration in a string and turn it into an ESMF_TimeInterval. It is just that this is in not in a release yet. As soon we have a release with this we can get, then yes we can use this, until just have to parse it ourselves. I even made an issue to remind us to do this: https://github.com/GEOS-ESM/MAPL/issues/2736

gridcomps/History3G/HistoryCollectionGridComp.F90
```diff
@@ -98,8 +108,21 @@ subroutine run(gridcomp, importState, exportState, clock, rc)
...
 
       integer :: status
+      type(HistoryCollectionGridComp), pointer :: collection_gridcomp
+      character(*), parameter :: PRIVATE_STATE = "HistoryCollectionGridComp"
+      logical :: time_to_write, run_collection
+      type(ESMF_Time) :: current_time
+    
+      call ESMF_ClockGet(clock, currTime=current_time, _RC) 
+      _GET_NAMED_PRIVATE_STATE(gridcomp, HistoryCollectionGridComp, PRIVATE_STATE, collection_gridcomp)
+      time_to_write = ESMF_AlarmIsRinging(collection_gridcomp%write_alarm, _RC)
```
tclune, line 118: There is a subtle issue lurking here.  Namely,  I had envisioned that the outer/generic gridcomp would manage the basic process of when should a run method be called.

At the same time, we expect this information to be processed a bit differently in HistoryCollection than it will in other gridcomps.

The middle path might be that History adds a `mapl` section to the config that it passes down to HistoryCollection and that lets the generic layer intercede appropriately.

But we can keep your implementation here while we debate it in the more general context.

tclune (approved):
But please answer my inline questions before merging.

gridcomps/History3G/HistoryCollectionGridComp.F90
```diff
@@ -15,6 +16,8 @@ module mapl3g_HistoryCollectionGridComp
    type :: HistoryCollectionGridComp
 !#      class(Client), pointer :: client
       type(ESMF_FieldBundle) :: output_bundle
+      type(ESMF_Alarm) :: write_alarm
```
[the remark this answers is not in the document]
bena-nasa: @tclune, you are right, you can retrieve an individual alarm from a clock. Did not even cross my mind as this not something we have made use of in GEOS, we always store the alarm. I guess I could just pull it from the clock, I can try this.

gridcomps/History3G/HistoryCollectionGridComp_private.F90
bena-nasa: @clune, look more carefully, I do, if the frequency is not in the hconfig, time_interval has already been set to the timeStep of the clock.

gridcomps/History3G/HistoryCollectionGridComp.F90
bena-nasa: @tclune, I had no idea that was your vision

bena-nasa: > @tclune, I had no idea that was your vision
@tclune
And there still this issue related to when things run.
https://github.com/GEOS-ESM/MAPL/issues/2792

gridcomps/History3G/HistoryCollectionGridComp_private.F90
tclune: Ah.  Was looking too hard for `time_interval = ...`.

Generally speaking, I really prefer to put those 2 pieces of logic closer together so that the 2 ways of defining the variable are clear to the reader.   At the very least,  `int_time = 0` should be moved down to the next code chunk where it is overridden in the hconfig.      Then you could still move the `ESMF_ClockGet` call down to just before the `has_frequency = ...` line.

Then, as far as I can tell, startTime is not used and could be eliminated.

gridcomps/History3G/HistoryCollectionGridComp.F90
tclune: Not a vision baked in stone (too mix multiple metaphors), but it seems like it would be good for the generic layer to handle this and avoid repetitive code in each gridcomp.   Esp. since the gridcomp might get it wrong which could be a subtle bug.

gridcomps/History3G/HistoryCollectionGridComp.F90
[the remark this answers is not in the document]
bena-nasa: @tclune, I changed it to eliminate storing alarm in private state.

gridcomps/History3G/HistoryCollectionGridComp_private.F90
bena-nasa: @tclune startTime is used later, line 183

tclune (dismissed):

PR #2814, opened 2024-05-07
IO prototype for History (and checkpoint)

bena-nasa:
Following @tclune slides about the GeomIO, propose an abstract GEOM_IO layer that works with PFIO and a concrete grid class.

Also add a geomCatagorizer function to instantiate the right GEOM_IO concrete class (that only supports grid cause there's nothing else we have besides grids).

Also adds a IOShared layer that has a bunch of utilities procedures that can be used for both history and checkpointing. For example a procedure to take a bundle and create a file metadata object (what we can so far...)

In the future I envision the we can also create a GEOM_IO layer that can implement the old gather/scatter strategy rather than using PFIO as the disk access for checkpoints.

Names are probably are wrong, just wanted to get something committed and a draft.

Note this does work, I can now output multiple timestep files in History3G albeit only 2D gridded files.

## Checklist


GeomIO/Geom_PFIO.F90
```diff
@@ -0,0 +1,92 @@
...
+   use PFIO
+   use mapl3g_geom_mgr
+   use mapl3g_SharedIO
+   implicit none
+   private
+
+   public :: GeomPFIO
+   type, abstract :: GeomPFIO
+      private
+      integer :: collection_id
```
tclune, line 15: Will checkpoint/restart have a collection_id?

tclune: Or rather - do you expect to use this layer with checkpoint restart?

GeomIO/Geom_PFIO.F90
```diff
@@ -0,0 +1,92 @@
...
+      class(GeomPFIO), intent(inout) :: this
+      character(len=*), intent(in) :: filename
+      real, intent(in) :: times
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+      type(ArrayReference) :: ref
+
+      ref = ArrayReference(times)
+      call o_Clients%stage_nondistributed_data(this%collection_id, filename, 'time', ref)
```
tclune, line 69: `o_clients` will need to be accessed differently in MAPL3.   But if you can get this to work now, we can fix it later.

GeomIO/Grid_PFIO.F90
```diff
@@ -0,0 +1,63 @@
...
+      call ESMF_FieldBundleGet(bundle, fieldNameList=field_names, _RC)
+      do i=1,num_fields
+         call ESMF_FieldBundleGet(bundle, field_names(i), field=field, _RC)
+         ! all this logic needs to be generalized
+         call ESMF_FieldGet(field, farrayPtr=ptr2d, _RC)
+         allocate(global_start, source=[1,1,time_index])
+         call ESMF_FieldGet(field, grid=grid, _RC)
+         call MAPL_GridGet(grid, globalCellCountPerDim=global_dim, _RC)
+         allocate(global_count, source=[global_dim(1),global_dim(2),1])
+         call MAPL_GridGetInterior(grid, i1, in, j1, jn)
```
tclune, line 51: For MAPL3 I want to change the argument names here.   `in` is particularly confusing.

GeomIO/Grid_PFIO.F90
```diff
@@ -0,0 +1,63 @@
...
+      integer :: global_dim(3), i1, j1, in, jn
+
+      collection_id = this%get_collection_id()
+      call ESMF_FieldBundleGet(bundle, fieldCount=num_fields, _RC)
+      allocate(field_names(num_fields))
+      call ESMF_FieldBundleGet(bundle, fieldNameList=field_names, _RC)
+      do i=1,num_fields
+         call ESMF_FieldBundleGet(bundle, field_names(i), field=field, _RC)
+         ! all this logic needs to be generalized
+         call ESMF_FieldGet(field, farrayPtr=ptr2d, _RC)
```
tclune, line 46: I think @darianboggs has a procedure that will return  a 1D array ptr regardless of actual rank of the field.   This should avoid conditional logic on the rank.   But more thought is needed about avoiding conditional logic on type (R4, R8, I4, I8, L).

GeomIO/Grid_PFIO.F90
```diff
@@ -0,0 +1,63 @@
...
+         allocate(global_start, source=[1,1,time_index])
+         call ESMF_FieldGet(field, grid=grid, _RC)
+         call MAPL_GridGet(grid, globalCellCountPerDim=global_dim, _RC)
+         allocate(global_count, source=[global_dim(1),global_dim(2),1])
+         call MAPL_GridGetInterior(grid, i1, in, j1, jn)
+         allocate(local_start, source=[i1, j1,1])
+         ref = ArrayReference(ptr2d)
```
tclune, line 53: Maybe @darianboggs can create an interface that returns a `ref` object directly.  At least hides the SELECT CASE deeper down and keeps logic here cleaner.

tclune (approved):
Looks like a good start.

My inline comments are all about long term evolution of this layer.  Nothing actionable at the moment.

GeomIO/Grid_PFIO.F90
bena-nasa: @tclune, yes this needs thoughts, however we also have the issue that for certain grid classes (aka cube the pointer we need to pass the array reference gets that extra dimensions, hence the proliferation of methods in the old grid factory...

GeomIO/Geom_PFIO.F90
bena-nasa: @tclune what's the right way to do this in MAPL3? Do you have an example somewhere?

GeomIO/Grid_PFIO.F90
bena-nasa: @tclune, remember the cube, the reference needs to get an extra dimension for certain grid classes so there are some things to throw this for a loop. This absolutely needs generalization as the comments said but there are lurking issues.

GeomIO/Geom_PFIO.F90
tclune: No.  :-)

The issue is that in MAPL2 we ended up with a global (module) variable for the clients.  And I would like to avoid that.

At the very least we can move this to an accessor on the MaplFramework singleton that I started as part of the GEOS.F90 PR.   Am doing work on that this week to allow our remote server (and an MIT server).   Which brings me closer to the issue I'm highlighting.

Long term, I may pass the MAPL_OBJECT down through the hierarchy which will necessitate an extra argument in lots of places.    I may be satisfied instead by limiting us to a single global variable (THE_MAPL_OBJECT).

GeomIO/Grid_PFIO.F90
tclune: The idea is that you would separately determine shape (and offsets).   But not add any conditional logic on the rank/type of the Field pointer itself.    I think I see how that will work well for rank.  But type is more problematic, or at least requires a possibly unforgivable sin.

GeomIO/Grid_PFIO.F90
tclune: Agreed.

tclune (approved):

PR #2828, opened 2024-05-13
Break the module mapl3g_ESMF_HConfigUtilities into submodules

JulesKouatchou:
## Types of change(s)

## Checklist

## Description

The module `mapl3g_ESMF_HConfigUtilities` was broken into submodules that are now inside the folder `ESMF_HConfigUtilities`.


tclune (commented):
Unfortunately this one is probably useless because it has too many inner procedures.

Let's discuss if maybe it should be restructured first.

tclune (approved):

PR #2830, opened 2024-05-14
Add debug information to fglob function

metdyn:
## Types of change(s)

## Checklist

## Description
Added _FAIL to sub. fglob, if pathlen exceeds upper bound of filename length (ESMF_MAXSTR). 
I find this could be a potential problem and we need the debug information.


tclune (dismissed):

mathomp4:
Hmm. If we are using `ESMF_MAXSTR` (which is 256) anywhere for a path, you might want to find those and change them to `ESMF_MAXPATHLEN` (which is 1024).

@bena-nasa, @atrayano, and myself can regale you with fun that has happened in the past with paths longer than 256!

tclune (approved):

PR #2831, opened 2024-05-15
Added timer to sampler codes in history directory

metdyn:
## Types of change(s)

## Checklist

## Description
Adding timer to history_gc reviews station sampler slows down the full sampler code. 
This PR completes the Timer task. 
Next PR will correct station sampler code.

### DIFF
--- a/gridcomps/History/MAPL_HistoryGridComp.F90
+++ b/gridcomps/History/MAPL_HistoryGridComp.F90
@@ -2423,7 +2423,6 @@ ENDDO PARSER
           end if
           if (list(n)%timeseries_output) then
              list(n)%trajectory = HistoryTrajectory(cfg,string,clock,genstate=GENSTATE,_RC)
-             !!list(n)%trajectory = HistoryTrajectory(cfg,string,clock,_RC)
              call list(n)%trajectory%initialize(items=list(n)%items,bundle=list(n)%bundle,timeinfo=list(n)%timeInfo,vdata=list(n)%vdata,_RC)
              IntState%stampoffset(n) = list(n)%trajectory%epoch_frequency
           elseif (list(n)%sampler_spec == 'mask') then
@@ -3526,7 +3525,7 @@ ENDDO PARSER
                ! it's tempting to use the variable "oneMonth" but it does not work
                ! instead we compute the differece between
                ! thisMonth and lastMonth and as a new timeInterval
-
+               !
                call ESMF_ClockGet(clock,currTime=current_time,_RC)
                call ESMF_TimeIntervalSet( oneMonth, MM=1, _RC)
                lastMonth = current_time - oneMonth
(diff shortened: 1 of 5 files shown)


gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3451,7 +3451,7 @@ subroutine Run ( gc, import, export, clock, rc )
   ! swath only
    epoch_swath_grid_case: do n=1,nlist
       call MAPL_TimerOn(GENSTATE,trim(list(n)%collection))
-      call MAPL_TimerOn(GENSTATE,"SwathGen")
+      call MAPL_TimerOn(GENSTATE,"SwathRun")
```
tclune, line 3454: What does "Run" mean here?    You are already inside of the Run method of history.  Ideally there would be some more specific term.    How about "SwathRegridAccum"  of if that feels too long,  "SwathRgrdAcc"?

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3543,6 +3543,7 @@ subroutine Run ( gc, import, export, clock, rc )
                list(n)%unit = -1
             end if
          elseif (list(n)%sampler_spec == 'station') then
+            call MAPL_TimerOn(GENSTATE,"Station_preRun")
```
tclune, line 3546: This timer appears to be timing a lgr output and two assignments.  That seems unnecessary.  Please eliminate unless I've misunderstood.

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3692,11 +3695,15 @@ subroutine Run ( gc, import, export, clock, rc )
          end if
 
          if (list(n)%sampler_spec == 'station') then
+            call MAPL_TimerOn(GENSTATE,"StationRun")
```
tclune, line 3698: As above, I suspect "Run" is redundant here and can be dropped.

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3692,11 +3695,15 @@ subroutine Run ( gc, import, export, clock, rc )
...
             call ESMF_ClockGet(clock,currTime=current_time,_RC)
             call list(n)%station_sampler%append_file(current_time,_RC)
+            call MAPL_TimerOff(GENSTATE,"StationRun")
          elseif (list(n)%sampler_spec == 'mask') then
+            call MAPL_TimerOn(GENSTATE,"MaskRun")
```
tclune, line 3703: Again "Run" --> ...

gridcomps/History/MAPL_HistoryGridComp.F90
tclune: Looking at some of the other work below, maybe what you want here is 2 timers.
1. "Swath" for context
2. "RegridAccum" as the nested timer for this specific action, as SwatchRun appears again later.

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -3724,8 +3731,8 @@ subroutine Run ( gc, import, export, clock, rc )
   ! swath only
    epoch_swath_regen_grid: do n=1,nlist
       call MAPL_TimerOn(GENSTATE,trim(list(n)%collection))
-      call MAPL_TimerOn(GENSTATE,"Swath regen")
       if (index(trim(list(n)%output_grid_label), 'SwathGrid') > 0) then
+         call MAPL_TimerOn(GENSTATE,"SwathRun")
```
tclune, line 3735: Is enough work happening here to be worth timing?

tclune, line 3735: 1.  I think the timer should be inside the alarm check
2. "Run" is redundante.
3. Maybe what you want is "Swath" as the outer timer and "destroy" and "create_bundle" subtimers

tclune (approved):

PR #2832, opened 2024-05-15
Add Cubed Sphere grid factory to MAPL3 geom factory

bena-nasa:
Implements a cubed-sphere geometry factory

Could probably use some more tests, but so could the LatLon, but its getting to be a big PR as it is.
## Types of change(s)

## Checklist


geom_mgr/CubedSphere/CubedSphereGeomSpec.F90
```diff
@@ -0,0 +1,118 @@
...
+   real(kind=ESMF_Kind_R8) :: undef_schmidt = 1d15
+   private
+
+   public :: CubedSphereGeomSpec
+   public :: make_CubedSphereGeomSpec
+
+   type, extends(GeomSpec) :: CubedSphereGeomSpec
+      private
+      integer :: im_world
+      type(ESMF_CubedSphereTransform_Args) :: schmidt_parameters
```
tclune, line 17: I wonder if we should point out to the ESMF team that these actually have nothing to do with CS.   We just happen to want to use them for the CS.   I can imagine someone wanting to do the same with an Icosahedral mesh.   (LatLon would seem a bit wonky, but even then could be used to move the pole problem to a location of choice.)

geom_mgr/CubedSphere/CubedSphereGeomSpec_smod.F90
```diff
@@ -0,0 +1,290 @@
...
+         ifound = ifound + 1
+      end if
+      is_stretched = all([has_sfac, has_tlon, has_tlat])
+      consistent = (ifound .eq. 3) .or. (ifound .eq. 0)
+      _ASSERT(consistent, "specfied partial stretch parameters")
+      if (.not. is_stretched) then
+         schmidt_parameters%stretch_factor = undef_schmidt 
+         schmidt_parameters%target_lon= undef_schmidt 
+         schmidt_parameters%target_lat= undef_schmidt 
+      end if
```
tclune, line 111: Not worth the change,  but as an exercise in software engineering choices...

 I suspect it would be a little bit clearer if the default (undef) block is moved to the top and just always performed.  (Inefficient, but ... it is just 3 assignments.   Then for each parameter you return if not found.   The 1st one without an error.  Subsequent ones return an error due to inconsistency.    Eliminates the `ifound = ifound + 1` and such.

geom_mgr/CubedSphere/CubedSphereGeomSpec_smod.F90
```diff
@@ -0,0 +1,290 @@
...
+   end function supports_hconfig_
+
+   logical module function supports_metadata_(this, file_metadata, rc) result(supports)
+      class(CubedSphereGeomSpec), intent(in) :: this
+      type(FileMetadata), intent(in) :: file_metadata
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+
+      supports = file_metadata%has_dimension("Xdim", _RC)
```
tclune, line 283: I'm slightly concerned that this is a bit too lax, at least in theory.  But I'm sure it is fine until we hit the exception.

tclune (dismissed):
I'm approving, but please resolve my inline comments when you've read them.

geom_mgr/tests/Test_CubedSphereGeomFactory.pf
```diff
@@ -0,0 +1,34 @@
...
+   subroutine test_make_from_hconfig(this)
+      class(MpiTestMethod), intent(inout) :: this
+
+      type(ESMF_HConfig) :: hconfig
+      integer :: status
+      type(CubedSphereGeomFactory) :: factory
+      class(GeomSpec), allocatable :: geom_spec
+      type(ESMF_Geom) :: geom
+
+      hconfig = ESMF_HConfigCreate(content="{im_world: 12, nx: 1, ny: 1, target_lon: 34.0}", rc=status)
```
tclune, line 23: You expect this case to fail, if I've understood your implementation.

Did you intend the test to check that it gets the right error message?  If so I can try to point you at an example.  Basically you instead assert a failed message that must coincide.

tclune (dismissed):

geom_mgr/CubedSphere/CubedSphereGeomSpec_smod.F90
bena-nasa: yes, it should have been at least nf. I'll clean this up later

tclune (approved):

PR #2835, opened 2024-05-17
Generic IO for server

bena-nasa:
This makes the server interaction totally generic in the new IO layer for History. Will work for any grid + any ungridded dimensions for any supported type.

Had to make new constructor for array reference. Note the new helper routines are needed and are a bit ugly because this was logic previously stuffed in the grid factory but per @tclune we are trying to avoid a proliferation of methods on the factories. So instead I say, if the grid has 6 tiles I do the messiness needed to the write to our cube format via the server. Still a bit kludgy but I don't see how there wouldn't be a kluge somewhere short of adding the methods that have to do this to the factories or something that is cube specific. This logic is unavoidable somewhere.

At the very least this shows how the actual call to the server can be made to require not a single conditional for rank and type where as the old GriddedIO layer had a if rank==2, if rank==3, and that was it.
This should make writing checkpoints via the server with this layer easy too @pchakraborty 

Note I'm also still using a few helper functions in MAPL_Base (MAPL_GridGet and MAPL_GridGetInterior), both are still are needed for basic operations and introspection of grids.

I have confirmed I can write files on the cubed-sphere with this. This just seemed like a good spot to stop and make a PR on a Friday. Obviously more work to do, those helper procedures should be generalized to work with an ESMF Geom, grid specific now, need to work on this. Other details still to implement and units tests needed.

## Types of change(s)

## Checklist


GeomIO/SharedIO.F90
```diff
@@ -13,9 +14,128 @@ module mapl3g_SharedIO
...
 
+   function create_file_shape(grid, field_shape, rc) result(file_shape)
+      integer, allocatable :: file_shape(:)
+      type(ESMF_Grid), intent(in) :: grid
+      integer, intent(in) :: field_shape(:)
+      integer, intent(out), optional :: rc
+
+      integer :: status, sz, ungr, tile_count
+      call ESMF_GridGet(grid, tileCount=tile_count, _RC)
+      sz = size(field_shape)
```
tclune, line 33: If find `sz` to be a very confusing name here.   We usually use that for the size of arrays, but the size of the _shape_ of an array is its _rank_.     Perhaps you could see your way to replacing it with `n_dims` or something like that?

GeomIO/SharedIO.F90
```diff
@@ -13,9 +14,128 @@ module mapl3g_SharedIO
...
+      ungr = sz - 2
```
tclune, line 34: Perhaps I'm missing the constraints on the call side, but is this always true?

pfio/ArrayReference.F90
```diff
@@ -25,6 +25,7 @@ module pFIO_ArrayReferenceMod
    end type ArrayReference
 
    interface ArrayReference
+      module procedure new_ArrayReference_from_param
```
tclune, line 28: While we are here,  `module` is now superfluous for these.  Please remove in next pass.

GeomIO/SharedIO.F90
```diff
@@ -13,9 +14,128 @@ module mapl3g_SharedIO
...
+      else if (tile_count == 1) then
+         file_shape = field_shape
+      else 
+         _FAIL("unsupported grid")
+      end if
+
+      _RETURN(_SUCCESS)
+   end function create_file_shape
+
+   function create_global_start(grid, field_shape, time_index, rc) result(global_start)
```
tclune, line 48: There is enough duplication in logic in the procedures below that I'm wondering if it might be useful to instead do the following:
1. Introduce a new type that contains global bounds and local offset
2. Have one function that produces this type.

GeomIO/SharedIO.F90
```diff
@@ -13,9 +14,128 @@ module mapl3g_SharedIO
...
+         allocate(global_start(sz+1+tm))
+         global_start(1:sz+1) = 1
+         if (present(time_index)) global_start(sz+2) = time_index
+      else if (tile_count == 1) then
+         allocate(global_start(sz+tm))
+         global_start(1:sz) = 1
+         if (present(time_index)) global_start(sz+1) = time_index
+      else 
+         _FAIL("unsupported grid")
+      end if
```
tclune, line 71: As you might expect, this trips my `else` klaxon ...

Probably none of the options are great, but the duplicated checks on the presence of `time_index` likely pushes toward some sort of refactor.    Could possibly due this as a pair-programming exercise to reduce latency in the likely event multiple iterations are needed.

tclune (changes requested):
See inline comments.

GeomIO/SharedIO.F90
bena-nasa: For grids yes, the rank - 2 is the number ungridded dimensions 

If this were for a locstream the constraint would be -1, for example, so this needs generalization

GeomIO/SharedIO.F90
bena-nasa: Seems reasonable. And needs to be generalized eventually for other ESMF geom types. I'll combine into a single type.

GeomIO/SharedIO.F90
tclune: Ah.  Yes.  I think this will be clearer if "2" is replaced by a parameter "NUM_GRIDDED_DIMS".  Then clearly 
```
num_ungridded_dims = num_dims - NUM_GRIDDED_DIMS
```
Make it obvious to the next person ...

GeomIO/pFIOServerBounds.F90
```diff
@@ -0,0 +1,120 @@
...
+   integer, parameter :: grid_dims = 2
+
+   type :: pFIOServerBounds
+      private
+      integer, allocatable :: local_start(:)
+      integer, allocatable :: global_start(:)
+      integer, allocatable :: global_count(:)
+      integer, allocatable :: file_shape(:)
+   contains
+      procedure :: create_server_bounds
```
tclune, line 23: This is really a factory method.   Probably should not be a type-bound procedure.   (I've made a similar "mistake" throughout my own code.)

If we do want to leave it as a TBP, then I suggest the name "initialize" is more correct.

GeomIO/pFIOServerBounds.F90
```diff
@@ -0,0 +1,120 @@
...
+      integer :: i1, in, j1, jn, tile
+
+      call ESMF_GridGet(grid, tileCount=tile_count, _RC)
+      call MAPL_GridGetInterior(grid, i1,in, j1, jn)
+      call MAPL_GridGet(grid, globalCellCountPerDim=global_dim, _RC)
+      n_dims = size(field_shape)
+      ungrid_dims = n_dims - grid_dims
+      tm = 0
+      if (present(time_index))  tm = 1
+
```
tclune, line 73: I think you want another local variable:  "total_dims".   With that, the ALLOCATE's can be refactored out of the conditionals.   

But this is just the tip of potential refactorings to eliminate what is now a long conditional (with an else).   I'll set up a meeting to pair-refactor this.

tclune (changes requested):
See inline

GeomIO/pFIOServerBounds.F90
bena-nasa: changed to initialize

GeomIO/pFIOServerBounds.F90
bena-nasa: see latest commit

GeomIO/pFIOServerBounds.F90
```diff
@@ -0,0 +1,117 @@
...
+      else if (tile_count == 1) then
+         
+         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2)]
+         this%global_count(1:new_grid_dims) = [global_dim(1), global_dim(2)]
+         this%local_start(1:new_grid_dims) = [i1,j1]
+
+      else
+         _FAIL("unsupported grid")
+      end if
```
tclune, line 110:
```suggestion
      select case (tile_count)
      case(6) ! assume Cubed-Sphere
         tile = 1 + (j1-1)/global_dim(1)
         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2) ,1]
         this%global_count(1:new_grid_dims) =[global_dim(1), global_dim(1), tile_count]
         this%local_start(1:new_grid_dims) = [i1, j1-(tile-1)*global_dim(1), tile]

      case (1) ! non cubed sphere  case
         
         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2)]
         this%global_count(1:new_grid_dims) = [global_dim(1), global_dim(2)]
         this%local_start(1:new_grid_dims) = [i1,j1]

      case default
         _FAIL("unsupported grid")
      end if
```

tclune, line 110:
```suggestion
      end select
```

GeomIO/pFIOServerBounds.F90
```diff
@@ -0,0 +1,117 @@
...
+
+      else if (tile_count == 1) then
+         
+         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2)]
+         this%global_count(1:new_grid_dims) = [global_dim(1), global_dim(2)]
+         this%local_start(1:new_grid_dims) = [i1,j1]
+
+      else
+         _FAIL("unsupported grid")
+      end select
```
tclune, line 110:
```suggestion
      select case (tile_count)
      case (6) then ! Assume cubed-sphere

         tile = 1 + (j1-1)/global_dim(1)
         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2) ,1]
         this%global_count(1:new_grid_dims) =[global_dim(1), global_dim(1), tile_count]
         this%local_start(1:new_grid_dims) = [i1, j1-(tile-1)*global_dim(1), tile]

      case (1) then
         
         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2)]
         this%global_count(1:new_grid_dims) = [global_dim(1), global_dim(2)]
         this%local_start(1:new_grid_dims) = [i1,j1]

      case default
         _FAIL("unsupported grid")
      end select
```

tclune (dismissed):

GeomIO/pFIOServerBounds.F90
```diff
@@ -0,0 +1,118 @@
...
+
+      select case (tile_count)
+      case (6) then ! Assume cubed-sphere
+
+         tile = 1 + (j1-1)/global_dim(1)
+         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2) ,1]
+         this%global_count(1:new_grid_dims) =[global_dim(1), global_dim(1), tile_count]
+         this%local_start(1:new_grid_dims) = [i1, j1-(tile-1)*global_dim(1), tile]
+
+      case (1) then
```
tclune, line 103:
```suggestion
      case (6) ! Assume cubed-sphere

         tile = 1 + (j1-1)/global_dim(1)
         this%file_shape(1:new_grid_dims) = [field_shape(1), field_shape(2) ,1]
         this%global_count(1:new_grid_dims) =[global_dim(1), global_dim(1), tile_count]
         this%local_start(1:new_grid_dims) = [i1, j1-(tile-1)*global_dim(1), tile]

      case (1) 
```

tclune (approved):

Issue #2850, opened 2024-05-28
Stretched grid runs are noisy and error filled...but they run?


mathomp4:
In looking at UXarray stuff (see https://github.com/GEOS-ESM/MAPL/discussions/2844), I decided to try out a C270 stretched grid run. After fixing up some setup issues (see https://github.com/GEOS-ESM/GEOSgcm_App/pull/613), I ran C270 and...so many MAPL errors. But, the thing is, the model doesn't stop, just spits out errors at *every time step*.

For example:
```
 accurate_lat:   0.465029113217626
 edge_lat    :  -0.240701801723663
 edge point  :          238
 Error: It could be
   1)Grid is NOT gnomonic_ed;
   2)lats lons from MAPL_GridGetCorners are NOT accurate (single precision from
 ESMF)
   3)This is a stretched grid which is not yet supported
pe=00080 FAIL at line=02773    Base_Base_implementation.F90             <MAPL_GetGlobalHorzIJIndex cannot handle this grid>
```
and:
```
 accurate_lon:    5.32325421858271
 corner_lon  :    4.87329784856145
 Error: Grid should have pi/18 shift
pe=00000 FAIL at line=02773    Base_Base_implementation.F90             <MAPL_GetGlobalHorzIJIndex cannot handle this grid>
```

As the error says, this seems due to `MAPL_GetGlobalHorzIJIndex` not being able to handle a stretched grid. But, as I also said, the model does run and the world does look like the world:

![T_in_test-fixapp-c270 geosgcm_prog 20200415_180](https://github.com/GEOS-ESM/MAPL/assets/2982494/f70d688a-d3fe-442d-92ae-0b6da0008635)

In talking with @bena-nasa, he said this code is used by point emission code in GOCART (at least?), so I suppose if those aren't handled perfect, the climate won't explode. Aerosols might look funky, but things would still run.

I guess the question is (for @tclune and others to decide): What should MAPL be doing here? The code indicates this is a FAIL and yet the model runs. So we could trap the FAIL, but when someone (@wmputman?) next runs stretched grid, would we essentially "break" that by stopping the run at step one? (Though maybe these error conditions only apply at low-resolution stretched grid?)

The "right" thing is obviously to fix the algorithm to support stretched-grid, but that might take time. 

Maybe changing this from a FAIL to a logger WARNING? I don't know. It's confusing to be spitting out FAIL messages at every step yet things run.

bena-nasa:
[...]

PR #2851, opened 2024-05-28
bubble up error from MAPL_getHorzIJindex function

weiyuan-jiang:
## Types of change(s)

## Checklist

## Description
Stop the program when the grid is not a normal CS grid but compute the IJindex. It is partially address issue #2850 . Eventually, I think the Grid class should have information about stretched and the ij index is computed according to the info the grid provides

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Propgated the error message from MAPL_Horzijindex function
 - fixed a bug in generate_newnxy in MAPL_SwathGridFactory.F90 (NX*NY=Ncore)
 - pFIO Clients don't send "Done" message when there is no request
 - Update `components.yaml`
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -2629,7 +2629,7 @@ contains
     call ESMF_AttributeGet(grid, name='GridType', value=grid_type, _RC)
     if(trim(grid_type) == "Cubed-Sphere") then
 
-      call MAPL_GetGlobalHorzIJIndex(npts, II, JJ, lon=lon, lat=lat, lonR8=lonR8, latR8=latR8, Grid=Grid, rc=rc)
+      call MAPL_GetGlobalHorzIJIndex(npts, II, JJ, lon=lon, lat=lat, lonR8=lonR8, latR8=latR8, Grid=Grid, _RC)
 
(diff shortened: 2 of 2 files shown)


mathomp4:
The main problem is, if we commit this, then we can't run the stretched grid with GOCART, right? And, well, I'm not sure if that's desired. 

We might need to support some sort of "slower"/"dumber" IJ method if running stretched grid.

mathomp4:
I've drafted this PR until @tclune can have a think on it as well.

mathomp4:
I can confirm this definitely works! I ran a stretched case and failed as expected.

mathomp4:
Hmm. GNU is not happy:
```
/root/project/GEOSgcm/src/Shared/@MAPL/base/Base/Base_Base_implementation.F90:3458:33:

 3458 |        latRe = half_pi*sign(1.0, Zz)
      |                                 1
Error: 'b' argument of 'sign' intrinsic at (1) must be the same type and kind as 'a'
make[2]: *** [src/Shared/@MAPL/base/CMakeFiles/MAPL.base.dir/build.make:1094: src/Shared/@MAPL/base/CMakeFiles/MAPL.base.dir/Base/Base_Base_implementation.F90.o] Error 1
```

weiyuan-jiang:
The function for reverse Schmidt transform is added and the indices can be calculated now.  But I have not run the gcm with stretched grid. How can I do the setup and run? @mathomp4

mathomp4:
I just tried a C270 run and:
```
 accurate_lon:    5.32325421858271
 accurate_lat:  -0.483265549030082
 edge_lat    :   0.350599277377354
 edge point  :           30
 Error: It could be
   1)Grid is NOT gnomonic_ed;
   2)lats lons from MAPL_GridGetCorners are NOT accurate (single precision from
 ESMF)
```
and I also saw:
```
 Error: Grid should have pi/18 shift
```

weiyuan-jiang:
This PR also addresses issue #3060

mathomp4:
Update: maybe this is on me. My Debug build worked. I'm going to nuke my release build and build from scratch.

mathomp4:
Okay. It was me. I can run C270 now without errors, etc.

mathomp4:
Oh. I guess I should test with v11. I just did it with v12. 

Also, ping @mmanyin that the stretch grid might be okay again...

tclune (approved):
If it works, it works.

mathomp4:
For posterity, I ran GEOS at C270 before and after this fix. This is the output of `SO2EMVN` before the fix:

stock_volc

and then after the fix:

weiyuan_volc

PR #2856, opened 2024-06-04
Fixes for NVHPC 24.5

mathomp4:
## Types of change(s)

## Checklist

## Description

This PR is collecting changes needed for NVHPC work. Note that none of this has been tested yet. The PR here is to make sure I don't lose it!

Changes:

- Update to `MAPL_GeosatMaskMod` code from @metdyn to echo the style of submodules being done by @JulesKouatchou. Not sure why just this code NVHPC tripped over, but it seems to be okay with this.


tclune (approved):
Looks like most of the issues were against my preferred style of submodule use anyway.

Issue #2847, opened 2024-05-24
Need Update of ACG for Proper Handling of Aerosol Longnames


acollow:
# the issue
For carbonaceous aerosols the State Spec reads 
    *MASS        | kg kg-1      | xyz   | C     |                                | * Aerosol Mass Mixing Ratio

which creates the short name BCMASS and long name CA.bc Aerosol Mass Mixing Ratio. 
# the fix
Please modify the automatic code generator so that the long name reads as "BC Aerosol Mass Mixing Ratio".
Please consult with @amdasilva if any questions.

This is needed for M21C and GEOS FP.

mathomp4:
First, I've added @darianboggs as he is our ACG expert. 

But, I think this is a deliberate choice by GOCART2G (and I hope @amdasilva can weigh in).

This line (say) in [`CA2G_StateSpecs.rc`](https://github.com/GEOS-ESM/GOCART/blob/main/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_StateSpecs.rc):
```
 *EXTCOEF     | m-1          | xyz   | C     | size(self%wavelengths_profile) | * Aerosol Extinction Coefficient
```
when processed by the ACG substitutes the component name for the `*`. And per the instance definitions in [`GOCART2G_GridComp.rc`](https://github.com/GEOS-ESM/GOCART/blob/main/ESMF/GOCART2G_GridComp/GOCART2G_GridComp.rc), the component names are `CA.bc`, `CA.oc`, and `CA.br`.[^1]

The only reason, you get `BCEXTCOEF` in a history output is because it is renamed in `HISTORY.rc` from `CA.bcEXTCOEF` to `BCEXTCOEF`:
```
         'CA.bcEXTCOEF'     , 'CA.bc'    , 'BCEXTCOEF'  ,
```
But, as the ACG currently works, the long name is based on the component name so you get:
```
                BCEXTCOEF:long_name = "CA.bc Aerosol Extinction Coefficient" ;
```

I suppose we could put in a translation table into the ACG so that it knows `CA.bc` is `BC`. But to me that seems to be outside the purview of MAPL as we'd be encoding GOCART-specific information into a generic MAPL script. GOCART itself decided that the name of the black carbon instance of the Carbonaceous aerosol gridded component is `CA.bc` and MAPL is honoring that. 

[...]

PR #2858, opened 2024-06-05
Allow alternative mangle for LONG_NAME

darianboggs:
## Types of change(s)

## Checklist

## Description

Based on user request, this PR adds the capability to supply an alternative prefix for mangling `LONG_NAME` instead of the default `comp_name`. In spec files, `*` can be added to the `SHORT_NAME` and `LONG_NAME`, and this PR allows the alternative prefix.

In order to use this in GEOS, you'll need to update your `CMakeLists.txt` with the new option. For example, this:

```cmake
mapl_acg (${this}   CA2G_StateSpecs.rc
          IMPORT_SPECS EXPORT_SPECS INTERNAL_SPECS
          GET_POINTERS DECLARE_POINTERS
          LONG_NAME_PREFIX "GCsuffix")
```

would tell the ACG to use the variable `GCsuffix` for the `long_name`. 

NOTE1: This variable must be in the GC where this is being used. This works for CA because of [this code](https://github.com/GEOS-ESM/GOCART/blob/6ea78fd79037b31a1dcdd30d8a315f6558d963e4/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90#L128-L134):

```fortran
    if (comp_name(1:5) == 'CA.oc') then
       GCsuffix = 'OC'
    else if (comp_name(1:5) == 'CA.bc') then
       GCsuffix = 'BC'
    else if (comp_name(1:5) == 'CA.br') then
       GCsuffix = 'BR'
    end if
```
which fills in `GCsuffix` for each instance.

NOTE2: This is 0-diff in the sense that the state of the model does not change if you enable this option. But it *will* change both restarts and history output due to the change in metadata.

## Related Issue
#2847


darianboggs:
@mathomp4 will modify the `CMakeLists.txt` to call `MAPL_GridCompSpecs_ACG.py` with the new command-line option as appropriate.

tclune (dismissed):
This seems to be an awfully large change for an otherwise stable bit of code.   Is all of this really just for the parametization or did other bits sneak in here?

darianboggs:
It is a significant change for `LONG_NAME`. Previously, the command-line options were previously for global values, filenames. There was no mechanism to pass those options to into the output process itself. We assumed the process was determined by the code. The `*` was hard-coded to `trim(comp_name)`. One possibility was a global variable that is checked if the part (*option*/*column*) is `Option[LONG_NAME]`. I am trying to avoid checking global variables and to treat each part of the spec (`SHORT_NAME`, `DIMS`, `LONG_NAME`, etc.) the same by giving each an output function (`emit`), with a default emit function and a small set of non-default emit functions that are shared by multiple options. `SHORT_NAME` and `LONG_NAME` shared an emit function. `LONG_NAME` needed it's own emit for this PR, and it needed to use a command-line value for which there was no facility. So I added a parameterized emit function. If another user comes to us to say they want something similar to what Allie requested, we have the capacity now. Only `LONG_NAME` uses this new feature.

mathomp4 (approved):

PR #2860, opened 2024-06-07
Allow alternative mangle for LONG_NAME (R21C)

mathomp4:
## Types of change(s)

## Checklist

Based on user request, this PR adds the capability to supply an alternative prefix for mangling `LONG_NAME` instead of the default `comp_name`. In spec files, `*` can be added to the `SHORT_NAME` and `LONG_NAME`, and this PR allows the alternative prefix.

In order to use this in GEOS, you'll need to update your `CMakeLists.txt` with the new option. For example, this:

```cmake
mapl_acg (${this}   CA2G_StateSpecs.rc
          IMPORT_SPECS EXPORT_SPECS INTERNAL_SPECS
          GET_POINTERS DECLARE_POINTERS
          LONG_NAME_PREFIX "GCsuffix")
```

would tell the ACG to use the variable `GCsuffix` for the `long_name`. 

NOTE1: This variable must be in the GC where this is being used. This works for CA because of [this code](https://github.com/GEOS-ESM/GOCART/blob/6ea78fd79037b31a1dcdd30d8a315f6558d963e4/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90#L128-L134):

```fortran
    if (comp_name(1:5) == 'CA.oc') then
       GCsuffix = 'OC'
    else if (comp_name(1:5) == 'CA.bc') then
       GCsuffix = 'BC'
    else if (comp_name(1:5) == 'CA.br') then
       GCsuffix = 'BR'
    end if
```
which fills in `GCsuffix` for each instance.

NOTE2: This is 0-diff in the sense that the state of the model does not change if you enable this option. But it *will* change both restarts and history output due to the change in metadata.

## Related Issue
#2847


mathomp4:
@acollow Can you try this branch out? You'll want to change the GOCART as said above.

Once you do, please make sure history looks like what you want. Especially look at:

> @mathomp4 , Note that M21C will also need the MAPL patch that corrects the "001" bug.

Please see if it is doing what you want or not. If not, tell us what you want and then @atrayano can get the appropriate fix in

acollow:
@mathomp4, @darianboggs Is it possible to clarify which CMakeLists.txt to add this too?

mathomp4:
> @mathomp4, @darianboggs Is it possible to clarify which CMakeLists.txt to add this too?


@acollow It's the Cmake for whatever component needs this. So in this case, CA. So edit `.../CA2G_GridComp/CMakeLists.txt` and change:
```cmake
mapl_acg (${this}   CA2G_StateSpecs.rc 
          IMPORT_SPECS EXPORT_SPECS INTERNAL_SPECS 
          GET_POINTERS DECLARE_POINTERS)
```
to:
```cmake
mapl_acg (${this}   CA2G_StateSpecs.rc
          IMPORT_SPECS EXPORT_SPECS INTERNAL_SPECS
          GET_POINTERS DECLARE_POINTERS
          LONG_NAME_PREFIX "GCsuffix")
```
where that last line is added.

acollow:
I am confirming that I tested the branch and the carbon long names are now being written in an ideal manner. Thank you! The branch however still has the bug for the dust and sea salt bins. I copied below the output from ncdump on a file I generated showing what should be SS001 and SS002.

	float SS(time, lev, nf, Ydim, Xdim) ;
		SS:_FillValue = 1.e+15f ;
		SS:add_offset = 0.f ;
		SS:coordinates = "lons lats" ;
		SS:fmissing_value = 1.e+15f ;
		SS:grid_mapping = "cubed_sphere" ;
		SS:long_name = "Sea Salt Mixing Ratio (bin 001)" ;
		SS:missing_value = 1.e+15f ;
		SS:regrid_method = "identity" ;
		SS:scale_factor = 1.f ;
		SS:standard_name = "Sea Salt Mixing Ratio (bin 001)" ;
		SS:units = "kg kg-1" ;
		SS:valid_range = -1.e+15f, 1.e+15f ;
		SS:vmax = 1.e+15f ;
		SS:vmin = -1.e+15f ;
	float SS002(time, lev, nf, Ydim, Xdim) ;
		SS002:_FillValue = 1.e+15f ;
		SS002:add_offset = 0.f ;
		SS002:coordinates = "lons lats" ;
		SS002:fmissing_value = 1.e+15f ;
		SS002:grid_mapping = "cubed_sphere" ;
		SS002:long_name = "Sea Salt Mixing Ratio (bin 002)" ;
		SS002:missing_value = 1.e+15f ;
		SS002:regrid_method = "identity" ;
		SS002:scale_factor = 1.f ;
		SS002:standard_name = "Sea Salt Mixing Ratio (bin 002)" ;
		SS002:units = "kg kg-1" ;
		SS002:valid_range = -1.e+15f, 1.e+15f ;
		SS002:vmax = 1.e+15f ;
		SS002:vmin = -1.e+15f ;

atrayano:
@acollow Allie, could you please point me to your History.rc file. I just want to make sure I fully understand the requirements. My dilemma is when the user provides a single alias, do we always append 00n to the alias, or we take the alias value and leave it as is. One quick workaround (but not a fix), is to always provide an alias for the field as in the following example 'DUEM'           , 'DU'        , 'DUEM001;DUEM002;DUEM003;DUEM004;DUEM005'

acollow:
@atrayano, you're right! Neither my HISTORY.rc nor the one for M21C have the aliases. I just ran a quick test and that solved it. Thanks!

atrayano:
@acollow I am happy now! Thanks! The rules for split names are a bit confusing. If you do not have alias(es), then MAPL tries to keep the field name in the output the same. But in case of multiple bins, having two different fields with the same name is not possible, and to disambiguate, we append 00n, but starting at 2 to keep the original name the same in case of a single bin. On the other hand, if you provide enough split-field-aliases, we keep them as is.

mathomp4:
Since @acollow says this works for her I will undraft. Once all is good, I can work with @sdrabenh to pull in and make a new R21C release

elakkraoui (approved):
@mathomp4 , @acollow, I approved this PR but I unless I missed something, I see that Allie still needs to submit another PR with the Cmake update for the change to take effect.

mathomp4:
> @mathomp4 , @acollow, I approved this PR but unless I missed something, I see that Allie still needs to submit another PR with the Cmake update for the change to take effect.

@elakkraoui They can be taken separately, so we can merge this in at any time (same for https://github.com/GEOS-ESM/MAPL/pull/2852).

This just adds the ability for @acollow to make a change in GOCART to enable it.

elakkraoui:
> > @mathomp4 , @acollow, I approved this PR but unless I missed something, I see that Allie still needs to submit another PR with the Cmake update for the change to take effect.
> 
> @elakkraoui They can be taken separately, so we can merge this in at any time (same for #2852).
> 
> This just adds the ability for @acollow to make a change in GOCART to enable it.

Sounds good.

acollow:
Totally spaced that I needed to add the PR to GOCART! Sorry about that! I just added it - https://github.com/GEOS-ESM/GOCART/pull/278

tclune (approved):

PR #2898, opened 2024-07-11
Restart/checkpoint

pchakraborty:
## Types of change(s)

## Checklist

## Description

This PR adds ability to read restarts and write checkpoints. Changes include

- Appropriate hooks in
  - `gridcomps/cap3g/Cap.F90`
  - `generic3g/ComponentDriver.F90`
  - `generic3g/GenericGridComp.F90`
  - `generic3g/GriddedComponentDriver.F90`
  - `generic3g/OuterMetaComponent.F90`

- New code in
  - `generic3g/Restart.F90` - new file
  - `generic3g/OuterMetaComponent/<read/write>_restart.F90`
  - `GeomIO/Grid_PFIO.F90` - new routine `request_data_from_file`

- Added new test for writing restarts - `gridcomps/cap3g/tests/write_restart`

- Replaced `initialize_simple_oserver` in `mapl3g/MaplFramework.F90` with `initialize_simple_servers` that initializes both simple *iserver* and simple *oserver*.

- Rearranged files to remove circular dependency between `GeomIO` and `generic3g`

## TODO

- Add test for reading restarts (need to store restart files)
- Move tests to `generic3g/tests`
- ? Rename Ext/Hist in `pfio` directory to something more generic
- Bootstrap variables if restart files are not found. Currently, a message is printed if a particular restart file is missing
- Test the case of non-simple servers
## Related Issue


mathomp4:
Usual GSI race condition with ADAS.

pchakraborty:
@tclune The restart code now uses the current `GriddedComponentDriver` to read/write restarts instead of the child's. I also extended the `recurse.F90` to include recursion for read/write restarts. A better way may be to create a base class for recursion and extend from it.

GeomIO/pFIOServerBounds.F90
```diff
@@ -114,5 +115,18 @@ subroutine initialize(this, grid, field_shape, time_index, rc)
...
+
+      deallocate(this%file_shape)
+      deallocate(this%global_start)
+      deallocate(this%global_count)
+      deallocate(this%local_start)
+
+      _RETURN(_SUCCESS)
+
+   end subroutine finalize
+
```
tclune, line 130: Do we need an explicit finalizer?  Fortran should deallocate these when the object goes out of scope.

generic3g/OuterMetaComponent.F90
```diff
@@ -277,6 +280,16 @@ module recursive subroutine recurse_(this, phase_idx, rc)
          integer, optional, intent(out) :: rc
       end subroutine recurse_
    
+      module recursive subroutine recurse_read_restart_(this, rc)
```
tclune, line 283: I think the prefix `recurse_` is a bit overkill here.   Almost every procedure in the module should be renamed if this is the style you prefer. (In which case we'll discuss the merits either way.)

generic3g/OuterMetaComponent.F90
```diff
@@ -415,6 +428,14 @@ end subroutine set_entry_point
       module procedure recurse_
    end interface recurse
 
+   interface recurse_read_restart
+      module procedure recurse_read_restart_
+   end interface recurse_read_restart
+
```
tclune, line 434: If you need these to be public then do they need to be type-bound as well?

generic3g/OuterMetaComponent/read_restart.F90
```diff
@@ -14,6 +14,31 @@ module subroutine read_restart(this, importState, exportState, clock, unusable,
...
+      character(:), allocatable :: name
+      type(MultiState) :: states
+      type(ESMF_State) :: internal_state, import_state
+      type(ESMF_Geom) :: geom
+      type(Restart) :: rstrt
+      integer :: status
+
+      driver => this%get_user_gc_driver()
+      name = driver%get_name()
+      if ((name /= "cap") .and. (name /= "HIST")) then
```
tclune, line 29: Need generic method for this conditional.   And in any event the logic should be reversed:
``` if (name == 'cap' .or. name == 'HIST') then
         ... return?
      end if
      ! no else...

File: generic3g/OuterMetaComponent.F90 (hunk 3 above)
Comment (tclune) [reply generic3g/OuterMetaComponent.F90:434]: I think you may have duplicated superfluous bits of code in the existing source.   Please try deleting this and similar constructs and see if things still work.

File: generic3g/OuterMetaComponent.F90 (hunk 2 above)
Comment (tclune) [reply generic3g/OuterMetaComponent.F90:283]: Ignore that comment.  I see now where the name comes from.

The question we should consider is whether we want/need a different "recurse" procedure for each ESMF "run" method, or if we should just generalize the existing one.  I.e., add arguments to the interface for the ESMF_METHOD_FLAG and phase.

File: generic3g/Restart.F90 [hunk 5]
```diff
@@ -0,0 +1,195 @@
...
+      character(len=ESMF_MAXSTR) :: file_name
+      logical :: file_exists
+      integer :: item_count, status
+
+      call ESMF_StateGet(state, itemCount=item_count, _RC)
+      if (item_count > 0) then
+         file_name = trim(this%gc_name) // "_" // trim(state_type) // "_rst.nc4"
+         inquire(file=trim(file_name), exist=file_exists)
+         if (file_exists) then
+            print *, "Reading restart: ", trim(file_name)
```
tclune, line 97: If you think we want the prints long term, please convert to use pFlogger calls.

generic3g/Restart.F90
```diff
@@ -0,0 +1,195 @@
...
+            call this%read_fields_(file_name, state, _RC)
+         else
```
tclune, line 99: Bonus points for eliminating else.

tclune (changes requested):
I think one more iteration.   Can meet tomorrow to discuss.

generic3g/OuterMetaComponent/read_restart.F90
pchakraborty: This is temporary solution till we have a better way to identify a gridcomp that reads/writes a restart.
We are not returning at this point, instead recursing over children.

generic3g/OuterMetaComponent.F90
pchakraborty: Yes, we need a way to generalize the `recurse` procedure. The existing one would only call `initialize` recursively. That can be a separate issue and PR.

generic3g/Restart.F90
tclune: I understand.   I'm tending to agree with a lot of developers that early return is preferable to `else` in terms of code clarity.  When this is not the case, it usually indicates other improvements that can be made to the procudere. (New helper procedure for example.)

generic3g/OuterMetaComponent.F90
tclune: OK - but do not forget ...

generic3g/OuterMetaComponent/read_restart.F90
tclune: I am thinking that this one needs further discussion with @bena-nasa and @atrayano 

Our usual rule-of-thumb is to checkpoint all imports.   But HIST makes a mockery of this decision.   The MAPL2 solution (ignoring the fact that MAPL2 implemented history quite differently), would be for the individual collections to explicitly say not to checkpoint their imports when they advertise them.      The generic layer would then not need to do anything special.

Note that "cap" will generally not have any imports (or internal state) and therefore the check here is actually not needed even now. (If I've understood correctly.)

GeomIO/pFIOServerBounds.F90
pchakraborty: The way to avoid this would be to convert the `initialize` subroutine into a constructor.

pchakraborty: The `initialize` is indeed being used in a loop over fields.

generic3g/Restart.F90
pchakraborty: This is done.

tclune (approved):

Issue #2896, opened 2024-07-11
Untrapped errors in MPI calls


tclune:
Some calls to MPI procedures inside MAPL do not have any followup checks on their ierror return status.
Just add `_VERIFY(ierror)` on following lines.

JulesKouatchou:
Do I have to work with MAPL-v3? I assume that I need to check all the MPI calls and make sure that they are followed with `_VERIFY(ierror)`.

tclune:
No this should be a change on the `develop` branch.

PR #2910, opened 2024-07-16
MAPL-V2: Error checking after mpi calls

JulesKouatchou:
## Types of change(s)

## Checklist

## Description

Whenever possible, we add a `_VERIFY(ierror)` statement after each MPI call.

## Related Issue
This is to address Issue #2896

### DIFF
--- a/shared/MAPL_Throw.F90
+++ b/shared/MAPL_Throw.F90
@@ -60,6 +60,7 @@ contains
       logical :: is_mpi_initialized
 
       call MPI_Initialized(is_mpi_initialized,ierror)
+      _VERIFY(ierror)
   
       base_name = get_base_name(filename)
       if (len(base_name) > FIELD_WIDTH) then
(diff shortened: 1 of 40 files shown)


shared/MAPL_Throw.F90
```diff
@@ -56,10 +56,11 @@ subroutine MAPL_Fail(filename, line, message)
       character(3) :: prefix
       character(:), allocatable :: base_name
       
-      integer :: rank, ierror
+      integer :: rank, ierror, rc
```
tclune, line 59: Yes - this file should be reverted.   `_VERIFY` indirectly calls throw, so cannot use it here ...

shared/SimpleCommSplitter.F90
```diff
@@ -315,6 +315,7 @@ subroutine free_sub_comm(this)
      integer :: ierror
      if (this%is_split) then
         call MPI_Comm_free(this%sub_comm, ierror)
+        !_VERIFY(ierror)
```
tclune, line 318: Why is this one commented out?  I would think it could be used here.

shared/SimpleCommSplitter.F90
```diff
@@ -327,6 +328,7 @@ subroutine assign(this, from)
 
      if (from%is_split) then
        call MPI_Comm_rank(comm, rank, ierror)
+       !_VERIFY(ierror)
```
tclune, line 331: Same comment as above.

PR #2918, opened 2024-07-17
fixes #2909

### LINKED ISSUE #2909: Clean up FieldSpec.F90 in MAPL3 to use helper function to set field to constant
This block of code can be replaced, we already have a helper function to do this for fields in the field utilities.
https://github.com/GEOS-ESM/MAPL/blob/release/MAPL-v3/generic3g/specs/FieldSpec.F90#L242

bena-nasa:
Just trivial change to use the field utilities to replace some code Tom that was redundant with the utilities
## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/specs/FieldSpec.F90
+++ b/generic3g/specs/FieldSpec.F90
@@ -162,7 +162,7 @@ contains
       integer :: status
 
       this%payload = ESMF_FieldEmptyCreate(_RC)
-      _RETURN_UNLESS(allocated(this%geom))  ! mirror 
+      _RETURN_UNLESS(allocated(this%geom))  ! mirror
       call MAPL_FieldEmptySet(this%payload, this%geom, _RC)
 
       _RETURN(ESMF_SUCCESS)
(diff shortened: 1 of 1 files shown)


generic3g/specs/FieldSpec.F90
```diff
@@ -236,60 +236,13 @@ subroutine allocate(this, rc)
       call ESMF_FieldGet(this%payload, status=fstatus, _RC)
       _ASSERT(fstatus == ESMF_FIELDSTATUS_COMPLETE, 'ESMF field status problem.')
       if (allocated(this%default_value)) then
-         call set_field_default(_RC)
+         call FieldSet(this%payload, this%default_value, _RC)
```
tclune, line 239: This looks weird to me.   I usually have camel case just for constructors and this is a subroutine call.

tclune: Am guessing this comes from the "Darian" layer and I did not force camel case there when I should have ...  (will open separate ticket for that.)

tclune (approved):

PR #2924, opened 2024-07-26
Initial placeholder for extdata3g

bena-nasa:
Just very simple start for extdata3g. Not much does at least make the directory and get a few files in place.
Just got some code in place to merge all the individual files into a single hconfig that can be processed. Would have done this in old code but could be done with yafyaml but now why not, will make parsing it easier.
Added a test. That's about it. 
That's as short as I could make the routine to do the merge given the limitations of HConfigSet and HConfigAdd, believe me I didn't it to be that long either but it is what it has to be.

Not a whole lot more that can be done until this "bracket spec" and the ability to have each spec have its own geometry is implemented. 

## Types of change(s)

## Checklist


gridcomps/ExtData3G/tests/Test_ExtDataGridComp.pf
```diff
@@ -0,0 +1,57 @@
+module Test_ExtDataGridComp
```
tclune, line 1: @bena-nasa I have created a new include file that lets us use `_RC` in pFUnit tests.  See `./include/MAPL_TestErr.h`

tclune (approved):

gridcomps/ExtData3G/tests/Test_ExtDataGridComp.pf
bena-nasa: ok, I'll take a look and cleanup in subsequent PR

PR #2929, opened 2024-07-30
Fixes #2928 - Simple refactoring.

### LINKED ISSUE #2928: Fix dependency between StateRegistry and Connection (MAPL3)
Currently StateRegistry has a method `add_connection` which passes itself back to the connection objects `connect` method.   This looks like the Visitor design pattern and is unnecessarily complicated.  The largest smell is that it forces the StateRegistry and Connection (abstract) types to be defined in the same module to allow access to each others' methods.

Should be an easy fix.

tclune:
Lots of small changes - mostly changing which module provides Connection type.

## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/ComponentSpecParser.F90
+++ b/generic3g/ComponentSpecParser.F90
@@ -7,10 +7,10 @@ module mapl3g_ComponentSpecParser
    use mapl3g_UserSetServices
    use mapl_ErrorHandling
    use mapl3g_VariableSpec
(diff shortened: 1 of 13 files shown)


mathomp4 (approved):
Simple for @tclune that is 😄  (My simple refactoring is "changed variable foo to bar")

atrayano (approved):

PR #2940, opened 2024-07-31
Fixes #2926. Update quantization properties

### LINKED ISSUE #2926: MAPL History quantization metadata should echo NCO (aka CF!)
@czender recently released NCO 5.2.7. From the announcement:

```
A. Version 5.2.7 implements the final (we think) CF quantization
metadata convention. The main change is that the keywords and
container variable are now "quantization" instead of
"lossy_compression". 
ncks -7 -v ps,ts --qnt_alg=btr --qnt default=9 --qnt ps=13 --cmp='shf|zst' in.nc out.nc
ncks -m -C -v ps,ts,quantization_info out.nc
netcdf out {
...
   float ps(time,lat,lon) ;
      ps:standard_name = "surface_air_pressure" ;
      ps:units = "Pa" ;
      ps:quantization = "quantization_info" ;
      ps:quantization_nsb = 13 ;
      ps:quantization_maximum_relative_error = 6.103516e-05f ;

    char quantization_info ;
      quantization_info:algorithm = "bitround" ;
      quantization_info:implementation = "NCO version 5.2.7" ;

    float ts(time) ;
      ts:standard_name = "surface_temperature" ;
      ts:units = "K" ;
      ts:quantization = "quantization_info" ;
      ts:quantization_nsb = 9 ;
      ts:quantization_maximum_relative_error = 0.0009765625f ;
} // group /
http://nco.sf.net/nco.html#qnt_alg
http://nco.sf.net/nco.html#qnt
```

A quick look around the CF site leads to:

* Issue: https://github.com/cf-convention/cf-conventions/issues/403
* PR: https://github.com/cf-convention/cf-conventions/pull/519

as being the discussions about this. Reading time!

Since something like this will be in the CF Conventions "soon" (on the scale of CF Convention changes), MAPL should work toward it as well. 

Now, at the moment, if we set, say:
```
  geosgcm_prog.quantize_level: 4,
  geosgcm_prog.quantize_algorithm: 'GranularBR',
```
in History, we get out what I think are the "what comes from netCDF" metadata:
```
        float T(time, lev, lat, lon) ;
                T:_QuantizeGranularBitRoundNumberOfSignificantDigits = 4 ;
...
```

So not quite the same. I mean, some info is there but we should issue our own metadata writes to match the CF Draft including `quantization_info`...container? (Not sure.)

Some questions:

1. What do we do about our home-grown bit-shaving? It is sort of like BitGroom, but I think the implementation is slightly different. Does CF have a prescribed way of specifying a new algorithm? 
2. At the moment, MAPL can only apply quantization on a per-collection basis, i.e., if like above, *everything* in geosgcm_prog gets GranularBR at 4. This is often not ideal (some fields need more precision than others). Eventually this should probably be at a per-*field* basis but something like that will probably need to wait for the yaml-based MAPL3/History3G. But, once we can do per-field quantization, the `nsd` is per-variable metadata, but if a collection uses different types of quantization[^1], I assume one just makes `quantization_info1`, `quantization_info2`, etc? We'll need to read the draft conventions to figure it out.

[^1]: This might be moot as perhaps using different quantization algorithms per file might just be something we just don't want to support? It does seem odd one might use BitGroom and GranularBR in the same file for different variables.

mathomp4:
## Types of change(s)

## Checklist

## Description

This is an attempt to add CF quantization properties echoing the work done by @czender in NCO.

With these changes you get out, say, for `bitround`:

```
        float H(time, lev, lat, lon) ;
                H:_QuantizeBitRoundNumberOfSignificantBits = 4 ;
...
                H:quantization = "quantization_info" ;
                H:quantization_maximum_relative_error = 0.03125f ;
                H:quantization_nsb = 4 ;
...
        char quantization_info ;
                quantization_info:algorithm = "bitround" ;
                quantization_info:implementation = "libnetcdf 4.9.2 of Apr 24 2024 11:29:22 $" ;
```

Now for, say, `granular bitround` you get:
```
        float H(time, lev, lat, lon) ;
                H:_QuantizeGranularBitRoundNumberOfSignificantDigits = 4 ;
...
                H:quantization = "quantization_info" ;
                H:quantization_maximum_absolute_error = 0.0004882812f ;
                H:quantization_nsd = 4 ;
```

The max absolute error (MAE) in that is probably not quite right. It's based on Table 3 of https://gmd.copernicus.org/articles/12/4099/2019/. I freely admit I might have read that paper wrong, but I wanted to get something in there. I'll monitor the CF PRs and when they have a formula for the MAE, I'll update MAPL.

We also fix some constant names, update some ASSERTs, and clean up some weird defaults that really should have wreaked havoc with History (i.e., they would seem to always turn on quantization!), but they are never used as a default. Still, now at least the defaults are sane.

All testing shows this is zero-diff for state and history except for if you enable the netcdf quantization.

## Related Issue

Closes #2926


griddedio/GriddedIO.F90
```diff
@@ -442,6 +474,87 @@ subroutine CreateVariable(this,itemName,rc)
...
+
+        mae_base = 4.0 * (1.0/16.0)**floor(real(nsd)/2.0) * (1.0/8.0)**ceiling(real(nsd)/2.0)
+
+        if (nsd > 2 .and. mod(nsd, 2) == 0) then
+           correction = 2
+        else if (nsd == 7) then
+           correction = 2
+        else
+           correction = 1
+        end if
```
tclune, line 496: Would love to eliminate the ELSE's here.   Usually one could use SELECT CASE, but that does not quite work here.   In theory one could do:
```
   correction = 2
   if (.not. ((nsd > 2 .and. mod(nsd,2) == 0) .or. nsd == 7) ) then
      correction = 1
   end if
```
But this is a dubious improvement.

griddedio/GriddedIO.F90
```diff
@@ -442,6 +474,87 @@ subroutine CreateVariable(this,itemName,rc)
 
      end subroutine CreateVariable
 
+     function calculate_mae(nsd) result(mae)
+
+        ! This function is based on Table 3 of doi 10.5194/gmd-12-4099-2019
+        ! The algorithm is weird, but it does duplicate the table
+
+        implicit none
```
tclune, line 482: unnecessary
```suggestion
```

griddedio/GriddedIO.F90
```diff
@@ -442,6 +474,87 @@ subroutine CreateVariable(this,itemName,rc)
...
+        netcdf_version = 'libnetcdf ' // nf90_inq_libvers()
+        call v%add_attribute('implementation', netcdf_version)
+
+        ! NOTE: In the future if we add the MAPL bit-shaving
+        ! to use the quantization parts of the code, it will
+        ! need a different implementation string
+
+        call factory%append_variable_metadata(v)
+        call this%metadata%add_variable(trim(varName),v,rc=status)
+        _VERIFY(status)
```
tclune, line 552:
```suggestion
        call this%metadata%add_variable(trim(varName),v,_RC)
```

griddedio/GriddedIO.F90
```diff
@@ -442,6 +474,87 @@ subroutine CreateVariable(this,itemName,rc)
...
+        integer, optional, intent(out) :: rc
+
+        integer :: status
+
+        class (AbstractGridFactory), pointer :: factory
+        character(len=:), allocatable :: varName, netcdf_version
+        type(Variable) :: v
+
+        factory => get_factory(this%output_grid,rc=status)
+        _VERIFY(status)
```
tclune, line 512:
```suggestion
        factory => get_factory(this%output_grid,_RC)
```

griddedio/GriddedIO.F90
```diff
@@ -190,19 +192,24 @@ subroutine CreateFileMetaData(this,items,bundle,timeInfo,vdata,ogrid,global_attr
...
 
-
         order = this%metadata%get_order(rc=status)
         _VERIFY(status)
         metadataVarsSize = order%size()
 
+        ! If quantize algorithm is set, create a quantization_info variable
+        if (this%quantizeAlgorithm /= MAPL_NOQUANTIZE) then
+           call this%CreateQuantizationInfo(rc=status)
+           _VERIFY(status)
```
tclune, line 209:
```suggestion
           call this%CreateQuantizationInfo(_RC)
```

griddedio/GriddedIO.F90
mathomp4: Hmm. Interesting. I'm not sure what is better. I guess we could go the other way:
```fortran
correction = 1
if ( (nsd > 2 .and. mod(nsd, 2) == 0) .or. nsd == 7 ) then
   correction = 2
end if
```

tclune (approved):

PR #2956, opened 2024-08-09
Add future ability to test ifx in CI

mathomp4:
## Types of change(s)

## Checklist

## Description

This is an update to CI to eventually allow the ability to add `ifx`. At the moment `ifx` does not run the MAPL tests (though it can build), it can build GEOSgcm but not run it, and it just cannot build GEOSldas (due to missing HDF4 Fortran headers, see https://github.com/GEOS-ESM/GEOSldas_GridComp/issues/55).


mathomp4:
Oh yeah. `ifx` + MAPL means test failures at the moment. See:

https://app.circleci.com/pipelines/github/GEOS-ESM/MAPL/9508/workflows/69823f8f-67cf-4525-a489-5a4adfd828aa/jobs/78296

Perhaps if I can work with @tclune to fix the `MAPL.base.tests` failure, we'd solve the `ExtData2G_case24` case fail as well?

tclune:
Hmm.  That crash is basically in a bit of code that is extracting a C address of an array.   The problem must be farther upstream.  Also interesting that the stacktrace itself is missing the intermediary lines between GriddedIO and CubedSphereGridFactory.

Quite possibly a compiler bug, but has enough noise that it could also be some latent uninitialized thingo(TM) in our code.

tclune (approved):

PR #2960, opened 2024-08-12
Clean up NAG + OpenMP CMake

mathomp4:
## Types of change(s)

## Checklist

## Description

This PR cleans up the CMake in MAPL to avoid the ugly:
```cmake
# CMake has an OpenMP issue with NAG Fortran: https://gitlab.kitware.com/cmake/cmake/-/issues/21280
if (NOT CMAKE_Fortran_COMPILER_ID MATCHES "NAG")
  target_link_libraries(Regrid_Util.x PRIVATE OpenMP::OpenMP_Fortran)
endif ()
```
sprinkled all around. 

NOTE: This update *REQUIRES* ESMA_cmake v3.50.0 (or v4.5.0) when compiling with NAG. Without it, MAPL will not build. Other compilers do not care.

### DIFF
--- a/benchmarks/io/checkpoint_simulator/CMakeLists.txt
+++ b/benchmarks/io/checkpoint_simulator/CMakeLists.txt
@@ -6,11 +6,6 @@ ecbuild_add_executable (
   SOURCES checkpoint_simulator.F90
   DEFINITIONS USE_MPI)
 
-target_link_libraries (${exe} PRIVATE MAPL.shared MPI::MPI_Fortran FARGPARSE::fargparse ESMF::ESMF )
+target_link_libraries (${exe} PRIVATE MAPL.shared MPI::MPI_Fortran FARGPARSE::fargparse ESMF::ESMF OpenMP::OpenMP_Fortran)
 target_include_directories (${exe} PUBLIC $<BUILD_INTERFACE:${MAPL_SOURCE_DIR}/include>)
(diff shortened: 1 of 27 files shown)


benchmarks/io/checkpoint_simulator/CMakeLists.txt
```diff
@@ -6,11 +6,6 @@ ecbuild_add_executable (
   SOURCES checkpoint_simulator.F90
   DEFINITIONS USE_MPI)
 
-target_link_libraries (${exe} PRIVATE MAPL.shared MPI::MPI_Fortran FARGPARSE::fargparse ESMF::ESMF )
+target_link_libraries (${exe} PRIVATE MAPL.shared MPI::MPI_Fortran FARGPARSE::fargparse ESMF::ESMF OpenMP::OpenMP_Fortran)
```
tclune, line 9: Not to change now, but I'm wondering if we should start putting one argument per line to keep the CMake lines readable?   Just a style thing.

tclune (approved):

benchmarks/io/checkpoint_simulator/CMakeLists.txt
mathomp4: Maybe. I know that gersemi:

https://github.com/BlankSpruce/gersemi

is a thing. Let me fiddle with it and see.

PR #2965, opened 2024-08-14
Add geom to VariableSpec

pchakraborty:
## Types of change(s)

## Checklist

## Description

Added `geom` as an optional variable in `VariableSpec`. The case where this `geom` exists and someone specifies a `geom` via `make_itemspec` is not allowed.

### DIFF
--- a/generic3g/specs/VariableSpec.F90
+++ b/generic3g/specs/VariableSpec.F90
@@ -21,7 +21,6 @@ module mapl3g_VariableSpec
    use mapl3g_ActualPtVector
    use mapl_ErrorHandling
    use mapl3g_StateRegistry
-   use mapl3g_GeomUtilities, only: MAPL_SameGeom
    use esmf
    use gFTL2_StringVector
    use nuopc
--- a/generic3g/specs/VariableSpec.F90
+++ b/generic3g/specs/VariableSpec.F90
@@ -205,7 +205,7 @@ contains
       type(ActualPtVector) :: dependencies
 
       if (present(geom) .and. allocated(this%geom)) then
-         _ASSERT(MAPL_SameGeom(geom, this%geom), "specified geom is different from existing one")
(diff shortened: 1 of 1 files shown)


tclune (changes requested):
Found some issues with scenarios where no geom is passed down.   Discussed with @pchakraborty offline.

generic3g/specs/VariableSpec.F90
```diff
@@ -199,11 +203,14 @@ function make_ItemSpec_new(this, geom, vertical_grid, registry, rc) result(item_
 
       integer :: status
       type(ActualPtVector) :: dependencies
+      type(ESMF_Geom), allocatable :: geom_local
+
+      call this%pick_geom_(geom, geom_local, _RC)
```
tclune, line 208: One subtle problem.   If `pick_geom` fails, then the allocatable function result (`item_spec`) is not allocated.  This creates errors upstream that cannot be trapped.

The procedure now needs a temporary assignment of `item_spec = NullSpec()` except ... that triggers a known problem with gfortran and reallocation of poly items that means you have to deallocate it again before each of the allocations in the SELECT CASE construct.  Ugh.

I'll approve but please open a ticket to fix this upon your return.

tclune (approved):

generic3g/specs/VariableSpec.F90
pchakraborty: Created the issue.

I have been struggling with these Fortran subtleties all of last week :-(. My understanding of Fortran is not what I thought it was.

tclune: > I have been struggling with these Fortran subtleties all of last week :-(. My understanding of Fortran is not what I thought it was.

It is not that your understanding was inadequate, but rather that the language (and my use of it) has opened new ways for things to fail.   Functions that can fail are really a bad idea.  But I like functional programming, and keep trying to apply it in a software ecosystem that is saturated with failure conditions.

Issue #2959, opened 2024-08-12
MAPL changes to support CICE6 rewind for replay


zhaobin74:
Hi,

Some code needs to be implemented to support rewinding CICE6 states when coupled model runs in replay mode. 
Specifically, the following patch is required in subroutine MAPL_GenericRefresh.

```
       character(len=ESMF_MAXSTR)                  :: CHILD_NAME
       character(len=14)                           :: datestamp ! YYYYMMDD_HHMMz
       integer                                     :: status
+      integer                                     :: UserRC
       integer                                     :: I
       type (MAPL_MetaComp), pointer               :: STATE
       character(len=1)                            :: separator
@@ -2920,6 +2921,12 @@ contains
          ! call the actual record method
          call MAPL_StateRefresh (GC, IMPORT, EXPORT, CLOCK, RC=status )
          _VERIFY(status)
+
+         if (associated(STATE%phase_coldstart)) then
+            call ESMF_GridCompReadRestart(GC, importState=import, &
+                    exportState=export, clock=CLOCK, phase=MAPL_MAX_PHASES+1, userRC=userRC, 
_RC)
+         endif
+
       endif
       call MAPL_TimerOff(STATE,"GenRefreshMine",_RC)
       call MAPL_TimerOff(STATE,"GenRefreshTot",_RC)

```

The above code was tested in V11.4.0 and shown to be working. 
Could SI group add the implementation in the next MAPL version? This is required by @Dooruk for his coupled DA work with CICE6.

Thanks

tclune:
@zhaobin74 We can certainly do this, but it also fine if you want to do a PR yourself (onto the develop branch).

tclune:
I thought @atrayano added this to his plate before he departed for Bulgaria.  I think we wanted to do a slight variation ...

I'll email him separately to get a status update.  (Not sure if he links his personal email to GitHub.)

Dooruk:
Thanks @tclune!

mathomp4:
Yes, @atrayano said in https://github.com/GEOS-ESM/MAPL/pull/2982#issuecomment-2307907700 that:


[...]

PR #2982, opened 2024-08-22
support for CICE6(or others) to rewind

zhaobin74:
This PR closes https://github.com/GEOS-ESM/MAPL/issues/2959

## Types of change(s)

## Checklist

## Description

Allows a GC's refresh method (if present)to be called. Necessary for some GC's to rewind states.


tclune (commented):
This is ok by me, but I want @atrayano to review.

It is odd to use phase=MAPL_MAX_PHASES + 1, and am wondering if since this is not a run phase, it could use a different value.

zhaobin74:
> This is ok by me, but I want @atrayano to review.
> 
> It is odd to use phase=MAPL_MAX_PHASES + 1, and am wondering if since this is not a run phase, it could use a different value.

@tclune, I don't like hardcoded 1 either. But it is the only one I can come up with which works. Maybe @atrayano can fix it.

mathomp4:
@zhaobin74 How crucial is this? 

If you'd like a release but can wait until next week, we can look at getting out a new MAPL release. `develop` has been accumulating a lot.

But, if you need it urgently, I can make a hotfix and release a new MAPL 2.47

zhaobin74:
> @zhaobin74 How crucial is this?
> 
> If you'd like a release but can wait until next week, we can look at getting out a new MAPL release. `develop` has been accumulating a lot.
> 
> But, if you need it urgently, I can make a hotfix and release a new MAPL 2.47

@mathomp4 , it is not urgent. @Dooruk is on PTO and he can already run his DA tests using my build (v1160) with the MAPL branch I made these changes on. I think it should be more vigorously vetted and hopefully @atrayano could further refine it.

atrayano:
@zhaobin74 I have several concerns regarding this PR. The most trivial one is that instead of hardcoding MAPL_MAX_PHASES+1, we could introduce a local variable "phase", which we should get by calling ESMG_GridCompGet(gc, currentPhase=phase, _RC). A far more serious concern is the fact that we are using cold_start method which was supposed to be only as a fallback strategy during Initialize. Luckily for us, thus method is not used anywhere in GEOS (with a possible exception of the dycore), but if another component legitimately tries to use it, this PR has the potential to break that imaginary component. Let's touch base on Monday to brainstorm and find a better solution. 

For the time being, I will add "do not approve" label.

generic/MAPL_Generic.F90
```diff
@@ -351,11 +351,33 @@ module MAPL_GenericMod
...
+         use mapl_KeywordEnforcerMod
+         use ESMF
+         implicit none
+         type(ESMF_GridComp), intent(inout) :: gc
+         type(ESMF_State), intent(inout) :: import_state
+         type(ESMF_State), intent(inout) :: export_state
+         type(ESMF_Clock), intent(inout) :: clock
+         class(KeywordEnforcer), optional, intent(in) :: unusable
+         integer, optional, intent(out) :: rc
+      end subroutine i_Run
```
tclune, line 365: @atrayano This interface is mandated by ESMF.   Since it works with their C func pointers it cannot have the extra arguments and intents.
```suggestion
      subroutine i_Run(gc, import_state, export_state, clock, rc)
         use mapl_KeywordEnforcerMod
         use ESMF
         implicit none
         type(ESMF_GridComp) :: gc
         type(ESMF_State):: import_state
         type(ESMF_State) :: export_state
         type(ESMF_Clock) :: clock
         integer, intent(out) :: rc
      end subroutine i_Run
```

generic/MAPL_Generic.F90
```diff
@@ -425,6 +447,10 @@ module MAPL_GenericMod
       integer                        , pointer :: phase_final(:)    => null()
       integer                        , pointer :: phase_record(:)   => null()
       integer                        , pointer :: phase_coldstart(:)=> null()
+      integer                        , pointer :: phase_refresh(:)=> null()
+      procedure(), public, nopass    , pointer :: customRefresh => null()
```
tclune, line 451: Did you mean to use `I_Run()` here?    An implicit interface `()` seems wrong.

tclune (changes requested):
See inline comments.

generic/MAPL_Generic.F90
atrayano: Adding I_run to line 451 works now

tclune (approved):

PR #2998, opened 2024-08-27
fixes #2996

### LINKED ISSUE #2996: Fix MAPL_HConfigMatch
This procedure needs to use YAML tags to produce correct results.   

Currently there are 2 disable tests in Test_HConfigMatch.pf that are disabled.  (See `@disable`)  These should be activated and the procedure should be fixed.

bena-nasa:
Fixes #2996 

## Types of change(s)

## Checklist


generic3g/ESMF_HConfigUtilities.F90
```diff
@@ -9,6 +9,11 @@ module mapl3g_ESMF_HConfigUtilities
    public :: write(formatted)
    public :: MAPL_HConfigMatch
 
+   character(*), parameter :: CORE_SCHEMA_INT_TAG = 'tag:yaml.org,2002:int'
+   character(*), parameter :: CORE_SCHEMA_FLOAT_TAG = 'tag:yaml.org,2002:float'
+   character(*), parameter :: CORE_SCHEMA_STR_TAG = 'tag:yaml.org,2002:str'
+   character(*), parameter :: CORE_SCHEMA_BOOL_TAG = 'tag:yaml.org,2002:bool'
+
```
tclune, line 16: Am confused?   These are not used anywhere.

generic3g/ESMF_HConfigUtilities/MAPL_HConfigMatch.F90
```diff
@@ -95,36 +95,35 @@ recursive logical function MAPL_HConfigMatchScalar(a, b, rc) result(match)
...
-         _RETURN_UNLESS(a_is .eqv. b_is)
+         a_tag = ESMF_HConfigGetTag(a, _RC)
+         b_tag = ESMF_HConfigGetTag(b, _RC)
+         _RETURN_UNLESS(a_tag == b_tag)
+ 
 
-         if (a_is) then
+         if (a_tag == CORE_SCHEMA_BOOL_TAG) then
+            a_as_bool  = ESMF_HConfigAsLogical(a, _RC)
+            b_as_bool  = ESMF_HConfigAsLogical(b, _RC)
```
tclune, line 112: Perhaps this would be clearer now using `SELECT CASE (a_tag)`?

tclune (dismissed):
Approving.  But see my comment about SELECT CASE.  If you concur, I can approve again after the change.

tclune (changes requested):
The various `_RETURN_SUCCESS` statements are not required once you switch to SELECT CASE.   Sorry - should have hinted at that a bit more clearly.


Also, since there is now one more iteration ... please change the DEFAULT clause to check instead that it is a string tag.  Then the default clause should be a FAIL mesasge about an unsupported tag.   Future proofing us against an exotic tag from a clever end user.

bena-nasa:
> The various `_RETURN_SUCCESS` statements are not required once you switch to SELECT CASE. Sorry - should have hinted at that a bit more clearly.
> 
> Also, since there is now one more iteration ... please change the DEFAULT clause to check instead that it is a string tag. Then the default clause should be a FAIL mesasge about an unsupported tag. Future proofing us against an exotic tag from a clever end user.

changed

tclune (approved):

Issue #2981, opened 2024-08-21
Enable import items to specify regrid parameter


tclune:
FieldSpec already has a component for regrid parameter.  Logic needs to be added to allow specifying it for VariableSpec.

Logic in FieldSpec that gets method from FieldDicitionary should be moved into VariableSpec.

pchakraborty:
Done #3006

PR #3006, opened 2024-09-06
Enable import items to specify regrid parameter

pchakraborty:
## Types of change(s)

## Checklist

## Description
Addresses #2981 
- Added member variable `regrid_param` to VariableSpec
- Moved reading `regrid_method` from field dictionary, from FieldSpec to VariableSpec


generic3g/specs/FieldSpec.F90
```diff
@@ -181,7 +181,7 @@ function new_FieldSpec_geom(unusable, geom, vertical_grid, vertical_dim_spec, ty
 
       ! regrid_param
       field_spec%regrid_param = EsmfRegridderParam() ! use default regrid method
-      regrid_method = get_regrid_method_(field_spec%standard_name)
+      regrid_method = get_regrid_method_from_field_dict(field_spec%standard_name)
```
tclune, line 184:
```suggestion
```
1. Method can return an error so does not belong in a constructor.
2. The value should be passed explicitly in the interface (optional arg)

generic3g/specs/FieldSpec.F90
```diff
@@ -37,7 +37,7 @@ module mapl3g_FieldSpec
    use mapl3g_geom_mgr, only: MAPL_SameGeom
    use mapl3g_FieldDictionary
    use mapl3g_GriddedComponentDriver
-   use mapl3g_VariableSpec
+   use mapl3g_VariableSpec, only: VariableSpec, get_regrid_method_from_field_dict
```
tclune, line 40:
```suggestion
   use mapl3g_VariableSpec, only: VariableSpec
```

generic3g/specs/VariableSpec.F90
```diff
@@ -14,14 +14,16 @@ module mapl3g_VariableSpec
...
+   use mapl3g_FieldDictionary
    use esmf
    use gFTL2_StringVector
    use nuopc
 
    implicit none
    private
 
-   public :: VariableSpec
+   public :: VariableSpec, get_regrid_method_from_field_dict
```
tclune, line 26:
```suggestion
   public :: VariableSpec
   public :: get_regrid_method_from_field_dict
```
This method should ultimately go into a different module that extracts things from the field dict.   The name will improve on that change.

generic3g/specs/VariableSpec.F90
```diff
@@ -111,6 +117,12 @@ function new_VariableSpec( &
       _SET_OPTIONAL(bracket_size)
       _SET_OPTIONAL(dependencies)
 
+      ! regridding parameter
+      var_spec%regrid_param = EsmfRegridderParam() ! use default regrid method
+      regrid_method = get_regrid_method_from_field_dict(var_spec%standard_name)
```
tclune, line 122: This does not handle the case where the dictionary does not have a regrid method.   Do we need to handle an RC?   We have a default in the line above, so it would seem that we should allow the dictionary to have missing items.

Have to run.  But we can discuss later how to address this.

tclune (changes requested):
See inline comments.

generic3g/specs/FieldSpec.F90
pchakraborty: So now it is
```Fortran
! regrid_param
field_spec%regrid_param = EsmfRegridderParam() ! use default regrid method
if (present(regrid_param)) field_spec%regrid_param = regrid_param
```
If not specified, it defaults to the `regrid_param` based on the default regridding method.

generic3g/specs/FieldSpec.F90
pchakraborty: Done.

generic3g/specs/VariableSpec.F90
pchakraborty: Done. Now that FieldSpec is not calling `get_regrid_method_from_field_dict`, we don't need to make it public.

generic3g/specs/VariableSpec.F90
pchakraborty: I think we can handle this once NUOPC FieldDictionary is available for use.

tclune (approved):

PR #3019, opened 2024-09-09
Feature/bmauer/fixes #2986

### LINKED ISSUE #2986: Enable FixedLevelsVerticalGrid in ComponentSpecParser
We have placeholder logic for BasicVerticalGrid in ComponentSpecParser.    Similar logic should be placed for FixedLevelsVerticalGrid.     But be mindful of #2985 as it may inform how this is to be done.

bena-nasa:
Fixes #2986 
Didn't know whether all connections to fixedVerticalGrid should fail at this point or allow sleazy connections. I can make all fail for now if wanted.

## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/tests/Test_FixedLevelsVerticalGrid.pf
+++ b/generic3g/tests/Test_FixedLevelsVerticalGrid.pf
@@ -30,5 +30,20 @@ contains
 
    end subroutine test_equals
    
+   @test
+   subroutine test_not_equals()
+      type(FixedLevelsVerticalGrid) :: vgrid1, vgrid2, vgrid3
+
+      real, parameter :: levels1(*) = [1.,5.,7.]
+      real, parameter :: levels2(*) = [.01,4.]
+
+      vgrid1 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels1)
+      vgrid2 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='mb', levels=levels1)
+      vgrid3 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels2)
+      @assert_that(vgrid1 /= vgrid2, is(.true.))
+      @assert_that(vgrid1 /= vgrid3, is(.true.))
+
+   end subroutine test_not_equals
+   
    
 end module Test_FixedLevelsVerticalGrid
--- a/generic3g/tests/Test_FixedLevelsVerticalGrid.pf
+++ b/generic3g/tests/Test_FixedLevelsVerticalGrid.pf
@@ -26,7 +26,7 @@ contains
 
       vgrid1 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
       vgrid2 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
-      @assert_that(vgrid1==vgrid2, is(.true.))
+      @assert_that(vgrid1==vgrid2, is(true()))
 
    end subroutine test_equals
    
--- a/generic3g/tests/Test_FixedLevelsVerticalGrid.pf
+++ b/generic3g/tests/Test_FixedLevelsVerticalGrid.pf
@@ -13,9 +13,22 @@ contains
 
       real, parameter :: levels(*) = [1.,5.,7.]
 
-      vgrid = FixedLevelsVerticalGrid(standard_name='air_pressure', levels=levels)
+      vgrid = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
       @assert_that(vgrid%get_num_levels(), is(size(levels)))
 
    end subroutine test_num_levels
+
+   @test
+   subroutine test_equals()
+      type(FixedLevelsVerticalGrid) :: vgrid1, vgrid2
+
+      real, parameter :: levels(*) = [1.,5.,7.]
+
+      vgrid1 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
+      vgrid2 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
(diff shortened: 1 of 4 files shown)


generic3g/tests/Test_FixedLevelsVerticalGrid.pf
```diff
@@ -13,9 +13,22 @@ contains
 
       real, parameter :: levels(*) = [1.,5.,7.]
 
-      vgrid = FixedLevelsVerticalGrid(standard_name='air_pressure', levels=levels)
+      vgrid = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
```
tclune, line 16: In theory the units can come from the FIeldDictionary, but this is fine for now.

generic3g/tests/Test_FixedLevelsVerticalGrid.pf
```diff
@@ -13,9 +13,22 @@ contains
...
+
+   @test
+   subroutine test_equals()
+      type(FixedLevelsVerticalGrid) :: vgrid1, vgrid2
+
+      real, parameter :: levels(*) = [1.,5.,7.]
+
+      vgrid1 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
+      vgrid2 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels)
+      @assert_that(vgrid1==vgrid2, is(.true.))
```
tclune, line 29: Please add another test (or 2 or 3) that checks that grids with different levels, different units, etc. are different.  Those are the interesting tests.

generic3g/vertical/FixedLevelsVerticalGrid.F90
```diff
@@ -79,5 +91,22 @@ logical function can_connect_to(this, src, rc)
        _UNUSED_DUMMY(src)
     end function can_connect_to
 
+    logical function equal_FixedLevelsVerticalGrid(a, b) result(equal)
```
tclune, line 94:
```suggestion
    impure elemental logical function equal_FixedLevelsVerticalGrid(a, b) result(equal)
```

tclune: These equality operations often find themselves inside array expressions.   Probably not in this case, and I'd not ask for the change were it not for the request to add more unit tests.

tclune (changes requested):
Just some minor edits and a request to add more unit tests.

generic3g/tests/Test_FixedLevelsVerticalGrid.pf
bena-nasa: @tclune Not in this case, the primary use for the will be History, the collection will define the vertical coordinate, up to the user what units they want to specify the levels in that they are going to type in their History file

generic3g/vertical/FixedLevelsVerticalGrid.F90
bena-nasa: done

generic3g/vertical/FixedLevelsVerticalGrid.F90
```diff
@@ -79,5 +91,22 @@ logical function can_connect_to(this, src, rc)
...
+       type(FixedLevelsVerticalGrid), intent(in) :: a, b
+
+       equal = a%standard_name == b%standard_name
+       if (.not. equal) return       
+       equal = a%units == b%units
+       if (.not. equal) return       
+       equal = all(a%levels == b%levels)
+    end function equal_FixedLevelsVerticalGrid 
+
+    logical function not_equal_FixedLevelsVerticalGrid(a, b) result(not_equal)
```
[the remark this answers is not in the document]
bena-nasa: done

generic3g/vertical/FixedLevelsVerticalGrid.F90
```diff
@@ -79,5 +91,24 @@ logical function can_connect_to(this, src, rc)
...
+    impure elemental logical function equal_FixedLevelsVerticalGrid(a, b) result(equal)
+       type(FixedLevelsVerticalGrid), intent(in) :: a, b
+
+       equal = a%standard_name == b%standard_name
+       if (.not. equal) return       
+       equal = a%units == b%units
+       if (.not. equal) return       
+       equal = size(a%levels) == size(b%levels)
+       if (.not. equal) return       
+       equal = all(a%levels == b%levels)
```
tclune, line 103: Note to self (and others).  This pattern happens often enough throughout MAPL3 that I wonder if it is worth creating a special helper function.  The trick will be remembering what we called it.

generic3g/tests/Test_FixedLevelsVerticalGrid.pf
```diff
@@ -13,9 +13,37 @@ contains
...
+      type(FixedLevelsVerticalGrid) :: vgrid1, vgrid2, vgrid3
+
+      real, parameter :: levels1(*) = [1.,5.,7.]
+      real, parameter :: levels2(*) = [.01,4.]
+
+      vgrid1 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels1)
+      vgrid2 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='mb', levels=levels1)
+      vgrid3 = FixedLevelsVerticalGrid(standard_name='air_pressure', units='Pa', levels=levels2)
+      @assert_that(vgrid1 /= vgrid2, is(.true.))
+      @assert_that(vgrid1 /= vgrid3, is(.true.))
```
tclune, line 44: I don't think this works.   You probably need to change this to:
```suggestion
      @assert_that(vgrid1 /= vgrid2, is(true()))
      @assert_that(vgrid1 /= vgrid3, is(true()))
```
I can't remember the details at the moment, but there is some reason why the annotation fails (maybe because it uses `==` under the hood?).  Might work with some compilers, but 99% certain it fails with NAG.

tclune (approved):

PR #3040, opened 2024-09-19
new restart_simulator.x code

bena-nasa:
This is complementary code to the existing `checkpoint_simulator.x`, and as the name suggests something to mimic the restart reading of cubed-sphere fields. Laura Carriere wants an intern to do some IO benchmarking. They had been running ExtDataDriver.x, but for the restarts, the `checkpoint_simulator.x` provides finer grained detail and is the better tool. I had started creating  `restart_simulator.x` so this was impetus to finish it so they can test both read and write with this. Like the checkpoint code it has similar options to play with the number of readers and the number of trials. Once this is in I will provide instructions to the intern on how to run the `checkpoint_simulator.x`, followed by the `restart_simulator.x`. For now you have to tell in the input file the resolution, number of levels, and number of variables. This could be detected in a future enhancement.

## Types of change(s)

## Checklist


tclune (approved):
As much as I like HConfig for GEOS over using something like fArgParse, it is too bad that it does not have an equivalent of "-h" to give the options at the command line. 

Might be worth some investment to do something along those lines.   @mathomp4

mathomp4 (approved):

Issue #3021, opened 2024-09-09
Extract rank-3 array from ESMF_Field


darianboggs:
MAPL3 requires procedures to extract a rank-3 array pointer from an `ESMF_Field`. The rank-3 array provides a common array representation of `ESMF_Field` objects. It should support these typekinds:

- `ESMF_KIND_R8`
- `ESMF_KIND_R4`
- `ESMF_KIND_I4`
- `ESMF_KIND_I8`

The dimensions of the array are:

1. horizontal dimensions reshaped to a rank-1 array.
2. vertical dimension as a rank-1 array.
3. ungridded dimensions reshaped to a rank-1 array.

If there are no horizontal dimensions, the first dimension has size 1. Similarly, if there are no ungridded dimensions, the third dimension has size 1. Finally, if there is no vertical dimension, the second dimension has size 1.

The steps are:
1. Determine the `ESMF_Field` geometry, including the rank and size for horizontal, vertical, and ungridded dimensions.
2. Based on that geometry, extract a Fortran pointer to array in the `ESMF_Field`.

darianboggs:
Here are the notes of the Microsoft Teams meeting discussion for this issue:

0.   X(i,j,k)
1.   X(i,j,k,a,b)  H+V+ungrid   grid
2.   X(i,k,a,b)    H+V+ungrid   loc stream, mesh
3.   X(k)          ak, bk, Pref
4.   X(i,j,a,b)    surface + ungrid


0.   Y(ij,k,1)   [im*jm, km, na*nb]  ! dimCount=2; rank=3; nvert=
1.   Y(ij,k,ab)                      ! dimCount=2; rank=5; nvert=
2.   Y(i,k,ab)                       ! dimCount=1; rank=4
3.   Y(1,k,1)                        ! dimCount=0; rank=1
4 .  Y(ij,1,ab)                      ! dimcount=2; rank=4


I.  shp = GetFieldCondensedShape(F, rc)

FieldGetCondensedArray_r4(F, x, rc)
FieldGetCondensedArray_r8(F, x, rc)
FieldGetCondensedArray_i4(F, x, rc)
FieldGetCondensedArray_i8(F, x, rc)

  real(R4), pointer :: x(:,:,:)
  real(R8), pointer :: x(:,:,:)
  integer(I4), pointer :: x(:,:,:)
  integer(I8), pointer :: x(:,:,:)

  shp = FieldGetCondenedShape(F, _RC)
[...]

PR #3045, opened 2024-09-19
Get array size of a condensed rank-3 array pointer formed from ESMF_Field Fortran pointer

darianboggs:
## Types of change(s)

## Checklist

## Description
We need a rank-agnostic way to interface with `ESMF_Field` instances. (See #3021 for more details.) The resulting array pointer will have rank-3 with dimensions `(h, v, u)`.
* `h`: horizontal dimensions collapsed to rank-1
* `v`: vertical dimension
* `u`: ungridded dimensions collapsed to rank-1.

## Related Issue
#3021


darianboggs:
Currently, the rank-3 pointer is in a separate public procedure pending feedback on the location.

esmf_utils/OutputInfo.F90
```diff
@@ -123,37 +130,47 @@ function get_vertical_dim_spec_names_bundle_info(info, rc) result(names)
...
+         _ASSERT(spec_name /= EMPTY_STRING, 'No vertical dim spec found.')
+         if(find_index(names, spec_name) == 0) call names%push_back(trim(spec_name))
       end do
       _RETURN(_SUCCESS)
 
    end function get_vertical_dim_spec_names_bundle_info
 
    function get_vertical_dim_spec_name_field(field, rc) result(spec_name)
-      character(len=:), allocatable :: spec_name
+      character(len=ESMF_MAXSTR) :: spec_name
```
tclune, line 147:
```suggestion
      character(len=:), allocatable :: spec_name
```

esmf_utils/OutputInfo.F90
```diff
@@ -123,37 +130,47 @@ function get_vertical_dim_spec_names_bundle_info(info, rc) result(names)
...
       type(ESMF_Field), intent(in) :: field
       integer, optional, intent(out) :: rc
       integer :: status
       type(ESMF_Info) :: info
 
+      spec_name = EMPTY_STRING
```
tclune, line 153: `EMPTY_STRING` seems a bit verbose for this.    We don't define ZERO to be 0, so why not just use `""` here (and elsewhere?   Don't change.  Just curious if there is a specific motivation here.

esmf_utils/OutputInfo.F90
```diff
@@ -123,37 +130,47 @@ function get_vertical_dim_spec_names_bundle_info(info, rc) result(names)
...
       call ESMF_InfoGetFromHost(field, info, _RC)
       spec_name = get_vertical_dim_spec_info(info, _RC)
       _RETURN(_SUCCESS)
 
    end function get_vertical_dim_spec_name_field
 
    function get_vertical_dim_spec_info(info, rc) result(spec_name)
-      character(len=:), allocatable :: spec_name
+      character(len=ESMF_MAXSTR) :: spec_name
```
tclune, line 161:
```suggestion
      character(len=:), allocatable :: spec_name
```

esmf_utils/OutputInfo.F90
```diff
@@ -123,37 +130,47 @@ function get_vertical_dim_spec_names_bundle_info(info, rc) result(names)
...
-
-      call ESMF_InfoGetCharAlloc(info, key=KEY_VLOC, value=spec_name, _RC)
+      logical :: isPresent
+      character(len=ESMF_MAXSTR) :: raw
+      character, parameter :: error_message = 'Failed to get vertical dim spec name.'
+
+      spec_name = EMPTY_STRING
+      isPresent = ESMF_InfoIsPresent(info, key=KEY_VLOC, _RC)
+      _ASSERT(isPresent, error_message)
+      call ESMF_InfoGet(info, key=KEY_VLOC, value=spec_name, rc=status)
```
tclune, line 172:
```suggestion
      call ESMF_InfoGet(info, key=KEY_VLOC, value=raw, rc=status)
      spec_value = trim(raw)
```

esmf_utils/OutputInfo.F90
```diff
@@ -240,11 +257,11 @@ function make_ungridded_dim(info, n, rc) result(ungridded_dim)
       end if
       _ASSERT(is_present, 'Key ' // key // ' not found in ' // trim(json_repr))
```
tclune, line 258:
```suggestion
      _ASSERT(is_present, 'Key ' // key // ' not found in ' // trim(json_repr))
```
I thin this assertion should be moved into the conditional on `is_present`.  And then just use `_FAIL` instead of `_ASSERT`

field_utils/FieldCondensedArray_private.F90
```diff
@@ -0,0 +1,42 @@
...
+   public :: get_array_shape
+
+contains
+
+   function get_array_shape(gridToFieldMap, localElementCount, vert_dims, rc) &
+         &result(array_shape)
+      integer :: array_shape(3)
+      integer, intent(in) :: gridToFieldMap(:)
+      integer, intent(in) :: localElementCount(:)
+      integer, optional, intent(in) :: vert_dims(:)
```
tclune, line 18: Why is this optional?   Maybe it will be more obvious later in the code.

field_utils/FieldPointerUtilities.F90
```diff
@@ -151,6 +158,40 @@ subroutine assign_fptr_r8_rank2(x, fp_shape, fptr, rc)
...
+      integer(ESMF_KIND_I8), allocatable :: fp_shape(:)
+      integer :: status
+
+      fp_shape = get_array_shape(x, _RC)
+      call FieldGetCptr(x, cptr, _RC)
+      call c_f_pointer(cptr, fptr, fp_shape)
+
+      _RETURN(_SUCCESS)
+   end subroutine assign_fptr_r8_rank3
+
```
tclune, line 194: What is rank3 about these procedures?

field_utils/FieldPointerUtilities.F90
```diff
@@ -796,159 +837,185 @@ subroutine copy_r8_r8(cptr_x, cptr_y, n)
...
+         else if (tk == ESMF_TYPEKIND_R4 .and. rank == 4) then
+            call ESMF_FieldGet(Field,0,VR4_4D,_RC)
+            deallocate(VR4_3d,_STAT)
+         else if (tk == ESMF_TYPEKIND_R8 .and. rank == 4) then
+            call ESMF_FieldGet(Field,0,VR8_4D,_RC)
+            deallocate(VR8_3d,_STAT)
+         else
+            _FAIL( 'unsupported typekind+rank')
+         end if
+      end if
```
tclune, line 989: I know this is not part of this PR, but ... MAPL cannot delete this here for the same reason that ESMF cannot in their Destroy call.   When the user has allocated the memory, there is no way to know if something else is using the memory.    


Do not change now, but please discuss with me.  Let's be sure we have a use case for this.

field_utils/tests/Test_FieldCondensedArray_private.pf
```diff
@@ -0,0 +1,147 @@
...
+
+    @Test
+    subroutine test_get_array_shape_3D()
+       integer :: expected(3), actual(3)
+       integer, allocatable :: gridToFieldMap(:)
+       integer, allocatable :: localElementCount(:)
+       integer, allocatable :: vertical_dims(:)
+
+       gridToFieldMap = [1, 2]
+       localElementCount = [3, 5, 7]
```
tclune, line 21: This localElementCount is inconsistent with the specification on the line below it.   Makes the test suspect.

field_utils/tests/Test_FieldCondensedArray_private.pf
```diff
@@ -0,0 +1,147 @@
...
+    @Test
+    subroutine test_get_array_shape_general()
+       integer :: expected(3), actual(3)
+       integer, allocatable :: gridToFieldMap(:)
+       integer, allocatable :: localElementCount(:)
+       integer, allocatable :: vertical_dims(:)
+
+       gridToFieldMap = [1, 2]
+       vertical_dims = [3]
+       localElementCount = [2, 3, 5, 7, 11]
```
tclune, line 52: Again - this specification is inconsistent.

tclune: Same.

field_utils/tests/Test_FieldCondensedArray_private.pf
```diff
@@ -0,0 +1,147 @@
...
+
+       gridToFieldMap = [0, 0]
+       localElementCount = [5, 7, 11]
+       expected = [1, 1, product(localElementCount)]
+       actual = get_array_shape(gridToFieldMap, localElementCount)
+       @assertEqual(actual, expected, GENERIC_MESSAGE)
+
+    end subroutine test_get_array_shape_0D
+
+    subroutine test_get_array_shape_vert_only()
```
tclune, line 87: Missing `@test`

field_utils/tests/Test_FieldCondensedArray_private.pf
```diff
@@ -0,0 +1,147 @@
...
+       gridToFieldMap = [0, 0]
+       vertical_dims = [3]
+       localElementCount = vertical_dims
+       expected = [1, localElementCount(1), 1]
+       actual = get_array_shape(gridToFieldMap, localElementCount, vertical_dims)
+       @assertEqual(actual, expected, GENERIC_MESSAGE)
+
+    end subroutine test_get_array_shape_vert_only
+
+    subroutine test_get_array_shape_vert_ungrid()
```
tclune, line 102: Missing `@test`

esmf_utils/OutputInfo.F90
darianboggs: It's an aesthetic choice. All these being equal, I like to make code read well.

field_utils/tests/Test_FieldCondensedArray_private.pf
```diff
@@ -0,0 +1,147 @@
...
+    subroutine test_get_array_shape_2D_ungrid()
+       integer :: expected(3), actual(3)
+       integer, allocatable :: gridToFieldMap(:)
+       integer, allocatable :: localElementCount(:)
+
+       gridToFieldMap = [1, 2]
+       localElementCount = [3, 5, 7, 11]
+       expected = [product(localElementCount(1:2)), 1, product(localElementCount(3:))]
+       actual = get_array_shape(gridToFieldMap, localElementCount)
+       @assertEqual(actual, expected, GENERIC_MESSAGE)
```
tclune, line 127: `assert_that` will provide much better error messages for these.

tclune: Or at least better.

field_utils/FieldPointerUtilities.F90
```diff
@@ -796,159 +837,185 @@ subroutine copy_r8_r8(cptr_x, cptr_y, n)
...
+      call ESMF_FieldDestroy(Field,noGarbage = .true., rc=status)
+      _VERIFY(STATUS)
+      _RETURN(ESMF_SUCCESS)
+
+   end subroutine Destroy
+
+   function get_array_shape(f, rc) result(array_shape)
```
tclune, line 996: This needs a better name.  It is not the array shape.   It is compressed for a special purpose.

field_utils/FieldPointerUtilities.F90
darianboggs: I understand. I don't know why those are deleted. The only reason that it shows up as a change is the indentation. I ran a vim command to fix the indentation because it was incorrect and inconsistent. It should follow the 3 space convention. I am not familiar with that code.

field_utils/tests/Test_FieldCondensedArray_private.pf
darianboggs: fixed

field_utils/tests/Test_FieldCondensedArray_private.pf
darianboggs: fixed

field_utils/FieldPointerUtilities.F90
tclune: Partial answer is that `get_array_shape()` is size 3.    Might be clearer to avoid deferred shape (allocatable) in this context.    In any event I think these procedures should go.     I'll explain when we talk.

darianboggs: I was referring to the shape in `call c_f_pointer`. I see that you were using `rank` to refer to the rank of the `fptr`. These procedures have the same signature as the rank-1 procedures, so they could not be included in the generic `assign_fptr` funtion. The names are placeholders.

esmf_utils/OutputInfo.F90
darianboggs: This private function is used for two purposes:
1) to obtain the name for its own sake
2) to obtain the name for determining the number of vertical levels

In the second case, the name is need to know if the vertical dim spec is named `VERTICAL_DIM_NONE`. We don't set the num_levels info when the name is `VERTICAL_DIM_NONE`. Maybe we should. In any case, right now that's how that is. Since it is a simple string comparison, I don't need a trim or an allocatable string.

In the first case, I trim the string at the higher level call for the bundle case. For the case of a single field, I don't trim it because of performance hit seems greater than the memory used for the string.

tclune: For `VERTICAL_DIM_NONE` it should be an error to request number of levels.   Setting to 1 would be incorrect, but would lead to correct results is most (but not all) use cases.

Will discuss soon.

esmf_utils/OutputInfo.F90
darianboggs: I wasn't sure about the performance impact of using allocatable instead of fixed-length. I'll take a look.

esmf_utils/OutputInfo.F90
darianboggs: I'll take a look.

esmf_utils/OutputInfo.F90
darianboggs: I'll take a look.

field_utils/FieldCondensedArray_private.F90
darianboggs: The `get_array_shape` in `FieldPointerUtilities.F90` always calls the private function with an integer array, now, even when there is no vertical dimension. So this no longer needs to be optional.

field_utils/tests/Test_FieldCondensedArray_private.pf
darianboggs: What is the syntax for that? I didn't find that in the pFUnit documentation I have. Is that recent?

field_utils/FieldPointerUtilities.F90
darianboggs: I think `get_fptr_shape` would be the most descriptive name. This is a private utility function used in the assign_fptr specific function for the condensed array procedure. Since it is providing the fptr shape in the `c_f_pointer`, that would be the most direct.

field_utils/tests/Test_FieldCondensedArray_private.pf
tclune: Not new, but probably/possibly newer than documentation.  There is an entire "hamcrest" layer that provides a much more flexible manner of composing messages and asserting combinations of things.  Unfortunately, it is largely limited to working on Fortran intrinsic types due to lack of support for multiple inheritance.   Forces users to make their type extend a pFUnit base type which is generally ill advised.

tclune: Typical usage is:
```
@assert_that(found, is(expected))
```
or
```
@assert_that(message, found, is(expected))
```
or even the more verbose:
```
@assert_that(found, is(equal_to(expected)))
```
And there are `less_then` and variants.  As well as `nealy_equal_to` which takes a tolerance.  And advanced stuff like `all` and `any`.
```
And then the caveats.  Don't use it with strings - gfortran is broken.

File: esmf_utils/OutputInfo.F90 (hunk 5 above)
Comment (darianboggs) [reply esmf_utils/OutputInfo.F90:258]: Done.

**Comment by darianboggs** [comment]
I still need to fix the failing Exception test, and move the public interface for the fptr to FieldCondensedArray.F90 with the rank-3 return type.

File: esmf_utils/OutputInfo.F90 [hunk 15]
```diff
@@ -152,12 +156,33 @@ function get_vertical_dim_spec_info(info, rc) result(spec_name)
...
+!      logical :: isPresent
+!      character, parameter :: error_message = 'Failed to get vertical dim spec name.'
+!
+!      isPresent = ESMF_InfoIsPresent(info, key=KEY_VLOC, _RC)
+!      _ASSERT(isPresent, error_message)
+!      call ESMF_InfoGet(info, key=KEY_VLOC, value=spec_name, _RC)
+!      _RETURN(_SUCCESS)
+!
+!   end function get_vertical_dim_spec_info
+
```
tclune, line 185:
```suggestion
```
Delete.

esmf_utils/OutputInfo.F90
```diff
@@ -225,23 +250,26 @@ function make_ungridded_dim(info, n, rc) result(ungridded_dim)
...
       end if
-      _ASSERT(is_present, 'Key ' // key // ' not found in ' // trim(json_repr))
-      dim_info = ESMF_InfoCreate(info, key=key, _RC)
-      call ESMF_InfoGetCharAlloc(dim_info, key=KEY_UNGRIDDED_NAME, value=name, _RC)
-      call ESMF_InfoGetCharAlloc(dim_info, key=KEY_UNGRIDDED_UNITS, value=units, _RC)
+      dim_info = ESMF_InfoCreate(info, key=trim(adjustl(raw)), _RC)
+      call ESMF_InfoGet(dim_info, key=KEY_UNGRIDDED_NAME, value=raw, _RC)
+      name = trim(adjustl(raw))
+      call ESMF_InfoGet(dim_info, key=KEY_UNGRIDDED_UNITS, value=raw, _RC)
+      units = trim(adjustl(raw))
```
tclune, line 272: Have you attempted a reproducer to send to ESMF?

darianboggs:
Also, I am cleaning up comments and _HERE lines.

field_utils/FieldPointerUtilities.F90
```diff
@@ -151,6 +158,40 @@ subroutine assign_fptr_r8_rank2(x, fp_shape, fptr, rc)
...
+      integer, optional, intent(out) :: rc
+
+      ! local declarations
+      type(c_ptr) :: cptr
+      integer(ESMF_KIND_I8), allocatable :: fp_shape(:)
+      integer :: status
+
+      fp_shape = get_fptr_shape(x, _RC)
+      call FieldGetCptr(x, cptr, _RC)
+      call c_f_pointer(cptr, fptr, fp_shape)
```
tclune, line 173: This is a runtime failure unless `fp_shape` is size 1.  And we want size 3. (rank 3)

You need a test that covers this.

tclune (changes requested):
The 3D ptr routines are incorrect.  (See inline comment.)  

You will need at least one test that covers those which should expose the bug.

tclune:
For the test that is failing with an expected exception, please post more details here.  What error message do you actually get from the framework?

darianboggs:
Here is the message for the failing exception test:
```Start: <Test_FieldCondensedArray_private_suite.test_get_fptr_shape_wrong_order>
 Failure in <
 Test_FieldCondensedArray_private_suite.test_get_fptr_shape_wrong_order>
F   end: <Test_FieldCondensedArray_private_suite.test_get_fptr_shape_wrong_order
 >

Time:         0.034 seconds

Failure
 in:
Test_FieldCondensedArray_private_suite.test_get_fptr_shape_wrong_order
  Location:
[FieldCondensedArray_private.F90:37]
MAPL expects geom dims before ungridded.```

File: esmf_utils/OutputInfo.F90 (hunk 15 above)
Comment (darianboggs) [reply esmf_utils/OutputInfo.F90:185]: Done

File: field_utils/FieldPointerUtilities.F90 (hunk 17 above)
Comment (darianboggs) [reply field_utils/FieldPointerUtilities.F90]: Moved to `FieldCondensedArray.F90` and fixed with:
```
integer, parameter :: ARRAY_RANK = 3 ! FieldCondensedArray_private.F90
integer(ESMF_KIND_I8) :: fp_shape(ARRAY_RANK) ! FieldCondensedArray.F90
```
Now, it is not possible for fp_shape to not have the correct size.

File: esmf_utils/OutputInfo.F90 [hunk 18]
```diff
@@ -93,13 +93,11 @@ integer function get_num_levels_info(info, rc) result(num)
...
       integer, optional, intent(out) :: rc
       integer :: status
-      logical :: is_none
       character(len=:), allocatable :: spec_name
 
       num = 0
       spec_name = get_vertical_dim_spec_info(info, _RC)
-      is_none = (VERT_DIM_NONE == spec_name)
-      if(is_none) then
+      if(spec_name==VERT_DIM_NONE) then
```
tclune, line 100:
```suggestion
      _RETURN_IF(spec_name == VERT_DIM_NONE)
```

field_utils/FieldPointerUtilities.F90
tclune: No.  That was not the intent.   The intent was to fix the procedure so that its argument was 3D.

esmf_utils/OutputInfo.F90
darianboggs: I have not attempted a reproducer. With the `isPresent` check, `MAPL` throws an exception before the `ESMF_FieldGet` call. The original problem was that the `ESMF_Field` objects I used for the testing the History3G changes were missing the vertical dim spec. I wasn't using `isPresent`, so when I made the `ESMF_FieldGetCharAlloc` call I received an error about an allocatable character variable.

field_utils/FieldPointerUtilities.F90
tclune: Note that the rank2 procedure that is still there has the same bug.

esmf_utils/OutputInfo.F90
darianboggs: I created a reproducer. The problem was what I described above. If the key is not found, the rc code is nonzero and the value becomes unallocated. Therefore it is not sufficient to allocate the value to an empty string (or any string) before calling `ESMF_InfoGetCharAlloc'.

base/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ set (srcs
   MAPL_XYGridFactory.F90
   MAPL_NetCDF.F90                       Plain_netCDF_Time.F90
   MAPL_DateTime_Parsing_ESMF.F90	MAPL_ObsUtil.F90
-  MAPL_ESMF_InfoKeys.F90
+  #MAPL_ESMF_InfoKeys.F90
```
tclune, line 59:
```suggestion
```

esmf_utils/CMakeLists.txt
```diff
@@ -10,9 +10,10 @@ set(srcs
 
 esma_add_library(${this}
   SRCS ${srcs}
-  DEPENDENCIES MAPL.shared MAPL.base
+  DEPENDENCIES MAPL.shared
   TYPE SHARED
   )
+  #  DEPENDENCIES MAPL.shared MAPL.base
```
tclune, line 16:
```suggestion
```

darianboggs: Done

base/CMakeLists.txt
darianboggs: Done

tclune (changes requested):
Just one minor issue about a comment that requests its own deletion.

field_utils/FieldCondensedArray.F90
```diff
@@ -0,0 +1,76 @@
...
+      integer :: rank
+      integer, allocatable :: gridToFieldMap(:)
+      integer, allocatable :: localElementCount(:)
+      logical :: has_vertical
+      character(len=:), allocatable :: spec_name
+      character(len=*), parameter :: VERTICAL_DIM_NONE_NAME = 'VERTICAL_DIM_NONE'
+      !wdb fixme deleteme This seems fragile. We should probably make a utility function
+      !that selects the type(VerticalDimSpec) parameter based on a string. Perhaps a
+      !logical function in VerticalDimSpec.F90 that recogizes a VerticalDimSpec based on
+      !the string from the ESMF_Info.
```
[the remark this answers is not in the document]
darianboggs: Done

tclune (approved):

PR #3056, opened 2024-09-25
Fixes #1618. Add zstandard support

### LINKED ISSUE #1618: Add support for zstandard compression
I am hoping to build netCDF-C 4.9.0 and netCDF-Fortran 4.6.0 in Baselibs soon. In doing so, we have the capability to maybe try out zstd compression (see https://github.com/Unidata/netcdf-c/issues/2173)

As such, I'll probably need to ask @bena-nasa to help me figure out how to get a new flag added to History so that I can test it out.

mathomp4:
## Types of change(s)

## Checklist

## Description

This PR adds support for zstandard compression in History via `zstandard_level`. 

Much like netCDF quantize support, if netCDF-C wasn't built with zstandard support, then MAPL will not allow a user to set `zstandard_support`. We also prevent a user from setting both `deflate` and `zstandard_level` at the same time. Also, two new ExtData tests are added.

NOTE: At the moment, this support is only available via Spack builds of MAPL (and GEOS). The reason is that zstandard support in netCDF-C is only available via shared library (because it uses HDF5 plugins). For historic reasons, ESMA Baselibs uses static HDF5, netCDF, etc.

cc: @edwardhartnett

## Related Issue

Closes #1618


mathomp4:
Yup. This is currently not working. I've done something wrong, but not sure what yet. Turns out, enabling zstandard doesn't actually enable it! Bug hunting time.

tclune (approved):

PR #3069, opened 2024-10-03
Fixes #3065 - field reallocation

### LINKED ISSUE #3065: Reallocate field when relevant attributes change.
When attributes of an input stateitem in a coupler change,  the corresponding attributes in the output stateitem often need to change as well.  Some of these changes require the field to be reallocated.

We need an interface that looks at the changed attributes and reallocates the field if necessary.   To do this we unfortunately must circumvent ESMF interfaces for now to deallocate the data while leaving the field object itself in place.  Then we can use FieldComplete in the usual manner.

tclune:
These probably need to move to a different directory.

## Types of change(s)

## Checklist


generic3g/FieldUtilities.F90
```diff
@@ -0,0 +1,131 @@
...
+
+      typekind_ = old_typekind
+      if (present(typekind)) typekind_ = typekind
+
+      geom_ = old_geom
+      if (present(geom)) geom_ = geom
+
+      ungriddedUBound_ = old_ungriddedUBound
+      if (present(ungriddedUBound)) ungriddedUBound_ = ungriddedUBound
+      _ASSERT(size(ungriddedUBound_) == size(old_ungriddedUBound), 'MAPL does not allow the rank of a field to change after creation.')
```
darianboggs, line 59: Is it the _**total**_ number of dimensions of the field or the number of **_ungridded_** dimensions of the field you are checking?

tclune: Just the ungridded dimensions.    We only allow the number of levels to change.    Changes in the gridded dimensions happen through a change in the geometry, which indirectly changes the extents of the gridded dimensions.   MAPL3 controls the ungridded explicitly.

generic3g/FieldUtilities.F90
```diff
@@ -0,0 +1,131 @@
...
+      integer, allocatable :: localElementCount(:)
+      integer, allocatable :: old_ungriddedUBound(:)
+      integer, allocatable :: ungriddedUBound_(:), ungriddedLBound_(:)
+      integer :: i
+
+      skip_reallocate = .true.
+
+      call ESMF_FieldGet(field, typekind=old_typekind, geom=old_geom, ungriddedDimCount=ungriddedDimCount, rank=rank, _RC)
+      localElementCount = FieldGetLocalElementCount(field, _RC)
+      old_ungriddedUBound = localElementCount(rank-ungriddedDimCount+1:)
```
darianboggs, line 49: What value do you _**intend**_ to calculate here?

generic3g/FieldUtilities.F90
```diff
@@ -0,0 +1,131 @@
...
+   end interface operator(==)
+
+   interface operator(/=)
+      procedure :: ESMF_GeomNotEqual
+   end interface operator(/=)
+
+contains
+
+
+   subroutine reallocate(field, geom, typekind, ungriddedUBound, rc)
```
darianboggs, line 28: Does this subroutine reallocate the field empty, or does it reallocate the field with the data converted by typekind, shape, and geom? It looks like the former, but I want to confirm?

generic3g/FieldUtilities.F90
tclune: the UBound argument that was used in the original FieldEmptyComplete() call so that I can reuse in the new FieldEmptyComplete().  The ESMF calls are unfortunately not symmetric.  The complete() is just the ungridded, but the FieldGet() is all dims.

generic3g/FieldUtilities.F90
tclune: Has to be the former.   For geom you would need to do a regrid.  For a new shape - where would you get the values?  typekind could be done.

The idea is that the reallocate is done inside of a coupler.   The new field will obtain new values when the update() step happens.

darianboggs (dismissed):
Looks good. I wish the diffs in GitHub were smarter. Though you only added to FieldUtilities.F90, GitHub diffed it as if the existing code were added.

generic3g/FieldUtilities.F90
darianboggs: OK

generic3g/FieldUtilities.F90
darianboggs: OK.

tclune:
> Looks good. I wish the diffs in GitHub were smarter. Though you only added to FieldUtilities.F90, GitHub diffed it as if the existing code were added.

That's because I used emacs to properly indent the file while I was there.  Made almost every line have a small change.   GitHub has options to ignore whitespace for such, not that I know how to do that.

mathomp4:
New Error seems a bit more simple to track down:
```
Test_FieldUtilities.F90(2): #error: can't find include file: UNUSED_DUMMY.H
```

darianboggs (approved):
Looks good

darianboggs:
> GitHub has options to ignore whitespace for such, not that I know how to do that.

Yy26r

PR #3077, opened 2024-10-04
Fixes #3072 Feature/wdboggs/#3072 update get fptr shape

### LINKED ISSUE #3072: Update CondensedArrays
1. The names `get_fptr` procedures for Condensed Arrays should be replaced by specific names for the public and _MAPL private_ versions, with a generic interface for `get_fptr`.
2. The `gridToFieldMap` variable in `mapl3g_FieldCondensedArray:get_fptr_shape` needs to be a fixed shape array.

darianboggs:
## Types of change(s)

## Checklist

## Description

1. Change the name of the `get_fptr_shape` function in mapl3g_FieldCondensedArray_private.
2. Change the `gridToFieldMap` array to be allocated before running `ESMF_GridGet(field, gridToFieldMap, _RC)`

Previously, both `mapl3g_FieldCondensedArray` and `mapl3g_FieldCondensedArray_private` had a function get_fptr_shape. Since `get_fptr_shape` in `mapl3g_FieldCondensedArray` used the `mapl3g_FieldCondensedArray_private` `get_fptr_shape` function, the `use` statement aliased the private version of the function. Since the `get_fptr_shape` function in `mapl3g_FieldCondensedArray` is private, I changed function name in `mapl3g_FieldCondensedArray_private` to avoid the name collision, instead of creating a generic function with the specific functions in each module.

In `ESMF_FieldGet`, the `gridToFieldMap` dummy argument is `intent(out)`, but it is not allocated in the subroutine. This should have generated a compiler error since the actual argument was not allocated before calling `ESMF_FieldGet`. I allocated the `gridToFieldMap` actual argument before passing it to `ESMF_FieldGet`, by first checking dimCount, and then allocating the array. 

## Related Issue
#3072


field_utils/FieldCondensedArray.F90
```diff
@@ -1,7 +1,7 @@
 #include "MAPL_Generic.h"
 module mapl3g_FieldCondensedArray
    use, intrinsic :: iso_c_binding, only: c_ptr, c_f_pointer
-   use mapl3g_FieldCondensedArray_private, only: ARRAY_RANK, get_shape => get_fptr_shape
+   use mapl3g_FieldCondensedArray_private, only: ARRAY_RANK, get_fptr_shape_private
```
tclune, line 4: This change really does not accomplish anything.   Before it was renamed on the USE statement.  Now it is renamed manually here.  And the `_private` suffix suggests that it should not be used outside the module (except for testing).

I recommend undoing it.  But I defer to you.

darianboggs: I thought about using the generic function via, but it only included the public `get_fptr_shape` specific function. `get_fptr_shape_private` does not act on `ESMF_Field` objects.

PR #3084, opened 2024-10-08
Feature/tclune/#3079 interpolate action

tclune:
## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/actions/TimeInterpolateAction.F90
+++ b/generic3g/actions/TimeInterpolateAction.F90
@@ -5,6 +5,7 @@ module mapl3g_TimeInterpolateAction
    use mapl3g_regridder_mgr
    use mapl3g_InfoUtilities
    use MAPL_FieldUtils
+   use MAPL_Constants, only: MAPL_UNDEFINED_REAL
    use mapl_ErrorHandling
    use esmf
 
@@ -57,7 +58,7 @@ contains
       call ESMF_StateGet(importState, 'import[1]', itemType=itemType, _RC)
       _ASSERT(itemType == ESMF_STATEITEM_FIELDBUNDLE, 'Expected FieldBundle in importState.')
 
-      call ESMF_StateGet(importState, 'export[1]', itemType=itemType, _RC)
+      call ESMF_StateGet(exportState, 'export[1]', itemType=itemType, _RC)
       _ASSERT(itemType == ESMF_STATEITEM_FIELD, 'Expected Field in exportState.')
 
       call ESMF_StateGet(importState, itemName='import[1]', fieldbundle=bundle_in, _RC)
@@ -106,7 +107,11 @@ contains
       y = weights(1)
       do i = 1, fieldCount
          call assign_fptr(fieldList(i), xi, _RC)
-         y = y + weights(i+1) * xi
+         where (xi /= MAPL_UNDEFINED_REAL .and. y /= MAPL_UNDEFINED_REAL)
+            y = y + weights(i+1) * xi
+         elsewhere
+            y = MAPL_UNDEFINED_REAL
+         end where
       end do
 
       _RETURN(_SUCCESS)
--- a/generic3g/actions/TimeInterpolateAction.F90
+++ b/generic3g/actions/TimeInterpolateAction.F90
@@ -1,0 +1,116 @@
+#include "MAPL_Generic.h"
+
+module mapl3g_TimeInterpolateAction
+   use mapl3g_ExtensionAction
+   use mapl3g_regridder_mgr
+   use mapl3g_InfoUtilities
+   use MAPL_FieldUtils
+   use mapl_ErrorHandling
+   use esmf
+
+   implicit none
+   private
+
+   public :: TimeInterpolateAction
+
+   type, extends(ExtensionAction) :: TimeInterpolateAction
+   contains
+      procedure :: initialize
+      procedure :: run
+   end type TimeInterpolateAction
+
+   interface TimeInterpolateAction
+      module procedure :: new_TimeInterpolateAction
+   end interface TimeInterpolateAction
+
+contains
+
+   function new_TimeInterpolateAction() result(action)
+      type(TimeInterpolateAction) :: action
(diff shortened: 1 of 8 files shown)


generic3g/actions/TimeInterpolateAction.F90
```diff
@@ -0,0 +1,121 @@
...
+
+   interface TimeInterpolateAction
+      module procedure :: new_TimeInterpolateAction
+   end interface TimeInterpolateAction
+
+contains
+
+   function new_TimeInterpolateAction() result(action)
+      type(TimeInterpolateAction) :: action
+   end function new_TimeInterpolateAction
```
darianboggs, line 31:
```suggestion

contains

```

darianboggs: If the constructor is a placeholder that will be updated, I understand. If not, I do not understand the purpose on a custom constructor.

tclune: It was not obvious at the outset that the class would have no components. (Mixing programming language idioms.)   But there is a script to follow ... (Indeed, I cut-and-paste from a sibling subclass.) And this applies to the client code what constructs things.   It could leave someone wondering why some Action subclasses use a constructor and others just allocate an uninitialized object.

Having a blank constructor also insulates from changes that may creep into the parent class, so I'm inclined to leave this here for now.

tclune: But I also admit that I find the compiler message "function result 'action' has not been given a value." to be rather annoying.

generic3g/tests/Test_InfoUtilities.F90
darianboggs, line 1: I do not understand why the extension is `F90` instead of `pf`

tclune: Egads!    Am guessing (hoping) that this file was just accidentally committed.  Investigating ...

generic3g/tests/Test_InfoUtilities.F90
```diff
@@ -109,4 +105,46 @@ subroutine test_setPrivate_is_private()
...
+
+      state = ESMF_StateCreate(name='import', _RC)
+      field = ESMF_FieldEmptyCreate(name='f', _RC)
+      call ESMF_StateAdd(state, [field], _RC)
+
+      call MAPL_InfoSetInternal(state, short_name='f', key='a', values=[1., 2.], _RC)
+      call MAPL_InfoGetInternal(state, short_name='f', key='a', values=w, _RC)
+
+      @assert_that(w, is(equal_to([1.,2.])))
+
```
darianboggs, line 123:
```suggestion
      real(ESMF_KIND_R4), allocatable :: w(:)
      real(ESMF_KIND_R4), parameter = expected(2) = [1., 2.]

      state = ESMF_StateCreate(name='import', _RC)
      field = ESMF_FieldEmptyCreate(name='f', _RC)
      call ESMF_StateAdd(state, [field], _RC)

      call MAPL_InfoSetInternal(state, short_name='f', key='a', values=expected, _RC)
      call MAPL_InfoGetInternal(state, short_name='f', key='a', values=w, _RC)

      @assert_that(w, is(equal_to(expected)))

```

darianboggs: It helps to make the expected value a parameter rather than a literal.

generic3g/tests/Test_InfoUtilities.F90
tclune: OK - committed under the wrong name, and then used the wrong prefix in the cmake as well.  I'm astonished that this worked and ran the tests. (If it had not, I would have caught it sooner.)   

Will  fix.

generic3g/tests/Test_InfoUtilities.F90
tclune: Agreed.  I was lazy. Thank you.

generic3g/tests/Test_InfoUtilities.F90
```diff
@@ -109,4 +105,46 @@ subroutine test_setPrivate_is_private()
...
+      call ESMF_FieldDestroy(field, _RC)
+      call ESMF_StateDestroy(state, _RC)
+
+   end subroutine test_setInternal
+
+   @test
```
darianboggs, line 130: See comment for `test_setInternal`.

generic3g/tests/Test_InfoUtilities.F90
tclune: Done in sandbox.

tclune:
While I am here anyway, I'm going to look at using preprocessor to simplify extensions of InfoUtilities.

generic3g/tests/Test_TimeInterpolateAction.pf
```diff
@@ -0,0 +1,173 @@
...
+
+      do i = 1, 2
+         b(i) = ESMF_FieldEmptyCreate(name='b', _RC)
+         call ESMF_FieldEmptySet(b(i), geom=geom, _RC)
+         call ESMF_FieldEmptyComplete(b(i), typekind=ESMF_TYPEKIND_R4, _RC)
+         call assign_fptr(b(i), x, _RC)
+         x = 2. * i
+      end do
+
+      x(2) = MAPL_UNDEFINED_REAL
```
darianboggs, line 145: This appears to set the second element of the second `ESMF_Field` to `MAPL_UNDEFINED_REAL` since `x` still points to the internal Fortran array for `b(i)`. Is that your intention here?

darianboggs (approved):
Looks good.

generic3g/tests/Test_TimeInterpolateAction.pf
tclune: Yes - the idea is that an isolated UNDEF should result in an isolated UNDEF in the result.   (As you can see in the subsequent assertions.)    But a comment might be helpful to the reader.  I'll add that in my sandbox.

darianboggs (approved):
Looks good

PR #3092, opened 2024-10-10
Vertical regridding - step 1 - VerticalDimSpec

pchakraborty:
## Types of change(s)

## Checklist

## Description

`ModelVerticalGrid::get_coordinate_field` returns the appropriate vertical coordinate field (and a coupler if a different unit or typekind is specified) based on its short name. For now, the short name can be one of `PLE`/`PL`.


generic3g/actions/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ target_sources(MAPL.generic3g PRIVATE
   VerticalRegridAction.F90
   CopyAction.F90
   ConvertUnitsAction.F90
+  ConvertVerticalDimSpecAction.F90
```
tclune, line 11: I very much hope that this action is only involved when an export spec is on a different staggering than the import spec.   This case could probably even be disallowed.

I also suggest that "DimSpec" in the name is misleading here as it also allows for things like surface fields.   Something more like ConvertVerticalStaggerAction.F90 seems more correct.

generic3g/specs/FieldSpec.F90
```diff
@@ -916,33 +911,64 @@ logical function adapter_match_vertical_grid(this, spec) result(match)
...
-           if (.not. same_vertical_grid) then
-              select type(src_grid)
-              type is (BasicVerticalGrid)
-                 select type (dst_grid)
-                 type is (BasicVerticalGrid)
-                    same_vertical_grid = (src_grid%get_num_levels() == dst_grid%get_num_levels())
-                 end select
-              end select
-           end if
+            use mapl3g_BasicVerticalGrid
```
tclune, line 924: Let's try to eliminate the BLOCK in this PR.   I know it pre-existed, but ... BLOCK is bad for anything other than a temporary kludge.  The time has come ...

generic3g/specs/FieldSpec.F90
```diff
@@ -1019,15 +1045,23 @@ recursive function make_adapters(this, goal_spec, rc) result(adapters)
...
-         allocate(adapters(3)%adapter, source=TypeKindAdapter(goal_spec%typekind))
-         allocate(adapters(4)%adapter, source=UnitsAdapter(goal_spec%units))
+         allocate(adapters(2)%adapter, source=VerticalDimSpecAdapter(goal_spec%vertical_dim_spec))
+         vertical_grid_adapter = VerticalGridAdapter( &
+              goal_spec%vertical_grid, &
+              goal_spec%vertical_dim_spec, &
+              goal_spec%geom, &
+              goal_spec%typekind, &
+              goal_spec%units, &
+              VERTICAL_REGRID_LINEAR)
```
tclune, line 1061: That's a lot of arguments.   I've not looked closely, but could we instead just pass `goal_spec` and "digest" it inside the constructor?

generic3g/specs/FieldSpec.F90
```diff
@@ -1019,15 +1045,23 @@ recursive function make_adapters(this, goal_spec, rc) result(adapters)
...
+         allocate(adapters(3)%adapter, source=vertical_grid_adapter)
+         allocate(adapters(4)%adapter, source=TypeKindAdapter(goal_spec%typekind))
+         allocate(adapters(5)%adapter, source=UnitsAdapter(goal_spec%units))
```
tclune, line 1064: Sigh - so much prettier if GNU would not choke on polymorphic assignment here.

generic3g/specs/VerticalDimSpec.F90
```diff
@@ -43,12 +48,43 @@ module mapl3g_VerticalDimSpec
...
+         dim_spec_str = "VERTICAL_DIM_NONE"
+      case(2)
+         dim_spec_str = "VERTICAL_DIM_CENTER"
+      case(3)
+         dim_spec_str = "VERTICAL_DIM_EDGE"
+      case(4)
+         dim_spec_str = "VERTICAL_DIM_MIRROR"
+      ! case default
+      !    _FAIL("Invalid vertical dim spec")
+      end select
```
tclune, line 81: I know it is tedious, but we probably should relocate these strings as parameters next to where the various spec parameters are defined.   Will make it much easier to maintain if we ever need to extend.

tclune: Can be a separate PR, but ... don't forget.

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -28,9 +30,9 @@ module mapl3g_ModelVerticalGrid
       integer :: num_levels = -1
       type(StringVector) :: variants
 
-!#      character(:), allocatable :: short_name
-!#      character(:), allocatable :: standard_name
-!#      type(ESMF_Field) :: reference_field
+      ! character(:), allocatable :: short_name
+      ! character(:), allocatable :: standard_name
+      ! type(ESMF_Field) :: reference_field
```
tclune, line 35: Do not modify comments starting `!#` except to fully delete them.  They are used by my editor to comment/uncomment sections.       Either delete the comment completely, or leave it as it is.
```suggestion
      !# character(:), allocatable :: short_name
      !# character(:), allocatable :: standard_name
      !# type(ESMF_Field) :: reference_field
```

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -65,16 +67,15 @@ module function can_connect_to(this, src, rc)
    function new_ModelVerticalGrid_basic(num_levels) result(vgrid)
       type(ModelVerticalGrid) :: vgrid
       integer, intent(in) :: num_levels
-!#      character(*), intent(in) :: short_name
-!#      character(*), intent(in) :: standard_name
-!#      type(StateRegistry), pointer, intent(in) :: registry
+      ! character(*), intent(in) :: short_name
+      ! character(*), intent(in) :: standard_name
+      ! type(StateRegistry), pointer, intent(in) :: registry
```
tclune, line 72: Same comment as above.

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -65,16 +67,15 @@ module function can_connect_to(this, src, rc)
...
 
       call vgrid%set_id()
       vgrid%num_levels = num_levels
-!#      vgrid%short_name = short_name
-!#      vgrid%standard_name = standard_name
-!#      vgrid%registry => registry
-
+      ! vgrid%short_name = short_name
+      ! vgrid%standard_name = standard_name
+      ! vgrid%registry => registry
```
tclune, line 78: And again.

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -127,9 +128,9 @@ subroutine get_coordinate_field(this, field, coupler, standard_name, geom, typek
        integer :: i
 
        v_pt = VirtualConnectionPt(state_intent='export', short_name=this%variants%of(1))
-       goal_spec = FieldSpec(geom=geom, vertical_grid=this, vertical_dim_spec=VERTICAL_DIM_EDGE, &
-            typekind=typekind, standard_name=standard_name, units=units, &
-            ungridded_dims=UngriddedDims())
+       goal_spec = FieldSpec( &
+            geom=geom, vertical_grid=this, vertical_dim_spec=vertical_dim_spec, &
+            typekind=typekind, standard_name=standard_name, units=units, ungridded_dims=UngriddedDims())
```
tclune, line 133: What I think is wanted here is that you select a different `short_name` depending on `vertical_dim_spec`.    Then it probably (hopefully) is not needed in `goal_spec`.

generic3g/actions/CMakeLists.txt
pchakraborty: > I very much hope that this action is only involved when an export spec is on a different staggering than the import spec. This case could probably even be disallowed.

Isn't that the case with HISTORY? It's currently happening here
https://github.com/GEOS-ESM/MAPL/blob/16073616abc2ca7e7b8ad37a2e50bbb5f954e4ed/base/MAPL_VerticalMethods.F90#L207

>I also suggest that "DimSpec" in the name is misleading here as it also allows for things like surface fields. Something more like ConvertVerticalStaggerAction.F90 seems more correct.

Not sure I understand. This 'action' is converting DimSpec.

generic3g/specs/FieldSpec.F90
pchakraborty: Yes! But not in this PR. I'll soon have another one where we regrid VerticalGrid. It'll be part of that.

generic3g/specs/VerticalDimSpec.F90
pchakraborty: Created new issue, just so I don't forget.

generic3g/vertical/ModelVerticalGrid.F90
pchakraborty: Interesting! My editor uses the standard `!` for commenting/un-commenting sections. I'll put the `#` back, but keep the indentation.

generic3g/actions/CMakeLists.txt
tclune: I misspoke about the situation where the action might be required.   I don't think we need it at all.   Let me explain.

The model (FV) will provide both PLE and PL.    History will either use "native" (which will choose between PLE and PL) or has an entirely different grid (fixed levels) in which case you can interpolate from either.  (Mathematically equivalent, but PLE is preferred due to edge-of-domain issues.)

At the other end, if ExtData files only provide say PLE, then rather than have an adaptor, I would require ExtData to also include PLE in its state and precompute that from PLE.   The hard case here is when a model quantity is say on centers - you will end up with a different sparse matrix than you would for edges.     But that's it.   No need to first regrid to edges and then interpolate to centers.    I can be convinced otherwise, but I think it would add needless expense and complexity.

Vertical staggering should always just result in selecting a different field for specifying the vertical coordinates when "same source", and otherwise just directly compute the needed sparse matrix.

generic3g/tests/Test_ModelVerticalGrid.pf
```diff
@@ -35,48 +39,58 @@ module Test_ModelVerticalGrid
...
+      case default
+         _FAIL("var_name should be one of PLE/PL, not" // trim(var_name))
+      end select
+
       rc = 0
       ! Inside user "set_geom" phase.
       geom = make_geom(_RC)
       vgrid = ModelVerticalGrid(num_levels=LM)
-      call vgrid%add_variant(short_name='PLE')
+      call vgrid%add_variant(short_name=var_name)
```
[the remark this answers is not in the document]
tclune: "category" is probably `standard_name`.

generic3g/tests/Test_ModelVerticalGrid.pf
```diff
@@ -211,9 +262,40 @@ contains
...
+
+      call vgrid%get_coordinate_field( &
+           vcoord, coupler, &
+           standard_name='air_pressure', geom=geom, typekind=ESMF_TYPEKIND_R4, units='hPa', _RC)
+      @assert_that(associated(coupler), is(false()))
+
+      call vgrid%get_coordinate_field( &
+           vcoord, coupler, &
+           standard_name='air_pressure', geom=geom, typekind=ESMF_TYPEKIND_R4, units='hPa', _RC)
+      @assert_that(associated(coupler), is(false()))
```
tclune, line 298: How are these 2 calls and assertions different?

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -120,16 +120,28 @@ subroutine get_coordinate_field(this, field, coupler, standard_name, geom, typek
...
+       v_pt = VirtualConnectionPt(state_intent='export', short_name=short_name)
+       select case (short_name)
+       case ("PLE")
+          vertical_dim_spec = VERTICAL_DIM_EDGE
+       case ("PL")
+          vertical_dim_spec = VERTICAL_DIM_CENTER
+       case default
+          _FAIL("short name should be one of PL/PLE, not" // trim(short_name))
+       end select
+
```
tclune, line 141: Ah - ok - this needs to be generalized.  It can go in now, but please open a ticket to generalize and that will be your next priority after we merge this PR.

The idea is that someone else might provide a dyncore that uses a different short name and we don't want to be tied to FV's decision here.   Instead we rely on (a) standard name and (b) staggering.   As we register fields as coodinates we specify the standard name and the staggering.    The "get" procedure then uses staggering and standard name to "find" the appropriate coordinates.

In theory this involves creating a gFTL map with key: {string, vert stagger}. A bit heavyweight, but there's just enough going on here that you probably don't want to reinvent the wheel.

tclune (changes requested):
Looks pretty good.  Most of my comments were about what is needed **next**.

But there were 1 or 2 questions /suggestions to address.

generic3g/tests/Test_ModelVerticalGrid.pf
pchakraborty: Ah, holdover from a previous version that include a vertical_dim_spec argument. Removed it.

generic3g/vertical/ModelVerticalGrid.F90
pchakraborty: > Instead we rely on (a) standard name and (b) staggering. As we register fields as coodinates we specify the standard name and the staggering.I'll address that.

One concern here is that `standard_name` is optional in FieldSpec and also VariableSpec. Is there someplace where standard name is enforced?

tclune: It is not optional for a vertical grid. Note that run-of-the-mill fields do not need to specify a standard name, but FixedLevels absolutely must indicate whether to use PLE vs ZLE.    My thinking is that standard_name will fill that role nicely, but am open to other solutions on that front.

Enforcing that standard name is specified for things in a vertical grid spec should be added in due course.

tclune: > It is not optional for a vertical grid. 

I mean not optional in the sense that we need it.   Might be optional in the Fortran sense, in which case we check.

PR #3095, opened 2024-10-13
Fixes #3075 - propagate time-varying aspects of fields

### LINKED ISSUE #3075: Create procedure to update a coupled field
We now have FieldReallocate(), but couplers need some additional adjacent functionality.

A source field will specify changed attributes in its info.   The new `FieldUpdate(f, f_src, ignore)` should copy changes from f_src to f, but only for attributes that do not match `ignore`.  E.g., A RegridAction should copy changes in units, but not changes in grid.  (But those should trigger an update in the regridder object itself.)

tclune:
Required updating various things to use new MAPL_Info interfaces.

Need to be careful that keys sent to ESMF do not end in "/" and yet we don't want users to have to prepend their keys with "/".  Fix for now is to check and prepend with "/" if not present.

If performance is found to be inadequate the issue will be revisited.

## Types of change(s)

## Checklist

### DIFF
--- a/GeomIO/SharedIO.F90
+++ b/GeomIO/SharedIO.F90
@@ -1,7 +1,7 @@
 #include "MAPL_Generic.h"
 module mapl3g_SharedIO
    use mapl_ErrorHandlingMod
-   use esmf
+   use mapl3g_InfoUtilities
    use pfio
    use gFTL2_StringVector
    use mapl3g_geom_mgr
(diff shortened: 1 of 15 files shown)


mathomp4:
CI is weird. Seems like a Github problem. I'll re-run

tclune:
> CI is weird. Seems like a Github problem. I'll re-run

Probably not the issue.   I keep forgetting to run the cap tests that also exercise the generic layer.  So my bad for pushing before all tests at least pass in my development environment.

But - now that I want to build on this, I'm going to use SU to get it into the code base.

PR #3098, opened 2024-10-15
Turn off zstandard with Baselibs

mathomp4:
## Types of change(s)

## Checklist

## Description

As found by @bena-nasa, Bucy is showing Baselibs supports zstandard compression because, I guess, HDF5 or netCDF or something found it. But, since Baselibs builds HDF5 and netCDF as static libraries, Baselibs can never *run* zstandard code. So, by fiat, if Baselibs is used, we just turn off zstandard.

Also, for consistency, set the ESMF required version in Baselibs builds to match that of Spack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


mathomp4:
I was able to test on Bucy and I needed a fix. (Had to `FORCE` the false on.). But it works for me.

bena-nasa (approved):

PR #3099, opened 2024-10-16
Allow update offsets based on timestep for extdata2g

### LINKED ISSUE #3064: ExtData2G needs variant specification for update_offset
Currently ExtData2g supports:
```
update_offset: 900T
```

We need an option to instead specify a heartbeat offset like:
```
update_offset: heartbeat
```
(Extdata knows the heartbeat_dt.)

I'm thinking that we probably need to be able to specify multiples of heartbeat, so need a bit of exploration on best way to do this.

darianboggs:
## Types of change(s)

## Checklist

## Description
Currently, the YAML configuration file for ExtData2 allows an offset for updates using an ISO8601 time duration string, such at `T09:00:00`. A user requested the capability to specify the offset as (+/-) timestep. This PR implements that capability. If map value for the `update_offset` is `HEARTBEAT`, the offset is set to an `ESMF_TimeInterval` equal to 1 timestep. If the map value is `-HEARTBEAT`, the offset is set to an `ESMF_TimeInterval` equal to -1 timestep. This may be extended to allow offsets of &#177;`n timestep`, where is a natural number.

## Related Issue
resolve #3064

### DIFF
--- a/gridcomps/ExtData2G/ExtDataUpdatePointer.F90
+++ b/gridcomps/ExtData2G/ExtDataUpdatePointer.F90
@@ -49,14 +49,11 @@ contains
       _RETURN(_SUCCESS)
    end function
 
-   function get_offset(this, rc) result(offset)
+   function get_offset(this) result(offset)
       type(ESMF_TimeInterval) :: offset
       class(ExtDataPointerUpdate), intent(in) :: this
-      integer, optional, intent(out) :: rc
-      integer :: status
 
       offset = this%offset
-      _RETURN(_SUCCESS)
 
    end function get_offset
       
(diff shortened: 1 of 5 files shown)


gridcomps/ExtData2G/ExtDataUpdatePointer.F90
```diff
@@ -71,22 +88,72 @@ subroutine create_from_parameters(this,update_time,update_freq,update_offset,tim
...
+         if(i > 1) before_string = string(:i-1)
+      end if
+      if(present(after_string)) then
+         after_string = ''
+         if(i + len(substring) <= len(string)) after_string = string(i+len(substring):)
+      end if
+
+   end subroutine split_on
+      
+   function to_upper(s) result(u)
```
bena-nasa, line 139: ESMF has a built in function to upper case strings so you can just use that.

darianboggs: Thanks. Done.

bena-nasa (approved):

darianboggs:
@bena-nasa and @tclune I added two tests to check if the result from using `HEARTBEAT` (string) matches the result from using ISO8601 Duration (string). They match for positive and negative `HEARTBEAT`.

Also, I added the capacity to include a `+` in front of the `update_offset` string. With the capability to use `HEARTBEAT`, I forsee a user using `+HEARTBEAT`, and I'd rather fix it now.

bena-nasa (approved):

Issue #3108, opened 2024-10-18
Add invalidate procedure and change name of run procedure to update for ExtensionAction objects


darianboggs:
For MAPL3, the requirements for the `ExtensionAction` derived type have evolved. We need to change the name of the existing `run` method to `update` and add a `invalidate` method. This is required for several MAPL3 development tasks.

PR #3112, opened 2024-10-18
Feature/wdboggs/change extension action methods 3108

darianboggs:
## Types of change(s)

## Checklist

## Description
Change `run` to `update` and `invalidate` method for `ExtensionAction`.

## Related Issue
Closes issue #3108


darianboggs:
Since I added `invalidate` method to `ExtensionAction`, all the types derived from it need an `invalidate` method also. Rather than add an no-op method to each of the extended types, I added a no-op `invalidate` method to the abstract type. Extended types can overload the method as needed. If we want the abstract type to have only deferred methods, we could create a "base" or "implementation" abstract type extended from `ExtensionAction` that the current extended types would extend instead. Open to suggestions.

darianboggs:
ifort is compiling. gfortran is not.

tclune (approved):

PR #3118, opened 2024-10-23
Make `get_fptr_shape` public and a bug fix

pchakraborty:
## Types of change(s)

## Checklist

## Description
- Excluding negative values of gridToFieldMap
- Making `mapl3g_FieldCondensedArray::get_fptr_shape` public

### DIFF
--- a/field_utils/FieldCondensedArray.F90
+++ b/field_utils/FieldCondensedArray.F90
@@ -56,13 +56,13 @@ contains
       logical :: has_vertical
       character(len=:), allocatable :: spec_name
       character(len=*), parameter :: VERTICAL_DIM_NONE_NAME = 'VERTICAL_DIM_NONE'
-      integer :: dimCount
+      integer :: geomDimCount
 
-      call ESMF_FieldGet(f, dimCount=dimCount, rank=rank, _RC)
+      call ESMF_FieldGet(f, geomDimCount=geomDimCount, rank=rank, _RC)
       _ASSERT(.not. rank < 0, 'rank cannot be negative.')
-      _ASSERT(.not. dimCount < 0, 'dimCount cannot be negative.')
+      _ASSERT(.not. geomDimCount < 0, 'geomDimCount cannot be negative.')
       allocate(localElementCount(rank))
-      allocate(gridToFieldMap(dimCount))
+      allocate(gridToFieldMap(geomDimCount))
       call ESMF_FieldGet(f, gridToFieldMap=gridToFieldMap, _RC) 
       !  Due to an ESMF bug, getting the localElementCount must use the module function.
       !  See FieldGetLocalElementCount (specific function) comments.
--- a/field_utils/FieldCondensedArray.F90
+++ b/field_utils/FieldCondensedArray.F90
@@ -11,7 +11,6 @@ module mapl3g_FieldCondensedArray
    implicit none
    private
    public :: assign_fptr_condensed_array
-   public :: get_fptr_shape
 
    interface assign_fptr_condensed_array
       module procedure :: assign_fptr_condensed_array_r4
@@ -72,6 +71,7 @@ contains
       has_vertical = spec_name /= VERTICAL_DIM_NONE_NAME
       fptr_shape = get_fptr_shape_private(gridToFieldMap, localElementCount, has_vertical, _RC)
 
+      _RETURN(_SUCCESS)
    end function get_fptr_shape
 
 end module mapl3g_FieldCondensedArray
--- a/field_utils/FieldCondensedArray.F90
+++ b/field_utils/FieldCondensedArray.F90
@@ -11,6 +11,7 @@ module mapl3g_FieldCondensedArray
    implicit none
    private
    public :: assign_fptr_condensed_array
+   public :: get_fptr_shape
 
    interface assign_fptr_condensed_array
       module procedure :: assign_fptr_condensed_array_r4
--- a/field_utils/FieldCondensedArray_private.F90
+++ b/field_utils/FieldCondensedArray_private.F90
@@ -26,9 +26,9 @@ contains
       
       vert_dim = 0
       vert_size = 1
-      
+
       rank = size(localElementCount)
-      grid_dims = pack(gridToFieldMap, gridToFieldMap /= 0)
+      grid_dims = pack(gridToFieldMap, gridToFieldMap > 0)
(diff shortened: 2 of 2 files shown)


field_utils/FieldCondensedArray_private.F90
```diff
@@ -26,9 +26,9 @@ function get_fptr_shape_private(gridToFieldMap, localElementCount, has_vertical,
       
       vert_dim = 0
       vert_size = 1
-      
+
       rank = size(localElementCount)
-      grid_dims = pack(gridToFieldMap, gridToFieldMap /= 0)
+      grid_dims = pack(gridToFieldMap, gridToFieldMap > 0)
```
tclune, line 31: Can you elaborate on the circumstances where elements of gridToFieldMap are negative?   My memory is that the documentation says "0" or the index of ...

field_utils/FieldCondensedArray.F90
```diff
@@ -11,6 +11,7 @@ module mapl3g_FieldCondensedArray
    implicit none
    private
    public :: assign_fptr_condensed_array
+   public :: get_fptr_shape
```
tclune, line 14: The intent is that the vertical regridder use the higher level procedure that simply returns a pointer to the correctly shaped 3D array.

We will need to understand why this does not work in your use case.

field_utils/FieldCondensedArray_private.F90
pchakraborty: For the field below,  `gridToFieldMap` is [ 1, 2, -1]
```
Field Print Starts ====>
 Field base status = Ready
   Name = 'Field034'
 Base name    = Field034
 Status: Base = Ready,  object = Ready
 Proxy        = no
 Root Info (Attributes):
{
  "MAPL": {
    "internal": {
      "long_name": "unknown",
      "standard_name": "E",
      "ungridded_dims": {
        "num_ungridded_dimensions": 0
      },
      "units": "m",
      "vertical_dim": {
        "vloc": "VERTICAL_DIM_CENTER"
      },
      "vertical_grid": {
        "num_levels": 3
      }
    }
  }
} Field status =            4
 gridrank =            2
--- ESMCI::Array::print start ---
 Base name    = Field034
 Status: Base = Ready,  object = Ready
 Proxy        = no
 Root Info (Attributes):
{}Array typekind/rank: Real*4 / 3
~ lower class' values ~
DistGrid dimCount = 2
deCount = 1
localDeCount = 1
~ local data in LocalArray for DE 0 ~
LocalArrayPrint: Array at address 0x23fe6f0:
            rank = 3, typekind = 5, base_addr = 0x14aab7580100
            dim[0] = 12  dim[1] = 13  dim[2] = 3
  Real, *4, Dim 3, Data values:
( 1, 1, 1) =  17
  skipping to end ...
(12,13, 3) =  17
dim 1: [0]: [1 [1 [1, 12] 12] 12]
dim 2: [1]: [1 [1 [1, 13] 13] 13]
dim 3: undistLBound[0]=1            undistUBound[0]=3
--- ESMCI::Array::print end ---
 arrayrank =            3
 gridToFieldMap ungriddedLBound ungriddedUBound totalLWidth totalUWidth
           1           1           3                0           0
           2          -1          -1                0           0
          -1          -1          -1               -1          -1
          -1          -1          -1               -1          -1
          -1          -1          -1               -1          -1
          -1          -1          -1               -1          -1
          -1          -1          -1               -1          -1
 Field Print Ends   ====>
```

field_utils/FieldCondensedArray.F90
pchakraborty: This doesn't need to be public when using the higher level function `assign_fptr_condensed_array`. I'll make the change.

field_utils/FieldCondensedArray_private.F90
tclune: `gridToFieldMap` should only have 2 elements in for this field:

List with number of elements equal to the grid’s dimCount.

This suggests the bug is elsewhere when allocating the array.

tclune: I think the problem is in the higher level routine.   `dimCount` was used to allocate the array but it should be `geomDimCount`.    Not that the man page makes the distinction particularly clear.

tclune: @darianboggs - please determine why the existing tests did not cover this case (and then help @pchakraborty to remedy the tests as part of this PR.)

pchakraborty: > I think the problem is in the higher level routine. `dimCount` was used to allocate the array but it should be `geomDimCount`. Not that the man page makes the distinction particularly clear.

I have pushed this change.

tclune (approved):
I would still like to see a unit test that guards against this issue.   This is the 2nd bug that has been found in this interface.  First was that the gridToFieldMap was not even being allocated, and the unit tests only covered the lower level procedures.  (@darianboggs)

pchakraborty:
> I would still like to see a unit test that guards against this issue. This is the 2nd bug that has been found in this interface. First was that the gridToFieldMap was not even being allocated, and the unit tests only covered the lower level procedures. (@darianboggs)

Created issue #3127

darianboggs (commented):
It looks good.

I'm not sure what the use case for `get_fptr_shape` being public is, but I'm not opposed. The intention was that `assign_fptr_condensed_array` would get the fptr without the user needing to know the shape. `get_fptr_shape` is used to get the shape according to the Condensed Array convention. `assign_fptr_condensed_array` uses `assign_fptr` in `FieldPointerUtilities.F90` to get the fptr. Once you have the fptr,  you can get the shape with the Fortran `shape` function.

In other words, `FieldCondensedArray.F90` depends on `FieldPointerUtilities.F90`, so you do not need to use both to get the fptr, and you don't need `get_fptr_shape` to get the shape of the fptr outside of `FieldCondensedArray.F90`

We could move all of `FieldCondensedArray.F90` to `FieldPointerUtilities.F90` with `assign_fptr_condensed_array` public and `get_fptr_shape` private. 

`FieldCondensedArray_private.F90` is internal to `field_utils` with public methods for the sake of unit testing.

Let me know if there is an unmet need.

tclune:
> I'm not sure what the use case for `get_fptr_shape` being public is, but I'm not opposed.

I think you are out of date.  That was removed.

PR #3129, opened 2024-10-28
Fixes #3128 - support copy shared attrs

### LINKED ISSUE #3128: enable propagation of shared info attributes
An export stateitem may have info attribute that is to be accessed by connected import items.   To this end, couplers must copy shared info attributes from their imports to their exports.

Currently we do not allow shared attrs to vary with time, so copying once is sufficient.  Nominally this is during initialization of the coupler, but this does make certain assumptions about ordering.   Namely, we must call the coupler init **after** the component that sets the value of the attribute, and **before** the initialization of the component(s) that consumes it.

tclune:
## Types of change(s)

## Checklist

## Description

Added interface to copy all shared attributes in one info into the shared address space of another.   Current implementation assumes that the target info object does not have a shared space already.   If that assumption is later shown to be invalid, we must instead loop over each item in shared space and then copy.

### DIFF
--- a/esmf_utils/InfoUtilities.F90
+++ b/esmf_utils/InfoUtilities.F90
@@ -36,9 +36,11 @@ module mapl3g_InfoUtilities
    public :: MAPL_InfoSet
 
    public :: MAPL_InfoCreateFromInternal
+   public :: MAPL_InfoCreateFromShared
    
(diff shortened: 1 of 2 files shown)


darianboggs (approved):
Looks good

PR #3132, opened 2024-10-28
fixes #3131

### LINKED ISSUE #3131: Timer bug in History with coupler loop in run
Just making issue for a PR. Ricardo found a bug, there was loop in the run method of History (where the couplers are applied) where the cycle should have been right at the beginning of the loop to avoid starting a timer but not stopping it.

bena-nasa:
Fixes issue #3131 
Appeared to only affect Ricardo in GEOSagcmPert.x and caused a timer to be turned on but not off causing issues up the chain in MAPL_Generic

## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Fixed bug with cycle plament in coupler loop in History
+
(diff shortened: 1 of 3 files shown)


tclune (dismissed):

mathomp4 (approved):

mathomp4:
I've borrowed this PR to get it into shape for a release.

PR #3143, opened 2024-11-01
Feature/tclune/#3140 create field factory

tclune:
## Types of change(s)

## Checklist

## Description

See commit message

### DIFF
--- a/field/FieldGet.F90
+++ b/field/FieldGet.F90
@@ -1,0 +1,53 @@
+#include "MAPL_Generic.h"
+
+module mapl3g_FieldGet
+   use mapl3g_FieldInfo
+   use mapl_KeywordEnforcer
+   use mapl_ErrorHandling
+   use esmf
+   implicit none (type, external)
+   private
+
+   public :: MAPL_FieldGet
+
+   interface MAPL_FieldGet
+      procedure field_get
+   end interface MAPL_FieldGet
+
+contains
+
+!#   subroutine field_get (field, unusable, &
+!#        ! pass thru to ESMF
+!#        status, geomtype, geom, typekind,  rank, dimCount, staggerloc, name, vm, &
+!#        ! allocatable in MAPL
+!#        minIndex,  maxIndex, elementCount, &
+!#        localMinIndex, localMaxIndex, &
+!#        ! MAPL specific
+!#        units, standard_name, long_name, &
+!#        rc)
+!#
+!#   end subroutine field_get
+
+   subroutine field_get(field, unusable, &
+        units, &
+        rc)
+
+      type(ESMF_Field), intent(in) :: field
+      class(KeywordEnforcer), optional, intent(in) :: unusable
+      character(len=:), optional, allocatable, intent(out) :: units
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+
+      if (present(units)) then
+         call MAPL_FieldInfoGetInternal(field, units=units, _RC)
+      end if
+
+      _RETURN(_SUCCESS)
+   end subroutine field_get
+      
+
+end module mapl3g_FieldGet
+        
+        
+   
--- a/field/FieldInfo.F90
+++ b/field/FieldInfo.F90
@@ -1,0 +1,147 @@
+#include "MAPL_Generic.h"
+
+module mapl3g_FieldInfo
+   use mapl3g_esmf_info_keys, only: INFO_INTERNAL_NAMESPACE
+   use mapl3g_InfoUtilities
+   use mapl3g_UngriddedDims
+   use mapl3g_VerticalStaggerLoc
+   use mapl_KeywordEnforcer
+   use mapl_ErrorHandling
+   use esmf, only: ESMF_Field
+   use esmf, only: ESMF_Info, ESMF_InfoGetFromHost, ESMF_InfoCreate
+   implicit none(type,external)
+   private
+
+   public :: MAPL_FieldInfoSetInternal
+   public :: MAPL_FieldInfoGetInternal
+
+   public :: KEY_UNITS
+   public :: KEY_LONG_NAME
+   public :: KEY_STANDARD_NAME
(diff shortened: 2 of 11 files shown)


field/FieldGet.F90
```diff
@@ -0,0 +1,53 @@
...
+!#        ! pass thru to ESMF
+!#        status, geomtype, geom, typekind,  rank, dimCount, staggerloc, name, vm, &
+!#        ! allocatable in MAPL
+!#        minIndex,  maxIndex, elementCount, &
+!#        localMinIndex, localMaxIndex, &
+!#        ! MAPL specific
+!#        units, standard_name, long_name, &
+!#        rc)
+!#
+!#   end subroutine field_get
```
darianboggs, line 29: Will this block be uncommented and used eventually, or it is a remnant?

field/FieldInfo.F90
darianboggs, line 1: It appears that this is a move in location. If it is more than that, I will review more carefully.

field/VerticalStaggerLoc.F90
```diff
@@ -0,0 +1,95 @@
...
+
+   public :: make_VerticalStaggerLoc
+
+   type :: VerticalStaggerLoc
+      private
+      integer :: id
+   contains
+      ! TODO: Convert to DTIO once compilers support allocatable internal files
+      procedure :: to_string
+   end type VerticalStaggerLoc
```
darianboggs, line 22: Could you move the name string into the derived type?
```suggestion
   type :: VerticalStaggerLoc
      private
      integer :: id
      character(len=:), allocatable :: name
   contains
      ! TODO: Convert to DTIO once compilers support allocatable internal files
      procedure :: to_string
   end type VerticalStaggerLoc
```

field/VerticalStaggerLoc.F90
```diff
@@ -0,0 +1,95 @@
...
+   end interface operator(==)
+
+   interface operator(/=)
+      procedure are_not_equal
+   end interface operator(/=)
+
+   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_NONE = VerticalStaggerLoc(1)
+   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_EDGE = VerticalStaggerLoc(2)
+   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_CENTER = VerticalStaggerLoc(3)
+   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_INVALID = VerticalStaggerLoc(4)
```
darianboggs, line 35: If you move the name string into the type, the instantiation would look like this:
```suggestion
   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_NONE = VerticalStaggerLoc(1, "VERTICAL_STAGGER_NONE")
   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_EDGE = VerticalStaggerLoc(2, "VERTICAL_STAGGER_EDGE")
   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_CENTER = VerticalStaggerLoc(3, "VERTICAL_STAGGER_CENTER")
   type(VerticalStaggerLoc), parameter :: VERTICAL_STAGGER_INVALID = VerticalStaggerLoc(4, "VERTICAL_STAGGER_INVALID")
```

field/VerticalStaggerLoc.F90
```diff
@@ -0,0 +1,95 @@
...
+
+   character(*), parameter :: VERTICAL_STAGGER_NONE_NAME = "VERTICAL_STAGGER_NONE"
+   character(*), parameter :: VERTICAL_STAGGER_EDGE_NAME = "VERTICAL_STAGGER_EDGE"
+   character(*), parameter :: VERTICAL_STAGGER_CENTER_NAME = "VERTICAL_STAGGER_CENTER"
```
darianboggs, line 39: These would go away the changes above.
```suggestion
```

field/VerticalStaggerLoc.F90
```diff
@@ -0,0 +1,95 @@
...
+         return
+      end if
+
+      if (this == VERTICAL_STAGGER_CENTER) then
+         s = VERTICAL_STAGGER_CENTER_NAME
+         return
+      end if
+
+      s = "VERTICAL_STAGGER_INVALID"
+   end function to_string
```
darianboggs, line 81: These would be changed in line with the changes suggested above.

field/VerticalStaggerLoc.F90
darianboggs: This would allow more easy lookup by name string.

darianboggs (approved):
All suggestions are optional.

field/VerticalStaggerLoc.F90
tclune: Unfortunately, this would prevent declaring constants (parameters).   This is an annoying Fortran feature that you can't use a structure constructor to fill in an allocatable string, and at the same time you cannot use a user-define constructor in a constant expression.

I think the priority is for having defined constants. Hence the implementation here.

field/VerticalStaggerLoc.F90
tclune: It would be nice, but ...

field/VerticalStaggerLoc.F90
tclune: Agreed.  They would ...

tclune:
> All suggestions are optional.

It was a good suggestion - other than that it does not work.

I should add that non-allocatable strings would work, but then we have to pad them and use trim and all that.    I may look at that for the next revision.

field/FieldGet.F90
tclune: I've deleted in my branch where I'm extending and integrating this.

field/FieldInfo.F90
tclune: Next PR does a lot more here.  Save your energy for then.

PR #3162, opened 2024-11-07
Further progress towards integrating support for info at Field and

tclune:
FieldBundle.

- introduced FieldBundleType_Flag to check whether a bundle is a bracket bundle.
- introduced FieldundleInfo.F90 - analog to FieldInfo.F90

Changes unfortunately leaked into lots of other code.

## Types of change(s)

## Checklist


GeomIO/SharedIO.F90
```diff
@@ -188,23 +189,41 @@ subroutine add_vertical_dimensions(bundle, metadata, rc)
...
+      call MAPL_FieldBundleGet(bundle, fieldList=fieldList, _RC)
+
+      vertical_names = StringVector()
+      do i = 1, size(fieldList)
+         _HERE, i, size(fieldList)
+         call MAPL_FieldGet(fieldList(i), vert_staggerloc=vert_staggerloc, _RC)
+         dim_name = vert_staggerloc%get_dimension_name()
+         if (dim_name == "") cycle
+         
+         call MAPL_FieldGet(fieldList(i), num_vgrid_levels=num_vgrid_levels, _RC)
```
darianboggs, line 212: Is num_vgrid_levels used?

GeomIO/SharedIO.F90
```diff
@@ -188,23 +189,41 @@ subroutine add_vertical_dimensions(bundle, metadata, rc)
...
+         call vertical_names%push_back(dim_name)
+         _HERE, i, size(fieldList)
       end do
+
+      associate (e => vertical_names%ftn_end())
+        iter = vertical_names%ftn_begin()
+        do while(iter /= e)
+           call iter%next()
+           dim_name = iter%of()
+           num_levels = vert_staggerloc%get_num_levels(num_vgrid_levels)
```
darianboggs, line 222: This seems to be where num_vgrid_levels is used, but I don't think that it will have the right value here.

field/VerticalStaggerLoc.F90
```diff
@@ -80,4 +82,36 @@ elemental logical function are_not_equal(this, that)
...
+      case ("VERTICAL_STAGGER_NONE")
+         num_levels = 0
+      case ("VERTICAL_STAGGER_EDGE")
+         num_levels = num_vgrid_levels
+      case ("VERTICAL_STAGGER_CENTER")
+         num_levels = num_vgrid_levels - 1
+      case default
+         num_levels = -1
+      end select
+   end function get_num_levels
```
darianboggs, line 115: 1. Could you use this%id for the select case in both functions? You could use module private integer parameters for the id values so that the constructed instances and these functions are consistent.
2. Also, would it be worthwhile to make dim_name an attribute of VerticalStaggerLoc that is set at construction? It would add an argument to the constructor and attribute for the instance, but it would simplify the get_dimension_name.

generic3g/vertical/FixedLevelsVerticalGrid.F90
```diff
@@ -147,6 +147,44 @@ impure elemental logical function not_equal_FixedLevelsVerticalGrid(a, b) result
...
+   ! a 1D array at each point of the horizontal grid
+   function esmf_field_create_(geom, farray1d, rc) result(field)
+      type(ESMF_Field) :: field ! result
+      type(ESMF_Geom), intent(in) :: geom
+      real(kind=REAL32), intent(in) :: farray1d(:)
+!#      character(len=*), intent(in) :: vloc
+      integer, optional, intent(out) :: rc
+
+      integer, allocatable :: local_cell_count(:)
+      real(kind=REAL32), pointer :: farray3d(:, :, :)
```
darianboggs, line 160: Should the ESMF_KIND be used here like this?
```suggestion
      real(kind=REAL32), pointer :: farray3d(:, :, :)
```

darianboggs (approved):
Looks good. A few comments.

field/VerticalStaggerLoc.F90
tclune: You've got it right.  I did not use `id` because wanted to avoid magic numbers.  And introducing named parameters gets weird here given the string names.    Would have done it anyway, but for the fact the need for this was last minute (see previous response).

Taking a note to fix this in next PR.

field_bundle/tests/Test_FieldBundleDelta.pf
```diff
@@ -165,22 +158,22 @@ contains
...
+!#      do i = 1, FIELD_COUNT
+!#         call ESMF_FieldGet(fieldList(i), fArrayPtr=x_r8, _RC)
+!#         @assert_that(shape(x_r8), is(equal_to([4,4])))
+!#      end do
+!#
+!#      call MAPL_FieldBundleGet(bundle, interpolation_weights=weights, _RC)
+!#      @assert_that(weights, is(equal_to(DEFAULT_WEIGHTS)))
+!#
+!#      call teardown_bundle(bundle)
+!#      call teardown_geom(geom)
```
tclune, line 176: Oops - this should not hav been commented.  Am uncommenting.  If it breaks CI I'll put the other changes in as well.
```suggestion
      delta = FieldBundleDelta(FieldDelta(typekind=ESMF_TYPEKIND_R8))
      call delta%update_bundle(bundle, _RC)

      call MAPL_FieldBundleGet(bundle, fieldList=fieldList, _RC)
      @assert_that(size(fieldList), is(FIELD_COUNT))

      do i = 1, FIELD_COUNT
         call ESMF_FieldGet(fieldList(i), fArrayPtr=x_r8, _RC)
         @assert_that(shape(x_r8), is(equal_to([4,4])))
      end do

      call MAPL_FieldBundleGet(bundle, interpolation_weights=weights, _RC)
      @assert_that(weights, is(equal_to(DEFAULT_WEIGHTS)))

      call teardown_bundle(bundle)
      call teardown_geom(geom)
```

generic3g/vertical/FixedLevelsVerticalGrid.F90
tclune, line 160:
```suggestion
      real(kind=ESMF_KIND_R4), pointer :: farray3d(:, :, :)
```

tclune: Note - this code only shows up as mine because of a merge conflict.

field/VerticalStaggerLoc.F90
tclune: Regarding (2).   I don't like the idea of "overspecifying" the value of an object.    Becomes more difficult to ensure consistency if someone gets creative later on.    I agree that it would simplify the new method.  OTOH, once I introduce the integer named parameter, this just becomes a SELECT CASE that can be easily verified by eye.

But the dim names should be params as well.  Adding that to my TODO list.

tclune: fixed in next push.

PR #3166, opened 2024-11-08
Accumulator Additional Tests and Fixes

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR provides additional tests for AccumlatorAction and the types derived from it. Also, it includes fixes to AccumulatorAction and derived types in response to tests.


generic3g/actions/MinAccumulator.F90
```diff
@@ -7,10 +7,9 @@ module mapl3g_MinAccumulator
    use ESMF
    implicit none
    private
-   public :: AccumulatorAction
+   public :: MinAccumulator
```
tclune, line 10: I really prefer all of the subclasses to end in "Action".    Am open to reasons why that is not appropriate here.

generic3g/tests/Test_MaxAccumulator.pf
```diff
@@ -0,0 +1,55 @@
...
+
+   end subroutine test_max_accumulate_R4
+
+   @Before
+   subroutine set_up()
+      integer :: status
+
+      if(is_initialized()) return
+      call ESMF_Initialize(_RC)
+
```
tclune, line 52: No.   Cannot be done at the suite level.  We discussed this last week.

tclune: E.g., where will you put the finalize?

tclune (commented):
Have a few issues that I want to discuss before merging.

generic3g/actions/MinAccumulator.F90
darianboggs: Done

generic3g/tests/Test_MaxAccumulator.pf
darianboggs: Since I cannot initialize ESMF at the suite level, I initialize it before a test if and only if it has not been initialized. I cannot run finalize after a test, which is not a problem until after the last test runs. Then ESMF is not finalized, and then ... “Human sacrifice! Dogs and cats living together! Mass hysteria!”

I have my notes from our discussion about the ESMF Initialize/Finalize problem with pFUnit, and I will update the test once I have the solution in place.

tclune:
I understand.  But why not use the mechanism that the rest of MAPL tests are using in this regard?      I'm not doubting that it will work.

darianboggs:
I don't know why GEOSadas tests are failing, but I assume they are not fatal.

tclune:
> I don't know why GEOSadas tests are failing, but I assume they are not fatal.

They are random and unrelated to any MAPL3 code.     And yes, very annoying.

generic3g/actions/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ target_sources(MAPL.generic3g PRIVATE
 
   TimeInterpolateAction.F90
   AccumulatorAction.F90
-  MeanAccumulator.F90
-  MaxAccumulator.F90
-  MinAccumulator.F90
+  MeanAction.F90
+  MaxAction.F90
+  MinAction.F90
```
tclune, line 16: These may hold for now, but am concerned that they do not properly indicate that the mean/max/min are over time rather than space.   Hopefully clear in context, and invisible to external users in any event.

tclune (approved):

PR #3170, opened 2024-11-11
Fixes #3169 - handle duplicate dimensions

### LINKED ISSUE #3169: Allow FileMetadata::add_dimension() to allow duplicate dimension requests
When creating metadata for a FieldBundle, many/most variables share various dimensions.   Rather than having to digest these to find the unique dimensions, it will be more convenient to simply not flag it as an error when an identical dimension is requested to be added to a metadata object.

The intent here is to preserve the existing behavior, but add a new method that silently ignores duplicates.

tclune:
Issue #3169 was incorrect - code was not trapping duplicate dimensions
at all.   Now throws exception if duplicate dim name has different
extent.

## Types of change(s)

## Checklist

### DIFF
--- a/pfio/FileMetadata.F90
+++ b/pfio/FileMetadata.F90
@@ -123,11 +123,20 @@ contains
       class (FileMetadata), target, intent(inout) :: this
       character(len=*), intent(in) :: dim_name
       integer, intent(in) :: extent
-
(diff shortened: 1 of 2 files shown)


tclune:
Good enough on CI for trivial change.

PR #3173, opened 2024-11-12
Add a fake DYN grid comp

pchakraborty:
## Types of change(s)

## Checklist

## Description

- Added a fake DYN gridcomp, `FakeDynGridComp.F90`, whose `initialize` method runs during the phase `INIT_REALIZE`, and initializes the pressure and temperature fields.
- Updated the Scenarios test `vertical_regridding_3` to use this FakeDyn gridcomp.
- Added accessor `OuterMetaComponent::get_vertical_grid`


gridcomps/FakeDyn/FakeDynGridComp.F90
```diff
@@ -0,0 +1,163 @@
+#include "MAPL_Generic.h"
```
tclune, line 1: Do you really need a new directory for this?    We have gridcomps.

gridcomps/FakeDyn/FakeDynGridComp.F90
```diff
@@ -0,0 +1,163 @@
...
+      type(OuterMetaComponent), pointer :: outer_meta
+      type(ESMF_Geom) :: geom
+      class(VerticalGrid), allocatable :: vertical_grid
+      type(ESMF_Field) :: field1, field2
+      integer :: num_levels, status
+
+      outer_meta => get_outer_meta_from_inner_gc(gridcomp,_RC)
+      geom = outer_meta%get_geom()
+      vertical_grid = outer_meta%get_vertical_grid()
+      num_levels = vertical_grid%get_num_levels()
```
tclune, line 53: It might be useful to make this look even more like a true user component.   In particular, I don't want ordinary GEOS users to "see" the outer meta.

pchakraborty: Is there another way to retrieve `geom`? This is a special case, since this needs to execute before `INIT_USER` phase (where the coupler gets initialized), and at this stage an empty field has been create, but `geom` has not been set yet.

gridcomps/FakeDyn/FakeDynGridComp.F90
pchakraborty: Not really. I'll pull it out.

gridcomps/FakeDyn/FakeDynGridComp.F90
tclune: For (1) there is a special init phase for the purpose.

generic3g/MAPL_Generic.F90
```diff
@@ -239,6 +242,11 @@ subroutine gridcomp_get(gridcomp, unusable, &
       if (present(logger)) logger => outer_meta_%get_lgr()
       if (present(registry)) registry => outer_meta_%get_registry()
       if (present(geom)) geom = outer_meta_%get_geom()
+      if (present(num_levels)) then
```
tclune, line 245: This needs discussion.   I remain concerned about "num_levels" being ambiguous, though less so in this context.

Possibly we just want the user to retrieve the vertical grid, like they would with geom.

pchakraborty: It can be easily be done but is there a need for users to retrieve anything other than the `num_levels`? We can rename it to `num_vertical_levels`.

tclune: The ambiguity isn't whether it is "vertical" it is whether it is edges vs centers, grid vs field, ...

Nothing fundamental, but we want to get it "right" (TM) before we base too much code on it.

pchakraborty: Replaced the `num_level` accessor with `vertical_grid`.

gridcomps/FakeDyn/FakeDynGridComp.F90
pchakraborty: On second thought, the separate directory lets me create a separate `libMAPL.fakedyn.so` easily.

tclune: ?    The existing directory does that for the other test gridcomps.   We'll discuss tomorrow.

pchakraborty: Ah, I had it in the wrong place - in the top level `gridcomps` directory. Moved it to `generic3g/tests/gridcomps`.

tclune: Hmm - how did I miss that.   It looked like it was just one level too far down ...

generic3g/tests/Test_Scenarios.pf
```diff
@@ -131,7 +131,7 @@ contains
 #ifndef __GFORTRAN__
               , &
               ScenarioDescription('vertical_regridding_2',   'parent.yaml',      check_name, check_stateitem), &
-              ScenarioDescription('vertical_regridding_3',   'parent.yaml',      check_name, check_stateitem)  &
+              ScenarioDescription('vertical_regridding_3',   'AGCM.yaml',        check_name, check_stateitem)  &
```
tclune, line 134: We can't leave this `#ifndef __GFORTRAN__` forever.   Will approve for now.

generic3g/tests/gridcomps/FakeDynGridComp.F90
```diff
@@ -0,0 +1,155 @@
...
+
+      call MAPL_RunChildren(gridcomp, phase_name="run", _RC)
+
+      _RETURN(_SUCCESS)
+      _UNUSED_DUMMY(importState)
+      _UNUSED_DUMMY(exportState)
+      _UNUSED_DUMMY(clock)
+   end subroutine run
+
+   subroutine field_complete_(field, geom, num_levels, vertical_stagger, units, standard_name, rc)
```
tclune, line 77: I do not understand why this is needed.      User gridcomps should let the generic layer allocate fields.    the fact that this is needed here suggests a hole in the vgrid functionality that should be addressed.

generic3g/tests/gridcomps/FakeDynGridComp.F90
```diff
@@ -0,0 +1,155 @@
...
+      call assign_fptr_condensed_array(field, farr, _RC)
+      shape_ = shape(farr); num_levels = shape_(2)
+      do concurrent(horz = 1:shape_(1), ungrd = 1:shape_(3))
+         do vert = 1, num_levels
+            farr(horz, vert, ungrd) = real((num_levels - vert + 1) * 10)
+         end do
+      end do
+
+      _RETURN(_SUCCESS)
+   end subroutine set_pressure_
```
tclune, line 121: Fine for now.  But eventually we want this to look more like a User gridcomp where we use MAPL_GetPointer() (or MAPL3 equivalent).

generic3g/tests/gridcomps/FakeDynGridComp.F90
pchakraborty: Made the changes
- Linear interpolation matrix between src and dst vcoord are now done in the `update` step.
- That change lets us have the FakeDyn gridcomp's initialize step run during the `INIT_USER` phase where the fields are complete - so no need to allocate them.

Now we don't need to retrieve `vertical_grid` from the gridcomp (via `MAPL_GridCompGet`, but it may not be a bad idea to leave that capability in.

generic3g/tests/gridcomps/FakeDynGridComp.F90
pchakraborty: Using `MAPL_GetPointer` to access the data pointer from state.

tclune (approved):

PR #3174, opened 2024-11-12
Fixes #2845. Allow ESMF_CONFIG_FILE to specify ESMF control file

### LINKED ISSUE #2845: Specify ESMF Config File via environment variable?
In talking with @darianboggs about some ESMF stuff I was reminded of https://github.com/GEOS-ESM/MAPL/pull/2233 which fixed #2133, a request from @lizziel to allow the ability to set the ESMF logging level via a file rather than via command line option.

But, it is *hardcoded* that the file name is `ESMF.rc` and nothing else. Now, in MAPL3 land, this will need to be (or should be) a YAML file. So that means `ESMF.yml` or `ESMF.yaml`.

But but, @darianboggs reminded me that `get_environment_variable` does exist in Fortran.

So my thinking is maybe we should do something like:

1. Look for environment variable `ESMF_CONTROL_FILE` or whatever (@tclune will have good opinions on this). If found, check if the file ends in `.rc`, `.yml` or `.yaml`. If so, feed that in. (Note: I believe ESMF *depends* on a file ending in `.yml` or `.yaml` to trigger hconfig, so we would want to enforce that and then it makes our life easy if we say `ESMF_Config` files end in `.rc`.)
2. If no environment variable, look for `ESMF.rc`, `ESMF.yml` or `ESMF.yaml`, and use that
3. Check for command line argument.

Now, I suppose, the command line argument should maybe "win" this battle, but I also know the arguments might be going away in favor of yamls so... 🤷🏼 

I suppose the question is: would this be useful? I can sort of see maybe having a set of control files for scripting purposes with different names and then using the environment variable to select?

mathomp4:
## Types of change(s)

## Checklist

## Description

This PR adds the ability for users to set an environment variable, `ESMF_CONFIG_FILE`, that can point to a file where pre `ESMF_Initialize` values can be set. Previously, the file *had* to be called `ESMF.rc`. Now, it looks at `ESMF_CONFIG_FILE` and if that isn't set, it then looks for `ESMF.rc` (for backwards compatibility).

I have tested this with GEOS and it works as expected.

@tclune if you have time, let me know if you have preferences for what to call the environment variable, etc.

## Related Issue

Closes #2845

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Allow update offsets of &#177;timestep in ExtData2G
 - Minor revision (and generalization) of grid-def for GSI purposes
-- Trajectory sampler: fix a bug when group_name does not exist in netCDF file and a bug that omitted the first time point
+- Add ability to use an `ESMF_CONFIG_FILE` environment variable to specify name of file to pass in pre-`ESMF_Initialize` options to ESMF (see [ESMF Docs](https://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node4.html#SECTION04024000000000000000) for allowed flags.
 
 ### Changed
 
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed issue of some Baselibs builds appearing to support zstandard. This is not possible due to Baselibs building HDF5 and netCDF as static libraries
+- Trajectory sampler: fix a bug when group_name does not exist in netCDF file and a bug that omitted the first time point
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added macro _RETURN(_SUCCESS) to fetch_data
+- Added macro `_RETURN(_SUCCESS)` to fetch_data
 - Allow update offsets of &#177;timestep in ExtData2G
 - Minor revision (and generalization) of grid-def for GSI purposes
 - Add ability to use an `ESMF_CONFIG_FILE` environment variable to specify name of file to pass in pre-`ESMF_Initialize` options to ESMF (see [ESMF Docs](https://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node4.html#SECTION04024000000000000000) for allowed flags.
-- Trajectory sampler: fix a bug when group_name does not exist in netCDF file and a bug that omitted the first time point
 - Allow lat-lon grid factory to detect and use CF compliant lat-lon bounds in a file when making a grid
 - PFIO/Variable class, new procedures to retrieve string/reals/int attributes from a variable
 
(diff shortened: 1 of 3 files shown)


818Nawaf (approved):

CHANGELOG.md
```diff
@@ -50,7 +51,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed issue of some Baselibs builds appearing to support zstandard. This is not possible due to Baselibs building HDF5 and netCDF as static libraries
+<<<<<<< HEAD
+- Trajectory sampler: fix a bug when group_name does not exist in netCDF file and a bug that omitted the first time point
+=======
```
tclune, line 56: Failed merge ...

gridcomps/Cap/MAPL_Cap.F90
```diff
@@ -288,16 +289,40 @@ subroutine run_model(this, comm, unusable, rc)
...
+      esmfConfigFileLen = len(esmfConfigFile)
+
+      ! Step two: get the length of the environment variable
+      call get_environment_variable('ESMF_CONFIG_FILE', length=esmfConfigFileLen, status=status)
+      ! Step three: if the environment variable exists, get the value of the environment variable
+      if (status == 0) then
+         ! We need to deallocate so we can reallocate
+         deallocate(esmfConfigFile)
+         allocate(character(len = esmfConfigFileLen) :: esmfConfigFile)
+         call get_environment_variable('ESMF_CONFIG_FILE', value=esmfConfigFile, status=status)
```
tclune, line 308: Still need to check status again.

CHANGELOG.md
mathomp4: Well that is weird. I've fixed that.

gridcomps/Cap/MAPL_Cap.F90
mathomp4: Ahhh. Very true. okay, I'll add that and test!

gridcomps/Cap/MAPL_Cap.F90
```diff
@@ -288,16 +289,40 @@ subroutine run_model(this, comm, unusable, rc)
...
+
+      lgr => logging%get_logger('MAPL')
 
       ! If the file exists, we pass it into ESMF_Initialize, else, we
       ! use the one from the command line arguments
-      if (file_exists) then
-         call ESMF_Initialize (configFileName='ESMF.rc', mpiCommunicator=comm, vm=vm, _RC)
+      if (esmfConfigFileExists) then
+         call lgr%info("Using ESMF configuration file: %a", esmfConfigFile)
+         call ESMF_Initialize (configFileName=esmfConfigFile, mpiCommunicator=comm, vm=vm, _RC)
```
tclune, line 325: If `esmfConfigFile` is allocatable, then you do not need a separate call to ESMF_Initialized.  An unallocated actual appears as a non-present dummy.  So this conditional can be simplified.   (You may still want the conditional for the lgr bit though.)

gridcomps/Cap/MAPL_Cap.F90
```diff
@@ -288,16 +289,40 @@ subroutine run_model(this, comm, unusable, rc)
...
+      ! we use the default of "ESMF.rc" for backward compatibility
+
+      ! Step one: default to ESMF.rc
+
+      esmfConfigFile = 'ESMF.rc'
+      esmfConfigFileLen = len(esmfConfigFile)
+
+      ! Step two: get the length of the environment variable
+      call get_environment_variable('ESMF_CONFIG_FILE', length=esmfConfigFileLen, status=status)
+      ! Step three: if the environment variable exists, get the value of the environment variable
```
tclune, line 303: I'd like to say that this back-and-forth is unneccesary, but I don't immediately see a way to simplify this.

tclune (changes requested):
One easy change.  One slightly less so.

tclune (approved):

PR #3178, opened 2024-11-13
check tableEnd efficiently

weiyuan-jiang:
## Types of change(s)

## Checklist

## Description
The "::" will not be returned so it is not necessary to check it every time

### DIFF
--- a/gridcomps/Cap/MAPL_CapGridComp.F90
+++ b/gridcomps/Cap/MAPL_CapGridComp.F90
@@ -979,12 +979,11 @@ contains
 
     cap_import = ""
     if (present) then
-
-       do while(trim(cap_import) /= "::")
+       do while( .true.)
           call ESMF_ConfigNextLine(config, tableEnd=tableEnd, _RC)
           if (tableEnd) exit
           call ESMF_ConfigGetAttribute(config, cap_import, _RC)
-          if (trim(cap_import) /= "::") call vec%push_back(trim(cap_import))
+          call vec%push_back(trim(cap_import))
        end do
     end if
     _RETURN(_SUCCESS)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- refactored tableEnd check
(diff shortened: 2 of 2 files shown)


gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -979,12 +979,11 @@ function get_vec_from_config(config, key, rc) result(vec)
 
     cap_import = ""
     if (present) then
-
-       do while(trim(cap_import) /= "::")
+       do while( .true.)
```
mathomp4, line 982: Hmm. A query for @tclune . As far as my reading of the Standard goes, I think the loop-control bit of do is optional. So this:
```fortran
do
  ...
end do
```
seems legal. 

But is that considered "better" than:
```fortran
do while (.true.)
   ...
end do
```
for readable code?

tclune: Both are legal.  Modest preference for the simpler form,  but both are rather rare in practice.

But there is a more important problem here.   Why is `cap_import` being set if it is not referenced?   Before it was, but was effectively the same as `.true.` in that reference.       But the real fix should be to eliminate the variable entirely so that people like me don't go look at the loop to see how it is being used/modified there.   The original conditional really screamed that it would get updated in the loop.  The newer form reduces that assumption.  Elimination of the variable makes it crystal clear.

tclune: Sorry - ignore that last rant.   I got distracted and failed to see that it is actually used in the loop.   Just does not play a role in termination.   However  it does look like the assignment before the loop is unnecessary.

weiyuan-jiang: I removed the assignment. The name 'cap_import' is also changed to 'value'. This call can be used in any case. At least it was used to read cap_export

gridcomps/Cap/MAPL_CapGridComp.F90
```diff
@@ -967,24 +967,22 @@ end function get_CapGridComp_from_gc
...
-    type(StringVector) :: vec
+    character(len=ESMF_MAXSTR) :: value
 
     call ESMF_ConfigFindLabel(config, key//":", isPresent = present, _RC)
 
-    cap_import = ""
     if (present) then
-
-       do while(trim(cap_import) /= "::")
+       do while(.true.)
```
tclune, line 981:
```suggestion
       do
```

tclune (approved):

mathomp4 (approved):

mathomp4 (approved):

PR #3182, opened 2024-11-15
Select short name of variable defining ModelVerticalGrid

pchakraborty:
## Types of change(s)

## Checklist

## Description

- A Model VerticalGrid is now instantiated with `standard_name`, `units` and `num_levels`. The method `add_short_name` is used to specify variable short names corresponding to edge and center vertical staggerings, which is selected at runtime using the method `get_short_name`.
- Updated `ComponentSpecParse::parse_geometry_spec` to use the new interface.
- Updated `Test_ModelVerticalGrid.pf` to use the new interface.
- Updated vertical regridding tests to specify `standard_name` instead of `short_name` for Model VerticalGrids.
- Cleanup of FixedLevels VerticalGrid - we didn't need `MAPL_BaseMod::MAPL_GridGet` after all.

### DIFF
--- a/generic3g/ComponentSpecParser/parse_geometry_spec.F90
+++ b/generic3g/ComponentSpecParser/parse_geometry_spec.F90
@@ -107,13 +107,12 @@ contains
             levels = ESMF_HConfigAsR4Seq(vertical_grid_cfg, keyString='levels' ,_RC)
             vertical_grid = FixedLevelsVerticalGrid(standard_name, levels, units)
          case('model')
-            num_levels = ESMF_HConfigAsI4(vertical_grid_cfg, keyString='num_levels', _RC)
+            standard_name = ESMF_HConfigAsString(vertical_grid_cfg, keyString='standard_name', _RC)
             units = ESMF_HConfigAsString(vertical_grid_cfg, keyString='units', _RC)
-            vertical_grid = ModelVerticalGrid(num_levels=num_levels, units=units)
-            short_name = ESMF_HConfigAsString(vertical_grid_cfg, keyString='short_name', _RC)
+            num_levels = ESMF_HConfigAsI4(vertical_grid_cfg, keyString='num_levels', _RC)
+            vertical_grid = ModelVerticalGrid(standard_name=standard_name, units=units, num_levels=num_levels)
             select type(vertical_grid)
             type is(ModelVerticalGrid)
-               call vertical_grid%add_variant(short_name=short_name)
                call vertical_grid%set_registry(registry)
             end select
          case default
--- a/generic3g/ComponentSpecParser/parse_geometry_spec.F90
+++ b/generic3g/ComponentSpecParser/parse_geometry_spec.F90
@@ -33,7 +33,7 @@ contains
       type(GeomManager), pointer :: geom_mgr
       class(GeomSpec), allocatable :: geom_spec
       integer :: num_levels
-      character(:), allocatable :: vertical_grid_class, standard_name, units, short_name
+      character(:), allocatable :: vertical_grid_class, standard_name, units
       class(VerticalGrid), allocatable :: vertical_grid
       real, allocatable :: levels(:)
 
@@ -114,6 +114,7 @@ contains
             select type(vertical_grid)
             type is(ModelVerticalGrid)
                call vertical_grid%set_registry(registry)
+               call vertical_grid%add_short_names(edge="PLE", center="PL")
             end select
          case default
             _FAIL('vertical grid class '//vertical_grid_class//' not supported')
--- a/generic3g/ComponentSpecParser/parse_geometry_spec.F90
+++ b/generic3g/ComponentSpecParser/parse_geometry_spec.F90
@@ -114,7 +114,13 @@ contains
             select type(vertical_grid)
             type is(ModelVerticalGrid)
                call vertical_grid%set_registry(registry)
-               call vertical_grid%add_short_names(edge="PLE", center="PL")
+               if (standard_name == "air_pressure") then
+                  call vertical_grid%add_short_names(edge="PLE", center="PL")
+               else if (standard_name == "height") then
+                  call vertical_grid%add_short_names(edge="ZLE", center="ZL")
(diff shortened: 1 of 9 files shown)


generic3g/ComponentSpecParser/parse_geometry_spec.F90
```diff
@@ -107,14 +107,20 @@ module function parse_geometry_spec(mapl_cfg, registry, rc) result(geometry_spec
...
             type is(ModelVerticalGrid)
-               call vertical_grid%add_variant(short_name=short_name)
                call vertical_grid%set_registry(registry)
+               if (standard_name == "air_pressure") then
+                  call vertical_grid%add_short_names(edge="PLE", center="PL")
+               else if (standard_name == "height") then
+                  call vertical_grid%add_short_names(edge="ZLE", center="ZL")
+               else
+                  _FAIL("unsupported standard name ["//standard_name//"]")
+               end if
```
tclune, line 123: These cannot be hardwired here.  At least not in the long term.   How about requiring them in the yaml file when you try to specify the vertical grid that way:
```yaml
vertical_grid:
    coordinate:
        dimension: pressure
        stagger: edge
        field: PLE
    coordinate:
        dimension: height
        stagger: center
        field: ZL
...
```

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -65,89 +63,99 @@ module function can_connect_to(this, src, rc)
...
-      !# vgrid%registry => registry
    end function new_ModelVerticalGrid_basic
 
    integer function get_num_levels(this) result(num_levels)
       class(ModelVerticalGrid), intent(in) :: this
       num_levels = this%num_levels
    end function get_num_levels
 
-   subroutine add_variant(this, short_name)
+   subroutine add_short_names(this, edge, center)
```
tclune, line 83: Not sure that this is what we want here.  It is conceivable that someone could provide just the edge and not the center. or vice versa.     The center can in theory be derived from the edge (with some assumptions that appear to be nigh universal.)

It can stay for now, but be prepared for something where instead you specify name and staggerloc.

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -65,89 +63,99 @@ module function can_connect_to(this, src, rc)
...
       class(ModelVerticalGrid), intent(inout) :: this
-      character(*), intent(in) :: short_name
+      character(*), optional, intent(in) :: edge
```
tclune, line 85: You need an "unusable" argument here to separate out the options.   Both arguments are the same type (character), so it is dangerously ambiguous which one is intended.

generic3g/vertical/ModelVerticalGrid.F90
```diff
@@ -65,89 +63,99 @@ module function can_connect_to(this, src, rc)
...
+
+      if (vertical_dim_spec == VERTICAL_DIM_EDGE) then
+         short_name = this%short_name_edge
+         _RETURN(_SUCCESS)
+      else if (vertical_dim_spec == VERTICAL_DIM_CENTER) then
+         short_name = this%short_name_center
+         _RETURN(_SUCCESS)
+      else
+         _FAIL("unsupported vertical_dim_spec")
+      end if
```
tclune, line 106: War on nesting ...   (Makes it more obvious that a final `_RETURN(_SUCCESS)` is not needed also)
```suggestion
      if (vertical_dim_spec == VERTICAL_DIM_EDGE) then
         short_name = this%short_name_edge
         _RETURN(_SUCCESS)
      end if
      if (vertical_dim_spec == VERTICAL_DIM_CENTER) then
         short_name = this%short_name_center
         _RETURN(_SUCCESS)
      end if
      _FAIL("unsupported vertical_dim_spec")
```

tclune (changes requested):
A couple of minor changes requested.  See inline comments.

generic3g/vertical/ModelVerticalGrid.F90
pchakraborty: There is no nesting here. Also I don't see how separate if blocks is any more readable than the if/else if/else block.

generic3g/vertical/ModelVerticalGrid.F90
pchakraborty: Not sure I understand. Both `edge` and `center` are optional arguments. Did you mean something like
```Fortran
   subroutine add_short_names(this, edge, center, unusable)
      class(ModelVerticalGrid), intent(inout) :: this
      character(*), optional, intent(in) :: edge
      class(KeywordEnforder), optional, intent(in) :: unusable
      character(*), optional, intent(in) :: center

      if (present(edge)) this%short_name_edge = edge
      if (present(center)) this%short_name_center = center
      _UNUSED_DUMMY(unusable)
   end subroutine add_short_names
```

generic3g/vertical/ModelVerticalGrid.F90
tclune: I drank the proverbial cool-aid provided by some YouTube videos on "never nesting".   (Relevant text post here: https://dev.to/dglsparsons/write-better-code-and-be-a-better-programmer-by-never-using-else-statements-4dbl)

I will grant that this form is nearly like a SELECT CASE construct which is acceptable by that standard.

I strongly suspect that if we refactor just a bit further we get something that we both agree is more readable.    In any event this is not an example worth spending more time on.  I yield.

generic3g/vertical/ModelVerticalGrid.F90
tclune: No  - I mean "unusable" should be even before edge.   Force the user to be explicit since there is no obvious ordering of "center" vs "edge" in this context.

pchakraborty: Got it. I had seen the `unusable` argument, but was not familiar with its usage. Pushed the change.

818Nawaf (approved):

tclune (approved):
Looks good.

PR #3189, opened 2024-11-20
Update MAPL3 time averaging to use a pointwise counter

darianboggs:
## Types of change(s)

## Checklist

## Description
Time Averaging in MAPL3 needs a counter for each grid point so that the counts can vary points to account for a variable number of updates to the accumulation field over the field.

### DIFF
--- a/field/FieldPointerUtilities.F90
+++ b/field/FieldPointerUtilities.F90
@@ -33,6 +33,8 @@ module MAPL_FieldPointerUtilities
       module procedure assign_fptr_r8_rank2
       module procedure assign_fptr_r4_rank3
       module procedure assign_fptr_r8_rank3
+      module procedure assign_fptr_i4_rank1
+      module procedure assign_fptr_i8_rank1
    end interface assign_fptr
 
    interface FieldGetCptr
@@ -989,5 +991,43 @@ contains
       _RETURN(ESMF_SUCCESS)
 
    end subroutine Destroy
+
+   subroutine assign_fptr_i4_rank1(x, fptr, rc)
+      type(ESMF_Field), intent(inout) :: x
+      integer(kind=ESMF_KIND_I4), pointer, intent(out) :: fptr(:)
+      integer, optional, intent(out) :: rc
+
+      ! local declarations
+      type(c_ptr) :: cptr
+      integer(ESMF_KIND_I8), allocatable :: fp_shape(:)
+      integer(ESMF_KIND_I8) :: local_size
+      integer :: status
+
+      local_size = FieldGetLocalSize(x, _RC)
+      fp_shape = [ local_size ]
+      call FieldGetCptr(x, cptr, _RC)
+      call c_f_pointer(cptr, fptr, fp_shape)
+
+      _RETURN(_SUCCESS)
+   end subroutine assign_fptr_i4_rank1
+
+   subroutine assign_fptr_i8_rank1(x, fptr, rc)
+      type(ESMF_Field), intent(inout) :: x
+      integer(kind=ESMF_KIND_I8), pointer, intent(out) :: fptr(:)
+      integer, optional, intent(out) :: rc
+
+      ! local declarations
+      type(c_ptr) :: cptr
+      integer(ESMF_KIND_I8), allocatable :: fp_shape(:)
+      integer(ESMF_KIND_I8) :: local_size
+      integer :: status
+
+      local_size = FieldGetLocalSize(x, _RC)
+      fp_shape = [ local_size ]
+      call FieldGetCptr(x, cptr, _RC)
+      call c_f_pointer(cptr, fptr, fp_shape)
+
+      _RETURN(_SUCCESS)
+   end subroutine assign_fptr_i8_rank1
 
 end module MAPL_FieldPointerUtilities
--- a/field/FieldPointerUtilities.F90
+++ b/field/FieldPointerUtilities.F90
@@ -848,8 +848,11 @@ contains
 
       real(kind=ESMF_KIND_R4), pointer :: r4_1d(:),r4_2d(:,:),r4_3d(:,:,:),r4_4d(:,:,:,:)
       real(kind=ESMF_KIND_R8), pointer :: r8_1d(:),r8_2d(:,:),r8_3d(:,:,:),r8_4d(:,:,:,:)
+      integer(kind=ESMF_KIND_I4), pointer :: i4_1d(:),i4_2d(:,:),i4_3d(:,:,:),i4_4d(:,:,:,:)
+      integer(kind=ESMF_KIND_I8), pointer :: i8_1d(:),i8_2d(:,:),i8_3d(:,:,:),i8_4d(:,:,:,:)
 
       call ESMF_FieldGet(field,rank=rank,typekind=tk,_RC)
+      _ASSERT(rank > 0 .and. rank < 5, "Unsupported rank")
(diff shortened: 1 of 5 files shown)


generic3g/actions/AccumulatorAction.F90
```diff
@@ -188,4 +206,49 @@ subroutine accumulate_R4(this, update_field, rc)
...
+      class(AccumulatorAction), intent(inout) :: this
+      integer, optional, intent(out) :: rc
+      
+      integer :: status
+
+      _RETURN(_SUCCESS)
+      _UNUSED_DUMMY(this)
+      
+   end subroutine clear_post
+
```
tclune, line 232: Starting to be a bit of a code smell that there are so many no-op methods on the base class.   Not for now,  but I may eventually (ha) explore use of has-a vs is-a (current) implementation for the individual cases.

generic3g/actions/MeanAction.F90
```diff
@@ -70,50 +85,30 @@ subroutine calculate_mean(this, rc)
...
-   end subroutine invalidate_mean_accumulator
+   end subroutine mean_update_pre
 
    subroutine calculate_mean_R4(this, rc)
       class(MeanAction), intent(inout) :: this
       integer, optional, intent(out) :: rc
 
       integer :: status
       real(kind=ESMF_KIND_R4), pointer :: current_ptr(:) => null()
+      real(kind=ESMF_KIND_R4), pointer :: counter(:) => null()
```
tclune, line 105: I thought we had agreed that `counter` would be I4 rather than R4?    Or is that to be deferred until later?

tclune (commented):
See inline comments.  (I know this is draft.)

generic3g/actions/MeanAction.F90
darianboggs: I am working on the I4 right now. Parts of mapl3g_FieldPointerUtilities only support R4 and R8. I am extending them to support I4 (and I8 for good measure.)

tclune: Thanks.   My memory is just poor so I needed that confirmation.

generic3g/actions/AccumulatorAction.F90
tclune: Yes, that's pretty much what I'm hinting at.  But stay the course for now.  Functionality before perfection.

field/FieldPointerUtilities.F90
```diff
@@ -877,8 +880,34 @@ subroutine MAPL_FieldGetLocalElementCount(field,local_count,rc)
...
+         else if (rank ==2) then
+            call ESMF_FieldGet(field,0,farrayptr=i4_2d,_RC)
+            local_count = shape(i4_2d)
+         else if (rank ==3) then
+            call ESMF_FieldGet(field,0,farrayptr=i4_3d,_RC)
+            local_count = shape(i4_3d)
+         else if (rank ==4) then
+            call ESMF_FieldGet(field,0,farrayptr=i4_4d,_RC)
+            local_count = shape(i4_4d)
+         end if
```
tclune, line 897: Please replace with SELECT CASE construct.   If nothing else this will make it more obvious what happens if rank is 0 or larger than 4 because it will have a DEFAULT block.

Same for block of IFs below.

generic3g/actions/AccumulatorAction.F90
```diff
@@ -62,23 +63,45 @@ subroutine initialize(this, importState, exportState, clock, rc)
...
+      same_typekind = .FALSE.
       call get_field(importState, import_field, _RC)
+      call ESMF_FieldGet(import_field, typekind=typekind, _RC)
+      _ASSERT(typekind==ESMF_TYPEKIND_R4, 'Only ESMF_TYPEKIND_R4 is supported.')
       call get_field(exportState, export_field, _RC)
+      conformable = FieldsAreConformable(import_field, export_field, _RC)
+      _ASSERT(conformable, 'Import and export fields are not conformable.')
+      same_typekind = FieldsAreSameTypeKind(import_field, export_field, _RC)
+      _ASSERT(same_typekind, 'Import and export fields are different typekinds.')
+      this%typekind = typekind
```
tclune, line 80: I think some (well-chosen) empty lines would improve readability in this procedure.   I would not bother but since there are changes to use SELECT CASE above, might as well do it.

generic3g/actions/AccumulatorAction.F90
```diff
@@ -92,19 +115,30 @@ subroutine update(this, importState, exportState, clock, rc)
...
    end subroutine update
 
+   subroutine update_result(this, rc)
+      class(AccumulatorAction), intent(inout) :: this
+      integer, optional, intent(out) :: rc
+      
+      integer :: status
+
+      call FieldCopy(this%accumulation_field, this%result_field, _RC)
+      this%update_calculated = .TRUE.
```
tclune, line 137:
```suggestion
      this%update_calculated = .true.
```
stylistic preference

generic3g/actions/MeanAction.F90
```diff
@@ -70,78 +83,63 @@ subroutine calculate_mean(this, rc)
...
+   end subroutine update_result_mean
 
    subroutine calculate_mean_R4(this, rc)
       class(MeanAction), intent(inout) :: this
       integer, optional, intent(out) :: rc
 
       integer :: status
-      real(kind=ESMF_KIND_R4), pointer :: current_ptr(:) => null()
+      real(kind=ESMF_KIND_R4), pointer :: current_ptr(:)
+      integer(kind=COUNTER_KIND), pointer :: counter(:)
```
tclune, line 104: I like `COUNTER_KIND`.   In hindsight it is unfortunate that ESMF did not do `ESMF_R4_KIND` instead of `...KIND_R4`.

tclune (changes requested):
See comments inline.

generic3g/actions/MeanAction.F90
darianboggs: Agreed.

darianboggs:
Changes to `field/FieldPointerUtilities.F90`:
* Replace if block with select block.

Changes to `generic3g/actions/AccumulatorAction.F90`:
* comments
* blank lines
* reorder (slight)

field/FieldPointerUtilities.F90
```diff
@@ -846,40 +848,78 @@ subroutine MAPL_FieldGetLocalElementCount(field,local_count,rc)
...
+            local_count = shape(i8_2d)
+         case(3)
+            call ESMF_FieldGet(field,0,farrayptr=i8_3d,_RC)
+            local_count = shape(i8_3d)
+         case(4)
+            call ESMF_FieldGet(field,0,farrayptr=i8_4d,_RC)
+            local_count = shape(i8_4d)
+         case default
+            _FAIL("Unsupported rank")
+         end select
```
tclune, line 922: You still need a final `_RETURN(_SUCCESS)` at the end of the procedure.  (Can't make a comment down there or I'd fix it for you.)  The reason is just in case anyone ever adds some logic that might skip the above.  If procedure was shorter I would not worry.

tclune (changes requested):
One very small change.

field/FieldPointerUtilities.F90
darianboggs: OK. I'll add a final _RETURN(_SUCCESS).

tclune (approved):

darianboggs:
Merging despite the adas failures

PR #3195, opened 2024-11-27
Resolve gfortran segfault issue with vertical regridding tests

pchakraborty:
## Types of change(s)

## Checklist

## Description

This resolves gfortran segfault issue with running Scenarios tests vertical_regridding_2/3. The issue was caused by the type bound write overload - deactivated that in case of gfortran builds.


generic3g/specs/BracketSpec.F90
```diff
@@ -47,7 +47,9 @@ module mapl3g_BracketSpec
 
       procedure :: make_adapters
       procedure :: set_geometry
+! #ifndef __GFORTRAN__
```
tclune, line 50:
```suggestion
```
I think these should all be eliminated at this point.   Just keep the one for generic ...

tclune (changes requested):
ELiminate the unneeded `#ifndef ...`

generic3g/specs/BracketSpec.F90
pchakraborty: Done.

tclune (approved):

PR #3203, opened 2024-12-02
fixes #3201

### LINKED ISSUE #3201: nf90_get_att returning character variables with c null terminator
While trying to use a NetCDF4 file provided by a user I was hitting an error that we ultimately traced to the fact that `nf90_get_att` was returning a character string with a C null terminator at the end. We are not sure if it is this file but we do need to protect against this. I will make a PR.

bena-nasa:
Checks if the last character of a string retrieved from `nf90_get_att` is the `C_NULL_CHAR` and removes if it is when adding attributes to a `FileMetadata` object in `NetCDF4_FileFormatter.F90`

## Types of change(s)

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,7 +161,8 @@ workflows:
           fixture_branch: feature/mathomp4/mapldevelop
           checkout_mapl_branch: true
           mepodevelop: false
-          rebuild_procs: 1
+          rebuild_procs: 4
+          build_type: Release
 
   build-and-publish-docker:
     when:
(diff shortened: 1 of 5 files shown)


weiyuan-jiang (commented):
If this works, we should replace all pfio_get_att_str with nf90_get_att and remove pfio_get_att_string

mathomp4:
I made a push to fix the ADAS CI issue.

weiyuan-jiang (approved):

PR #3211, opened 2024-12-03
Feature/bmauer/fixes #3210

### LINKED ISSUE #3210: Refactor Variable class to that you can retrieve specify attribute types
I doing some work and it was helpful to have specific procedures in the `Variable` class in PFIO to retrieve attributes of specific types. Note there was a helper procedure in `FileMetaDataUtilities.F90` that given a `FileMetaDataUtilities` class, a variable name, and an attribute name you could retrieve an attribute of a specific type but no way, given a `Variable` to retrieve. 

It makes sense to do a simple refactor and move some of this into the `Variable` class so that this class can have the functionality to get attributes of specific types and the existing interfaces  to `FileMetadataUtilities` can remain unchanged.

I'm just making an issue for the PR I'm going to do as I've already done the refactor.

Will make more sense when one sees the PR.

bena-nasa:
Moved some code so now one can retrieve a string/real/int attribute from a Variable instance via helper functions in the Variable class. Used these to simplify the helper functions in FIleMetaDataUtilities.F90

## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow update offsets of &#177;timestep in ExtData2G
 - Minor revision (and generalization) of grid-def for GSI purposes
 - Trajectory sampler: fix a bug when group_name does not exist in netCDF file and a bug that omitted the first time point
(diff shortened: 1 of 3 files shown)


tclune (approved):
I had no idea that these were not already implemented.

I strongly suspect we should be able to avoid the duplication, but not worth the trouble just now.

PR #3212, opened 2024-12-03
Feature/fixes #3175

### LINKED ISSUE #3175: Grid detection heuristics in MAPL_LatLonGridFactory need improvement, allow periodic grid when does not span full latitude
I will add more as I figure out more, but @mfmehari and @zhaobin74 reported they had files that were being regridded and results in "missing" points at the dateline. The files are evenly spaced in longitude and clearly are `GC` in our convention. The latitude is less clear, but when regridded to cube beyond some resolution it would indeed have the dateline full of missing values when I reproduced this.

I put a bunch of prints in. It decides it is a `GC` in longitude and `XY` in latitude. Now because of this the heuristic set the grid to non-periodic, so now if a destination point falls "outside" the grid which is that last half delta, it will not be mapped to any source cells.

I need to look at this, we clearly need allow for periodic grids, even if they do not span the full latitude.

bena-nasa:
Fixes #3175 
Address 2 issues.
1. The pole should have no effect on the grid periodicity in the longitude direction
2. If the file has CF compliant bounds for the coordinate we should use those to define the edges rather than the heuristics and so forth.

I have confirmed that this is zero-diff at least with the files we are using for ExtData now. This will allow Bin and Mike to use these files with bounds.

## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow update offsets of &#177;timestep in ExtData2G
 - Minor revision (and generalization) of grid-def for GSI purposes
 - Trajectory sampler: fix a bug when group_name does not exist in netCDF file and a bug that omitted the first time point
+- Allow lat-lon grid factory to detect and use CF compliant lat-lon bounds in a file when making a grid
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


tclune (approved):
Made a small change to simplify conditional logic for presence of bnds.

mfmehari:
@bena-nasa @tclune Great work! Thank you very much for fixing the bug!

PR #3221, opened 2024-12-05
Feature/wdboggs/component timestep

darianboggs:
## Types of change(s)

## Checklist

## Description
MAPL3 will support component-level timestep (`run_dt`). This PR adds an `ESMF_TimeInterval` (`run_dt`) to the `ComponentSpec` derived type, and it provides a parsing function for reading `run_dt` as an ISO8601 Duration string from an `ESMF_HConfig` and converting it into an `ESMF_TimeInterval`. If `run_dt` is not found in the `ESMF_HConfig`, the resulting `ESMF_TimeInterval` is not set. In this case, `run_dt` cannot be used. This is not treated as an error condition in the parsing, but it should be handled at a higher level.

### DIFF
--- a/generic3g/ComponentSpecParser.F90
+++ b/generic3g/ComponentSpecParser.F90
@@ -62,6 +62,7 @@ module mapl3g_ComponentSpecParser
    character(*), parameter :: KEY_UNGRIDDED_DIM_COORDINATES = 'coordinates'
    character(*), parameter :: KEY_VERTICAL_DIM_SPEC = 'vertical_dim_spec'
    character(*), parameter :: KEY_ACCUMULATION_TYPE = 'accumulation_type'
+   character(*), parameter :: KEY_RUN_DT = 'run_dt'
 
    !>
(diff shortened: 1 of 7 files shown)


generic3g/ComponentSpecParser.F90
```diff
@@ -62,6 +63,7 @@ module mapl3g_ComponentSpecParser
    character(*), parameter :: KEY_UNGRIDDED_DIM_COORDINATES = 'coordinates'
    character(*), parameter :: KEY_VERTICAL_DIM_SPEC = 'vertical_dim_spec'
    character(*), parameter :: KEY_ACCUMULATION_TYPE = 'accumulation_type'
+   character(*), parameter :: KEY_RUN_DT = 'run_dt'
```
tclune, line 66: This is fine for the PR, but we should review with Atanas for long-term "how do we want to spell it".

darianboggs: Got it

PR #3240, opened 2024-12-10
Feature/wdboggs/add run dt to field specs

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR adds `run_dt` to `FieldSpec`. `run_dt` for a `FieldSpec` is determined by `run_dt` for the component the `Field` is associated with. It is needed on the `FieldSpec` for setting `ExtensionAction`. `run_dt` is not an attribute for `VariableSpec`, so when a `FieldSpec` is constructed from a `VariableSpec`, `run_dt` must be set after construction.


darianboggs:
I added `run_dt` to `make_itemSpec` to set `run_dt` for `FieldSpec`. Because `make_itemSpec` produces a `StateItemSpec`, which does not have a `run_dt` variable, I used a `select type(itemSpec) ... end select` block to cast `itemSpec` to `FieldSpec`. It's rather ugly, but it's better than adding `run_dt` to `StateItemSpec`. There could be another place to set `run_dt` for `FieldSpec`, but barring that, we are stuck with the `select type` :pig:

tclune:
My apologies - I had forgotten that recent changes had relocated where we set the geom:

https://github.com/GEOS-ESM/MAPL/blob/39d4278dbebe1624fe35b3b31eb0872f33d3647e/generic3g/registry/StateRegistry.F90#L630-L654

It should never be part of VariableSpec.

tclune:
DT should be passed alongside geom and vertical grid.

darianboggs:
Adding `run_dt` to `set_blanket_geometry` required adding it to `set_geometry` for `StateItemSpec` and the types derived from it. We may rename both procedures to be more inclusive for geometry and run_dt.

tclune (approved):

PR #3241, opened 2024-12-11
Test Ninja Builds of GCM in CI

mathomp4:
I want to see if the GCM builds any faster with Ninja in CI.


mathomp4:
Welp, looks like no big advantage with Ninja. I'll close.

PR #3249, opened 2024-12-16
Fixes #3247 - Initial implementation of StateItemAspect.

### LINKED ISSUE #3247: Introduce StateItemAspect
Introduce a base class that represents a single adaptable property of state items.  E.g.:
   - geom
   - units
   - precision
   - ...

Ideally this abstraction will simplify introduction of (fine-grained) time dependent specs into the framework as well as improve consistency for how various properties support mirroring.

tclune:
- Some simple tests have been implemented for generic Aspect logic.
- Some subclasses have been implemented - but not tested.
- High level logic has been added to exercise aspects, but it is not currently activated as the default aspect list is empty.
- Next step is to convert adapters to aspects one at a time and see if things still work.  Undoubtedly this will expose problems with new logic.
- Everything compiles and tests pass, such as they are.

## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/specs/StateItemSpec.F90
+++ b/generic3g/specs/StateItemSpec.F90
@@ -5,6 +5,8 @@ module mapl3g_StateItemSpec
    use mapl_ErrorHandling
    use mapl3g_ActualPtVector
    use mapl3g_ExtensionAction
+   use mapl3g_StateItemAspect
+   use mapl3g_AspectMap
    use gftl2_stringvector
    implicit none
    private
@@ -39,6 +41,8 @@ module mapl3g_StateItemSpec
       type(StringVector) :: raw_dependencies
       type(ActualPtVector) :: dependencies
 
+      type(AspectMap) :: aspects
+
    contains
 
       procedure(I_create), deferred :: create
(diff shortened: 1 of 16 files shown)


generic3g/specs/StateItemSpec.F90
```diff
@@ -49,6 +53,10 @@ module mapl3g_StateItemSpec
       procedure(I_can_connect), deferred :: can_connect_to
       procedure(I_make_adapters), deferred :: make_adapters
 
+      procedure :: get_aspect_order ! as string vector
+!#      procedure(I_get_aspect_priorities), deferred :: get_aspect_priorities ! as colon-separated string
```
darianboggs, line 57: Could this be deleted?

generic3g/specs/StateItemSpec.F90
```diff
@@ -187,6 +198,13 @@ function I_make_adapters(this, goal_spec, rc) result(adapters)
          integer, optional, intent(out) :: rc
       end function I_make_adapters
 
+      function I_get_aspect_priorities(src_spec, dst_spec) result(aspect_order)
+         import StateItemSpec
+         character(:), allocatable :: order
+         class(StateItemSpec), intent(in) :: src_spec
+         class(StateItemSpec), intent(in) :: dst_spec
+      end function I_get_aspect_priorities
+
```
darianboggs, line 207: Is this still needed?

generic3g/specs/StateItemSpec.F90
tclune: The line after it is temporary for the bootstrapping process.   Need a default implementation for subclasses to compile.    Once bootstrapping is complete, that line will be uncommented and the next line will be deleted.

generic3g/specs/StateItemSpec.F90
tclune: Yes - as per comment above.    Note - that I would love a better name.    Basically I wan the base method to call the subclass method (to avoid tedium of creating a long vector manually in the subclasses).   So the two procedures do  the same thing except that they return the information in a different form (string vector vs colon-separated string).

Or make them subroutines and then we could overload the same name.   Suggestions welcome.

darianboggs (approved):
Approved with suggestions for deleting unused code

mathomp4 (approved):

PR #3263, opened 2024-12-20
Fixes #3250 - integration of aspects

### LINKED ISSUE #3250: Complete integration of StateItemAspect
Complete the integration of StateItemAspect into higher level code and demonstrate with at least one aspect in existing StateItems. (Probably GEOM.)

tclune:
## Types of change(s)

## Checklist


tclune:
@darianboggs Found a workaround for GFortran.    I tested to see if it had any impact on the earlier NAG issue, but ... I could not make NAG fail either way.  So mystery already lost in the midst of time.

generic3g/specs/StateItemSpec.F90
```diff
@@ -205,6 +208,15 @@ function I_get_aspect_priorities(src_spec, dst_spec) result(aspect_order)
          class(StateItemSpec), intent(in) :: dst_spec
       end function I_get_aspect_priorities
 
+!#      function I_make_extension(this, aspect_name, aspect, rc) result(new_spec)
```
atrayano, line 211: Why do we have a function that is commented out tottaly?

generic3g/specs/UnitsAspect.F90
```diff
@@ -14,7 +14,7 @@ module mapl3g_UnitsAspect
 
 
    type, extends(StateItemAspect) :: UnitsAspect
-      private
```
atrayano, line 17: Do we really want a public UnitsAspect?

atrayano (approved):

generic3g/specs/UnitsAspect.F90
tclune: For now yes.     I will likely add accessors when the dust settles, but inserting getters everywhere that units are currently referenced is counter productive.  Most will not be needed in the end.

generic3g/specs/StateItemSpec.F90
tclune: Ah - this is another one of those temporary expediencies.  Eventually we want an abstract interface here, but that would force me to first provide a stub function in all the concrete subclasses.    Instead I provide a stub function in the base class (further down).  Again the intent is to clean up later on when each subclass has been properly updated.  The trick will be to remember and go back to fix-up the base class.

I'll try to be better about adding TODO's but ... I've never actually found myself going back to find "TODO" in an implementation.  The cleanup tends to happen more organically than that I've found.

PR #3269, opened 2024-12-23
fixes #3262

### LINKED ISSUE #3262: Allow MAPL to overwrite checkpoint restarts
I am running into a problem that leads me to believe that the checkpoint restarts cannot be overwritten … I recall there was a similar issue w/ history sometime back, and you pointed me to a flag in history that allows history to (use its old behavior) to overwrite its nc4 output files, this is the toggle:
 
Allow_Overwrite: .true.
 
I am wondering if the checkpoint writing suffers from the same restriction, or rule, that then cannot be overwritten by default. If that is so, is there an equivalent statement that can be put in AGCM.rc that tells MAPL to allow for overwrite?

CONTEXT: when the ADAS ensemble crashes, all that's typically needed is for the job to be re-submitted and the ensemble picks up from where it left;  it re-does only what hasn't completely successful in the original run. In case a member GCM crashes past the time it had written checkpoint files ... a re-submit will try to overwrite the checkpoint files previously written; with the present MAPL restriction of not allow restarts to be overwritten it is impossible to use the checkpoint mechanisms of the ensemble ADAS ... it requires re-running the who cycle from scratch.

The version of MAPL used in the 5.41.x ADAS is: v2.47.1.2  it would be nice to have the option added on top of this; and carried forward to more recent versions of MAPL.

bena-nasa:
Fixes #3262 
Adds new option, `overwrite_checkpoint` to the `AGCM.rc` to allow checkpoints to be clobber/overwritten
## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added loggers when writing or reading weight files
+- Added new option to AGCM.rc `clobber_checkpoint` to allow checkpoint files to be overwritten. By default still will not clobber checkpoints
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


rtodling:
This will be nice to have. Thank you for working on it.

mathomp4 (approved):

PR #3279, opened 2025-01-04
Revise location_index for IODA file in Trajectory sampler 

metdyn:
## Types of change(s)

## Checklist

## Description
`location_index_ioda` in Trajectory sampler is creating problems (found by @mjkagnes123), when the code is using the previously designed index mapping scheme.  For the 29 IODA files (NPW platform), about 6 of them shows either missing points and/or extra points that does not belong to a single IODA file.  This is because, we did not expect to see, e.g., 
aircraft.20190801T210000Z.nc4  contains time points: [20 points at 15 Z,  mid,  20 points at 21 Z ] 
aircraft.20190802T030000Z.nc4  contains time points: [10 points at 21 Z,  mid, 20 points at 03 Z ]

Our time slice scheme follows the python convention and will create a file:
aircraft.jedi.20190802_0300z.nc4  contains time points:  [30 points at 21 Z, mid, zero points at 03 Z].

In this case, we would not be able to recover the IODA location_index in aircraft.20190802T030000Z.nc4.  In other words, we have extra points at lower bound and missing points at upper bound. This does not occur, if observation files does not contain exactly points at 21 Z or 03Z.   This explains why we had the right `location_index_ioda` for about 60% of the NPW platforms (IODA files), but wrong index for the rest. 

Now, `location_index_ioda` in Trajectory sampler is revised to allow for reading in a single observation input file, such as `sondes.20190801T210000Z.nc4`, then generate the sampled GEOVAL values along the trajectory (in time ordered sequence) and return an array index (location_index_ioda) mapping back to the location points in the input IODA file, with the exact number of points.  This is controlled by setting `jedi.use_NWP_1_file: .true.` in HISTORY.rc.   

Note, the default (`use_NWP_1_file=0`) is our standard method, using our designed PYTHON style for time slicing:  [ T1 ,  T1 + Epoch), with T1=21Z,  Epoch=6 hr, for example.    Here we will analyze an expanded time window [T1 - Epoch, T1 + 2*Epoch).

### DIFF
--- a/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
+++ b/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
@@ -57,7 +57,7 @@ module HistoryTrajectoryMod
      character(len=ESMF_MAXSTR)     :: var_name_lon_full
      character(len=ESMF_MAXSTR)     :: datetime_units
      character(len=ESMF_MAXSTR)     :: Location_index_name
-     integer                        :: use_NWP_1_file
+     logical                        :: use_NWP_1_file = .false.
      integer                        :: epoch        ! unit: second
      integer(kind=ESMF_KIND_I8)     :: epoch_index(2)
      real(kind=ESMF_KIND_R8), pointer:: obsTime(:)
--- a/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
+++ b/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
@@ -58,6 +58,7 @@ module HistoryTrajectoryMod
      character(len=ESMF_MAXSTR)     :: datetime_units
      character(len=ESMF_MAXSTR)     :: Location_index_name
      integer                        :: use_NWP_1_file
+     integer, dimension(2), parameter :: use_NWP_1_file_param = [0, 1]
      integer                        :: epoch        ! unit: second
      integer(kind=ESMF_KIND_I8)     :: epoch_index(2)
      real(kind=ESMF_KIND_R8), pointer:: obsTime(:)
--- a/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
+++ b/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
@@ -40,7 +40,6 @@ module HistoryTrajectoryMod
      type(VerticalData) :: vdata
      logical :: do_vertical_regrid
 
-
      type(LocstreamRegridder) :: regridder
      type(TimeData)           :: time_info
      type(ESMF_Clock)         :: clock
@@ -48,13 +47,7 @@ module HistoryTrajectoryMod
      type(ESMF_Time)          :: RingTime
      type(ESMF_TimeInterval), public  :: epoch_frequency
 
-
      integer                        :: nobs_type
-!     character(len=ESMF_MAXSTR)     :: nc_index
-!     character(len=ESMF_MAXSTR)     :: nc_time
-!     character(len=ESMF_MAXSTR)     :: nc_latitude
-!     character(len=ESMF_MAXSTR)     :: nc_longitude
-
      character(len=ESMF_MAXSTR)     :: index_name_x
      character(len=ESMF_MAXSTR)     :: var_name_time
      character(len=ESMF_MAXSTR)     :: var_name_lat
--- a/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
+++ b/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
@@ -58,7 +58,6 @@ module HistoryTrajectoryMod
      character(len=ESMF_MAXSTR)     :: datetime_units
      character(len=ESMF_MAXSTR)     :: Location_index_name
      integer                        :: use_NWP_1_file
-     integer, dimension(2), parameter :: use_NWP_1_file_param = [0, 1]
      integer                        :: epoch        ! unit: second
      integer(kind=ESMF_KIND_I8)     :: epoch_index(2)
      real(kind=ESMF_KIND_R8), pointer:: obsTime(:)
--- a/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
+++ b/gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
@@ -40,12 +40,14 @@ module HistoryTrajectoryMod
      type(VerticalData) :: vdata
      logical :: do_vertical_regrid
 
+
      type(LocstreamRegridder) :: regridder
      type(TimeData)           :: time_info
      type(ESMF_Clock)         :: clock
      type(ESMF_Alarm), public :: alarm
      type(ESMF_Time)          :: RingTime
      type(ESMF_TimeInterval), public  :: epoch_frequency
+
 
      integer                        :: nobs_type
 !     character(len=ESMF_MAXSTR)     :: nc_index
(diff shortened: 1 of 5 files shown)


gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
```diff
@@ -71,6 +72,15 @@
...
               label=trim(string) // 'var_name_time:', _RC)
+         call ESMF_ConfigGetAttribute(config, value=traj%use_NWP_1_file, default=0, &
+              label=trim(string)//'use_NWP_1_file:', _RC)
+         if (mapl_am_I_root()) then
+            if (traj%use_NWP_1_file == 1) then
+               write(6,105) 'WARNING: Traj sampler: use_NWP_1_file is ON'
+               write(6,105) 'WARNING: USER needs to check if observation file is fetched correctly'
+            end if
+            _ASSERT ( (traj%use_NWP_1_file==1).OR.(traj%use_NWP_1_file==0), 'use_NWP_1_file: wrong input')
+         end if
```
tclune, line 83: 1. The value of "1" for `use_NWP_1_file` is a "magic" number.  Please introduce named parameters for the allowed values of `use_NWP_1_file`.
2. The assertion can be made slightly clearer (and more extensible) as:
`_ASSERT( any(traj%use_NWP_1_file == [0, 1]), 'use_NWP_1_file: wrong input')`
(Again - the [0,1] should be replaced with parameter values.)

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
```diff
@@ -632,6 +657,14 @@
                      call lgr%debug('%a %a',   'exist: true filename       :', trim(filename))
                      call get_ncfile_dimension(filename, tdim=num_times, key_time=this%index_name_x, _RC)
                      len = len + num_times
+                     jj2 = jj2 + num_times
+                     if (j==this%obsfile_Ts_index) then
+                        this%obs(k)%count_location_until_matching_file = jj2
+                        write(6,'(2x,a,2x,i10)') 'this%obs(k)%count_location_until_matching_file', this%obs(k)%count_location_until_matching_file
```
tclune, line 663: Do you really want an unguarded output here?    I/O to the console can slow the code.   Is this just for debugging?

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
```diff
@@ -721,7 +763,15 @@
...
-            jx1 = real ( j1, kind=ESMF_KIND_R8)
+            if (this%use_NWP_1_file == 1) then            
+               ! IODA case:  Epoch + 1 second : for ioda location index recover to match observation files
+               ! this works because all IODA files has same end <= Epoch
+               ! and we use 1 file only
+               jx1 = real ( j1 + 1, kind=ESMF_KIND_R8)
+            else
+               ! normal case: strict design with cutoff time at Epoch
+               jx1 = real ( j1, kind=ESMF_KIND_R8)
+            end if
```
tclune, line 774: As above - invert the logic so that an else clause is not needed here.

I'm also starting to wonder if "1" is not magic but is actually meant as a counter of something.  If that is the case maybe the problem is in the name `use_NWP_1_file`?   Maybe that needs a "num_" prefix or something like that?

tclune (changes requested):
1. Lots of unguarded output to console.   If just for debugging and meant to be deleted later, just let me know.  Otherwise add conditionals or delete.
2. I cannot be sure if `use_NWP_1_file` is meant as a control or as a counter.  If the former then the values should be named constants rather than literals "0", "1".  If the latter, then a better name would make this clear.
3.

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
metdyn: Actually,  use_NPW_1_file is true (1) or false (0). 
j1 is 6 hr, and j1+1 is 6 hr 1 sec for bisect purpose.   I think the IODA input was generated by R2D2 python code, which did not follow a clean [ is, ie ) python rule to cut the observation time.  So we have to do a special treatment here at the end of 6 hr Epoch.

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
```diff
@@ -827,8 +871,53 @@
...
+                  if (j/=0) &
+                     write(6, *) trim(this%obs(k)%name)//'count_location_in_matching_file diff from cutted nobs_epoch'
+                  j = minval(this%obs(k)%location_index_ioda(:))
+                  if (j/=1) &
+                       write(6,*) trim(this%obs(k)%name)//' min =', j, &
+                       ' this%obs(k)%location_index_ioda(:) start diff from 1'
+                  j = maxval(this%obs(k)%location_index_ioda(:))
+                  if (j/=this%obs(k)%count_location_in_matching_file) &
+                       write(6,*) trim(this%obs(k)%name)//' max =', j, &
+                       ' this%obs(k)%location_index_ioda(:) end diff from obs file max pts'
```
metdyn, line 920: I realize that I should change this as debug for the NWP IODA file only, in order to prevent the general trajectory sampler case from printing out the information.

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
```diff
@@ -921,6 +1010,8 @@
          ! defer destroy fieldB at regen_grid step
          !
 
+         ! debug
+         ! write(6,*) 'ip, npt=',   ip, size(this%obsTime)
```
metdyn, line 1014: I will delete this

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
tclune: Why not make it a LOGICAL instead of INTEGER then?

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
tclune: ok

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
tclune: ok

gridcomps/History/Sampler/MAPL_TrajectoryMod.F90
```diff
@@ -58,6 +58,7 @@ module HistoryTrajectoryMod
      character(len=ESMF_MAXSTR)     :: datetime_units
      character(len=ESMF_MAXSTR)     :: Location_index_name
      integer                        :: use_NWP_1_file
+     integer, dimension(2), parameter :: use_NWP_1_file_param = [0, 1]
```
tclune, line 61:
```suggestion
     integer, dimension(2), parameter :: USE_NWP_1_FILE_PARAM = [0, 1]
```
MAPL style is to use all caps for constants.

gridcomps/History/Sampler/MAPL_TrajectoryMod_smod.F90
metdyn: You are right, thank you @tclune. I change this variable to logical and I found quite a few examples of using logical in the HISTORY.rc in MAPL, which should be and I should have searched for that. Now the code logic is as clear as it should be.

metdyn:
> 1. Lots of unguarded output to console.   If just for debugging and meant to be deleted later, just let me know.  Otherwise add conditionals or delete.
> 2. I cannot be sure if `use_NWP_1_file` is meant as a control or as a counter.  If the former then the values should be named constants rather than literals "0", "1".  If the latter, then a better name would make this clear.

You are right. I removed these write statements, except the warning when the `use_NWP_1_file: .true.` is activiated.

PR #3282, opened 2025-01-06
Feature/bmauer/geos simulator

bena-nasa:
I realized my "GEOS" simulator, a script I had to generate inputs for ExtDataDriver.x to simulate an ExtData configuration  that a real model run is using was not quite working right last week. I spent some time fixing it, and I don't want to loose it and it is useful enough I want to get it into version control.

## Types of change(s)

## Checklist

### DIFF
--- a/Tests/generate_extdatadriver_input.py
+++ b/Tests/generate_extdatadriver_input.py
@@ -5,6 +5,8 @@ import os
 import yaml
 import glob
 
+dims_dict = {"2":"xy", "3":"xyz"}
+
 def get_dims(component_map, name):
     for comp in component_map:
         if name in component_map[comp]:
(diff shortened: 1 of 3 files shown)


Tests/generate_extdatadriver_input.py
```diff
@@ -0,0 +1,171 @@
...
+
+   for item in vars_needed:
+      if item in extdata_def:
+         long_name = "NA"
+         units = "NA"
+         dims = get_dims(component_map, item) 
+         if dims == "2":
+            cdims = "xy"
+         elif dims == "3":
+            cdims = "xyz"
```
tclune, line 120: This might be marginally improved by having having a dictionary:  `cdims = cdims_dict[dims]`.  I would not have even bothered mentioning it, except that the logic is repeated in the export block so it is a 2x benefit.

Feel free to ignore though.

tclune (dismissed):
I don't pretend to have followed all the details, but looks quite clean and thus easy to understand/improve when needed.

I did make a small code-quality suggestion - at the discretion of the implementor.

Tests/generate_extdatadriver_input.py
bena-nasa: I made that change, easy enough

tclune (approved):

PR #3313, opened 2025-01-13
Feature/tclune/#3312 attributes aspect

tclune:
## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/specs/AttributesAspect.F90
+++ b/generic3g/specs/AttributesAspect.F90
@@ -1,0 +1,105 @@
+#include "MAPL_Generic.h"
+
+! We require that an export provides all attributes that an import
+! specifies as a shared attribute.  Some attributes of the export may
+! be unused and/or correspond to attributes needed by other imports.
+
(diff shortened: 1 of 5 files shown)


generic3g/specs/AttributesAspect.F90
```diff
@@ -0,0 +1,105 @@
...
+
+   end function matches
+
+   function make_action(src, dst, rc) result(action)
+      class(ExtensionAction), allocatable :: action
+      class(AttributesAspect), intent(in) :: src
+      class(StateItemAspect), intent(in)  :: dst
+      integer, optional, intent(out) :: rc
+
+      action = NullAction()
```
atrayano, line 100: Is this a placeholder? src and dst are not used and we return NullAction

tclune: This is an aspect for which we do not support any transforms.    But the interface is mandated by inheritance.   The procedure should never be called.   We just use NullAction() to prevent the compiler warning that we did not produce a value for the function.

Issue #3288, opened 2025-01-07
accumulator couplers need to trigger update on dependency during invalidate()


tclune:
Currently Accumulator actions accumulate the current value of the input during the invalidate() step.  This works when the coupler is directly connected to the user output.  But if there is an intervening coupler the "lazy" strategy for updating values interferes with the functioning of Accumulator.

Probably the fix is to add a method to GenericCoupler that controls whether to call update() on dependencies during invalidate().  Who knows, some other coupler might need the same.

darianboggs:
## Capturing Thoughts/Questions
- How to set `stale`?
- When to set `stale`?
- `invalidate` on a coupler may initiate `update` on **sources**
- `invalidate` on a coupler initiates `invalidate` on **consumers**
- `update` on a coupler initiates `update` on **sources**
- `update` on a coupler may initiate `invalidate` on on **consumers**
- Must avoid dependency loops
- Are dependency loops an issue for GridComps in general?
- Can we insist that a coupler chain be directed with regular GridComps at start and finish?

tclune:
Unless we also have update() generate invalidate() there is 0 chance of loops.    Couplers grow like a tree.  (Technically a DAG because of weirdness with vertical).

Worst case scenario - invalidate() always triggers update() and we waste computations.  (We very much want to avoid that though.)

darianboggs:
Closed by #3332

PR #3332, opened 2025-01-16
Run updates on invalidate

darianboggs:
## Types of change(s)

## Checklist

## Description
We need the `CouplerMetaComponent.invalidate` method it to force updates for its sources before running the `ExtensionAction.invalidate` method. This PR modifies `CouplerMetaComponent` to check if the associated `ExtensionAction` needs to run `invalidate`. If it needs to run `invalidate`, it forces updates for all the sources, and then runs the `ExtensionAction.invalidate` method. The PR adds the `ExtensionAction.runs_invalidate` method, which returns FALSE, and `AccumulatorAction.runs_invalidate` override.

## Related Issue
See #3288.


generic3g/tests/Test_ExtensionAction.pf
```diff
@@ -0,0 +1,20 @@
...
+contains
+
+   @Test(type=ESMF_TestMethod, npes=[1])
+   subroutine test_runs_invalidate(this)
+      class(ESMF_TestMethod), intent(inout) :: this
+      type(NullAction) :: action
+
+      @assert_that(action%runs_invalidate(), is(false()))
+
+   end subroutine test_runs_invalidate
```
tclune, line 18: Debatable that this is worthwhile as a test.   OTOH, certainly not worth deleting either.

tclune (approved):

generic3g/tests/Test_ExtensionAction.pf
darianboggs: It may be overkill. I was checking that the default ExtensionAction.runs_invalidate() returns .FALSE. which skips the update_sources() and ExtensionAction.invalidate(). I added a similar test of AccumulatorAction.runs_invalidate() to make sure it would run.

PR #3347, opened 2025-01-17
Fix for ExtData2G unit test on discover GNU

mathomp4:
## Types of change(s)

## Checklist

## Description

In testing MAPL on Discover with GNU, it was found the ExtData2G unit tests were failing. There were a few issues:

1. The overall test uses ESMF, but wasn't run with `mpirun`. On discover, all MPI code must be executed via `mpirun`
2. The test had an issue with a couple unit tests internally not handling an expected failure.
3. When initializing ESMF, we didn't actually specify a calendar.
4. An odd bug(?) with an internal write not being zero-padded.

These fixes were due to the kind help of @darianboggs and @tclune


mathomp4:
Sigh. I guess it fixed in one place and broke everywhere else. I'll take a look...

tclune (approved):

Issue #3274, opened 2025-01-02
use_threads set to True for openMp


weiyuan-jiang:
If use_threads set to ture, there would be some issues:
1) the call ESMF_GridCompGet(gc, grid=grid) crashes  immediately  after the gc is created [here](https://github.com/GEOS-ESM/MAPL/blob/1c55973c244834ed0d78c39964d2738c3d8f8a8b/generic/OpenMP_Support.F90#L546)
3) The im_world cannot be retrieved by the  call MAPL_GridGet(grid, globalCellCountPerDim=dims) where grid is created under OpenMP condition

@aoloso

PR #3350, opened 2025-01-17
Changes to return correct dimensions for the MPI grid from the OpenMP subgrids

aoloso:
## Types of change(s)

## Checklist

## Description
Changes were made to add attributes to the subgrids (i.e. created by dividing the MPI subdomain into smaller subdomains equal to the number of OpenMP threads) such that the correct dimensions for the MPI subdomain could be retrieved from the subgrids where ever needed.
## Related Issue
use_threads set to True for openMp Issue #3274


weiyuan-jiang:
I can confirm this fix works on Hercules with openmp

weiyuan-jiang (approved):

mathomp4:
@aoloso Can you add a changelog entry for this?

mathomp4:
Note: I did try this out with a simple AMIP run (so not exercising it probably) and it was zero-diff. So it doesn't seem to affect our way of running

tclune (dismissed):
How is the test coming along?   It will be more difficult to verify the test once the fix is merged.   You'll have to checkout a pre-fixed version but that version also won't have the test ...

mathomp4 (approved):

base/Base/Base_Base_implementation.F90
metdyn, line 1: Hi @bena-nasa,

There is a bug from your modification here
commit ef5e0245d69388c971174b34faab7a66db820958
Date:   Wed Jan 15 11:31:47 2025 -0500

My subroutine MAPL_MaskMod_smod.F90 calls MAPL_GetHorzIJIndex.. The MPI for gcc is paused here forever.  
This issue does not occur in the previous version of`base/Base/Base_Base_implementation.F90`

My call is in 
MAPL_MaskMod_smod.F90, 
 640 │      allocate ( II(nx), JJ(nx), _STAT )
 641 │      call MAPL_GetHorzIJIndex(nx,II,JJ,lonR8=obs_lons,latR8=obs_lats,grid=grid,_RC)

PR #3367, opened 2025-01-27
Fixes #3305 - support for Bracket exports

### LINKED ISSUE #3305: Implement BracketAspect
This aspect is to be used to indicate a spec is a "Bracket" and must use TimeInterpolateAction to transform into a Field.

tclune:
- This update involved a great many changes to fully integrate ClassAspect from the previous PR.  Many tests required updating and some existing logic required updates.  (Very little though.)
- VerticalGridAspect is also simplified.
- Obsolete code/tests have been removed.
- Remaining work is to establish a scenario that uses brackets.  For this an extension of the ComponentSpecParser is required.
- Because make_action() needs the src geom (and typekind), the previous implementation needed these stored in VerticalGridAspect. This was a dangerous duplication.  Now, the interface of make_action() is extended to have an argument that provides aux aspects such as geom and typekind.

## Types of change(s)

## Checklist


atrayano (approved):
Some diagnostic prints are still left in the code. When all tested, we should remove these.

mathomp4 (approved):

PR #3368, opened 2025-01-28
Feature/wdboggs/outer meta component datetime compatibility 3357

### LINKED ISSUE #3357: Check timestep and reference_time compatibility in OuterMetaComponent
We need to check that `timestep` and `reference_time` in `OuterMetaComponent` are compatible with the user component.

The use cases are:
1. _ordinary_ for uniform datetime durations. These durations include days and time durations, but it does not include years and months.
2. _monthly_ for datetime durations that include months since the number of days in a month varies, but it does not include years.
3. _yearly_ for datetime durations that include years and any other date or time durations since the number of days in a year varies.

darianboggs:
## Types of change(s)

## Checklist

## Description
As explained in #3357, the timestep and reference time for the user component, and therefore all the user alarms, must be compatible with the timestep and reference time for the OuterMetaComponent. Because the reference times may not be the same, it is insufficient to check the timesteps. This PR checks all user alarms, as determined by the user timestep and reference time, are compatible with the timestep and reference time of the OuterMetaComponent. It closes #3357.

## Related Issue
#3357

### DIFF
--- a/esmf_utils/ESMF_Time_Utilities.F90
+++ b/esmf_utils/ESMF_Time_Utilities.F90
@@ -9,7 +9,6 @@ module mapl3g_ESMF_Time_Utilities
    public :: zero_time_interval
    public :: intervals_are_compatible
    public :: times_and_intervals_are_compatible
-   public :: zero_time_interval
 
    interface zero_time_interval
       module procedure :: get_zero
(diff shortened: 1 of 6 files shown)


esmf_utils/ESMF_Time_Utilities.F90
```diff
@@ -0,0 +1,96 @@
...
+
+contains
+
+   logical function can_compare_intervals(larger, smaller, rc)
+      type(ESMF_TimeInterval), intent(inout) :: larger
+      type(ESMF_TimeInterval), intent(inout) :: smaller
+      integer, optional, intent(out) :: rc
+      integer :: status
+      integer(kind=I4) :: lyy, lmm, ld, lh, lm, ls
+      integer(kind=I4) :: syy, smm, sd, sh, sm, ss
```
tclune, line 35: Presumably one of these was meant to be I8?

I really do think you would benefit from compiling as you go ...

darianboggs: It compiles and passes the MAPL tests on bucy for ifort, nagfor, and gfortran. The tests on GitHub are still running, and sometimes, they don't pass even if they pass on bucy. It may be that on GitHub it won't compiler. Do we want to use I4 or I8 for getting and setting TimeInterval and Time? Our current code seems to use I4 for datetime units, but it is always safer to use I8.

tclune (approved):

PR #3370, opened 2025-01-28
fixes #3369

### LINKED ISSUE #3369: Changes for threading to GetHorzij have broken sampler code
@metdyn was testing the sampler code with the develop branch found the code was hanging in the GetHorzij routine which it uses.

I traced the problem to these protections
https://github.com/GEOS-ESM/MAPL/blob/v2.53.0/base/Base/Base_Base_implementation.F90#L2635
https://github.com/GEOS-ESM/MAPL/blob/v2.53.0/base/Base/Base_Base_implementation.F90#L2793

The issue is that in a procedure in the code we had to change to use the MAPL call to get the grid corners to allow for the threading:
https://github.com/GEOS-ESM/MAPL/blob/v2.53.0/base/Base/Base_Base_implementation.F90#L2918

where as before this was getting directly from the grid. 
The problem is in the non-threaded case, (i.e. not when the grid is a mini-grid) it calls the routine that returns the full corners which involves a halo, so if every doesn't call it we have a problem. And @metdyn had some processors that were not passing in points it seams.

So the simplest solution is just to remove this protection or move it after this call. I'll see  if I can move it. I've already confirmed that just removing lines 2635 and 2973 fixes @metdyn sampler issue but I'll just look at this a few more minutes before making a PR

bena-nasa:
Fixes issue #3369 
## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed but with MAPL_GetHorzijIndex from previous release when not points are passed on a processor causing a deadlock
+
 ### Removed
 
 ### Deprecated
--- a/base/Base/Base_Base_implementation.F90
+++ b/base/Base/Base_Base_implementation.F90
@@ -2632,7 +2632,6 @@ contains
     type(ESMF_CoordSys_Flag) :: coordSys
     character(len=ESMF_MAXSTR) :: grid_type
 
-    _RETURN_IF(npts == 0 )
     ! if the grid is present then we can just get the prestored edges and the dimensions of the grid
     ! this also means we are running on a distributed grid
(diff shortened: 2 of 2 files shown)


weiyuan-jiang:
I remember I used the non-collective call to get the coords. If it can't be use any more, then the call here may have problem too.
(https://github.com/GEOS-ESM/MAPL/blob/572a6442643efd6149e2378f0b127b9ccbd67b10/base/Base/Base_Base_implementation.F90#L2677-L2680)

bena-nasa:
> MAPL/base/Base/Base_Base_implementation.F90

Do we have a use case that even uses this non cubed-sphere branch? I agree that is we had non-cubed sphere mini-grids this would be a problem, but to my knowledge this is not a use case right now. I guess I could fix it to use the MAPL version of this call.

tclune (approved):

PR #3401, opened 2025-02-10
Test get horz ij index

aoloso:
## Types of change(s)

## Checklist

## Description
Added tests to ensure MAPL_GetHorzIJIndex works correctly
## Related Issue
3274


generic/OpenMP_Support.F90
```diff
@@ -177,7 +177,7 @@ function make_subgrids_from_bounds(primary_grid, bounds, unusable, rc) result(su
            call ESMF_AttributeSet(subgrids(i), name='GridCornerLats:', &
                 itemCount = count, valueList=lats1d, _RC)
            block
-              integer :: global_grid_info(10)
+              integer :: global_grid_info(11)
```
tclune, line 180: This is a magic number - a particularly surprising one at that.   Were it not for ESMF interface, a derived type would be much clearer here.

Maybe something like:
```
integer, parameter :: GRID_INFO_SIZE_GLOBAL = 3
integer, parameter :: GRID_INFO_SIZE_LOCAL= 3
...
integer, parameter :: GRID_INFO_SIZE = GRID_INFO_SIZE_GLOBAL + GRID_INFO_SIZE_LOCAL 
```

The extra bits are not entirely obvious to me. E.g. how are entries 4 & 5 different that the entries 7 & 8.  And why isn't there a "j" analog for entry 11?

tclune (commented):
See inline question.   I think the code can be made clearer, but depends on what is really going on there.

generic/OpenMP_Support.F90
tclune: I want this in, so accepting it.

tclune (approved):

Issue #3358, opened 2025-01-22
Check timestep and reference_time for compatibility in FrequencyAspect


darianboggs:
We need to check `timestep` and `reference_time` in `FrequencyAspect` to verify compatibility with other derived types and variables.

tclune:
(Mentioning @atanas so that he watches this issue.)

There are 2 sub-issues here (both are small so don't necessarily need to split into separate issues.)
1. Is the specified timestep and reference time for a user component compatible with those of the parent in terms of running.
2. Are the specified timestep/ref time compatible as FrequencyAspect's.

For each of these we must distinguish between the "ordinary" and the exceptional cases.   Ordinary cases are those where the timestep does not involve months/years.   In these cases we expect ESMF `MOD()` to be our friend.

My initial thoughts:

Case 1:
   - ordinary:  I think we want to require that each user clock tick coincides with some clock tick of the parent:
                    `MOD(UserRefTime - ParentRefTime, ParentTimeStep) == 0`
                    `MOD(UserTimestep,ParentTimestep) == 0`
Case 2:
   - ordinary:  I think we want to require that each import clock tick coincides with some clock tick of the export:
                    `MOD(ImportRefTime - ExportRefTime, ExportTimeStep) == 0`
                    `MOD(ImportTimestep,ExportTimestep) == 0`

More thought is needed for the extraordinary cases.

darianboggs:
Draft PR #3368 is pushed. It checks cases where the timesteps do not contain `yy` and `mm` units. Before merging this PR, that condition will be relaxed so that the user component timestep can contain any units, and the OuterMetaComponent timestep can contain `yy` and `mm` units if and only the user component timestep contains only `yy` and `mm` units, which avoids the nonuniform lengths of years and months.

darianboggs:
The `times_and_intervals_are_compatible` subroutine in `mapl3g_ESMF_Time_Utilities` allows comparison of (`ESMF_TimeInterval`, `ESMF_Time`) pairs, and assumes this relationship. So it can be used to compare the pairs of (`timestep`, `reference_time`) for the user and parent components.

darianboggs:
I pushed up a draft PR (#3404) that addresses subissue 2. (See above.) I believe PR #3368 addresses subissue 1.

PR #3404, opened 2025-02-10
Feature/wdboggs/mapl3 freq aspect ts reftime 3358

darianboggs:
## Types of change(s)

## Checklist

## Description
FrequencyAspect has a timestep and reference time now. Conversion between FrequencyAspect objects requires a check that considers `timestep` and `reference_time`. This PR implements a fuller test of compatibility in the `supports_conversion_specifc` function for FrequencyAspect.

## Related Issue
#3358 (part 2)


generic3g/specs/FrequencyAspect.F90
```diff
@@ -28,22 +30,18 @@ module mapl3g_FrequencyAspect
       procedure :: set_timestep
       procedure :: get_accumulation_type
       procedure :: set_accumulation_type
+      procedure :: get_reference_time
+      procedure :: set_reference_time
```
tclune, line 34: Does this class need any of these accessors?  (Including the ones that precede this PR.)

The tests don't use the accessors.  And external code certainly does not.

tclune (commented):
See one inline question.   Otherwise is fine.

Note that there will be some conflicts with my big PR that I started today.    (#3403) Ideally yours will go in first.    One thing that changed is that I used `refTime` in most places in deference to ESMF argument spelling.  (Also tried to use esmf capitalization for `timeStep`, but that is not crucial to be consistent.)

tclune:
Alas, my PR got in first (#3403).  Ping me if you need a second set of eyes to merge yours.

generic3g/specs/FrequencyAspect.F90
darianboggs: I added the accessors because the reference time is not in release/MAPLv3. I added it for this PR. I didn't add it to the constructor yet. Is reference time known at instantiation of the FrequencyAspect? If so and we don't need to change it after instantiation, then I can add it to the constructor and remove the setter. The getter is needed because the field for reference time is private. I don't know the use cases, so I don't know what is needed and/or appropriate.

I didn't want to break existing code by adding reference time to the constructor, and I needed to set and get reference time.

darianboggs:
> See one inline question. Otherwise is fine.
> 
> Note that there will be some conflicts with my big PR that I started today. (#3403) Ideally yours will go in first. One thing that changed is that I used `refTime` in most places in deference to ESMF argument spelling. (Also tried to use esmf capitalization for `timeStep`, but that is not crucial to be consistent.)

Do we want the reference time procedure argument to be called `refTime`? The ESMF variable names make sense to me once I know what the variables are, but when I look for a variable without knowing the variable name, the names are not obvious. I am learning the conventions, like `timeStep` and `refTime`, but I still use the ESMF documentation.

As a _general_ policy, are we using the ESMF conventions where they overlap with MAPL?

tclune:
I don't know these ESMF procedures any better.  And I've definitely been inconsistent where "my" naming style conflicts with ESMF argument names.    But when the variable really is just for the purpose of interacting with ESMF as these are it seems like the compromise should be towards ESMF.  (E.g., I left the yaml key that you defined alone as it was not identical to the ESMF argument no matter what.)

I'm not worried about 100% consistency - but if you are going back in to fix other things, then yes - as much as possible follow the changes that I made.

generic3g/specs/FrequencyAspect.F90
tclune: Yes - refTime and timeStep are both known prior to constructing the varspec.   If we use the yaml mechanism all is good, but this likely needs work if we use the ACG, because we don't want the users to manually include these in their tables.  Burn that bridge when we cross it. (Favorite mixed metaphor.)

Still not sure I see why you need any getter's.   The FrequencyAspect makes the action and can pass refTime and timeStep to the action constructor without an accessor.

tclune:
> As a _general_ policy, are we using the ESMF conventions where they overlap with MAPL?

Just to be more specify.  Yes - general policy is to use ESMF conventions when a variable is substantially only used as an argument to ESMF procedures.  This is the case for things like `importState` (where I would otherwise prefer `import_state`).    I just don't want to be overly strict here.  For now we won't go sweeping the code for consistency, but I had the energy when making related changes this weekend and propagated it as best I could.

Was hoping that this PR would go in before mine as then I would have the responsibility for completing the exercise. Instead this PR is paying the price.

generic3g/specs/FrequencyAspect.F90
```diff
@@ -55,9 +55,9 @@ function new_FrequencyAspect(timeStep, refTime, accumulation_type) result(aspect
       call aspect%set_time_dependent(.FALSE.)
       call aspect%set_accumulation_type(INSTANTANEOUS)
       call aspect%zero_timestep()
-      if(present(timestep)) aspect%timestep_ = timestep
-      if(present(refTime)) aspect%refTime_ = refTime
-      if(present(accumulation_type)) aspect%accumulation_type_ = accumulation_type
+      if(present(timestep)) call aspect%set_timestep(timestep)
+      if(present(refTime)) call aspect%set_reference_time(refTime)
+      if(present(accumulation_type)) call aspect%set_accumulation_type(accumulation_type)
```
tclune, line 60: I do not understand the verbosity here.  The other setters (e.g., `set_mirror()`) are necessary because the corresponding data component is PRIVATE in the base type. 

```suggestion
      if(present(timestep)) aspect%timeStep = timeStep
      if(present(refTime)) aspect%refTime = refTime
      if(present(accumulation_type)) aspect%accumulation_type = accumulation_type
```

darianboggs:
> > As a _general_ policy, are we using the ESMF conventions where they overlap with MAPL?
> 
> Just to be more specify. Yes - general policy is to use ESMF conventions when a variable is substantially only used as an argument to ESMF procedures. This is the case for things like `importState` (where I would otherwise prefer `import_state`). I just don't want to be overly strict here. For now we won't go sweeping the code for consistency, but I had the energy when making related changes this weekend and propagated it as best I could.
> 
> Was hoping that this PR would go in before mine as then I would have the responsibility for completing the exercise. Instead this PR is paying the price.

Got it. I might as well change them while I have them.

generic3g/specs/FrequencyAspect.F90
darianboggs: I made the change for time step and reference time. The accumulation type cannot be set simply. The setter checks the actual accumulation type to make sure it is a valid accumulation type, and it 

The setter updates the instance accumulation type only if the actual accumulation type argument if it is valid, so that the constructor does not return an instance with an invalid accumulation type.

generic3g/specs/FrequencyAspect.F90
```diff
@@ -170,50 +183,40 @@ end function supports_conversion_general
...
       class(FrequencyAspect), intent(in) :: src
       class(StateItemAspect), intent(in) :: dst
+      integer :: status
+      logical :: compatible
 
       select type(dst)
       class is (FrequencyAspect)
-         supports = src .divides. dst
+         call check_compatibility(dst, src, compatible, rc=status)
+         supports = compatible .and. status == _SUCCESS
```
tclune, line 192: Yeah - it was unfortunate that the first cases I looked at for this could not return a failure.   It meant that the interface for `...specific` did not need a return code.  

Probably the right thing to do is to modify the interface throughout the subclasses to have a return code.  But **NOT** as part of this PR.    Should be a separate PR.

tclune: Once you do that though, it it might obviate the need to have `check_compatibility` as that is just a thin wrapper on a procedure that you could just call directly here.

generic3g/specs/FrequencyAspect.F90
tclune: Got it.

So you'll also remove the 2 setter procedures, right?

tclune (approved):
I still think that the unnecessary accessors should be removed, and don't think that here should be trailing underscores on the component names.  But I don't want to slow things down.

PR #3407, opened 2025-02-11
Fixes #3405

### LINKED ISSUE #3405: Add switch to force allocation of all exports
This functionality is desired for testing gridcomps in isolation.  The switch will avoid the need to artificially connect all exports to imports in some other gridcomp.

Longer term we may want to allow VariableSpec to specify that any given export should be allocated regardless of connection status.  E.g., the export might be used for computing other exports.    OTOH, this could also be handled via "export dependencies" which already have some implementation foothold in MAPL3.

tclune:
Needs more testing of course.

## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/ComponentSpecParser/parse_component_spec.F90
+++ b/generic3g/ComponentSpecParser/parse_component_spec.F90
@@ -27,10 +27,34 @@ contains
       spec%connections = parse_connections(mapl_cfg, _RC)
       spec%children = parse_children(mapl_cfg, _RC)
 
+      call parse_misc(spec, mapl_cfg, _RC)
+
       call ESMF_HConfigDestroy(mapl_cfg, _RC)
 
       _RETURN(_SUCCESS)
    end function parse_component_spec
+
+   ! TODO - we may want a `misc` section in the mapl section, but
+   ! should wait to see what else goes there.  Or maybe a `test`
+   ! section?
+   
+   subroutine parse_misc(spec, hconfig, rc)
+      type(ComponentSpec), intent(inout) :: spec
+      type(ESMF_HConfig), intent(in) :: hconfig
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+      logical :: has_activate_all_exports
+
+      has_activate_all_exports = ESMF_HConfigIsDefined(hconfig,keyString=COMPONENT_ACTIVATE_ALL_EXPORTS, _RC)
+      if (has_activate_all_exports) then
+         spec%activate_all_exports = ESMF_HConfigASLogical(hconfig, keyString=COMPONENT_ACTIVATE_ALL_EXPORTS, _RC)
+      end if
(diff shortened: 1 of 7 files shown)


generic3g/ComponentSpecParser/parse_component_spec.F90
```diff
@@ -27,10 +27,34 @@ module function parse_component_spec(hconfig, registry, timeStep, refTime, rc) r
       spec%connections = parse_connections(mapl_cfg, _RC)
       spec%children = parse_children(mapl_cfg, _RC)
 
+      call parse_misc(spec, mapl_cfg, _RC)
```
pchakraborty, line 30: Not sure if it has already been discussed, but can't we add `activate_all` to the `states/export` section directly, and have it parsed in `parse_var_specs`? The config file then would look like
```yaml
mapl:
  states:
    export:
      activate_all: true
      E1:
        ...
      E2:
        ...
```
Or, going a step further, something like
```yaml
mapl:
  states:
    export:
      E1:
        activate: true
        ...
      E2:
        ...
```

generic3g/ComponentSpecParser/parse_component_spec.F90
```diff
@@ -27,10 +27,34 @@ module function parse_component_spec(hconfig, registry, timeStep, refTime, rc) r
...
       call ESMF_HConfigDestroy(mapl_cfg, _RC)
 
       _RETURN(_SUCCESS)
    end function parse_component_spec
 
+   ! TODO - we may want a `misc` section in the mapl section, but
+   ! should wait to see what else goes there.  Or maybe a `test`
+   ! section?
+   
+   subroutine parse_misc(spec, hconfig, rc)
```
pchakraborty, line 41: If we were to keep the `activate` switch localized, this would go in `parse_var_specs`.

pchakraborty (commented):
Please see inline comments.

generic3g/ComponentSpecParser/parse_component_spec.F90
tclune: No.  User GC's will not generally use that mechanism for their import/export states.  Useful for our testing of the generic layer, but users will use the ACG.     We could add something to the ACG, but that begs more questions.

@atrayano suggested possible use of ENV variables.  Creates other issues but at least lets the testing framework control this.

Will now spend some time thinking about a "mirror component" to avoid this approach.

generic3g/ComponentSpecParser/parse_component_spec.F90
tclune: How is that more localized than in the MAPL section itself?   The distinction seems at best minor, and at worst misleading as it needs to impact variables that are specified with the ACG / source code and not just those in the `states` section.

pchakraborty (approved):

PR #3434, opened 2025-02-14
Feature/bmauer/identify vertical coord extdata2g

bena-nasa:
This is the first PR to get a beta version of proper vertical constituent regridding into ExtData2G. It seemed like it was time for a pause to get this in now that it is getting really big and I have a few cases working.

Basically the idea as follows.
I have redone the detection of the vertical level in the file, and identify it according to CF as either hybrid sigma, fixed pressures/heights etc, or just "simple" coordinate if it doesn't follow either of those according to the letter of CF in new code.

If it is hybrid sigma, then that means there is a surface pressure and ak/bk in the file (note that I'm not entirely satisfied with the format for the file I coded to, I tried to follow CF, but might want to change; see plan at end of issue). I add a new rule, borrowing the original items parameters to read PS like any other item, so it gets regridded, time interpolated etc.
Then after the time interpolation I check if hybrid sigma, then regrid.
This so far has limited functionality and supports regridding from a file on hybrid sigma levels as defined by the CF conventions with a time varying pressure and ak/bk in the file. It only supports constituents in either mass mixing ratios: `kg kg-1` or mass per area per second. I.E. `kg m-2 s-1`. This uses the ppm code (brought a version into MAPL) from FV3 to conservatively regrid and also has corrections at the surface to ensure that mass is conserved.

I can go into greater detail what is being done here if needed.

In addition, in order for this feature to be usable ExtData must import PLE. This can be accomplished using the version Kyle added to the Cap by specify this in the CAP.rc file
```
CAP_EXPORTS:
PLE,DYN
::

EXTDATA_IMPORTS:
PLE,DYN
::
```
I have been testing this all along for 0 diff.
This is the first hopefully what will be much more frequent PR's now once this is in. The plan is as follows:

* Implement Volume mixing regridding `mol mol-1` (will require importing Q from model, and reading time varying Q from file), and reading species molecular weight from file.
* Decide if we are happy with the metadata requirements, it not change it at this step I don’t want to have to change it later. What input types we need to support:
    1. Hybrid sigma levels defined by CF
    2. Fixed  levels (either pressure or height) as defined by CF. Can also define edge boundaries via 7.1 of CF convention
* Add support for hybrid sigma levels with a file containing a time dependent PLE rather than PS/AK/BK
* Implement fixed pressure level regridding
* Add support for fixed heights (regrid via ZLE)

## Types of change(s)

## Checklist


Tests/ExtDataDriverGridComp.F90
```diff
@@ -403,6 +403,11 @@ subroutine initialize_gc(gc, import_state, export_state, clock, rc)
              _VERIFY(status)
           end if
        END DO
+       ! now add PS from Root if available
+       call ESMF_StateGet(cap%exports(cap%root_id), 'PLE', field, rc=status)
```
tclune, line 407: I know this comment applies more generally in MAPL, but doesn't this trigger an entry in the ESMF Log if `PLE` is not found?  I did a quick check of the API and don't see any other way to do this.  If you agree, please open a ticket with ESMF to create some query capability for this.

gridcomps/ExtData2G/ExtDataAbstractFileHandler.F90
```diff
@@ -19,6 +19,7 @@ module MAPL_ExtdataAbstractFileHandler
    private
    public :: ExtDataAbstractFileHandler
 
+   integer, parameter :: max_trials = 10
```
tclune, line 22:
```suggestion
   integer, parameter :: MAX_TRIALS = 10
```

gridcomps/ExtData2G/ExtDataAbstractFileHandler.F90
```diff
@@ -166,4 +168,39 @@ subroutine make_metadata(this,file,metadata,rc)
...
+        useable_time = this%valid_range(1)
+     end if
+     call fill_grads_template(trial_file, this%file_template, time=useable_time, _RC)
+     inquire(file=trim(trial_file),exist=file_found)
+
+     if (file_found) then
+        filename = trial_file
+        _RETURN(_SUCCESS)
+     end if
+     do i=1, max_trials
```
tclune, line 193:
```suggestion
     do i=1, MAX_TRIALS
```

vertical/Eta2Eta.F90
```diff
@@ -0,0 +1,1940 @@
+!#include "unused_dummy.H"
```
tclune, line 1: What is going on here?  This is a very large new program, but looks to be copied from somewhere else (inferred by style).

tclune (dismissed):
Please answer the one inline question I asked before merging.  And the very small suggestion edits.  (capitalize params)

And ... you are now not allowed to leave NASA until your knowledge is captured in MAPL3 on this front.

vertical/Eta2Eta.F90
bena-nasa: Yes, this this "SJ" code, in particular a version where Mike Manyin cleaned up the interface and did tests to make sure it conserved to the needs of the chemistry folks this would ultimately be fore. So yes, it is copied from somewhere else. 

The alternative would be to write our own conservative pressure regridding code. Doable and probably cleaner, but one more thing to do to get this all working.

Tests/ExtDataDriverGridComp.F90
bena-nasa: Yes, it would trigger an entry in the log. There is another way to do this. I can query the type of `PLE`. One option they have is NO_TYPE, if not found. I'll change it to that.

vertical/Eta2Eta.F90
tclune: Thanks - more-or-less what I expected, but wanted to be certain.   Part of a large tier of low-level "FV" code that needs to be shared across multiple repos/orgs.

Tests/ExtDataDriverGridComp.F90
tclune: Good.  And now that you've stated it, I suspect that is precisely how they expect us to handle this situation.

tclune (approved):

PR #3445, opened 2025-02-21
v3: Fixes for easy tests

### LINKED ISSUE #3443: Odd issue with tests target
As shown to me today by @darianboggs, if you do `make tests` without a `make install` some tests fail. I don't think this is necessarily required. I'll look at this.

mathomp4:
## Types of change(s)

## Checklist

## Description

As discovered by @darianboggs, if you checkout MAPL2 or MAPL3 and run CMake and then just run `make tests` without running `make install`, some of the tests crash. The reason is that CMake didn't know to build some of the executables needed for the tests.

So here we add the needed CMake to make that happen.

## Related Issue

Closes #3443


tclune (approved):
Thank you both for pursuing this.   I've gotten good at remembering that extra step to make the tests work, but I'm sure I would have forgotten again whenever I go a while without running the full suite.

PR #3449, opened 2025-02-24
Fixes #3488 - Control timestep of collection

tclune:
In MAPL3g, timestep of a component is specified by the parent. However the schema for History3g has the timestep being set by the collection as a "time_spec".  This adds a bit of extra logic to pull out timestep at the parent level.

We should consider whether the History schema should be modified to make this more natural.

## Types of change(s)

## Checklist

### DIFF
--- a/gridcomps/History3G/HistoryGridComp.F90
+++ b/gridcomps/History3G/HistoryGridComp.F90
@@ -3,6 +3,8 @@
 module mapl3g_HistoryGridComp
    use mapl3g_HistoryGridComp_private
    use mapl3g_HistoryCollectionGridComp, only: collection_setServices => setServices
+   use MAPL_TimeStringConversion
+   use mapl3g_ChildSpec
    use generic3g
    use mapl_ErrorHandling
    use pFlogger, only: logger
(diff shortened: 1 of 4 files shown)


gridcomps/History3G/HistoryGridComp.F90
```diff
@@ -53,13 +58,35 @@ subroutine setServices(gridcomp, rc)
          collection_name = ESMF_HConfigAsString(iter, _RC)
          child_hconfig = make_child_hconfig(hconfig, collection_name, _RC)
          child_name = make_child_name(collection_name, _RC)
-         call MAPL_GridCompAddChild(gridcomp, child_name, user_setservices(collection_setServices), child_hconfig, _RC)
-         !call ESMF_HConfigDestroy(child_hconfig, _RC)
+
+         call get_child_timestep(child_hconfig, timeStep, _RC)
+         child_spec = ChildSpec(user_setservices(collection_setServices), hconfig=child_hconfig, timeStep=timeStep, refTime=refTime)
```
bena-nasa, line 63: @tclune, `refTime` appears to be uninitialized.

bena-nasa:
@tclune 
You didn't remove the logic in the collection grid comp that is now redundant if the run if being controlled at this level.

gridcomps/History3G/HistoryGridComp.F90
tclune: Yes - by design.    We don't have a test that covers that case yet.    The variable is unallocated for now, which is equivalent to "use default from parent" which is 99% of what we want anyway.  So no harm.

bena-nasa (dismissed):

bena-nasa (approved):

Issue #1750, opened 2022-10-20
MAPL_GridCompSpecs_ACG.py adds superfluous brackets in varspec


darianboggs:
`MAPL_GridCompSpecs_ACG.py` adds brackets ("[ ]") to "array-like" varspec options (which includes only "ungridded_dims", currently.) If the option value already has brackets, it adds an additional pair, which is incorrect.

mathomp4:
@darianboggs Did you fix this? It's been around a while. Maybe this is something the work @sanAkel is doing in the ocean can test to see.

sanAkel:
> @darianboggs Did you fix this? It's been around a while. Maybe this is something the work @sanAkel is doing in the ocean can test to see.

Thanks @mathomp4, but I don't think I have any imports/exports with ungridded dims in the "ocean" repo. That will be the case for tiled variables that I won't touch now.

darianboggs:
This is fixed by PR #3462.

PR #3462, opened 2025-03-04
Feature/acg extra rm brackets 1750

darianboggs:
## Types of change(s)

## Checklist

## Description
When a column in a StateSpecs file is array-valued, and the user mistakenly puts array brackets (`[` and `]`) around the column value. The resulting include file is double brackets (`[[` and `]]`) which is wrong. This PR handles this more gracefully. With it, the column value can have array brackets, but they are not required. In the resulting include file, ACG writes a single set of array brackets. Because this problem has not been reported, I did not make it to ACG for MAPL2. It is worth correcting it in ACG3.

## Related Issue
#1750


tclune (dismissed):

darianboggs:
The only change was resolving the conflict in CHANGELOG.md.

tclune (approved):

PR #3474, opened 2025-03-11
Feature/wdboggs/add time accumulation to history3g

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR adds time accumulation to History3G.


darianboggs:
This PR does not include unit tests, and it does not change the new version of *register_imports*.

generic3g/specs/VariableSpec.F90
```diff
@@ -224,4 +224,92 @@ function get_regrid_method_from_field_dict_(standard_name, rc) result(regrid_met
       _RETURN(_SUCCESS)
    end function get_regrid_method_from_field_dict_
 
+   function make_AspectMap(standard_name, geom, units, typekind, vertical_dim_spec, &
```
tclune, line 227: I'm a bit confused by what you are trying to do here.   We need one factory method that just takes in a set of arguments that are aspects (with a few stragglers like intent, short name, etc).   Then we need another that matches the current interface with a much longer list of lower-level (user-level) items.   That procedure constructs aspects and calls the former.

darianboggs: So you want the current **make_VariableSpec** modified so that it calls a second method that constructs a **VariableSpec** from state_intent, short_name, standard_name, itemtype, service_items, default_value, bracket_size, dependencies, and aspects?

Should both methods be public, and if not, which one should be public?
Should the aspects argument be an **AspectMap** or an array of **StateItemAspect** objects?

I made it an **AspectMap** to make sure that the correct **AspectID** maps to the correct **StateItemAspect**. Currently **make_VariableSpec** creates the **AspectMap**,  but I don't think there is any control on which **AspectID** is mapped to which **StateItemAspect**. We may want a standard method to create an **AspectMap** from a **StateItemAspect** array to guarantee that the mapping is correct.

tclune: I think we need an in person discussion.  Either I'm misunderstanding you, or you are misunderstanding me.

generic3g/specs/VariableSpec.F90
```diff
@@ -224,4 +205,97 @@ function get_regrid_method_from_field_dict_(standard_name, rc) result(regrid_met
...
+         & result(var_spec)
+      type(VariableSpec) :: var_spec
+      type(ESMF_StateIntent_Flag), intent(in) :: state_intent
+      character(*), intent(in) :: short_name
+      class(KeywordEnforcer), optional, intent(in) :: unusable
+      character(*), optional, intent(in) :: standard_name
+      type(ESMF_StateItem_Flag), optional, intent(in) :: itemtype
+      type(StringVector), optional :: service_items
+      real, optional, intent(in) :: default_value
+      integer, optional, intent(in) :: bracket_size
```
tclune, line 221: you are missing `ClassAspect` which will eliminate several remaining arguments.  E.g., `bracket_size` is part of `BracketClassAspect`  Likewise, I think `service_items` is part of either Wildcard or Service.

Possibly some of this is complicated by the fact that `make_ItemSpec` is still doing some work downstream.   So maybe there is related refactoring to clean this up.   Or it may not make sense.  Give it some thought and then discuss with me.

darianboggs: Yes. I was wondering about the other Aspects. They weren't in **makeVariableSpec**, so I ignored them.

tclune (approved):

PR #3486, opened 2025-03-17
fixes #3485

### LINKED ISSUE #3485: NAG found issue when processing History collection list
Just making an issue for the PR. @mathomp4 reported that on the release/MAPL-v3 branch all the ExtData tests were failing with NAG. I traced down the line it was crashing. It was this line which is common to main/develop/MAPL3

https://github.com/GEOS-ESM/MAPL/blob/v2.54.1/gridcomps/History/MAPL_HistoryGridComp.F90#L580

Notice that if `nlist=1` you are assigning to `list(1:0)`. There is nothing to copy from the previous n-1 list. I found the failing tests were fixed by putting a protection to not call this if `nlist=1`. No idea why only NAG on the MAPL3 branch didn't like this, but it does seem to be a legitimate place we should protect.

bena-nasa:
Fixes #3485
This fixed the issue with the failing extdata tests on release/MAPL-v3 with NAG. Since it seemed like a bug @mathomp4 thought we should do this as a proper hot fix to get this propagated to the MAPL3 branch eventually

## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed bug when proccessing history collection list identified by NAG
-
 ### Removed
 
 ### Deprecated
+
+## [2.54.2] - 2025-03-18
+
+### Fixed
+
+- Fixed bug when proccessing history collection list identified by NAG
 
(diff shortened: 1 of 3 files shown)


tclune (approved):
I half expect that there is more to this story as to why it is crashing on one branch vs the other.   I suppose it could just be some out-of-bounds issue that only causes problems under some circumstances.

PR #3488, opened 2025-03-17
Replacing MPI_THREAD_MULTIPLE with MPI_THREAD_SINGLE everywhere

### LINKED ISSUE #3487: MPI_THREAD_MULTIPLE not supported by some MPI vendor implementation of P2P communication
This problem was reported by [NickSzapiro-NOAA](https://github.com/NickSzapiro-NOAA)

Please see [Enable GEFS ORTs on all platforms](https://github.com/ufs-community/ufs-weather-model/issues/2637#top) for details.

aoloso:
## Types of change(s)
- [X] Bug fix (non-breaking change which fixes an issue). Technically not a bug, rather a deficiency in the particular MPI implementation

## Checklist

## Description
Replace MPI_THREAD_MULTIPLE with MPI_THREAD_SINGLE. This is done to address https://github.com/ufs-community/ufs-weather-model/issues/2637#issuecomment-2722008171. This replacement appears not to cause any problem with the the different thread scenarios and the use of PFIO

## Related Issue
Closes #3487

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Replace MPI_THREAD_MULTIPLE with MPI_THREAD_SINGLE. This is done to address https://github.com/ufs-community/ufs-weather-model/issues/2637#issuecomment-2722008171. This replacement appears not to cause any problem with the the different thread scenarios and the use of PFIO
+
 ### Fixed
 
 - Fixed problem related to stale pointers to temp copies of dummy arguments in MAPL_Cap.F90.  Fix is to add TARGET attributein select locations.
--- a/gridcomps/Cap/MAPL_Cap.F90
+++ b/gridcomps/Cap/MAPL_Cap.F90
@@ -480,9 +480,9 @@ contains
       if (.not. this%mpi_already_initialized) then
 
          call ESMF_InitializePreMPI(_RC)
-         call MPI_Init_thread(MPI_THREAD_MULTIPLE, provided, ierror)
+         call MPI_Init_thread(MPI_THREAD_SINGLE, provided, ierror)
          _VERIFY(ierror)
-         _ASSERT(provided == MPI_THREAD_MULTIPLE, 'MPI_THREAD_MULTIPLE not supported by this MPI.')
+         _ASSERT(provided == MPI_THREAD_SINGLE, 'MPI_THREAD_SINGLE not supported by this MPI.')
       else
          ! If we are here, then MPI has already been initialized by the user
          ! and we are just using it. But we need to make sure that the user
@@ -490,7 +490,7 @@ contains
          call MPI_Query_thread(provided, ierror)
          _VERIFY(ierror)
       end if
-      _ASSERT(provided == MPI_THREAD_MULTIPLE, 'MPI_THREAD_MULTIPLE not supported by this MPI.')
+      _ASSERT(provided == MPI_THREAD_SINGLE, 'MPI_THREAD_SINGLE not supported by this MPI.')
 
       call MPI_Comm_rank(this%comm_world, this%rank, status)
       _VERIFY(status)
--- a/pfio/pfio_collective_demo.F90
+++ b/pfio/pfio_collective_demo.F90
@@ -322,7 +322,7 @@ program main
 !$   integer :: num_threads
    type (FakeExtData), target :: extData
 
-   required = MPI_THREAD_MULTIPLE
+   required = MPI_THREAD_SINGLE
    call MPI_init_thread(required, provided,  ierror)
    _VERIFY(ierror)
    call MPI_Comm_rank(MPI_COMM_WORLD, rank,  ierror)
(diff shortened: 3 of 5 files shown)


weiyuan-jiang:
The problem is if the user initializes MPI with MPI_THREAD_MULTIPLE, do we need this ASSERT ?
[
https://github.com/GEOS-ESM/MAPL/blob/0d334bc2879ef8c500438e6646d12d8308a81154/gridcomps/Cap/MAPL_Cap.F90#L493

mathomp4:
@aoloso @tclune I want to make sure `MPI_THREAD_SINGLE` is what we need. Per [ESMF docs for `ESMF_InitializePreMPI()`](https://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node4.html#SECTION04024200000000000000):

> This method is only needed for cases where MPI is initialized explicitly by user code. In most typical cases `ESMF_Initialize()` is called before MPI is initialized, and takes care of all the internal initialization, including MPI.
> 
> There are circumstances where it is necessary or convenient to initialize MPI before calling into `ESMF_Initialize()`. This option is supported by ESMF, and for most cases no special action is required on the user side. However, for cases where `ESMF_*CompSetVM*()` methods are used to move processing elements (PEs), i.e. CPU cores, between persistent execution threads (PETs), ESMF uses POSIX signals between PETs. In order to do so safely, the proper signal handlers must be installed before MPI is initialized. This is accomplished by calling `ESMF_InitializePreMPI()` from the user code prior to the MPI initialization.
> 
> Note also that in order to use any of the advanced resource management functions that ESMF provides via the `ESMF_*CompSetVM*()` methods, the MPI environment must be thread-safe. `ESMF_Initialize()` handles this automatically if it is in charge of initializing MPI. However, if the user code initializes MPI before calling into `ESMF_Initialize()`, it must do so via `MPI_Init_thread()`, specifying `MPI_THREAD_SERIALIZED` or above for the required level of thread support. 

Per the MPI Standard, when using the World Model, [the `MPI_Init_thread()` levels are (in order)](https://www.mpi-forum.org/docs/mpi-4.1/mpi41-report/node267.htm#Node267):

This call initializes MPI in the same way that a call to MPI_INIT would. In addition, it initializes the thread environment. The argument required is used to specify the desired level of thread support. The possible values are listed in increasing order of thread support.

> `MPI_THREAD_SINGLE`:
> Only one thread will execute.
> `MPI_THREAD_FUNNELED`:
> The process may be multithreaded, but the application must ensure that only the main thread makes MPI calls (for the definition of main thread, see `MPI_IS_THREAD_MAIN` on page [Starting MPI Processes](https://www.mpi-forum.org/docs/mpi-4.1/mpi41-report/node267.htm#Node267)).
> `MPI_THREAD_SERIALIZED`:
> The process may be multithreaded, and multiple threads may make MPI calls, but only one at a time: MPI calls are not made concurrently from two distinct threads (all MPI calls are ``serialized'').
> `MPI_THREAD_MULTIPLE`:
> Multiple threads may call MPI, with no restrictions.
> 
> These values are monotonic; i.e., `MPI_THREAD_SINGLE` < `MPI_THREAD_FUNNELED` < `MPI_THREAD_SERIALIZED` < `MPI_THREAD_MULTIPLE`.

My reading is that the lowest we can use for `MPI_Init_thread()` when passing the communicator into `ESMF_Initialize()` is `MPI_THREAD_SERIALIZED` if we call any of "the advanced resource management functions that ESMF provides via the `ESMF_*CompSetVM*()` methods". And there are [checks in ESMF](https://github.com/esmf-org/esmf/blob/bf628cb87505c2fc6892804b1a3054fd69e3d3c7/src/Infrastructure/VM/include/ESMCI_VMKernel.h#L496-L499) that seem to agree. @aoloso do we use those?

I might invoke the names of @oehmke and @theurich for their thoughts.

mathomp4:
Also, per the comment by @weiyuan-jiang, this:
```fortran
_ASSERT(provided == MPI_THREAD_MULTIPLE, 'MPI_THREAD_MULTIPLE not supported by this MPI.')
```
was "good" in that `MPI_THREAD_MULTIPLE` was the highest level of thread support. If we are supporting `MPI_THREAD_SINGLE` then I think asserts are moot. Since `MPI_THREAD_SINGLE` is the *lowest* level of thread support, I think *any* call to `MPI_Init_thread()` will at least return that as the `provided`:

> The level(s) of thread support that can be provided by `MPI_INIT_THREAD` will depend on the implementation, and may depend on information provided by the user before the program started to execute (e.g., with arguments to mpiexec). If possible, the call will return provided = required. Failing this, the call will return the least supported level such that provided > required (thus providing a stronger level of support than required by the user). Finally, if the user requirement cannot be satisfied, then the call will return in provided the highest supported level.
> 
> A thread compliant MPI implementation will be able to return `provided = MPI_THREAD_MULTIPLE`. Such an implementation may always return `provided = MPI_THREAD_MULTIPLE`, irrespective of the value of required.
> 
> An MPI library that is not thread compliant must always return `provided = MPI_THREAD_SINGLE`, even if `MPI_INIT_THREAD` is called on a multithreaded process. The library should also return correct values for the MPI calls that can be executed before initialization, even if multiple threads have been spawned.

tclune (approved):
I am still concerned about @mathomp4 comment about minimum needed for ESMF system image vs this new setting.    But if it works, it works.

weiyuan-jiang (changes requested):
For this PR, I still think we only need to remove the block of the original file 
https://github.com/GEOS-ESM/MAPL/blob/0d334bc2879ef8c500438e6646d12d8308a81154/gridcomps/Cap/MAPL_Cap.F90#L485-L493

mathomp4:
Closing superseded by #3493

PR #3503, opened 2025-03-20
Check that the Fortran pointer type and kind match the ESMF_Field typekind in assign_fptr 

darianboggs:
## Types of change(s)

## Checklist

## Description
**assign_fptr** does not check that the type and kind of the Fortran pointer match the typekind of the **ESMF_Field**, and this can result in incorrect casts. This PR adds a check to guarantee that the pointer and field are consistent so there is no casting.

## Related Issue
3492

### DIFF
--- a/field/FieldPointerUtilities.F90
+++ b/field/FieldPointerUtilities.F90
@@ -1094,25 +1094,12 @@ contains
       type(ESMF_TypeKind_Flag), intent(in) :: typekind
       integer, optional, intent(out) :: rc
       integer :: status
-      logical :: matches
-
-      matches = typekind_matches(field, typekind, _RC)
-      _ASSERT(matches, 'Field typekind does not match pointer type and kind.')
-      _RETURN(_SUCCESS)
-
-   end subroutine check_typekind
-
-   logical function typekind_matches(field, typekind, rc)
-      type(ESMF_Field), intent(in) :: field
-      type(ESMF_TypeKind_Flag), intent(in) :: typekind
-      integer, optional, intent(out) :: rc
-      integer :: status
       type(ESMF_TypeKind_Flag) :: actual_typekind
 
       call ESMF_FieldGet(field, typekind=actual_typekind, rc=status)
-      typekind_matches = actual_typekind == typekind
-      _RETURN(status)
+      _ASSERT(actual_typekind == typekind, 'Field typekind does not match pointer type and kind.')
+      _RETURN(_SUCCESS)
 
-   end function typekind_matches
+   end subroutine check_typekind
 
 end module MAPL_FieldPointerUtilities
--- a/field/FieldPointerUtilities.F90
+++ b/field/FieldPointerUtilities.F90
@@ -96,6 +96,7 @@ contains
       integer(ESMF_KIND_I8) :: local_size
       integer :: status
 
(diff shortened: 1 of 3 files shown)


field/FieldPointerUtilities.F90
```diff
@@ -1081,4 +1089,30 @@ subroutine assign_fptr_i8_rank1(x, fptr, rc)
...
+      type(ESMF_TypeKind_Flag), intent(in) :: typekind
+      integer, optional, intent(out) :: rc
+      integer :: status
+      logical :: matches
+
+      matches = typekind_matches(field, typekind, _RC)
+      _ASSERT(matches, 'Field typekind does not match pointer type and kind.')
+      _RETURN(_SUCCESS)
+
+   end subroutine check_typekind
```
tclune, line 1103: Why 2 procedures?   I can understand that you don't want an assert here, so just put the assert on the uses above.  (Will add comments there now that I see this issue.)

tclune: Ah.  Because the FieldGet call might fail for unrelated reasons.  Sigh.

field/FieldPointerUtilities.F90
```diff
@@ -1081,4 +1089,30 @@ subroutine assign_fptr_i8_rank1(x, fptr, rc)
...
+
+   logical function typekind_matches(field, typekind, rc)
+      type(ESMF_Field), intent(in) :: field
+      type(ESMF_TypeKind_Flag), intent(in) :: typekind
+      integer, optional, intent(out) :: rc
+      integer :: status
+      type(ESMF_TypeKind_Flag) :: actual_typekind
+
+      call ESMF_FieldGet(field, typekind=actual_typekind, rc=status)
```
tclune, line 1112:
```suggestion
      call ESMF_FieldGet(field, typekind=actual_typekind, _RC)
```

generic3g/tests/Test_MeanTransform.pf
```diff
@@ -40,9 +40,9 @@ contains
 
       ! counter 0 at one point
       call FieldSet(acc%accumulation_field, COUNTER*MEAN, _RC)
-      call assign_fptr(acc%counter_field, fptr, _RC)
-      fptr(n) = 0
-      mask = fptr /= 0
+      call assign_fptr(acc%counter_field, ifptr, _RC)
+      ifptr(n) = 0
+      mask = ifptr /= 0
```
tclune, line 45: Does this mean the PR actually found a bug in an existing test?  (Just want to be sure I understood.)

tclune (changes requested):
Minor change requested.

field/FieldPointerUtilities.F90
tclune: Still seems overkill to have 2 procedures here.  What is the downside of putting the assert in the other routine and deleting this one?

generic3g/tests/Test_MeanTransform.pf
darianboggs: Yes.

field/FieldPointerUtilities.F90
darianboggs: I'll make the changes we agreed on.

tclune (approved):

PR #3521, opened 2025-03-25
Fixes #3501: MAPL_GridCompAddFieldSpec - new public interface to add a field spec

### LINKED ISSUE #3501: New public interface to add a variable spec
Interface: `MAPL_GridCompAddVarspec`

NOTE: The interface accepts a string value (`xyz`/`xy`/`z`) for `dim_spec` and processes it before creating the `VariableSpec`.

pchakraborty:
## Types of change(s)

## Checklist

## Description

Fixes #3501

### DIFF
--- a/generic3g/MAPL_Generic.F90
+++ b/generic3g/MAPL_Generic.F90
@@ -22,13 +22,15 @@ module mapl3g_Generic
    use mapl3g_OuterMetaComponent, only: get_outer_meta
    use mapl3g_ChildSpec, only: ChildSpec
    use mapl3g_ComponentSpec, only: ComponentSpec
-   use mapl3g_VariableSpec, only: VariableSpec
+   use mapl3g_VariableSpec, only: VariableSpec, make_VariableSpec
    use mapl3g_Validation, only: is_valid_name
    use mapl3g_ESMF_Interfaces, only: I_Run
    use mapl3g_StateItemSpec
    use mapl3g_VerticalGrid
+   use mapl3g_VerticalStaggerLoc, only: VerticalStaggerLoc
    use mapl3g_StateRegistry, only: StateRegistry
    use mapl_InternalConstantsMod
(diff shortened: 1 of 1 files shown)


generic3g/MAPL_Generic.F90
```diff
@@ -402,6 +407,62 @@ subroutine gridcomp_add_varspec_basic(gridcomp, variable_spec, rc)
       _RETURN(_SUCCESS)
    end subroutine gridcomp_add_varspec_basic
 
+   subroutine gridcomp_add_fieldspec( &
```
tclune, line 410: Just a reminder that we may decide to generalize this to support e.g. TangentVector.  Basically the intent is to provide a simplified interface for the ACG.

And ... we have to keep in mind that there will always be some exceptional specs that require explicit user code which may place additional pressure on the args and/or name here.

tclune (approved):

generic3g/MAPL_Generic.F90
```diff
@@ -402,6 +407,62 @@ subroutine gridcomp_add_varspec_basic(gridcomp, variable_spec, rc)
...
+
+      type(ESMF_StateItem_Flag), parameter :: itemtype = ESMF_STATEITEM_FIELD
+      type(VariableSpec) :: var_spec
+      type(HorizontalDimsSpec) :: horizontal_dims_spec
+      type(OuterMetaComponent), pointer :: outer_meta
+      type(ComponentSpec), pointer :: component_spec
+      integer :: status
+
+      _ASSERT((dims=="xyz") .or. (dims=="xy") .or. (dims=="z"), "dims can be one of xyz/xy/z")
+      horizontal_dims_spec = HORIZONTAL_DIMS_GEOM
```
tclune, line 444: There is something interesting about the fact that no special logic is needed for vertical here, but horizontal needs it.   Not immediately seeing anything better, but bringing it to other intention in case they do.

pchakraborty: Indeed.

PR #3529, opened 2025-03-25
Fixes #3528 - Added an option to mapl_acg to use ACG v3 instead of ACG

### LINKED ISSUE #3528: Additional option for mapl_acg to use v3 instead
Currently, the cmake function `mapl_acg` executes `Apps/MAPL_GridCompSpecs_ACG.py`. With this new option, one would be able to executed `Apps/MAPL_GridCompSpecs_ACGv3.py` instead.

pchakraborty:
## Types of change(s)

## Checklist

## Description

Added the option `3g` for the cmake function `mapl_acg`. With this, one should be able to run `MAPL_GridCompSpecs_ACGv3.py`.

## Related Issue

#3528

### DIFF
--- a/Apps/CMakeLists.txt
+++ b/Apps/CMakeLists.txt
@@ -5,11 +5,13 @@
 # used.
 
 file (COPY MAPL_GridCompSpecs_ACG.py DESTINATION ${esma_etc}/MAPL)
+file (COPY MAPL_GridCompSpecs_ACGv3.py DESTINATION ${esma_etc}/MAPL)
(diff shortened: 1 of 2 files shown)


tclune (approved):
This works, but I think it would have been simpler to also introduce a new cmake function `mapl_acg3` rather than conditional logic.

pchakraborty:
> This works, but I think it would have been simpler to also introduce a new cmake function `mapl_acg3` rather than conditional logic.

I tried that as well. This is a significantly simpler solution.

PR #3534, opened 2025-03-26
Refactor ACG for MAPL3

darianboggs:
## Types of change(s)

## Checklist

## Description
The procedures for adding Fields is sufficiently different in MAPL3, so the ACG needed to be refactored to remove idiosyncratic behavior that was tightly coupled to MAPL. This PR:
1. implements the refactor
2. modifies the procedure created by ACG to match the FieldSpec addition procedure in MAPL3.
Further refactoring will be necessary with further MAPL3 developments.


tclune (commented):
How is this a breaking change?

darianboggs:
It's probably too soon to call it breaking. Existing ACG spec files for MAPL will need to change because of changes in columns and behavior. Some of that can be handled by aliases when there is a simple 1-1 relationship between column names which results in functionally similar behavior. Other columns, row values, and behavior will require changes to spec files. The files will obviously not be the same, but many of the changes will result in the results for the user.

pchakraborty:
Let me give this a try.

tclune:
> It's probably too soon to call it breaking. Existing ACG spec files for MAPL will need to change because of changes in columns and behavior. Some of that can be handled by aliases when there is a simple 1-1 relationship between column names which results in functionally similar behavior. Other columns, row values, and behavior will require changes to spec files. The files will obviously not be the same, but many of the changes will result in the results for the user.

For the moment I would like to stick to the requirement that ACG be backward compatible with MAPL2.  By this, I mean that a spec file from MAPL2 will work (and produce correct results) in MAPL3.    We can have extra alias's and extensions galore, but supporting existing spec files is extremely important.

Apps/MAPL_GridCompSpecs_ACGv3.py
```diff
@@ -103,97 +105,107 @@ def __call__(self, name, parameters):
...
+        'OPT'  : 'MAPL_RestartOptional',
+        'SKIP' : 'MAPL_RestartSkip',
+        'REQ'  : 'MAPL_RestartRequired',
+        'BOOT' : 'MAPL_RestartBoot',
+        'SKIPI': 'MAPL_RestartSkipInitial'
+    }},
+    'rotation': {'flags': PRINTABLE_ARGUMENT, },
+    'staggering': {'flags': PRINTABLE_ARGUMENT, },
+    'ungridded_dims': {'flags': PRINTABLE_ARGUMENT, 'writer': 'array'},
+    'vstagger': {'flags': PRINTABLE_ARGUMENT, 'writer': {
```
pchakraborty, line 167: These will be `VERTICAL_STAGGER_CENTER/EDGE/NONE`

pchakraborty (changes requested):
I see three issues

- `VERTICAL_SCATTER_CENTER/EDGE/NONE` should be `VERTICAL_STAGGER_CENTER/EDGE/NONE`
- Something is off with processing of the UNGRIDDED column
  - An `INTERNAL` spec with no ungridded dims still has an `ungridded_dims` argument
    ```Fortran
    call MAPL_GridCompAddFieldSpec(gc,&
        & short_name='SGH30', &
        & units='m', &
        & dims='xy', &
        & vstagger=VERTICAL_STAGGER_NONE, &
        & ungridded_dims=[ ], & ! <-- We shouldn't have this
        & standard_name='standard deviation of 30s elevation from 3km cube', &
        & state_intent=ESMF_STATEINTENT_INTERNAL, &
        & _RC)
    ```
  - The corresponding DeclarePointer line should be
    ```Fortran
     real, pointer, dimension(:,:) :: SGH30
     ```
     instead of the current `real, pointer, dimension(:,:,:) :: SGH30`
- The GetPointer call should be
  ```Fortran
  call MAPL_GetPointer(IMPORT, PLE, 'PLE', _RC)
  ```
   instead of `call MAPL_GetPointer(ESMF_STATEINTENT_IMPORT, PLE, 'PLE', _RC)

tclune:
> call MAPL_GetPointer(IMPORT, PLE, 'PLE', _RC)

This comment would have been more obvious had we already give the procedure its proper name.   What we really want is:
```
call MAPL_StateGetPointer(IMPORT, PLE, 'PLE', _RC)
```
Really too bad we cannot return "deferred rank pointers".  If Fortran were to support that then we could do:
```
PLE => MAPL_StateGetPointer(IMPORT, 'PLE', _RC)
```
wishes wishes

Apps/MAPL_GridCompSpecs_ACGv3.py
darianboggs: ungridded_dims is changing. I'll take a look at the issue of the argument. The issue is the generated subroutine call. If the subroutine is corrected, the declare will be corrected. Can you send me the specs file you used? If the column has all spaces, no ungridded_dims argument should be created. If it is created, the rank is determined by looking at the array and counting the dimensions from dimension delimiters. Under no circumstances will there be an empty array as an argument. I am working on exceptions for that case. So the number of delimiters + 1 is the rank. So, for some reason, the ACG read the ungridded_dims column, and it interpreted it as non-space.

Thanks for the correction on the VSTAGGER. It didn't make sense that the argument name is VSTAGGER and the values are VSCATTER... I was going to ask you if that was a typo.

So, in ** MAPL_GridCompAddFieldSpec**, the argument is **state_intent** with type **ESMF_StateIntentFlag**, and in **MAPL_GetPointer**, the argument is **state** with type **ESMF_State**. Is that correct?

If so, I will make the column 'StateIntent' which is what is used as the argument for **MAPL_GridCompAddFieldSpec**, and then in **MAPL_GetPointer**, the argument will be **state** with type **ESMF_State**.

pchakraborty: Here is the `INTERNAL` category of the specs file:
```
category: INTERNAL
#--------------------------------------------------------------------------------------------------------------------
#  VARIABLE    | DIMENSION   |          Additional Metadata
#--------------------------------------------------------------------------------------------------------------------
 NAME   | UNITS   | DIMS | VLOC | UNGRIDDED  | LONG NAME
#--------------------------------------------------------------------------------------------------------------------
 SGH30  | m       | xy   | N    |            | standard deviation of 30s elevation from 3km cube
 KWVRDG | km      | xy   | N    | [16]       | horizonal wwavenumber of mountain ridges
 EFFRDG | km      | xy   | N    | [16]       | efficiency of mountain ridge scheme
 GBXAR  | NA      | xy   | N    |            | grid box area
 HWDTH  | km      | xy   | N    | [16]       | width of mountain ridges
 CLNGT  | km      | xy   | N    | [16]       | width of mountain ridges
 MXDIS  | NA      | xy   | N    | [16]       | NA
 ANGLL  | NA      | xy   | N    | [16]       | NA
 ANIXY  | NA      | xy   | N    | [16]       | NA
```

pchakraborty: > So, in ** MAPL_GridCompAddFieldSpec**, the argument is **state_intent** with type **ESMF_StateIntentFlag**, and in **MAPL_GetPointer**, the argument is **state** with type **ESMF_State**. Is that correct?

Yes, that is correct.

tclune:
Just noticed that this PR is still open after 3 weeks.  I realize that we will have a steady stream of ACG changes in the near future, but let's try to get something resolved here and just open a new issue for extra changes.  Or ... if there is a reason that things cannot be merged now, please update me.

Apps/MAPL_GridCompSpecs_ACGv3.py
darianboggs: All changes made.

Apps/MAPL_GridCompSpecs_ACGv3.py
```diff
@@ -1,396 +1,246 @@
...
-                        default=None, const="{component}_Export___.h",
-                        help="override default output filename for AddExternalSpec() code")
+                    default=None, const="{component}_Export___.h",
+                    help="override default output filename for AddExternalSpec() code")
     parser.add_argument("-p", "--internal_specs", action="store", nargs='?',
-                        default=None, const="{component}_Internal___.h",
-                        help="override default output filename for AddImportSpec() code")
+                    default=None, const="{component}_Internal___.h",
+                    help="override default output filename for AddImportSpec() code")
     parser.add_argument("-g", "--get-pointers", action="store", nargs='?',
```
pchakraborty, line 230: Should be `get_pointers` instead

Apps/MAPL_GridCompSpecs_ACGv3.py
```diff
@@ -426,108 +280,145 @@ def dataframe(reader, columns):
...
-    else:
-        component = os.path.splitext(os.path.basename(args.input))[0]
-        component = component.replace('_Registry','')
-        component = component.replace('_StateSpecs','')
+    argdict = options[ARGDICT]
+    exit_code_ = ERROR
+
+    add_newline = lambda s: f"{s.rstrip()}{linesep}"
+
+    component, declare_pointers, get_pointers = (argdict.get(k) for k in ('name', 'declare_pointers', 'get-pointers'))
```
pchakraborty, line 387: Should be `get_pointers` instead

pchakraborty (changes requested):
We need a small change `get-pointers` -> `get_pointers` in a couple of places.

Apps/MAPL_GridCompSpecs_ACGv3.py
tclune: agreed

Apps/MAPL_GridCompSpecs_ACGv3.py
darianboggs: Fixed all the command-line options to use `-` and `_`. After reviewing the conventions for Unix-based operating systems, it appears that for command-line options, both are used, and the only real consistency is the `getopt` and `getopt_long` conventions, and after the `-` and `--`, there seem to be different approaches.

PR #3549, opened 2025-03-29
Fixes #3548 - Calling make_VariableSpec with UngriddedDims

### LINKED ISSUE #3548: Update MAPL_GridCompAddFieldSpec to create VariableSpec with UngriddedDims

pchakraborty:
## Types of change(s)

## Checklist

## Description

Calling `make_VariableSpec` with `UngriddedDims`. Also added an `UngriddedDims` constructor that takes an array of extents as an argument.

## Related Issue

Fixes #3548

### DIFF
--- a/esmf_utils/UngriddedDims.F90
+++ b/esmf_utils/UngriddedDims.F90
@@ -38,6 +38,7 @@ module mapl3g_UngriddedDims
       module procedure new_UngriddedDims_empty
       module procedure new_UngriddedDims_vec
       module procedure new_UngriddedDims_arr
+      module procedure new_UngriddedDims_extent_arr
    end interface UngriddedDims
 
    interface operator(==)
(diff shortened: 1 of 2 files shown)


esmf_utils/UngriddedDims.F90
```diff
@@ -79,6 +80,18 @@ function new_UngriddedDims_arr(dim_specs) result(spec)
...
+      integer, intent(in) :: extent_arr(:)
+
+      integer :: i
+      type(UngriddedDim) :: dim_spec
+
+      do i = 1, size(extent_arr)
+         dim_spec = UngriddedDim(extent_arr(i))
+         call spec%dim_specs%push_back(dim_spec)
+      end do
+   end function new_UngriddedDims_extent_arr
```
tclune, line 94: This does not support the scalar case, which is the most important.   Indeed, I'm tempted to disallow this shortcut for multiple ungridded dimensions, as names and such become so much more important when there is more than one.

tclune (changes requested):
See inline comment.

esmf_utils/UngriddedDims.F90
pchakraborty: I am not sure if we need a specific constructor for the scalar case. The scalar case can use the current interface by creating an array, as in `UngriddedDims([scalar_case])`. But I can always add one - `new_UngriddedDims_extent`.

Or we can go back to what I had originally, i.e. move the logic to `MAPL_Generic::gridcomp_add_fieldspec`

tclune (approved):

PR #3550, opened 2025-03-29
Update geomio for vert only fields

pchakraborty:
## Types of change(s)

## Checklist

## Description

Added a separate routine for the case when the field is vertical only.


GeomIO/pFIOServerBounds.F90
```diff
@@ -56,16 +56,22 @@ function get_file_shape(this) result(file_shape)
...
       integer, intent(in), optional :: time_index
       integer, intent(out), optional :: rc
       type(pFIOServerBounds) :: server_bounds
 
+      integer, allocatable :: field_shape(:)
       integer :: status, tile_count, n_dims, tm, global_dim(3)
       integer :: i1, in, j1, jn, tile, extra_file_dim, file_dims, new_grid_dims
 
+      field_shape = in_field_shape
+      if (size(in_field_shape)==1) then ! vertical-only field
```
tclune, line 71: I do wonder how fragile this logic is.    What if someone has a surface loc stream for their geom?

tclune: I think you need to access the metadata that is added during Field construction.  That, or you need to retrieve the geom and do analysis to determine if this 1 dimension is vertical or not.

pchakraborty: It might be alright in this case since the first argument to the routine is a grid. We will need a separate routine for a locstream.

GeomIO/SharedIO.F90
```diff
@@ -124,33 +128,40 @@ function get_variable_dim_names(field, geom, rc) result(dim_names)
...
+      integer :: grid_to_field_map(2), status
+
+      ! horizontal dimension
       call ESMF_FieldGet(field, geom=esmfgeom, _RC)
       mapl_geom => get_mapl_geom(esmfgeom, _RC)
       grid_variables = mapl_geom%get_gridded_dims()
-      dim_names = string_vec_to_comma_sep(grid_variables)
+      call ESMF_FieldGet(field, gridToFieldMap=grid_to_field_map, _RC)
+      vert_only = .false.
+      if (all(grid_to_field_map==0)) vert_only = .true.
```
tclune, line 146:
```suggestion
      vert_only = all(grid_to_field_map==0)
```

tclune (approved):

PR #3555, opened 2025-03-30
Partially fixes #3554 - create 1d field

### LINKED ISSUE #3554: Create 1D fields for VertOnly variables

pchakraborty:
## Types of change(s)

## Checklist

## Description

Ok, this is tricky. Ideally, one should be able to create a 1D field using the 'typekind' interface of `ESMF_FieldEmptyComplete` simply based on the argument `gridToFieldMap`, but there seems to be a bug in ESMF that prevents this (@bena-nasa has filed a bug report). So, we are using the 'farray' interface instead to create the 1D field, and that makes the code clunky. For simplicity, we restrict ourselves to the `ESMF_KIND_R4` case. Another restriction is that there are no additional ungridded dimensions in VertOnly variables.

## Related Issue

#3554

### DIFF
--- a/field/FieldCreate.F90
+++ b/field/FieldCreate.F90
@@ -1,20 +1,21 @@
 #include "MAPL_Generic.h"
 
 module mapl3g_FieldCreate
+
    use mapl3g_VerticalStaggerLoc
    use mapl3g_FieldInfo
    use mapl3g_UngriddedDims
+   use mapl3g_LU_Bound
    use mapl_KeywordEnforcer
    use mapl_ErrorHandling
-   use mapl3g_LU_Bound
+   use mapl_InternalConstantsMod, only: MAPL_UNDEFINED_REAL
    use esmf, MAPL_FieldEmptyCreate => ESMF_FieldEmptyCreate
+
    implicit none(type,external)
    private
 
-
    public :: MAPL_FieldCreate
    public :: MAPL_FieldEmptyComplete
-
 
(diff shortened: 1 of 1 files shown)


field/FieldCreate.F90
```diff
@@ -84,18 +85,45 @@ subroutine field_empty_complete( field, &
...
+              indexFlag=ESMF_INDEX_DELOCAL, &
+              datacopyFlag=ESMF_DATACOPY_VALUE, &
+              gridToFieldMap=grid_to_field_map, &
+              ungriddedLBound=bounds%lower, &
+              ungriddedUBound=bounds%upper, &
+              _RC)
+      else
+         call ESMF_FieldEmptyComplete(field, typekind=typekind, &
+              gridToFieldMap=gridToFieldMap, &
+              ungriddedLBound=bounds%lower, ungriddedUBound=bounds%upper, _RC)
```
tclune, line 123: I do not like this divergence in the creation arguments.   In particular, it is not at all obvious why vert only has DATACOPY and INDEX arguments.  Nor why it is returning an farray.

pchakraborty: I addressed the issue some in the description of the PR. Essentially, there seems to be a bug in the the type kind interface of `ESMF_FieldEmptyComplete`, so we use the `farray` interface. A bug report was filed, and once it is fixed, this extra bit of code goes away.

tclune (approved):

PR #3570, opened 2025-04-07
Feature/#3566 geom init phases

tclune:
## Types of change(s)

## Checklist


generic3g/specs/StateItemAspect.F90
```diff
@@ -166,7 +166,7 @@ logical function can_connect_to(src, dst)
            end if
         case (1)
            can_connect_to = .true.
-        case (2)
```
atrayano, line 169: The only difference in this file is the identation, and I like the older version better

generic3g/specs/WildcardClassAspect.F90
```diff
@@ -45,7 +45,7 @@ module mapl3g_WildcardClassAspect
 
    function new_WildcardClassAspect() result(wildcard_aspect)
       type(WildcardClassAspect) :: wildcard_aspect
-
+      _HERE
```
atrayano, line 48: Do we need the debugging print(s)?

tclune: I think the output from this one may have been buried in the stream of warning diagnostics that I cannot remove yet.
Fixed.

generic3g/specs/StateItemAspect.F90
tclune: Fixed.

atrayano (dismissed):

atrayano (approved):

atrayano (approved):

PR #3576, opened 2025-04-07
Fixed #3551 - new initialize phase - read_restart

### LINKED ISSUE #3551: Add restart reading to one of the initialize phases

pchakraborty:
## Types of change(s)

## Checklist

## Description

Created a new initialize phase, `GENERIC_INIT_READ_RESTART` to read restart files, when available, instead of adding restart reading to one of the existing phases.

Also cleaned up the `write_restart` routine, which, for now, remains an ESMF service.

## Related Issue

#3551


generic3g/GenericGridComp.F90
```diff
@@ -61,6 +61,7 @@ subroutine set_entry_points(gridcomp, rc)
          call ESMF_GridCompSetEntryPoint(gridcomp, ESMF_METHOD_INITIALIZE, initialize, phase=GENERIC_INIT_MODIFY_ADVERTISED2, _RC)
          call ESMF_GridCompSetEntryPoint(gridcomp, ESMF_METHOD_INITIALIZE, initialize, phase=GENERIC_INIT_REALIZE, _RC)
 !#         call ESMF_GridCompSetEntryPoint(gridcomp, ESMF_METHOD_INITIALIZE, initialize, phase=GENERIC_INIT_RESTORE, _RC)
+         call ESMF_GridCompSetEntryPoint(gridcomp, ESMF_METHOD_INITIALIZE, initialize, phase=GENERIC_INIT_READ_RESTART, _RC)
```
tclune, line 64: You should delete the lines just before this about "RESTORE".  (Unfortunately can't do the "suggestion" here.)

generic3g/GenericGridComp.F90
```diff
@@ -170,6 +170,8 @@ recursive subroutine initialize(gridcomp, importState, exportState, clock, rc)
          call outer_meta%initialize_realize(_RC)
 !#      case (GENERIC_INIT_RESTORE)
 !#         call outer_meta%initialize_realize(_RC)
+      case (GENERIC_INIT_READ_RESTART)
```
tclune, line 173: Same comment as above about deleting preceding lines.  (And these have a bad cut-and-paste-error as well.)

generic3g/OuterMetaComponent/initialize_read_restart.F90
```diff
@@ -0,0 +1,49 @@
...
+
+      character(*), parameter :: PHASE_NAME = 'GENERIC::INIT_READ_RESTART'
+      type(GriddedComponentDriver), pointer :: driver
+      character(len=:), allocatable :: name
+      type(MultiState) :: states
+      type(ESMF_State) :: import_state, internal_state
+      type(RestartHandler) :: restart_handler
+      integer :: status
+
+      driver => this%get_user_gc_driver()
```
tclune, line 28: This routine lacks recursion over children.

pchakraborty: The recursion is on line 41 - in the `if` block that checks if the name is `HIST` https://github.com/GEOS-ESM/MAPL/blob/e8e1d88b3302481a4333464240770fae2eb1971f/generic3g/OuterMetaComponent/initialize_read_restart.F90#L40-L42

generic3g/OuterMetaComponent/initialize_read_restart.F90
```diff
@@ -0,0 +1,49 @@
...
+         call states%get_state(import_state, "import", _RC)
+         call restart_handler%read("import", import_state, _RC)
+         call states%get_state(internal_state, "internal", _RC)
+         call restart_handler%read("internal", internal_state, _RC)
+      end if
+
+      call this%run_custom(ESMF_METHOD_INITIALIZE, PHASE_NAME, _RC)
+      if (name /= "HIST") then
+         call recurse(this, phase_idx=GENERIC_INIT_READ_RESTART, _RC)
+      end if
```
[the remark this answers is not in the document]
pchakraborty: I'll remove the check - that will print that restarts are not available for the children of `HIST`. I can figure out a better way of handling that later.

generic3g/GenericGridComp.F90
pchakraborty: ok.

generic3g/GenericGridComp.F90
pchakraborty: ok.

generic3g/OuterMetaComponent/initialize_read_restart.F90
tclune: Ah - ok.   The original implementation made the recursion look like the exceptional case.

  The alternative (if HIST was still needed as an exception) would be:
```
_RETURN_IF(name == "HIST")
call recurse(...)
```

Am starting to wonder if `recurse` should be made a method. ...

tclune (approved):

generic3g/OuterMetaComponent/initialize_read_restart.F90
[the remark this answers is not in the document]
pchakraborty: Removed the check for HIST

PR #3661, opened 2025-04-28
avoid using undefined comp_name in MAPL_GridCreate when gc is not present

weiyuan-jiang:
## Types of change(s)

## Checklist

## Description

When argument gc is not present, comp_name is not defined but used. That leads to unwanted label. Also,  there is typo in setting  prefix//NX:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- define comp_name in MAPL_GridCrate when GC is not present
+- define `comp_name` in `MAPL_GridCreate` when GC is not present
 - Fixed bug in profiler demo
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix a bug in MAPL_GridCreate to set "NX" and clean up when GC is not present
 - Fixed bug in profiler demo
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix a bug in MAPL_GridCreate to set "NX" and clean up when GC is not present
+- define comp_name in MAPL_GridCrate when GC is not present
 - Fixed bug in profiler demo
 
 ### Added
--- a/generic/MAPL_Generic.F90
+++ b/generic/MAPL_Generic.F90
@@ -10098,8 +10098,9 @@ contains
       nn = len_trim(gridname)
       dateline = gridname(nn-1:nn)
       if (dateline == 'CF') then
+         ! convert global NY to a local NY for each face
          call ESMF_ConfigGetAttribute(state%CF,ny,label=trim(Prefix)//'NY:', _RC)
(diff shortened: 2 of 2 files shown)


generic/MAPL_Generic.F90
```diff
@@ -10055,6 +10056,9 @@ subroutine MAPL_GridCreate(GC, MAPLOBJ, ESMFGRID, srcGC, rc)
          call ESMF_GridCompGet( GC, name=Comp_Name,   rc = status )
          _VERIFY(status)
          Iam = trim(Comp_Name)//Iam
+         Prefix = trim(comp_name)//MAPL_CF_COMPONENT_SEPARATOR
+      else
+         Prefix = ''
```
tclune, line 10061: Move this to before the `if` to eliminate the need for an `else` construct.

generic/MAPL_Generic.F90
```diff
@@ -10091,23 +10095,25 @@ subroutine MAPL_GridCreate(GC, MAPLOBJ, ESMFGRID, srcGC, rc)
          _FAIL('needs informative message')
       endif
 
-      call MAPL_ConfigPrepend(state%cf,trim(comp_name),MAPL_CF_COMPONENT_SEPARATOR,'NX:',rc=status)
-      _VERIFY(status)
-      call MAPL_ConfigPrepend(state%cf,trim(comp_name),MAPL_CF_COMPONENT_SEPARATOR,'NY:',rc=status)
-      _VERIFY(status)
+      if (trim(Prefix) /= '') then
+         call MAPL_ConfigPrepend(state%cf,trim(comp_name),MAPL_CF_COMPONENT_SEPARATOR,'NX:',rc=status)
```
tclune, line 10099: While here anyway,  please switch to use `_RC` in this section.

weiyuan-jiang:
I think we should respect  how users decompose the domain and should not  enforce NX=NY/6.
https://github.com/GEOS-ESM/MAPL/blob/2c5b3fb8e62cdfe255a3e20a989113bfb6e64f2a/generic/MAPL_Generic.F90#L10102

@tclune @bena-nasa

weiyuan-jiang:
It seems to me this is a bug ( NY should be NX)
https://github.com/GEOS-ESM/MAPL/blob/753a4de11302655d1d7c80a6537f476b0e109614/generic/MAPL_Generic.F90#L10106

But After I changed to NX, the tests failed.  Do you know what happens here ? @bena-nasa

bena-nasa:
> I think we should respect how users decompose the domain and should not enforce NX=NY/6.
> 
> https://github.com/GEOS-ESM/MAPL/blob/2c5b3fb8e62cdfe255a3e20a989113bfb6e64f2a/generic/MAPL_Generic.F90#L10102
> 
> @tclune @bena-nasa

The original code was correct, please restore

weiyuan-jiang:
> > I think we should respect how users decompose the domain and should not enforce NX=NY/6.
> > https://github.com/GEOS-ESM/MAPL/blob/2c5b3fb8e62cdfe255a3e20a989113bfb6e64f2a/generic/MAPL_Generic.F90#L10102
> > 
> > @tclune @bena-nasa
> 
> The original code was correct, please restore

Thanks. The original is restored and a comment is added

weiyuan-jiang:
This PR helps to move some functions from  GEOSgcm_GridComp's makebc to MAPL

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- define comp_name in MAPL_GridCrate when GC is not present
```
mathomp4, line 12:
```suggestion
- define `comp_name` in `MAPL_GridCreate` when GC is not present
```

PR #3662, opened 2025-04-28
Feature/bmauer/fixes #3634

### LINKED ISSUE #3634: MAPL_Filter
# The Problem

### Updated:  2025-04-25

In biblical times, creating passive instances with GOCART tagged tracers was a run time decision where each instance controlled the masking and customizations of its emissions via resource file. With the introduction of ExtData and more recently GOCART-2G, creating passive instances with tagged tracers became a run-time and compile-time decision as the state spec had to create its own imports qualified by the instance name. This is very undesirable. *Mea culpa*, I am afraid.

@pcolarco FYI only.

@tclune For your assignment..

# The Solution

The idea is to make all the nice masking and expression parsing capabilities of ```ExtData``` available at the component level. It turns out, the solution is a generalization of ```MAPL_GetPointer()``` or ```MAPL_StateGetPointer()``` as it will be called in MAPL-3.

## Use case to fix ideas

Let's say there is component that imports 2 fields:
- `EMIS`
- `SOURCE`

that one would like to *transform* before using it in the component.

## The rc file

One would like to enable expression parsing capabilities entirely specified in the rc file. Consider this rc file fragment as an example:
```yaml
FILTER.EMIS:         mask_region(@,'1-3,~4-5',inclusive,REGION_MASK)
FILTER.SOURCE:       log(PS * @ + 0.001)
```

In both cases, ```@``` should be replace with the variable name.  For the ```mask_region()``` function, the second argument specifies the regions to be included/excluded; Ben may already have a convention to specify this, we should keep that. The last argument, ```REGION_MASK```, is a variable with the mask to be found inside the *ESMF_State*, read on.

The second resource, `log(PS * @ + 0.001)` shows the example of an arithmetic expression with a supported intrinsic function `log`. As before, variable `PS` should be found in the state.

## The Component Code

Typically, in the `run()` method of the component we have code like this to extract variables from the import state:
``` f90     
            REAL, pointer :: EMIS(:,:) 
            call MAPL_GetPointer(IMPORT,EMIS,  'EMIS', __RC__)
```

where ```EMIS``` ends up pointing to the data array of the field ```EMIS``` in the Import state. (In reality this kind of code is often automatically generated with the ACG.)  The new feature I am looking for is to replace the ```call MAPL_GetPointer()``` with
``` f90    
            REAL, allocatable:: EMIS(:,:) 
            EMIS =  MAPL_Filter(IMPORT,CF, 'EMIS',__RC__)
```

where `CF` is the `MAPL_Config` object associated with the component.  The `EMIS` field on import should remain unmodified. Function `MAPL_Filter()`  should consult the CF to see if any filter operation is being requested. If no such resource is found it simply returns `EMIS` by default (see replacing the default default below).  If a filter operation is being requested, it evaluates the expression and returns a transformed version of `EMIS`. All rc file driven, during run-time.

## Replacing the default default

When a filter rule is omitted in the rc file, by default a copy of the original variable in the state is returned.  However,  one should be able to have this default behavior modified. Consider
```yaml
FILTER.@:  log(@ + 0.001) # this is the user specified default. 
````
This filter rule will be used for all variables when a specific rule is not provided.  The implied default is this:
```yaml
FILTER.@:   @    # this is the implied default, you can omit this from the rc file.
```

## Automatic Code Generator (ACG) enhancement

The ACG will need to be extended to support this.  A column named `FILTER` could be used to indicate whether *filtering* should be enabled. In principle, this feature could be applied to the export state as well.

Notice that the local array `EMIS` is automatically deallocated upon exit.

## Other design considerations

We may need to have a discussion with Ben about his implementation of the expression parser and masking in `ExtData`, and iterate this design.

## Instance Qualified Imports

A *Qualified Import* field is one that appends the instance name to the variable name, much like is currently done in the current implementation of GOCART-2G. As discussed above, this is to become obsolete for simple masking or any transformation afforded by the expression parser. However, there remains the need for *Qualified Imports*.  For example, the active instance could run with the CEDS emissions, while one would like to experiment with the EDGAR emissions as a passive instance.  In this case, one needs to explicit indicate  that for that particular instance, `EMIS` needs to be qualified. One possible solution is to add this to the instance rc file:
```yaml
IMPORT.EMIS.QUALIFY:     yes
````

This is tone for the instance rc file, not the ACG state spec.) If this is present, then the import `EMIS` ought to be qualified with the instance name. Of course, the ACG would take care of that by adding some small code fragment to consult the CF. The qualified ```EMIS`` would then  be added manually by the user to the ```ExtData``` resource file. The important point is this all done at run-time, not compilation time.

### PR DESCRIPTION
**Pull request by bena-nasa**
Contingent PR, just making to get feedback before using any more time on it
Fixes #3634

## Types of change(s)

## Checklist

### DIFF
--- a/state/StateFilter.F90
+++ b/state/StateFilter.F90
@@ -1,5 +1,5 @@
 #include "MAPL_ErrLog.h"
-module MAPL_StateFilterMod
+module MAPL_StateFilter
    use ESMF
    use MAPL_ExceptionHandling
    use MAPL_FieldUtils
@@ -9,17 +9,17 @@ module MAPL_StateFilterMod
    implicit none
    private
 
-   public MAPL_StateFilterItem
+   public StateFilterItem
    character(len=1), parameter :: var_placeholder = "@"
 
-   interface MAPL_StateFilterItem
-      module procedure MAPL_StateFilter_R4_2D
-      module procedure MAPL_StateFilter_R4_3D
+   interface StateFilterItem
+      procedure StateFilter_R4_2D
+      procedure StateFilter_R4_3D
    end interface
 
    contains
 
-   subroutine MAPL_StateFilter_R4_2D(state, config, itemName, array, rc)
+   subroutine StateFilter_R4_2D(state, config, itemName, array, rc)
       type(ESMF_State), intent(inout) :: state
       type(ESMF_Config), intent(inout) :: config
       character(len=*), intent(in) :: itemName
@@ -73,9 +73,9 @@ module MAPL_StateFilterMod
       call ESMF_FieldDestroy(new_field, noGarbage=.true., _RC)
       _RETURN(_SUCCESS)
 
-   end subroutine MAPL_StateFilter_R4_2D
+   end subroutine StateFilter_R4_2D
 
-   subroutine MAPL_StateFilter_R4_3D(state, config, itemName, array, rc)
+   subroutine StateFilter_R4_3D(state, config, itemName, array, rc)
       type(ESMF_State), intent(inout) :: state
       type(ESMF_Config), intent(inout) :: config
       character(len=*), intent(in) :: itemName
@@ -129,7 +129,7 @@ module MAPL_StateFilterMod
       call ESMF_FieldDestroy(new_field, noGarbage=.true., _RC)
       _RETURN(_SUCCESS)
 
-   end subroutine MAPL_StateFilter_R4_3d
+   end subroutine StateFilter_R4_3d
 
 
    function substitute_name(filter_expression, field_name, rc) result(processed_expression)
@@ -148,4 +148,4 @@ module MAPL_StateFilterMod
       _RETURN(_SUCCESS)
    end function 
       
-end module MAPL_StateFilterMod
+end module MAPL_StateFilter
--- a/state/StateFilter.F90
+++ b/state/StateFilter.F90
@@ -11,6 +11,7 @@ module MAPL_StateFilter
 
    public StateFilterItem
    character(len=1), parameter :: var_placeholder = "@"
+   character(len=1), parameter :: separator = "."
 
    interface StateFilterItem
       procedure StateFilter_R4_2D
@@ -44,13 +45,13 @@ module MAPL_StateFilter
       allocate(array( lbound(ptr2d_old,1):ubound(ptr2d_old,1) , lbound(ptr2d_old,2):ubound(ptr2d_old,2) ),  _STAT) 
       array = ptr2d_old
  
-      call ESMF_ConfigFindLabel(config, "FILTER%"//trim(itemName)//":", isPresent=name_Present, _RC)
+      call ESMF_ConfigFindLabel(config, "FILTER"//separator//trim(itemName)//":", isPresent=name_Present, _RC)
       if (name_Present) then
-         call ESMF_ConfigGetAttribute(config, filter_expression, label="FILTER%"//trim(itemName)//":", _RC)
+         call ESMF_ConfigGetAttribute(config, filter_expression, label="FILTER"//separator//trim(itemName)//":", _RC)
       else
-         call ESMF_ConfigFindLabel(config, "FILTER%*:", isPresent=default_Present, _RC)
+         call ESMF_ConfigFindLabel(config, "FILTER"//separator//"*:", isPresent=default_Present, _RC)
          _RETURN_UNLESS(default_present)
-         call ESMF_ConfigGetAttribute(config, filter_expression, label="FILTER%*:", _RC)
+         call ESMF_ConfigGetAttribute(config, filter_expression, label="FILTER"//separator//"*:", _RC)
       end if
 
       call FieldClone(old_field, new_field, _RC)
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: state/tests/Test_StateFilter.pf [hunk 1]
```diff
@@ -0,0 +1,218 @@
...
+      call ESMF_FieldGet(extra_2d, 0, farrayPtr=extra_ptr, _RC)
+      call ESMF_FieldGet(field_2d, 0, farrayPtr=ptr2d, _RC)
+      rval = 17.0 
+      ptr2d = rval
+      extra_ptr = 2.0*rval
+
+      rval=3.0*rval
+
+      allocate(result_array(3,3),_STAT)
+      result_array= reshape([rval, rval, rval, rval, rval, rval, rval, rval, rval],shape=[3,3])
```
tclune, line 95:
```suggestion
      result_array = rval
```

state/tests/Test_StateFilter.pf
```diff
@@ -0,0 +1,218 @@
...
+
+      hcf = ESMF_HConfigCreate(content='{FILTER%foo: "@+extra_2d"}', _RC)
+      cf = ESMF_ConfigCreate(hconfig=hcf, _RC)
+
+      call ESMF_FieldGet(field_2d, 0, farrayPtr=ptr2d, _RC)
+      rval = 17.0 
+      ptr2d = rval
+
+      allocate(result_array(3,3),_STAT)
+      result_array= reshape([rval, rval, rval, rval, rval, rval, rval, rval, rval],shape=[3,3])
```
tclune, line 122:
```suggestion
      result_array= rval
```

state/tests/Test_StateFilter.pf
```diff
@@ -0,0 +1,218 @@
...
+      call ESMF_FieldGet(extra_3d, 0, farrayPtr=extra_ptr, _RC)
+      call ESMF_FieldGet(field_3d, 0, farrayPtr=ptr3d, _RC)
+      rval = 17.0 
+      ptr3d = rval
+      extra_ptr = 2.0*rval
+
+      rval=3.0*rval
+
+      allocate(result_array(3,3,2),_STAT)
+      result_array=rval
```
tclune, line 211: Ah - you did it the easy way when it got hard enough!

state/tests/Test_StateMask.pf
```diff
@@ -0,0 +1,93 @@
...
+      class(MpiTestMethod), intent(inout) :: this
+      call ESMF_FieldDestroy(field_2d, noGarbage=.true.)
+      call ESMF_FieldDestroy(field_3d, noGarbage=.true.)
+      call ESMF_FieldDestroy(mask_field, noGarbage=.true.)
+      call ESMF_StateDestroy(state, noGarbage=.true.)
+   end subroutine teardown
+
+   @Test(npes=[1])
+   subroutine test_region_mask_2d(this)
+      class(MpiTestMethod), intent(inout) :: this
```
tclune, line 40:
```suggestion
      class(ESMF_TestMethod), intent(inout) :: this
```
Needed on all these tests.

state/tests/Test_StateMask.pf
```diff
@@ -0,0 +1,93 @@
...
+   @after
+   subroutine teardown(this)
+      class(MpiTestMethod), intent(inout) :: this
+      call ESMF_FieldDestroy(field_2d, noGarbage=.true.)
+      call ESMF_FieldDestroy(field_3d, noGarbage=.true.)
+      call ESMF_FieldDestroy(mask_field, noGarbage=.true.)
+      call ESMF_StateDestroy(state, noGarbage=.true.)
+   end subroutine teardown
+
+   @Test(npes=[1])
```
tclune, line 38: I think you need to make all of these ESMF_TestMethods to be safe.    If someone runs this suite on 4 cores, ESMF will try to spread these things across multiple processors.

```suggestion
   @test(type=ESMF_TestMethod, npes=[1])
```

state/tests/Test_StateFilter.pf
```diff
@@ -0,0 +1,218 @@
...
+
+      hcf = ESMF_HConfigCreate(content='{FILTER%field_2d: "regionmask(@,region_mask;2,5)"}', _RC)
+      cf = ESMF_ConfigCreate(hconfig=hcf, _RC)
+
+      call ESMF_FieldGet(mask_field, 0, farrayPtr=mask_ptr, _RC)
+      call ESMF_FieldGet(field_2d, 0, farrayPtr=ptr2d, _RC)
+      rval = 17.0 
+      ptr2d = rval
+      allocate(result_array(3,3),_STAT)
+      result_array= reshape([0.0, rval, 0.0, rval, rval, 0.0, rval, rval, 0.0],shape=[3,3])
```
tclune, line 62: I suggest renaming `result_array` to `expected_array`.   "result" is a bit ambiguous in this context.

state/StateFilter.F90
```diff
@@ -0,0 +1,151 @@
...
+   use, intrinsic :: iso_fortran_env, only: REAL32, REAL64
+   implicit none
+   private
+
+   public MAPL_StateFilterItem
+   character(len=1), parameter :: var_placeholder = "@"
+
+   interface MAPL_StateFilterItem
+      module procedure MAPL_StateFilter_R4_2D
+      module procedure MAPL_StateFilter_R4_3D
```
tclune, line 17:
```suggestion
      procedure filter_r4_2d
      procedure filter_r4_3d
```

state/StateFilter.F90
```diff
@@ -0,0 +1,151 @@
...
+   end interface
+
+   contains
+
+   subroutine MAPL_StateFilter_R4_2D(state, config, itemName, array, rc)
```
tclune, line 22:
```suggestion
   subroutine StateFilter_R4_2D(state, config, itemName, array, rc)
```

state/StateFilter.F90
```diff
@@ -0,0 +1,151 @@
...
+   use ESMF
+   use MAPL_ExceptionHandling
+   use MAPL_FieldUtils
+   use MAPL_StateArithmeticParserMod
+   use MAPL_StateMaskMod
+   use, intrinsic :: iso_fortran_env, only: REAL32, REAL64
+   implicit none
+   private
+
+   public MAPL_StateFilterItem
```
tclune, line 12:
```suggestion
   public StateFilterItem
```

state/StateFilter.F90
```diff
@@ -0,0 +1,151 @@
+#include "MAPL_ErrLog.h"
+module MAPL_StateFilterMod
```
tclune, line 2: Drop the `Mod` suffix in anticipation for merging with MAPL3

```suggestion
module MAPL_StateFilter
```

state/StateFilter.F90
```diff
@@ -0,0 +1,151 @@
...
+      call ESMF_ConfigFindLabel(config, "FILTER%"//trim(itemName)//":", isPresent=name_Present, _RC)
+      if (name_Present) then
+         call ESMF_ConfigGetAttribute(config, filter_expression, label="FILTER%"//trim(itemName)//":", _RC)
+      else
+         call ESMF_ConfigFindLabel(config, "FILTER%*:", isPresent=default_Present, _RC)
+         if (default_present) then
+            call ESMF_ConfigGetAttribute(config, filter_expression, label="FILTER%*:", _RC)
+         else
+            _RETURN(_SUCCESS)
+         end if
```
tclune, line 56:
```suggestion
         _RETURN_UNLESS(default_present)
         call ESMF_ConfigGetAttribute(config, filter_expression, label="FILTER%*:", _RC)
```

tclune: I realize some of this is probably just pasted code from elsewhere.   I'll try to refrain from doing too much of this.  But ... it really jumped out at me.

tclune (changes requested):
Mostly suggestions to minimize conflicts with MAPL3 style.

A few other simplifications suggested.

bena-nasa:
@tclune 
I took all your suggestions.  There are still some unit tests I want to add so will add those before removing the contingent label.

tclune (approved):

tclune (approved):

mathomp4:
I need to fix up the CI in MAPL. One moment...

tclune (approved):

Issue #3587, opened 2025-04-09
Create ifx reproducer


tclune:
generic3g tests currently fail with ifx on the following line in Test_Scenarios.pf:

```f90
      params = [params, add_params('item exist', check_item_type)]
```

Unfortunately, a trivial attempt at a reproducer does not show the problem.   Likely one has to include type extension from pFUnit and/or the use of a procedure pointer in the derived type.

weiyuan-jiang:
My tests succeed.  Start 11: MAPL.generic.tests
 4/48 Test #11: MAPL.generic.tests ...............   Passed    1.22 sec
      Start 12: MAPL.generic3g.tests
 5/48 Test #12: MAPL.generic3g.tests .............   Passed    3.18 sec

weiyuan-jiang:
I did get a  failure. I think the accuracy is not consistent somewhere.

Failure
 in: 
Test_LonAxis_suite.test_make_lon_axis_from_hconfig
  Location: 
[Test_LonAxis.pf:179]

Expected: is [<-180.0000000000000D>, <-90.00000000000000D>, <.000000000000000D>, <90.00000000000000D>]
     but: was [<-180.0000000000000D>, <-90.00000000000000D>, <-.9992007221626409E-14D>, <90.00000000000000D>]
  
 FAILURES!!!

tclune:
No - that is not the failure being discussed in this ticket.   You should see a crash around:
```
MAPL.generic3g.te  0000000000645009  get_parameters             99  Test_Scenarios.F90
```

Can you confirm which compiler version you are using?

mathomp4:
@weiyuan-jiang You probably need to be on Bucy for this. Discover doesn't have a new enough `ifx`. 

When you are on bucy, you can do:
```
ml ifx-stack
```
and that will get you ifx 2025.0:

```
> ifx --version
ifx (IFX) 2025.0.4 20241205
Copyright (C) 1985-2024 Intel Corporation. All rights reserved.
```

mathomp4:
Note: There is a newer `ifx` that I guess our CI is using. I'll look at getting that on Bucy in case that is the version with an issue.

weiyuan-jiang:
On discover: 
ifx -v
ifx version 2024.2.0

mathomp4:
Yeah. I think we need at least ifx 2025.0 to get past some issues with their compiler. So bucy (or NAS) are the only places with it at the moment.

[...]

PR #3670, opened 2025-05-01
assign default values to pass ifx 2025.1 tests

weiyuan-jiang:
## Types of change(s)

## Checklist
- [x ] Ran the Unit Tests (`make tests`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Workaround to pass ifx 2025.1 tests in debug mode
 - Profile reporting has been relocated into the `./profile` directory.
 - Improved diagnostic message for profiler imbalances at end of run.
   Now gives the name of the timer that has not been stopped when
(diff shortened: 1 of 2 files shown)


weiyuan-jiang:
Not a re-producer, but avoid the issue #3587

tclune (commented):
I want @darianboggs to examine this change.

The `n-3` has always been a bit confusing here.   But more importantly - the tests pass on other versions of Intel (and GNU, and NAG).   I want to know whether this is a subtle bug in our code (or test) or a compiler bug.

Of course, anything that makes the test more intuitive would also be welcome.

weiyuan-jiang:
> I want @darianboggs to examine this change.
> 
> The `n-3` has always been a bit confusing here. But more importantly - the tests pass on other versions of Intel (and GNU, and NAG). I want to know whether this is a subtle bug in our code (or test) or a compiler bug.
> 
> Of course, anything that makes the test more intuitive would also be welcome.

Here n=16. So the field (1:8) is not initialized. I don't know why ifx suddenly becomes picky

tclune:
> I don't know why ifx suddenly becomes picky

I'd prefer that all the compilers were so picky.   Detecting uninitialized thing is a major challenge for us.

mathomp4:
One thing. I just realized that the ifx docker image is using ifx 2025.0

We have seen that ifx 2025.1 gives different behavior. I might look at tomorrow building some ifx 2025.1 images. (I say tomorrow because getting new intel compilers is always "fun" and it takes a while to figure out the right magic to automate it.)

darianboggs (approved):
Good

PR #3691, opened 2025-05-07
Bugfix/wdboggs/add restart to acg3

darianboggs:
## Types of change(s)

## Checklist

## Description
ACG3 was not handling the `RESTART` column in the specs file. In current MAPL (MAPL2), `RESTART` allowed values from a finite set of constant values. In MAPL3, `RESTART` is a FORTRAN logical.


tclune (approved):

Apps/MAPL_GridCompSpecs_ACGv3.py
```diff
@@ -150,6 +150,10 @@ def get_options(args):
         'itemtype': {}, 
         'orientation': {}, 
         'regrid_method': {}, 
+        'restart': {MAPPING: dict(
```
pchakraborty, line 153: Does this cover the case where the `RESTART` value is set to `SKIP`? There are several import variables in GWD that look like
```
 PLE     |       | Pa          | xyz  | E    | SKIP    | air_pressure
 T       |       | K           | xyz  | C    | SKIP    | air_temperature
 Q       |       | kg kg-1     | xyz  | C    | SKIP    | specific_humidity
 U       |       | m s-1       | xyz  | C    | SKIP    | eastward_wind
 V       |       | m s-1       | xyz  | C    | SKIP    | northward_wind
 PHIS    |       | m+2 s-2     | xy   | N    | SKIP    | surface geopotential height
 SGH     |       | m           | xy   | N    | SKIP    | standard_deviation_of_topography
 VARFLT  |       | m+2         | xy   | N    | SKIP    | variance_of_the_filtered_topography
 PREF    |       | Pa          |   z  | E    | SKIP    | reference_air_pressure
 AREA    |       | m^2         | xy   | N    | SKIP    | grid_box_area
```

pchakraborty (commented):
Please see inline comment.

Apps/MAPL_GridCompSpecs_ACGv3.py
pchakraborty: Question for @tclune - currently `MAPL_GridCompAddFieldSpec` can accept a logical argument called `restart`, but then it ignores it since `make_VariableSpec` does not have any such option. We probably need to update `VariableSpec`.

tclune: This needs some thought.  Have been purposefully avoiding a decision about some problematic aspects of how we manage restarts.   At some point, I think it will be possible for MAPL itself to figure out which items are needed in the restart.

darianboggs: In `gridcomp_add_fieldspec` (`MAPL_GridCompAddFieldSpec`) in `generic3g/MAPL_Generic.F90` the restart argument is a Fortran logical. (If I am mistaken let me know.) In MAPL, the type is an enum value. Perhaps RESTART was a multi-value flag in MAPL, and in MAPL3 it is a simple logical flag.

tclune: At first glance,  in MAPL2 it is stored as an integer, but has only 2 legit values:
```
      if(present(RESTART)) then
         usableRESTART=RESTART
      else
         usableRESTART=MAPL_RestartOptional ! default
      endif
```

We need @atrayano to comment on what options were possibly envisioned.  Fodder for the Thu meeting.

PR #3698, opened 2025-05-08
initial check in

weiyuan-jiang:
## Types of change(s)

## Checklist

## Description
Add EASEGridFactory so the Regridder can use EASEGrid

### DIFF
--- a/base/MAPL_EASEGridFactory.F90
+++ b/base/MAPL_EASEGridFactory.F90
@@ -1,0 +1,1954 @@
+#include "MAPL_ErrLog.h"
+
+! overload set interfaces in legacy
+! Document PE, PC, DC, DE, GC
+
+! This module generates ESMF_Grids corresponding to _regular_ lat-lon coordinate grids.
+! I.e., spacing between lats (lons) is constant.
+
+module MAPL_EASEGridFactoryMod
+   use MAPL_AbstractGridFactoryMod
+   use MAPL_MinMaxMod
+   use MAPL_KeywordEnforcerMod
+   use mapl_ErrorHandlingMod
+   use MAPL_Constants
+   use ESMF
+   use pFIO
+   use MAPL_CommsMod
+   use, intrinsic :: iso_fortran_env, only: REAL32
+   use, intrinsic :: iso_fortran_env, only: REAL64
+   implicit none
+   private
+
+   public :: EASEGridFactory
+
+   integer, parameter :: NUM_DIM = 2
+
+   type, extends(AbstractGridFactory) :: EASEGridFactory
+      private
+      logical :: is_regular = .false.
+      character(len=:), allocatable :: grid_name
+      ! Grid dimensions
+      integer :: im_world = MAPL_UNDEFINED_INTEGER
+      integer :: jm_world = MAPL_UNDEFINED_INTEGER
+      integer :: lm = MAPL_UNDEFINED_INTEGER
+      real(kind=REAL64), allocatable :: lon_centers(:)
+      real(kind=REAL64), allocatable :: lat_centers(:)
+      real(kind=REAL64), allocatable :: lon_centers_degrees(:)
+      real(kind=REAL64), allocatable :: lat_centers_degrees(:)
+      real(kind=REAL64), allocatable :: lon_corners(:)
+      real(kind=REAL64), allocatable :: lat_corners(:)
+      logical :: force_decomposition = .false.
+
+      ! Domain decomposition:
+      integer :: nx = MAPL_UNDEFINED_INTEGER
+      integer :: ny = MAPL_UNDEFINED_INTEGER
+      integer, allocatable :: ims(:)
+      integer, allocatable :: jms(:)
+      ! Grid conventions:
+      character(len=:), allocatable :: pole
+      character(len=:), allocatable :: dateline
+      ! Regional vs global:
+      type (RealMinMax) :: lon_range = RealMinMax(MAPL_UNDEFINED_REAL,MAPL_UNDEFINED_REAL)
+      type (RealMinMax) :: lat_range = RealMinMax(MAPL_UNDEFINED_REAL,MAPL_UNDEFINED_REAL)
+      ! Used for halo
+      type (ESMF_DELayout) :: layout
+      integer :: px, py
+      logical :: is_halo_initialized = .false.
+      logical :: periodic = .true.
+      character(len=:), allocatable :: lon_bounds_name
+      character(len=:), allocatable :: lat_bounds_name
+   contains
+      procedure :: make_new_grid
+      procedure :: create_basic_grid
+      procedure :: get_longitudes_degrees
+      procedure :: get_longitudes
+      procedure :: get_latitudes_degrees
+      procedure :: get_latitudes
+      procedure :: compute_lon_centers
+      procedure :: compute_lat_centers
+      procedure :: get_lon_corners
+      procedure :: get_lat_corners
+      procedure :: compute_lon_corners
+      procedure :: compute_lat_corners
+      procedure :: add_horz_coordinates
+      procedure :: init_halo
+      procedure :: halo
+
+
+      procedure :: initialize_from_file_metadata
+      procedure :: initialize_from_config_with_prefix
+      procedure :: initialize_from_esmf_distGrid
+
+      procedure :: equals
+
+      procedure :: check_and_fill_consistency
+      procedure :: generate_grid_name
+      procedure :: to_string
+
+      procedure :: append_metadata
+      procedure :: get_grid_vars
+      procedure :: get_file_format_vars
+      procedure :: append_variable_metadata
+      procedure :: check_decomposition
+      procedure :: generate_newnxy
+      procedure :: generate_file_bounds
+      procedure :: generate_file_corner_bounds
+      procedure :: generate_file_reference2D
+      procedure :: generate_file_reference3D
+      procedure :: decomps_are_equal
+      procedure :: physical_params_are_equal
+   end type EASEGridFactory
+
+   character(len=*), parameter :: MOD_NAME = 'MAPL_EASEGridFactory::'
+
+   interface EASEGridFactory
+      module procedure EASEGridFactory_from_parameters
+   end interface EASEGridFactory
+
+   interface set_with_default
+      module procedure set_with_default_integer
+      module procedure set_with_default_real
+      module procedure set_with_default_character
+      module procedure set_with_default_range
+      module procedure set_with_default_logical
+   end interface set_with_default
+
+
+contains
+
+
+   function EASEGridFactory_from_parameters(unusable, grid_name, &
+        & im_world, jm_world, lm, nx, ny, ims, jms, &
+        & pole, dateline, lon_range, lat_range, force_decomposition, rc) result(factory)
+      type (EASEGridFactory) :: factory
+      class (KeywordEnforcer), optional, intent(in) :: unusable
+      character(len=*), optional, intent(in) :: grid_name
+
+      ! grid details:
+      integer, optional, intent(in) :: im_world
+      integer, optional, intent(in) :: jm_world
+      integer, optional, intent(in) :: lm
+      character(len=2), optional, intent(in) :: pole
+      character(len=2), optional, intent(in) :: dateline
+      type (RealMinMax), optional, intent(in) :: lon_range
+      type (RealMinMax), optional, intent(in) :: lat_range
+
+      ! decomposition:
+      integer, optional, intent(in) :: nx
+      integer, optional, intent(in) :: ny
+      integer, optional, intent(in) :: ims(:)
+      integer, optional, intent(in) :: jms(:)
+      logical, optional, intent(in) :: force_decomposition
+
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+
+      _UNUSED_DUMMY(unusable)
+
+      factory%is_regular = .true.
+      call set_with_default(factory%grid_name, grid_name, MAPL_GRID_NAME_DEFAULT)
+
+      call set_with_default(factory%nx, nx, MAPL_UNDEFINED_INTEGER)
+      call set_with_default(factory%ny, ny, MAPL_UNDEFINED_INTEGER)
+
+      call set_with_default(factory%im_world, im_world, MAPL_UNDEFINED_INTEGER)
+      call set_with_default(factory%jm_world, jm_world, MAPL_UNDEFINED_INTEGER)
+      call set_with_default(factory%lm, lm, MAPL_UNDEFINED_INTEGER)
+
+      ! default is unallocated
+      if (present(ims)) factory%ims = ims
+      if (present(jms)) factory%jms = jms
+
+      call set_with_default(factory%pole, pole, MAPL_UNDEFINED_CHAR)
+      call set_with_default(factory%dateline, dateline, MAPL_UNDEFINED_CHAR)
+
+      call set_with_default(factory%lon_range, lon_range, RealMinMax(MAPL_UNDEFINED_REAL,MAPL_UNDEFINED_REAL))
+      call set_with_default(factory%lat_range, lat_range, RealMinMax(MAPL_UNDEFINED_REAL,MAPL_UNDEFINED_REAL))
+      call set_with_default(factory%force_decomposition, force_decomposition, .false.)
+
+      call factory%check_and_fill_consistency(_RC)
+
+      ! Compute the centers and corners
+      factory%lon_centers = factory%compute_lon_centers(factory%dateline, _RC)
+      factory%lat_centers = factory%compute_lat_centers(factory%pole, _RC)
+      factory%lon_centers_degrees = factory%compute_lon_centers(factory%dateline, &
+            convert_to_radians = .false.,  _RC)
+      factory%lat_centers_degrees = factory%compute_lat_centers(factory%pole, &
+            convert_to_radians = .false.,  _RC)
+      factory%lon_corners = factory%compute_lon_corners(factory%dateline, _RC)
+      factory%lat_corners = factory%compute_lat_corners(factory%pole, _RC)
+
+      _RETURN(_SUCCESS)
+
+   end function EASEGridFactory_from_parameters
+
+
+   function make_new_grid(this, unusable, rc) result(grid)
+      type (ESMF_Grid) :: grid
+      class (EASEGridFactory), intent(in) :: this
+      class (KeywordEnforcer), optional, intent(in) :: unusable
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+
+      _UNUSED_DUMMY(unusable)
+      grid = this%create_basic_grid(_RC)
+
+      call this%add_horz_coordinates(grid, _RC)
+
+      _RETURN(_SUCCESS)
+
+   end function make_new_grid
+
+
+
+   function create_basic_grid(this, unusable, rc) result(grid)
+      type (ESMF_Grid) :: grid
+      class (EASEGridFactory), intent(in) :: this
+      class (KeywordEnforcer), optional, intent(in) :: unusable
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+      type(ESMF_PoleKind_Flag) :: polekindflag(2)
+
+      _UNUSED_DUMMY(unusable)
+
+      if (this%periodic) then
+         if (this%pole == "XY") then 
+            polekindflag = ESMF_POLEKIND_NONE
+         else
+            polekindflag = ESMF_POLEKIND_MONOPOLE
+         end if
+         grid = ESMF_GridCreate1PeriDim( &
+              & name = this%grid_name, &
+              & countsPerDEDim1=this%ims, &
+              & countsPerDEDim2=this%jms, &
+              & indexFlag=ESMF_INDEX_DELOCAL, &
+              & gridEdgeLWidth=[0,0], &
+              & gridEdgeUWidth=[0,1], &
+              & coordDep1=[1,2], &
+              & coordDep2=[1,2], &
+              & coordSys=ESMF_COORDSYS_SPH_RAD, &
+              & polekindflag=polekindflag, &
+              & _RC)
+      else
+         grid = ESMF_GridCreateNoPeriDim( &
+              & name = this%grid_name, &
+              & countsPerDEDim1=this%ims, &
+              & countsPerDEDim2=this%jms, &
+              & indexFlag=ESMF_INDEX_DELOCAL, &
+              & gridEdgeLWidth=[0,0], &
+              & gridEdgeUWidth=[1,1], &
+              & coordDep1=[1,2], &
+              & coordDep2=[1,2], &
+              & coordSys=ESMF_COORDSYS_SPH_RAD, &
+              & _RC)
+      end if
+
+      ! Allocate coords at default stagger location
+      call ESMF_GridAddCoord(grid, _RC)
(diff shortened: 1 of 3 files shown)


mathomp4:
@weiyuan-jiang It looks like there is some ambiguity in some GEOS code:
```
[ 83%] Building Fortran object src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/CMakeFiles/mkEASETilesParam.x.dir/mkEASETilesParam.F90.o
/root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkCatchParam.F90:193:23:

  193 |       call ease_extent(Gridname, nc_ease, nr_ease )
      |                       1
Error: Name 'ease_extent' at (1) is an ambiguous reference to 'ease_extent' from module 'mapl_easeconvmod'
make[2]: *** [src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/CMakeFiles/mkCatchParam.x.dir/build.make:75: src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/CMakeFiles/mkCatchParam.x.dir/mkCatchParam.F90.o] Error 1
```

Will this MAPL PR need to go with a GEOSgcm_GridComp PR?

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
+
+module MAPL_EASEConvMod
+  
+  ! =====================================================================================
+  !
+  ! EASE_conv.F90 - FORTRAN routines for conversion of Equal-Area Scalable Earth (EASE)
```
tclune, line 6:
```suggestion
  ! EASE_conv.F90 - Fortran routines for conversion of Equal-Area Scalable Earth (EASE)
```

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
...
+  !                 Implemented for global cylindrical ('Mxx') EASE grids only. 
+  !
+  !                 Works for EASE[v1] and EASEv2 grids. 
+  ! 
+  ! -------------------------------------------------------------------------------------
+  ! 
+  ! CHANGELOG (easeV1_conv.F90):
+  ! ============================
+  !
+  ! easeV1_conv.F90 - FORTRAN routines for conversion of azimuthal
```
tclune, line 18:
```suggestion
  ! easeV1_conv.F90 - Fortran routines for conversion of azimuthal 
```

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
...
+
+  ! =======================================================================
+  !
+  ! EASEv1 global constants
+
+  ! ***NEVER*** change these constants to GEOS MAPL constants!!!!
+    
+  ! radius of the earth (km), authalic sphere based on International datum 
+  
+  real*8, parameter :: easeV1_RE_km                    = 6371.228
```
tclune, line 97: `real*4` and `real*8` not permitted in MAPL.

```suggestion
  real(kind=REAL64), parameter :: easeV1_RE_km                    = 6371.228
```

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
+
+module MAPL_EASEConvMod
```
tclune, line 2: We have a convention for module names and file names.  Be sure to follow it.  no `Mod`

I also suggest expanding the abbreviation `Conv`: electrons are cheap.
```suggestion
module MAPL_EASEConversion
```

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
...
+       grid='M25'
+    else if (index(EASELabel,'M09') /=0 ) then
+       grid='M09'
+    else if (index(EASELabel,'M03') /=0 ) then
+       grid='M03'
+    else if (index(EASELabel,'M01') /=0 ) then
+       grid='M01'
+    else
+       print*,"ease_convert(): unknown grid projection and resolution: "//trim(EASELabel)//"  STOPPING."
+       stop
```
tclune, line 189: Use error handling.

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
...
+    character*(*), intent(in)  :: grid
+    real,          intent(in)  :: lat, lon
+    real,          intent(out) :: r, s
+
+    ! local variables
+    
+    integer :: cols, rows
+    real*8  :: Rg, phi, lam, rho, CELL_km, r0, s0
+    
+    real*8, parameter :: PI = easeV1_PI
```
tclune, line 343: How many values of pi does mapl need?

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
...
+    else if (grid(1:1).eq.'S') then
+       rho = 2 * Rg * cos(PI/4. - phi/2.)
+       r = r0 + rho * sin(lam)
+       s = s0 - rho * cos(lam)
+       
+    else if (grid(1:1).eq.'M') then
+       r = r0 + Rg * lam * easeV1_COS_PHI1
+       s = s0 - Rg * sin(phi) / easeV1_COS_PHI1
+       
+    endif
```
tclune, line 366: Missing default clause?

base/MAPL_EASEConv.F90
weiyuan-jiang: I have changed to MAPL's PI and all the REAL64  parameters' definition.  However,  I believe the change will make it non-zero diff, especially in BCS. I may have to change it back to the 'wrong' way for zero-diff.

base/MAPL_EASEConv.F90
```diff
@@ -0,0 +1,818 @@
...
+  
+  real*8, parameter :: N_map_reference_latitude        =  90.0  
+  real*8, parameter :: S_map_reference_latitude        = -90.0
+  
+  ! constants for 'M' (cylindrical) projection
+  
+  real*8, parameter :: M_map_reference_latitude        =   0.0
+  real*8, parameter :: M_map_second_reference_latitude =  30.0
+  
+  real*8, parameter :: M_sin_phi1 = sin(M_map_second_reference_latitude*easeV2_PI/180.)
```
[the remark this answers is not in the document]
weiyuan-jiang: I may make lead to non-zero diff

weiyuan-jiang:
@gmao-rreichle , I have moved the file EASE_Conv.F90 from makebcs to here ( MAPL_EASEConversion.F90). I also changed the parameters to a real double ( ie. add d0 at the end of a number)  and MAPL's PI is also used. But it may lead to non-zero-diff in BCS. Do you think it is worthy to make such changes? I can change back to the original 'wrong' way.

https://github.com/GEOS-ESM/MAPL/blob/60098729d5ab75fdf93ae0eb501b6487602a3fa6/base/MAPL_EASEConversion.F90#L101-L141

base/MAPL_EASEConv.F90
tclune: For the parameters, we have found that with optimization on, there usually is no difference.    The compiler apparently sees the advantage of not truncating just to promote.  But ... no guarantees.

For PI - if there is a difference that would be bad.   I hope not.

If this is a problem for @gmao-rreichle, we can relax this requirement for this PR and just add it to the list of things to fix in MAPL3.

base/MAPL_EASEConversion.F90
```diff
@@ -0,0 +1,828 @@
...
+  ! 2022-09-13, wjiang+reichle: 
+  !   merged easeV1_conv.F90 and easeV2_conv.F90 into EASE_conv.F90
+  !   - using different values for PI in easeV1 and easeV2 calcs as in old easeV[x]_conv.F90 modules; 
+  !       in contrast, LDAS_EASE_conv.F90 in GEOSldas used only a single value for PI.
+  !   - bug fix in easeV2_get_params() for EASEv2/M25 (to compute s0, divide by 2.0 not by integer 2)
+  !
+  !
+  ! ==========================================================================
+  use, intrinsic :: iso_fortran_env, only: REAL64
+  use MAPL_Constants, only: PI => MAPL_PI_r8
```
gmao-rreichle, line 81: We have to keep the v1- and v2-specific PI values.  Otherwise, we will lose backward compatibility for creating bcs, and we would also have lat/lons in published SMAP L4 products that aren't consistent with SMAP L1-3 products published by other institutions.

base/MAPL_EASEConversion.F90
```diff
@@ -0,0 +1,828 @@
...
+    else
+       _FAIL("ease_extent(): unknown grid version: "//trim(EASELabel)//"  STOPPING.")
+    endif
+
+    ! get lat/lon of corner grid cells
+    ! 
+    ! recall that EASE grid indexing is zero-based
+
+    if (present(ll_lat))  call ease_inverse(EASElabel, 0., rows-0.5, ll_lat, tmplon, _RC)     
+    if (present(ur_lat))  call ease_inverse(EASElabel, cols-1.0,     -0.5, ur_lat, tmplon, _RC)
```
gmao-rreichle, line 302: These two calls **only** get the **latitude** values.  I think we should stick with using 0. for the longitude index in both cases, as in the original code from EASE_conv.F90:
```
    if (present(ll_lat))  call ease_inverse(EASElabel, 0., rows-0.5, ll_lat, tmplon)     
    if (present(ur_lat))  call ease_inverse(EASElabel, 0.,     -0.5, ur_lat, tmplon)     
```
The **lon** values are set in the next two lines.

base/MAPL_EASEConversion.F90
```diff
@@ -0,0 +1,828 @@
...
+
+    ! get lat/lon of corner grid cells
+    ! 
+    ! recall that EASE grid indexing is zero-based
+
+    if (present(ll_lat))  call ease_inverse(EASElabel, 0., rows-0.5, ll_lat, tmplon, _RC)     
+    if (present(ur_lat))  call ease_inverse(EASElabel, cols-1.0,     -0.5, ur_lat, tmplon, _RC)     
+    
+    if (present(ll_lon))  ll_lon = -180.
+    if (present(ur_lon))  ur_lon =  180.
```
gmao-rreichle, line 305: The ur and ll values set here are for the "**global**" cylindrical EASE grid.  (I'm putting "global" into parentheses because the EASE min/max lat is near 85S/N.)  That is, the subroutine ease_extent does not work for "regional" grids.  There is some code related to "regional" grids in MAPL_EASEGridFactory.F90.  Is this cut-and-paste left over from the lat/lon grid that is not meant to work, or is the "regional" grid meant to work for the cylindrical EASE grids?  If it's meant to work, are we sure that subroutine ease_extent() is ok as is?

base/MAPL_EASEGridFactory.F90
```diff
@@ -0,0 +1,1866 @@
...
+      _UNUSED_DUMMY(unusable)
+
+      if (present(force_file_coordinates)) then
+         use_file_coords = force_file_Coordinates
+      else
+         use_file_coords = .false.
+      end if
+
+      ! Cannot assume that lats and lons are evenly spaced
+      this%is_regular = .false.
```
gmao-rreichle, line 631: "is_regular" may be misleading for the EASE grid.  In my understanding, the EASE grid is a "regular" grid.  It does not have even lat spacing, but it's still a regular grid (as opposed to, say, the cube-sphere tile space).    Maybe "is_evenspaced" (or something to that effect) would be a less confusing variable name.

base/MAPL_EASEConversion.F90
```diff
@@ -0,0 +1,828 @@
...
+       
+       _FAIL('easeV2_get_params(): unknown projection: '// grid)
+       
+    endif
+
+    _RETURN(_SUCCESS)       
+
+  end subroutine easeV2_get_params
+
+  function ease_grid_name(cols, rc) result(name)
```
gmao-rreichle, line 788: The function "ease_grid_name" is new and was not in the original EASE_conv.F90 module. I have a couple of concerns re. this new function:
1. The original design calls for knowing the grid name, and then subroutine "ease_extent" can be used to determine the number of columns and rows.  If we know the number of cols, we must already know the grid name (assuming we have dealt with the legacy "SMAP-EASEv[x]-M[xx]" grid name -- see my other comment).  So it doesn't make sense to me that we even need a function that converts the number of cols into the grid name. 
2. Hard-coding the relationship between the number of cols and the grid name here duplicates what's already hard-coded into ease_extent.  We shouldn't hard-code these things in two different places.
3. "ease_grid_name" requires that each EASE grid has a unique number of cols, which is ok with the handful of grids defined here but can't be generally assumed.

base/NCIO.F90
```diff
@@ -5569,6 +5571,11 @@ subroutine MAPL_ReadTilingASCII(layout, FileName, GridName, NT, im, jm, n_Grids,
           allocate(AVR(NT,9), STAT=STATUS) ! 9 columns for EASE grid
           _VERIFY(STATUS)
           allocate(AVR_transpose(9,NT))
+          ! WY note: In the old tile file, 
+          ! EASE grid name is in the form of SMAP-EASEvx-Mxx
+          ! It is changed to a standard form as EASEvx_Mxx
+          Correct_ease_name = ease_grid_name(IM(1))
+          GridNAME(1) = Correct_ease_name
```
gmao-rreichle, line 5578: I'm concerned about going through the number of cols to identify the grid.  See my comment near Line ~788 of MAPL_EASEConversion.F90.  Here, if we encounter the old naming convention ("SMAP-EASE..."),  it would be simple to directly convert the string SMAP-EASEv[x]-M[xx] into EASEv[x]_M[xx].

base/MAPL_EASEGridFactory.F90
```diff
@@ -0,0 +1,1866 @@
...
+            lon_name = 'longitude'
+            hasLongitude = file_metadata%has_dimension(lon_name)
+            if (hasLongitude) then
+               im = file_metadata%get_dimension(lon_name, _RC)
+            else
+               _FAIL('no longitude coordinate')
+            end if
+         end if
+
+         grid_name = ease_grid_name(im)
```
gmao-rreichle, line 648: I'm not sure why we need the EASE grid name here, and how it could be obtained without going through the number of cols.  Perhaps this is unavoidable, but I still like to think that we can obtain the name without going through the number of columns.  Same for the other "initialize_from...()" subroutine below.

gmao-rreichle (commented):
@weiyuan-jiang, @tclune : Thanks for the PR!  To answer your question about PI -- we absolutely can **not** change the original PI values, sorry.  See below for additional inline comments.

tclune:
> @weiyuan-jiang, @tclune : Thanks for the PR! To answer your question about PI -- we absolutely can **not** change the original PI values, sorry. See below for additional inline comments.

@gmao-rreichle  We should at least consider the forward path - are you locked into "wrong" values permanently?  E.g., what if we had to stop using Intel compiler in favor of say NVIDIA in the future?

base/MAPL_EASEConv.F90
tclune: @gmao-rreichle For PI - yes I don't really have an argument other than it is what it is.   But for the other parameters, I suspect you are getting different values with debug vs non-debug.   I've not looked closely though.  If all of the literals are real integers (e.g., 123.0) then it may not matter.    But ... in the past, MAPL had things like:
```
real(kind=REAL64), parameter :: pi = 3.14159265...
```
And the value depended on optimization until we fixed it as:
```
real(kind=REAL64), parameter :: pi = 3.14159265...d+0
```

base/MAPL_EASEGridFactory.F90
```diff
@@ -0,0 +1,1866 @@
...
+            max_coord = +180.d0 - delta
+         case ('DE')
+            min_coord = -180.d0 + delta/2
+            max_coord = +180.d0 - delta/2
+         case ('GC')
+            min_coord = 0.d0
+            max_coord = 360.d0 - delta
+         case ('GE')
+            min_coord = delta/2
+            max_coord = 360.d0 - delta/2
```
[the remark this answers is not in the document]
weiyuan-jiang: cleaned

base/MAPL_EASEGridFactory.F90
```diff
@@ -0,0 +1,1866 @@
+#include "MAPL_ErrLog.h"
+
+! overload set interfaces in legacy
+! Document PE, PC, DC, DE, GC
+
+! This module generates ESMF_Grids corresponding to _regular_ lat-lon coordinate grids.
+! I.e., spacing between lats (lons) is constant.
```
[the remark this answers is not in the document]
gmao-rreichle: It still says: "This module generates ESMF_Grids corresponding to _regular_ lat-lon coordinate grids. I.e., spacing between lats (lons) is constant."  This makes no sense.

base/MAPL_EASEConversion.F90
```diff
@@ -0,0 +1,828 @@
+#include "MAPL_ErrLog.h"
+
+module MAPL_EASEConversion
+  
+  ! =====================================================================================
+  !
+  ! EASE_conv.F90 - Fortran routines for conversion of Equal-Area Scalable Earth (EASE)
```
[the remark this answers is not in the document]
gmao-rreichle: I don't see such a change.

base/MAPL_EASEGridFactory.F90
weiyuan-jiang: At some point, if only cols is provided, we can construct the factory and thus EASE grid

base/MAPL_EASEConversion.F90
gmao-rreichle: Thanks for the explanation.  If the "global" cylindrical EASE grid is considered a "regional" grid for MAPL purposes, we need to add this info as a comment to MAPL_EASEGridFactory.F90.  There should then also be checks on whether variable "regional" is indeed "true".  E.g., in MAPL_EASEGridFactory.F90 near line 340, there's the following chunk of code 
```
      regional  = (dateline == 'XY')
      if (regional) then
         delta = (this%lon_range%max - this%lon_range%min) / this%im_world
         min_coord = this%lon_range%min + delta/2
         max_coord = this%lon_range%max - delta/2
      else
         delta = 360.d0 / this%im_world
         min_coord = -180.d0 + delta/2
         max_coord = +180.d0 - delta/2
      end if
``` 
Given the fact that the EASE grid must be regional, the "else" statement should exit with an error.  There are probably other places that need the same cleanup.

tclune (dismissed):

base/MAPL_EASEGridFactory.F90
```diff
@@ -0,0 +1,1845 @@
...
+            this%lon_corners(im+1) = (this%lon_centers(im) + this%lon_centers(1))/2 + 180
+         end if
+
+         ! This section about pole/dateline is probably not needed in file data case.
+         if (abs(this%lon_corners(1) + 180) < 1000*epsilon(1.0)) then
+            this%dateline = 'DE'
+         else ! assume 'XY'
+            this%dateline = 'XY' 
+            this%lon_range = RealMinMax(this%lon_centers(1), this%lon_centers(jm))
+         end if
```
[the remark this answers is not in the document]
weiyuan-jiang: I leave 'XY' there intentionally in case in the future we need regional longitudes

base/MAPL_EASEGridFactory.F90
```diff
@@ -0,0 +1,1845 @@
...
+
+        ! TODO: if 'lat' and 'lon' are not present then 
+        ! assume ... pole/dateline are ?
+
+        ! TODO: check radians vs degrees.  Assume degrees for now.
+
+
+        ! TODO: modify CoordinateVariable so that get_coordinate_data() is overloaded
+        ! for different types (as subroutine) to avoid casting here.
+        ! TODO: add get_coordinate_variable() interface to avoid the need to cast
```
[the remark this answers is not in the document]
weiyuan-jiang: NO. We assume if it is initialized from metadata,  the data we need is there.  I removed those "TODO" comments

base/MAPL_EASEGridFactory.F90
```diff
@@ -0,0 +1,1845 @@
...
+              extrap_lat = this%lat_centers(i_max-1) + d_lat
+              if (extrap_lat >= (90.0-(d_lat/20.0))) then
+                 this%lat_centers(i_max) =  90.0
+              end if
+           end if
+        end if
+
+         ! Corners are the midpoints of centers (and extrapolated at the 
+         ! poles for lats.)
+         if (.not. allocated(this%lon_corners)) then
```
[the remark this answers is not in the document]
weiyuan-jiang: indent corrected

biljanaorescanin:
@gmao-rreichle PR was tested for BCS package creation for `v12 EASE M09` and `v13 EASE M36` and it is zero diff. 
To run make_bcs needs 3 branches: 
`MAPL              | (t) v2.56.0 (DH)           | (b) feature/wjiang/ease_grid`
`GEOSldas_GridComp | (b) develop (DH, 4c7de26)  | (b) feature/wjiang/ease_grid`
`GEOSgcm_GridComp  | (b) develop (DH, 85b060ab) | (b) feature/wjiang/ease_grid`

biljanaorescanin:
> Thanks, @biljanaorescanin, for the test results. Did you run make_bcs from a build of GEOSgcm or GEOSldas?

@gmao-rreichle  I've tested from GEOSldas. Do you need me to do the same from GEOSgcm?

weiyuan-jiang:
> > Thanks, @biljanaorescanin, for the test results. Did you run make_bcs from a build of GEOSgcm or GEOSldas?
> 
> @gmao-rreichle I've tested from GEOSldas. Do you need me to do the same from GEOSgcm?

Do you test bcs generated by these PRs? That is important

biljanaorescanin:
> Do you test bcs generated by these PRs? That is important

Yes, few comments up I say I used those 3 branches for bcs creation. 
I can share experiment outputs on chat if you need to see it. 

This was my testing clone: 
https://github.com/GEOS-ESM/GEOSldas/blob/testing/borescan_ease_factory/components.yaml

gmao-rreichle:
> > > Thanks, @biljanaorescanin, for the test results. Did you run make_bcs from a build of GEOSgcm or GEOSldas?
> > 
> > 
> > @gmao-rreichle I've tested from GEOSldas. Do you need me to do the same from GEOSgcm?

No, make_bcs should be identical when run from either App.  I mostly wanted to know if you worked within GEOSldas, and how the GEOSldas_GridComp branch fits into all of this.  This has now been addressed with new PRs for GEOSldas and GEOSldas_GridComp.  I added a new test requirement to the intro comment above (0-diff for GEOSldas).

gmao-rreichle (approved):
@mathomp4, @weiyuan-jiang : The new MAPL EASE Grid Factory has been successfully  0-diff tested with GEOSldas and GEOSgcm.  The PR is ready to be merged as far as I'm concerned.  
To use the new functionality in GEOSldas, we'll need a new MAPL release.  Both GEOSldas and GEOSgcm are on MAPL v2.56, so the new MAPL release should work ok for both. 
I'm a bit unsure about where we are with GEOSadas (which now includes the LDAS in its build).  I know Sara was sorting out MAPL versions for ADAS and LDAS recently, so whatever the issue was then might still be an issue now.  But we can sort this out later. 
cc: @tclune @biljanaorescanin

mathomp4 (approved):

PR #3704, opened 2025-05-13
node and bracket

### LINKED ISSUE #3701: ExtData3G Base Classes, node/bracket
Starting ExtData3G
Since the MAPL3 interfaces are still in flux, makes sense to start from the bottom at the layers that do the work of finding the right files

I have the following assumptions about ExtData3G when starting based on the discussions we have had about MAPL3

1. A bunch of stuff is moving to the generic layer like the time interpolation and regridding
2. Since replay is being redone and we believe we will not have a clock running backwards now so we will not have to worry about that
3. We believe hack used for GAAS will not be used in MAPL3

The result of this is that several lower layers from ExtData2G can be rewritten to be simpler assuming the above, these are:

the file handler; decides what files and time index in said files to establish as the bracket at any given time, and assumptions about persistent, climatology, etc. this can become much simpler given 2 and 3 above.

the brackets  stores the left and right brackets or nodes as I called them, and in ExtData2G did the time interpolation (now just have to compute weights given 1 above, rather than do that and do the time interpolation on the fields)

the nodes themselves, basically what file it this, what time index and and time from the file is it and what time to use in the interpolation, pretty much the same, not much different here

I am going to start with these lower level classes. Should be straightforward given those assumptions.

The first PR will just be bracket and node

bena-nasa:
Fixes #3701 
Very, very, very simple PR to create now simple node and bracket classes (note that these are not definitive and may have some additions once we start interacting with files and fields) for ExtData3G as the first part of my plan.

The idea will be that there will be a file handling class(es) that will be responsible for actually filling these in each time based on the current time and other parameters etc... following MAPL2
That will be next on the agenda.

## Types of change(s)

## Checklist


gridcomps/ExtData3G/ExtDataBracket.F90
```diff
@@ -0,0 +1,116 @@
...
+         procedure :: get_right_node
+   end type ExtDataBracket
+
+contains
+
+   subroutine set_parameters(this, disable_interpolation, left_node, right_node)
+      class(ExtDataBracket), intent(inout) :: this
+      logical, intent(in) :: disable_interpolation
+      type(ExtDataNode), intent(in) :: left_node
+      type(ExtDataNode), intent(in) :: right_node
```
tclune, line 31: Should these be optional?  Or do you always need to set all 3 params?

gridcomps/ExtData3G/ExtDataBracket.F90
```diff
@@ -0,0 +1,116 @@
...
+      this%right_node = right_node
+   end subroutine
+
+   function time_in_bracket(this,time) result(in_bracket)
+      logical :: in_bracket
+      class(ExtDataBracket), intent(inout) :: this
+      type(ESMF_Time), intent(in) :: time
+      type(ESMF_Time) :: left_time, right_time
+      
+      left_time = this%left_node%get_interp_time()
```
tclune, line 44: I'm going to guess that `get_interp_time()` means get the time corresponding to that node?  Not sure "interp" is the best descriptor for that.    "ref_time" seems closer, but still not ideal.

gridcomps/ExtData3G/ExtDataBracket.F90
```diff
@@ -0,0 +1,116 @@
...
+      type(ExtDataNode) :: node
+      class(ExtDataBracket), intent(inout) :: this
+      integer, optional, intent(out) :: rc
+
+      node = this%left_node
+      _RETURN(_SUCCESS)
+
+   end function get_left_node
+
+   function get_bracket_weights(this,time,rc) result(weights)
```
tclune, line 88: maybe `compute_bracket_weights` is more accurate?

gridcomps/ExtData3G/ExtDataBracket.F90
```diff
@@ -0,0 +1,116 @@
...
+         weights(2) = 0.0
+      else
+         time1 = this%left_node%get_interp_time()
+         time2 = this%right_node%get_interp_time()
+         tinv1 = time - time1
+         tinv2 = time2 - time1
+         alpha = tinv1/tinv2
+         weights(1) = alpha
+         weights(2) = 1.0 - alpha
+      end if
```
tclune, line 111: Eliminate nested logic, and simplify.
```suggestion
      weights = [1.0, 0.0]
      _RETURN_IF(this%disable_interpolation)
    
      time1 = this%left_node%get_interp_time()
      time2 = this%right_node%get_interp_time()
      tinv1 = time - time1
      tinv2 = time2 - time1
      alpha = tinv1/tinv2
      weights = [alpha, 1.0 - alpha]
```

tclune (changes requested):
See inline comments.   I may have more concerns after a 2nd pass.

gridcomps/ExtData3G/ExtDataBracket.F90
```diff
@@ -0,0 +1,116 @@
...
+   use mapl3g_ExtDataNode
+   implicit none
+   private
+
+   public :: ExtDataBracket
+
+   type ExtDataBracket
+      type(ExtDataNode) :: left_node
+      type(ExtDataNode) :: right_node
+      logical          :: disable_interpolation = .false.
```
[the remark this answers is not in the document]
bena-nasa: Yes, that's why I said this was in no way definitive. If we are running backwards probably would want to reverse this. Also if we are persisting outside of a dataset there will only be one node. My plan in that case is to add a parameter saying this node is disabled.

tclune (approved):

PR #3709, opened 2025-05-15
Fix issue with GNU and MVAPICH 4

mathomp4:
## Types of change(s)

## Checklist

## Description

In trying to test GCC 14 with MVAPICH 4 on Discover, it was found that when you tried to run GEOS, the code would crash at `MPI_Init_thread()` with a SIGFPE. This is weird.

After talking with @FlorianDeconinck, he suggested something his group found which is that at times you have to turn off IEEE halting around a call. (I want to say @pchakraborty also mentioned something about this once...)

Should we need to do this? No. Can I run a simple `MPI_Init_thread()` example code with GCC 14 + MVAPICH 4? Yes. But this seems to be needed here.

I'm going to keep this as Draft until I can discuss with @tclune (and perhaps @FlorianDeconinck and @pchakraborty). I mean, it seems "harmless" but it is weird. I could (via CMake) make this an `#ifdef` sort of thing with the right detection...but is it worth it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- define `comp_name` in `MAPL_GridCreate` when GC is not present
+- Define `comp_name` in `MAPL_GridCreate` when GC is not present
 - Fixed bug in profiler demo
+- Fix for GNU + MVAPICH 4 disabling ieee halting around `MPI_Init_thread()`
 
 ### Added
 
--- a/gridcomps/Cap/MAPL_Cap.F90
+++ b/gridcomps/Cap/MAPL_Cap.F90
@@ -15,6 +15,7 @@ module MAPL_CapMod
    use MAPL_CapOptionsMod
    use MAPL_ServerManager
    use MAPL_ApplicationSupport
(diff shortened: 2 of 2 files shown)


pchakraborty:
Yeah, I struggled with this once. Numpy throws a SIGFPE when imported, and the code would crash. The good thing is that Fortran allows disabling of trapping of FPEs, so the solution that I ended up implementing was to disable the trapping temporarily
https://github.com/GEOS-ESM/FVdycoreCubed_GridComp/blob/714c2fe104b086d4c9e3865ed867d5a947d62d91/FV_StateMod.F90#L1157-L1163

tclune:
Amusingly, you have stumbled onto a use case that demonstrates pure functions (IEEE*) that have side-effects, while the intent of PURE is to not have side-effects!

mathomp4:
One note: I'm also keeping this draft because I want to make sure it is zero-diff with Intel/Intel MPI, say.

I mean, I am like 99.999999999999999999999% sure it is (if not, time to just give up), but, well, testing is good.

pchakraborty:
I agree with the "if not, time to just give up".

mathomp4:
Good news! It is zero-diff. 

I mean, Cap started up without crashing so, it had to be, but good to confirm!

pchakraborty (approved):

PR #3745, opened 2025-05-23
Feature/wdboggs/add restart enum 3730

### LINKED ISSUE #3730: Create enum flag for RESTART
Create a new enum flag to specify the behavior of ESMF_Field instances on restart.

darianboggs:
## Types of change(s)

## Checklist

## Description
MAPL3 needs to support more than two possibile values of the **restart** argument of 
**MAPL_GridCompAddFieldSpec**. This PR adds the capability.

## Related Issue
Closes #3730


darianboggs:
I added @bena-nasa and @atrayano explicitly as reviewers, but if anyone else has feedback, please add it. I based the MAPL3 enumerators on the values from MAPL2 as a starting point, but they will probably need to be changed for MAPL3.

pchakraborty:
There is a `RestartHandler.F90` in `generic3g`. Should the `enum` go there instead?

tclune:
> There is a `RestartHandler.F90` in `generic3g`. Should the `enum` go there instead?

That seems like a reasonable suggestion.    We always go back-and-forth between centralizing our "constants" and placing them next to the relevant logic.   My (weak) opinion is that purely internal things like this should be colocated with the relevant logic as per Purnendu's suggestion.

darianboggs:
enum moved to RestartHandler.F90

PR #3748, opened 2025-05-28
Update MAPL_EQsat to ramping version

mathomp4:
## Types of change(s)

## Checklist

## Description

This PR brings in the "ramping" version of MAPL_SatVapor I'd made in CVS long ago. This adds the RAMP functionality you see the [GEOS_Shared Qsat code](https://github.com/GEOS-ESM/GMAO_Shared/blob/deec45d81642cbf1cd0c8e9768d7b0b0c0a4b710/GEOS_Shared/GEOS_Utilities.F90#L501). 

Now, I found a few variants I had there, and this one is the "pascals" only version that is zero-diff in current GEOSgcm v11 and v12 (since it is functionally identical unless you use the new Ramping options).

Note this is *NOT* zero-diff if you convert GEOS_Shared qsat to this code, but this will allow this.

### DIFF
--- a/base/MAPL_SatVapor.F90
+++ b/base/MAPL_SatVapor.F90
@@ -185,6 +185,7 @@ contains
 
    function set_optional(flag) result(flag_)
       logical, optional, intent(in) :: flag
+      logical :: flag_
       flag_ = .false.
       if (present(flag)) flag_ = flag
    end function


tclune (approved):
Definitely want to revisit when compilers support automatic overloads.

mathomp4:
> Definitely want to revisit when compilers support automatic overloads.

It is sort of annoying that it's the same code over and over! :)

Issue #3630, opened 2025-04-22
ECMWF-like pressure level output


amdasilva:
# The problem

HISTORY now produces pressure level output where grid points below the surface (p>ps) receive an undefined value. This convention is at odds with meteorological practice where parameters are extrapolated below the surface.  This is no more unphysical than the notion for sea level pressure.  These missing values are particularly inconvenient for developing AI/ML models, making ERA-5 a much more convenient alternative.

@tclune I am assigning this to you so that you can decide who in the SI team should implement this.

# The solution 

Use a Fortran implementation of the so-called ECMWF-method to produce pressure level output with extrapolation below the surface.

## HISTTORY extension

In HISTORY.rc, where we have:
 
```inst3_3d_asm_Np.levels: 1000 975 950 925 900 875 850 … 0.1```
 
now support:
 
```inst3_3d_asm_Np.xlevels: 1000 975 950 925 900 875 850 … 0.1```
 
where **xlevels **stand for *extrapolated levels*. As far as MAPL is concerned, this is a new feature of HISTORY, an alternative to the standard pressure level implementation.

## Where to find the source code

On discover,

```/gpfsm/dswdev/adasilva/Notebooks/FP_regrid```

There 3 files of interest,

```
eta2xprs.F90
eta2xprs.py
build.sh
```

File ```eta2xprs.F90``` has the standard Fortran subroutine implementing this.  Take a look at the comments and get back to me if you have questions.  Notice that

- It takes ak/bk and ps on  input. It would be trivial to change the code to work with 3D pressures if you so wish.
- The ak/bk should match the field being interpolated: if interpolating mid-layer variables, ak./bk should be mid-layer.
- Notice that H and T are treated as particular cases. This is VERY IMPORTANT.
- I have been using LOG as the interpolation method. Try to match was is being done in the current plea interpolation. 

## Python interface

The Fortran file ```eta2xprs.F90``` can be compiled into module ```eta2xprs_.*.so```, see script ```build.sh``` (it is all self-contained). Install this under   $ESMDIR/Linux/lib/Python/MAPL.  Notice that you may need to change ```import eta``` to ```import .eta``` since they are both part of package ```MAPL```.

Please update ```eta.py``` in MAPL to include more recent options for vertical levels, like km=180. This Python module is needed to write scripts to create extrapolated pressure output from legacy output.

amdasilva:
There are Notebooks in the same directory with sample offline output.

tclune:
[...]

PR #3764, opened 2025-06-03
Feature/bmauer/below surf extrapolation ecmwf

bena-nasa:
This implements issue #3630 

As requested by the @amdasilva if you specify `xlevels` instead of `levels` in the collection it will do the ECMWF extrapolation to extrapolate below the surface, which from the code provided does the following:

If temperature use their formula
If height use their formula
If anything else, use the lowest level value from the input

I incorporated this into our existing vertical regridding scheme as it would be too hard to incorporate a whole new vertical regridding code in.

HOWEVER,
I cannot get the formula for height to give reasonable values so for now I have disabled that and am just using the lowest level for heights. I've stared the provided formula and can find no issue other than it gives complete unrealistic values.

 I'm rather stuck on it other than to say the formula does not give any semblance of realistic values.

## Types of change(s)

## Checklist


base/MAPL_VerticalMethods.F90
```diff
@@ -369,10 +381,156 @@ subroutine regrid_eta_to_pressure(this,ptrin,ptrout,rc)
...
+                       if (hgt <= 2500.0) then
+                          tprime0 = 0.002*((2500.0-hgt)*t0+(hgt-2000.0)*tplat)
+                       else
+                          tprime0 = tplat
+                       end if
+                       if (tprime0 < tstar) then
+                          alnp = 0.0
+                       else 
+                          alnp = MAPL_RDRY*(tprime0-tstar)/phis_ij*log(plevs_out(ii)/ps_in_mb)
+                       end if
```
tclune, line 523: Some of these conditionals are ripe for the never-elser approach.  

You've at least got 2 levels of conditionals inside a do loop inside a select case inside a conditional, and ... I did not scroll up further than that.     I hate to slow the PR down, but I think we will do our future selves a bit of a favor with some refactoring here.  Please think about it.

gridcomps/History/MAPL_HistoryGridComp.F90
```diff
@@ -1129,6 +1140,48 @@ subroutine Initialize ( gc, import, dumexport, clock, rc )
...
+             list(n)%field_set%nfields = list(n)%field_set%nfields + 1
+             allocate( fields(4,  list(n)%field_set%nfields), _STAT )
+             fields(1,1:list(n)%field_set%nfields-1) = list(n)%field_set%fields(1,:)
+             fields(2,1:list(n)%field_set%nfields-1) = list(n)%field_set%fields(2,:)
+             fields(3,1:list(n)%field_set%nfields-1) = list(n)%field_set%fields(3,:)
+             fields(4,1:list(n)%field_set%nfields-1) = list(n)%field_set%fields(4,:)
+             fields(1,  list(n)%field_set%nfields  ) = "TS"
+             fields(2,  list(n)%field_set%nfields  ) = "SURFACE"
+             fields(3,  list(n)%field_set%nfields  ) = "TS"
+             fields(4,  list(n)%field_set%nfields  ) = BLANK
```
tclune, line 1179: As above, depending on how you decide.

tclune (dismissed):
I'm marking as approve but I did leave some strongly recommended changes in the inline comments.  Up to you how much of that is to be pursued.

tclune:
> I cannot get the formula for height to give reasonable values so for now I have disabled that and am just using the lowest level for heights.

If you can create a figure - say including the Himalayas that may sharpen the discussion with @amdasilva .

base/MAPL_VerticalMethods.F90
bena-nasa: removed

gridcomps/History/MAPL_HistoryGridComp.F90
bena-nasa: done

tclune (approved):

tclune (approved):

Issue #3774, opened 2025-06-05
MAPL_GridCompAddFieldSpec - add support for ESMF_FieldBundle and ESMF_State


pchakraborty:

pchakraborty:
#3776 - Added `FieldBundleClassAspect` and `StateClassAspect`

PR #3775, opened 2025-06-05
Added optional argument `add2export` for MAPL_GridCompAddFieldSpec

### LINKED ISSUE #3773: MAPL_GridCompAddFieldSpec - new optional argument, add2export
For an `internal` spec, if `add2export` is present and set to `.true.`, re-export it

pchakraborty:
…added the optional argument add2export, with code for the case when an internal var gets exported

## Types of change(s)

## Checklist

## Description

Fixes #3773 - added optional argument `add2export` and code for the case when `add2export` is present and true.

Also, made item_type user specifiable, in preparation for #3774 

## Related Issue

#3773 
#3774

### DIFF
--- a/generic3g/ESMF_Utilities.F90
+++ b/generic3g/ESMF_Utilities.F90
@@ -9,6 +9,7 @@ module mapl3g_ESMF_Utilities
    public :: write(formatted)
    public :: get_substate
    public :: to_esmf_state_intent
+   public :: esmf_state_intent_to_string
    public :: MAPL_TYPEKIND_MIRROR
 
    type(ESMF_TypeKind_Flag), parameter :: MAPL_TYPEKIND_MIRROR = ESMF_TypeKind_Flag(200)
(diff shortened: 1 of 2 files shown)


generic3g/ESMF_Utilities.F90
```diff
@@ -198,5 +199,22 @@ function to_esmf_state_intent(str_state_intent, rc) result(state_intent)
       _RETURN(_SUCCESS)
    end function to_esmf_state_intent
 
+   function esmf_state_intent_to_string(state_intent, rc) result(state_intent_str)
```
tclune, line 202: Note, I'm fairly certain I've got this function somewhere else in MAPL3.   But let's not let that slow the PR down.

generic3g/MAPL_Generic.F90
```diff
@@ -489,6 +499,18 @@ subroutine gridcomp_add_fieldspec( &
       component_spec => outer_meta%get_component_spec()
       call component_spec%var_specs%push_back(var_spec)
 
+      if (present(add2export)) then
```
tclune, line 502: This is fine, but on a meta level, I see that this procedure is getting awfully large for a layer that was intended to be thin wrappers.   We may need to refactor soon.

tclune (approved):
See inline vents.

PR #3794, opened 2025-06-10
Add write_LZ_first option to sampler code

metdyn:
## Types of change(s)

## Checklist

## Description
In Fortran code, we will output to to the netCDF file, u[nz, Location], e.g., in trajectory sampler, with lev being the fast axis, different from usual GEOS output. This is required for AIST project (@amdasilva: to ease geoval calculations from CRTM)

3D variable output is set to [nz, Locations] in trajectory, mask, station sampler. 
Once this option is added to the griddedIO interface, the swath and geostationary samplers will have the same output format for 3D variables.


bena-nasa:
@metdyn maybe we could have a better/more descriptive keyword than `write_LZ_first`? The idea is we have a usual ordering of horizontal-vertical-time but we want this reversed?

metdyn:
> @metdyn maybe we could have a better/more descriptive keyword than `write_LZ_first`? The idea is we have a usual ordering of horizontal-vertical-time but we want this reversed?

Thanks for the question, @bena-nasa!  I only know @amdasilva mentioned for station and trajectory sampler and others,  Fortran code output  vertical-horizontal (e.g., horizontal is local stream points) is more favorable. I guess the time would be slow axis and maybe `write_vertical_first` is a better choice, albeit longer.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added new option to History, if you specify xlevels instead of levels, it will perform extrapolation below the surface, using ECMWF formulas for height and temperature, otherwise use lowest model level
 - Added `_USERRC` macro for use with ESMF commands that return both `rc` and `userrc`
-
+- Added write_LZ_first to trajectory and mask sampler code, default to true
```
tclune, line 17:
```suggestion
- Swapped order of output dimensions for tarjectory and mask samplers.   Now the leading dimension (Fortran convention) is level.
```

tclune (approved):
I strongly suspect that the conditional logic could be consolidated, rather than spread out across so many points in the code, but that would require some work.

I also question the value of leaving this as optional rather than requiring LZ to always be first.

Please open a follow-up ticket to update the documentation that you and Jules assembeled.

tclune:
I am also belatedly realizing that the new option has a default of `.true.`.   It is usually much better to have defaults be `.false.`.

So ... we need a new name for the option. We can then introduce another variable with `.not.` so that the conditionals can remain in place.

One idea:   make the option  `leading_dimension`  with default `lev`.   Then have
```
     this%write_lev_first = (leading_dimension == 'lev')
```

PR #3806, opened 2025-06-16
MAPL_ClockGet, bug fix in OuterMetaComponent::run_child, new FakeGocartGridComp

pchakraborty:
## Types of change(s)

## Checklist

## Description
- Added `MAPL_ClockGet` (and a test) to retrieve time step as a float
- Fixed a bug in `OuterMetaComponent::run_child` where the parent's run phases were being queried for index, instead of the child's
- Added a fake gridcomp that runs both the run methods of the child


tclune (changes requested):
See the inline change.   Need to make sure it still compiles, so at least wait for CI to run.

generic3g/MAPL_Generic.F90
```diff
@@ -985,4 +991,25 @@ subroutine gridcomp_reexport(gridcomp, unusable, src_comp, src_name, src_intent,
       _UNUSED_DUMMY(unusable)
    end subroutine gridcomp_reexport
 
+   subroutine clock_get(clock, unusable, timestep, dt, rc)
+      type(ESMF_Clock), intent(in) :: clock
+      class(KeywordEnforcer), optional, intent(in) :: unusable
+      type(ESMF_TimeInterval), optional, intent(out) :: timestep
```
tclune, line 997: Remove this argument.

tclune (approved):

PR #3808, opened 2025-06-16
Feature/bmauer/raw netcdf for hoot

bena-nasa:
I did this, so might as well save it...
Add an option to the `raw_bw.x` code to use netcdf files rather than Fortran binary.
## Types of change(s)

## Checklist


tclune (dismissed):
Hopefully Hoot is aware that we generally expect serial perf of netcdf is quite close to that of binary.

PR #3850, opened 2025-06-24
Replace single-underscore uppercase macros with lowercase

### LINKED ISSUE #3851: Need to replace double-underscore-uppercase macros
Per @tclune, the Fortran Standards committee is working on their own preprocessor. In that:

```
Macro names beginning with a leading underscore followed by an
      uppercase letter or a second underscore are reserved to the
      processor. Reserved names shall not be the subject of a #define
      or #undef preprocessing directive within a program unit. [Note:
      Any macro name matching the regular expression
      '/^_[A-Z_][A-Za-z0-9_]*$/' is considered reserved.]
```

This unfortunately applies to our macros like `_RC`

mathomp4:
## Types of change(s)

## Checklist

## Description

Per @tclune, the Fortran Standard will soon have its own preprocessor. And that preprocessor will reserve single-underscore variables.

So, this is an attempt at *adding* double-underscore versions of our macros in the hope we can remove support for single-underscore.

This is probably horribly broken, but I want to see what CI says.

## Related Issue

Closes #3851


tclune:
Is there an issue associated with this?

mathomp4:
> Is there an issue associated with this?

Oh I should make one. I was just trying it out...and failing badly! I did too much somewhere!

tclune:
The silver lining here is that the older macros are also not future proof, se we have good reason to force GEOS users to update.    But ... one more thing on the portability list.    

But .. this will make maintaining a separate MAPL3 branch for the GCM difficult.

mathomp4:
> I think we need a new token that is default false. Something like `_mapl_legacy_macros`. And then
> 
> ```
> #ifdef _mapl_legacy_macros
> #  include 'legacy_ErrLog.h'
> #else
> #  include 'new_ErrLog.h'
> #endif
> ```

Should this be an either/or? Or something like:
```
#ifdef _mapl_legacy_macros
#  include 'legacy_ErrLog.h'
#endif
#  include 'new_ErrLog.h'
```

So that the new ones are always around and if the macro is set, we also include the old ones?

PR #3857, opened 2025-07-02
A MAPL3 Version of ExtDataDriver.x

bena-nasa:
This is my first attempt at a mapl3 version of ExtDataDriver.x

Some background and things I've changed/need to still be done.
I have moved this to the Apps directory, and renamed it `MAPL_Component_Driver` because that is what is really is, a driver to exercise ExtData and/or History so that is a better name does not single out ExtData.

This is by no means complete but a start. I started with one of the Cap test components as the base and went from there to add some of the functionality in ExtDataDriver.x. There was some stuff about setting a default vertical profile in the template I started from, not sure what that is about but left it for now.

Since there is no functional ExtData3G yet I have not bothered with all the the run modes in the original. For now it can exercise History3G and fill in the exports from a "menu" of fields that are set in the internal state.
The "menu" of functions available to use to fill exports is not quite as many so far as in the MAPL2 ExtDataDriver.x
For now it is a time interval relative to a reference, the lats of the grid, the Lons of the grid, and random numbers.

I have changed it to use HCONFIG
Now that MAPL3 can dynamically create import and export variables at runtime from the component input file, I don't have to support this via a custom layer as I did in MAPL2.

For now I have made copies of the cap and capGridComp in the application directory. Future work will be to adapt those in the way I adapted the MAPL2 version for greater functionality in these layers. If this proves not necessary we can always revert to use the "stock" cap.

I have confirmed I can exercise History3G with this.

## Types of change(s)

## Checklist

### DIFF
--- a/Apps/MAPL_Component_Driver/CMakeLists.txt
+++ b/Apps/MAPL_Component_Driver/CMakeLists.txt
@@ -2,10 +2,10 @@ esma_set_this (OVERRIDE root_gridcomp)
 find_package (MPI REQUIRED)
 
 esma_add_library(${this}
-  SRCS RootGridComp.F90
+  SRCS RootGridComp.F90 time_support.F90
   DEPENDENCIES mapl3g MAPL
   TYPE SHARED)
 
-ecbuild_add_executable(TARGET MAPL_Component_Driver.x SOURCES MAPL_Component_Driver.F90 DriverCap.F90 DriverCapGridComp.F90 time_support.F90 DEPENDS MAPL mapl3g ESMF::ESMF)
+ecbuild_add_executable(TARGET MAPL_Component_Driver.x SOURCES MAPL_Component_Driver.F90 DriverCap.F90 DriverCapGridComp.F90 DEPENDS MAPL mapl3g ESMF::ESMF)
 target_link_libraries(MAPL_Component_Driver.x PRIVATE ${this})
 target_include_directories (MAPL_Component_Driver.x PRIVATE $<BUILD_INTERFACE:${MAPL_SOURCE_DIR}/include>)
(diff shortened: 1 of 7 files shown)


Apps/MAPL_Component_Driver/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+esma_set_this (OVERRIDE root_gridcomp)
```
tclune, line 1: This does not seem like the right name - too generic.  Will discuss when you return.

tclune (commented):
It looks good, but I want to do a thorough walk through when you return so that I can better understand how we might refactor with Cap/CapGridComp to reduce redundancy.

bena-nasa:
@tclune 
For now I had just made copies of cap and capgridcomp. My plan was to modify these to bring in the capabilities I had in extdatadriver in 2g.
These were
1. Run multiple configurations in the execution. I.E run one configuration of cap/history/extdata/root grid comp then, run a 2nd This was so you could say generate files, then read them back in via extdata in the next configuration
2. Run the clock forwards and backwards in cap to simulate replay
3. Control exactly when the hierarchy runs. Rather than run every step of the cap, run at specific times I choose. This is useful if say you want to generate a monthly climatology for the year for testing.

Apps/MAPL_Component_Driver/CMakeLists.txt
tclune: ok - you'll update the PR?

tclune: I'm thinking we may not even want a library here.   Maybe those few sources just go into the executable?   And then we can just leave the default name - has little role at that point.

tclune (approved):

PR #3863, opened 2025-07-10
Testing GCC 15 CI

mathomp4:
This is just me testing out GCC 15 in CI.


mathomp4:
This is un-good. I can now *duplicate* the ifort errors here with ifort 2021.13 on bucy. These do not happen with Baselibs 8.14.0.

The changes between 8.14.0 and 8.16.0 are presented at the end.

The scariest thing to me is that this is due to the update from ESMF 8.8.1 to 8.9.0b09. That update was made for GCC 15, but maybe something changed in ESMF 8.8.1 to 8.9.0b09?

I will do more testing. Next will be building Baselibs 8.15.0 with ifort 2021.13. 

cc: @tclune 

---

## [8.16.0] - 2025-07-08

### Updates

- curl 8.14.1
- NCO 5.3.4
- CDO 2.5.2
- nccmp 1.10.0.0

### Fixed

- Fix CMake issue with `GFE.check` target

### Changed

- Enable various `-Wno-error=` options when using GCC 14+

### Removed

- Removed patch for `nccmp` as it is no longer needed with the latest version

## [8.15.0] - 2025-05-27

### Updates

- ESMF v8.9.0b09 (needed for GCC 15)
- curl 8.13.0

### Fixed

- Fixes for GCC 15 Compatibility (https://github.com/GEOS-ESM/ESMA-Baselibs/issues/290)
  - Patches for szlib, nccmp, and hdf5
  - HDF5 patch is expected to be temporary until next version (v1.14.7 or v1.16)

Issue #3918, opened 2025-07-25
Add a logger argument to RestartHandler's constructor


pchakraborty:
This way, the gridcomp name is logged by RestartHandler, instead of the generic `mapl.restart`.

PR #3919, opened 2025-07-25
Fixed #3917 and #3918

### LINKED ISSUE #3917: Fix bug in initialize_read_restart
Currently the `import` and `internal` restart files of a gridcomp are read based on `checkpoint_controls` from ComponentSpec's `misc`. It should be `restart_controls` instead.

Also use the GriddedComponentDriver's `get_name` method to initialize the RestartHandler object instead of the OuterMetaComponent's, since the OuterMeta's `get_name` returns the gridcomp name in brackets.

pchakraborty:
## Types of change(s)

## Checklist

## Description
Fixed #3917 and #3918

### DIFF
--- a/generic3g/OuterMetaComponent/initialize_read_restart.F90
+++ b/generic3g/OuterMetaComponent/initialize_read_restart.F90
@@ -30,13 +30,18 @@ contains
 
       driver => this%get_user_gc_driver()
       states = driver%get_states()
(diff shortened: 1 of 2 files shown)


tclune (approved):
I am rather wary of the temptation to pass loggers around as arguments to procedures.  

We need to continue to think about the right path here, and possibly study how other logging systems handle these situations.   The examples that I'm familiar with seem to always use a locally specified logger.

Issue #3866, opened 2025-07-14
Add simple interface for MAPL_HConfigGet


darianboggs:
Earlier in MAPL3 design/implementation, we created a derived type, `HConfigParams`, to hold arguments: `type(ESMF_HConfig)` :: hconfig, `class(Logger), pointer :: logger`, `character(*) :: label`, `logical :: check_value_set`, and  `logical :: value_set`. An instance of this type is passed into `MAPL_HConfigGet` instead of the arguments themselves.

Now, there is a use case for passing in arguments instead of an instance of `HConfigParams`. We need to add a subroutine in the public interface for `hconfig3g` (in `hconfig_utils`) with individual arguments to `MAPL_HConfigGet`.

`MAPL_HConfigGet` is implemented through modules and an include file. This subroutine should use the existing hierarchy to minimize the new code necessary. It should create an `HConfigParams`, and then use the existing routines with that instance, and then set the `value` and `value_set` arguments. The `logger` will be updated in the process if present.

The interface for the new function should be:
```
subroutine MAPL_HConfigGet_with_args(hconfig, label, value, default, logger, value_set, rc)
   type(ESMF_HConfig), intent(in) :: hconfig
   character(len=*), intent(in) :: label
   <TYPE>, intent(out) :: value
   <TYPE>, optional, intent(in) :: default
   class(Logger), pointer, optional, intent(in) :: logger
   logical, optional, intent(out) :: value_set
   integer, optional, intent(out) :: rc
end subroutine MAPL_HConfigGet_with_args
```

darianboggs:
`MAPL_HConfigGet` was largely implemented in a PR (#2570) for `MAPL_GridCompGetResource`. This drove the design. It may be that there are issues with dependencies, especially the Logger. In that PR we chose **not** to provide a direct procedure using arguments corresponding to the members of the `HConfigParams` type. The lack of the direct procedure may be an oversight, but there may be reasons that complicate this issue.

darianboggs:
Meeting with Purnendu and Tom to discuss before proceeding.

PR #3929, opened 2025-07-30
Add simple MAPL_HConfigGet (specific) procedures

darianboggs:
## Types of change(s)

## Checklist

## Description
We want to change the generic `MAPL_HConfigGet` procedure to use arguments (`hconfig`, `label`, and `logger` directly instead of using the `HConfigParams` object.

## Related Issue
#3866

### DIFF
--- a/Apps/MAPL_Component_Driver/ComponentDriverGridComp.F90
+++ b/Apps/MAPL_Component_Driver/ComponentDriverGridComp.F90
@@ -30,6 +30,7 @@ module mapl3g_ComponentDriverGridComp
    character(*), parameter :: KEY_DEFAULT_VERT_PROFILE = "default_vertical_profile"
    character(len=*), parameter :: runModeGenerateExports = "GenerateExports"
    character(len=*), parameter :: runModeFillExportsFromImports = "FillExportsFromImports"
+   character(len=*), parameter :: runModeFillImports = "FillImports"
 
 contains
 
@@ -156,6 +157,8 @@ contains
          call fill_state_from_internal(exportState, internal_state, support, _RC)
       else if (support%runMode == "FillExportsFromImports") then
          call copy_state(exportState, importState, _RC)
+      else if (support%runMode == "FillImports") then
+         ! there's nothing to do here
       else
          _FAIL("no run mode selected")
       end if
--- a/GeomIO/CMakeLists.txt
+++ b/GeomIO/CMakeLists.txt
@@ -7,6 +7,9 @@ set(srcs
   Grid_PFIO.F90
   GeomCatagorizer.F90
   pFIOServerBounds.F90
+  DataCollection.F90
+  DataCollectionVector.F90
+  DataCollectionManager.F90
   )
 
 esma_add_library(${this}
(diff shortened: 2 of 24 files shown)


hconfig_utils/hconfig_get_private.F90
```diff
@@ -146,4 +164,104 @@ function make_fmt(descriptor) result(fmt)
...
 
+#include "mapl_hconfig_macros.h"
+
+#define _SUB_ mapl_get_value_I4
+#define _FTYPE_ integer(kind=ESMF_KIND_I4)
+#define _ESMF_FUNC_ ESMF_HConfigAsI4
+#define _TYPESTRING_ "I4"
+#define _EDIT_DESCRIPTOR_ "G0"
+#include "mapl_hconfig_get_value_template.h"
+#include "mapl_hconfig_get_value_template_undef.h"
```
tclune, line 175: Don't you want/need some #undef's after each block here?    There is some danger of definitions leaking through and would be very hard to debug.  Esp. concerned about _IS_LOGICAL_ leaking down.

hconfig_utils/hconfig_get_private.F90
```diff
@@ -146,4 +164,104 @@ function make_fmt(descriptor) result(fmt)
 
    end function make_fmt
 
+#include "mapl_hconfig_macros.h"
+
+#define _SUB_ mapl_get_value_I4
+#define _FTYPE_ integer(kind=ESMF_KIND_I4)
+#define _ESMF_FUNC_ ESMF_HConfigAsI4
+#define _TYPESTRING_ "I4"
+#define _EDIT_DESCRIPTOR_ "G0"
```
tclune, line 173: Since these are not user-level, we may want to future proof this layer by using tokens that are not "reserved'' by the language.  Fortran will soon reserve almost everything that starts with a leading underscore. (unfortunately)

No action now, but think about it.

tclune (dismissed):
Approved.  But consider some of my inline comments.

hconfig_utils/hconfig_get_private.F90
darianboggs: Makes sense. I want to use tokens that are not valid Fortran identifiers for macros, so that compile will fail if the tokens are not replaced. I guess I could use a valid identifier. However, if I use `FOO_` and there is a variable `foo_`, it might not fail even though I don't want it to use the variable `foo_`.

What would be a non-reserved, non-valid token pattern?

hconfig_utils/hconfig_get_private.F90
darianboggs: `_IS_LOGICAL_` is undefined in `mapl_hconfig_get_value_template_undef.h`. I kept the logical calls at the end just in case.

hconfig_utils/hconfig_get_private.F90
```diff
@@ -1,30 +1,48 @@
...
-      module procedure :: get_value_r8
-      module procedure :: get_value_string
-      module procedure :: get_value_logical
-      module procedure :: get_value_i4seq
-      module procedure :: get_value_i8seq
-      module procedure :: get_value_r4seq
-      module procedure :: get_value_r8seq
-      module procedure :: get_value_logical_seq
-   end interface get_value
+   public :: HConfigParams, DEFAULT_TAG, ELLIPSIS
```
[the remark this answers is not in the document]
darianboggs: Done

tclune (approved):

PR #3948, opened 2025-08-05
IO in ExtData3g

bena-nasa:
Hook actually IO into ExtData3g to read files.

With this I should be able to start implementing some the "ExtData tests" that we have now for 2G.

## Types of change(s)

## Checklist

### DIFF
--- a/gridcomps/ExtData3G/PrimaryExport.F90
+++ b/gridcomps/ExtData3G/PrimaryExport.F90
@@ -28,7 +28,7 @@ module mapl3g_PrimaryExport
          procedure :: get_export_var_name
          procedure :: get_bracket
          procedure :: update_my_bracket
-         procedure :: append_reader
+         procedure :: append_state_to_reader
    end type
 
    interface PrimaryExport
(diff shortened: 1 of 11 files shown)


gridcomps/ExtData3G/PrimaryExport.F90
```diff
@@ -49,6 +55,9 @@ function new_PrimaryExport(export_var, file_var, file_selector) result(primary_e
       call right_node%set_node_side(NODE_RIGHT)
       call primary_export%bracket%set_node(NODE_LEFT, left_node)
       call primary_export%bracket%set_node(NODE_RIGHT, right_node)
+      call file_selector%get_file_template(file_template)
+      primary_export%client_collection_id = i_clients%add_data_collection(file_template, _RC)
```
tclune, line 59: Are there issues if `file_template` is not allocated?  I see that in the getter in the line above that is a possibility, so wanted to double check.

tclune (approved):
Looks pretty good.  But do look at the questions/comments I inlined.  If I need to approve again, I will.

gridcomps/ExtData3G/PrimaryExport.F90
tclune: ok - but then the implementation of `get_file_template()` is a bit weird in that it checks for allocatable but does not fail if not allocated.      Feels only half guarded at that point.  But minor issue for now.

Issue #3969, opened 2025-08-08
Bug when time in `cap_restart.yaml` is later than the `start` in the `cap.yaml` file in MAPL3


bena-nasa:
As the description says, I was finding when the time in the `cap_restart.yaml` was after `start` in `cap.yaml` in an application using the Cap and CapGridComp, the run method of capGridComp and therefore all it's children was not getting executed at all. 

@tclune and I traced it to the fact that in `run_user.F90` in the outermetacomp directory the run alarm in `run_user` was not ringing.

This in turn was because in the end of `initialize_set_clock.F90`
https://github.com/GEOS-ESM/MAPL/blob/release/MAPL-v3/generic3g/OuterMetaComponent/initialize_set_clock.F90#L104

the run alarm for the component is turned off after it is created if the user runtime is less than the current time.

The user runtime is the `start` in `cap.yaml` and the current time is the contents of `cap_restart.yaml`

Clearly this behavior is a bug. Of course the time in the `cap_restart.yaml` will be greater than the start time in cap 99% of the time. Commenting out the line to return the alarm off after creation fixed it.

But neither Tom or I could remember why this turning off of the alarm was being done. So we need to understand why this was done in the first place.

tclune:
I think it may be due to the fact that the alarm starts in a ringing state when the start time is in the past or some such. (Which may no longer arise?).

The MAPL2 solution was to tick the clock back and then forward.  But in MAPL3 that's a bit clunky given the hierarchy of clocks.

PR #3980, opened 2025-08-12
Feature/bmauer/mapl histext systest

### LINKED ISSUE #3968: Bug in `ESMF_Time_Utilities.F90` in MAPL3
I ran into an issue when playing with History3G in my `MAPL_Component_Driver.x` when trying to set a 30 minute timestep in cap and hourly History output.

The issue is this routine in `ESMF_Time_Utilities.F90`

https://github.com/GEOS-ESM/MAPL/blob/release/MAPL-v3/esmf_utils/ESMF_Time_Utilities.F90#L42C19-L42C20

This `interval_and_offset_are_compatible` is called when setting up a components clocks. Essentially comparing the child to parents time interval.

In this function this `cannot_compare` routine has a bug. The code does an `ESMF_TimeGet` with all the options, year, month, day, hour etc... and puts each into an array, then creates logical arrays for which elements are zero, then compares them.

So if for example one time interval is 30 minutes and the 2nd is 1 hour, this routine will say that these 2 are not compatible which clearly not right.

For 1 hour ESMF_TimeGet returns
y=0,m=0,d=0,h=1,mn=0,s=0 ...
For 30 minutes
y=0,m=0,d=0,h=0,mn=30,s=0 ...

Now if we make arrays by comparing each element to 0 we get

[true, true, true, false, true, true ...] for 1 hour

and 

[true, true, true, true, false, true ...] for 30 minutes

And the routine says, these logical arrays are not equal, so the time intervals aren't compatible which they should be...

bena-nasa:
Implements case01 of the History/ExtData test cases with the MAPL3 component driver

Redid how time is handled in History in multiple step files. Simply append to the time vector in a file rather than try to compute the time vector.
This way history only adds a time to the file when History actually writes.

Have MAPL_Component_Driver use CapGridComp

Fixes a few issues:
Fixes #3968 
Sidesteps #3969 because ESMF alarms just don't work half the time

Also found a few other bugs in the cap and generic layers so fixed those.

Added new functionality to my component driver to run at specific times.

## Types of change(s)

## Checklist


tclune (approved):
glad to see that some of the cap duplication has been removed.

tclune (approved):

PR #4000, opened 2025-08-15
added pfaf index to MAPL_Locstream

weiyuan-jiang:
## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Added pfaf_index to MAPL_Losctream
+- Enforced present of grid when reading binary tile file
+
 ### Removed
 
 ### Deprecated
--- a/base/MAPL_LocStreamMod.F90
+++ b/base/MAPL_LocStreamMod.F90
@@ -31,6 +31,7 @@ use MAPL_CommsMod
 use MAPL_HashMod
 use MAPL_ShmemMod
 use MAPL_ExceptionHandling
+use MAPL_EASEConversion
 use, intrinsic :: iso_fortran_env, only: REAL64, INT64
 use mpi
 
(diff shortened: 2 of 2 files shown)


tclune:
I want @atrayano to look at this.  The change is very specific to Land and LocStream is meant to be somewhat generic.

gmao-rreichle:
> I want @atrayano to look at this. The change is very specific to Land and LocStream is meant to be somewhat generic.

@tclune, @atrayano : The change is needed for the addition of river routing (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1143) to the model.  For now, we're only targeting land tiles for routing (because we've always had the Pfafstetter network for land), but the idea is to also include runoff routing from landice tiles, now that GEOSldas can include landice tiles.  Lauren is working on creating Pfafstetter indices for landice tiles.  So the change isn't specific to land per se, although it is specific to Surface.  Admittedly, ocean tiles don't have a Pfafstetter index (pfaf_index).  I'm unsure how we'll handle lake tiles when we get to crossing that bridge.  Currently, there are "fake" catchments within large lakes such as the Great Lakes.  These "fake" catchments assist with routing water through the lakes.

tclune:
If I have understood correctly, the "ESMF way" that this would be handled would be for the Pfafstetter indices to be another ESMF_Field on the same ESMF_LocStream.    (I.e., the loc stream remains general, and it is just another field that is created and managed associated with that locstream.) 

 I am less familiar with how this analog is done with MAPL_LocStream, but am guessing that there is no analog to a "Field" here, and instead MAPL_LocStream is a fusion between geometry and data, and more hardwired in practice.   I'll try to discuss with @atrayano asap so that the question does not linger.
 
 Since MAPL3 has a better way to deal with this, I probably am not opposed to this change.   But if there is a better way even in MAPL2 I'd like to push for that.

tclune (approved):

weiyuan-jiang:
That PR is  not necessary.

PR #4008, opened 2025-08-20
Kludge to prevent crashes on NAG + OSX

tclune:
Clearly memory is corrupted, but the problems sometimes go away with mere print statements.

Typical error messages were:

```
Runtime Error:
/Users/tclune/swdev/GEOS-ESM/MAPL3/shared/ErrorHandling.F90, line 117:
Incorrect interface block - Function MAPL_ERRORHANDLING:MAPL_VERIFY
does not return a POINTER
```

Failing tests were scenarios:
  - history_wildcard
  - extdata_1
  - statistics

No idea what the actual underlying problem is after days of investigating.

## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/OuterMetaComponent/initialize_read_restart.F90
+++ b/generic3g/OuterMetaComponent/initialize_read_restart.F90
@@ -26,6 +26,7 @@ contains
       character(:), allocatable :: filename
       type(esmf_Time) :: currTime
       integer :: status
+      class(Logger), pointer :: user_logger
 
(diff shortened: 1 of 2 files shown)


tclune:
Can also add that the symptoms appeared with PR #3999 - but that appears to not be directly connected.

tclune:
> @tclune do you want to enable the `extdata_1` with this PR? It is still commented out in `Test_Scenarios.pf`

Yes - that was an oversight.

mathomp4 (approved):

PR #4033, opened 2025-09-04
Computing run alarm in MAPL3

bena-nasa:
This implements changes in how the user run alarms are created to support how History needs to operate. 

Also implements a simple alarm class because ESMF Alarms are just too buggy to work reliably.

Add a shift and ref_time option that can be used in the outermetacomp when creating alarms.

In practice History will set these in it's children's hconfig.

The point of this is so that History runs when we want and uses the right time since for normal model runs we want History to run -dt before the time the user asks for output because the clock ticks after history.

I had to introduce a way to not do this for my testing framework.

Finally probably need better names for these parameters, but I'm not good at thinking of names.

## Types of change(s)

## Checklist

### DIFF
--- a/alarm/CMakeLists.txt
+++ b/alarm/CMakeLists.txt
@@ -6,9 +6,6 @@ set(srcs
 
 list (APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")
 
-if (BUILD_WITH_PFLOGGER)
-  find_package (PFLOGGER REQUIRED)
-endif ()
 
 esma_add_library(${this}
   SRCS ${srcs}


generic3g/OuterMetaComponent/initialize_set_clock.F90
```diff
@@ -80,35 +81,61 @@ end subroutine set_children_outer_clock
...
+      type(ESMF_Time), intent(in) :: time
+      type(ESMF_TimeInterval), intent(in) :: time_interval
+      integer, optional, intent(out) :: rc
+
+      integer :: status, year, month, day
+
+      call ESMF_TimeGet(time, yy=year, mm=month, dd=day, _RC)
+      call ESMF_TimeSet(new_time, yy=year, mm=month, dd=day, h=0, m=0, s=0, _RC)
+      new_time=new_time+time_interval
+      _RETURN(_SUCCESS)
```
tclune, line 138:
```suggestion

      _RETURN(_SUCCESS)
```
(preferred style)

alarm/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+list (APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")
+
+if (BUILD_WITH_PFLOGGER)
+  find_package (PFLOGGER REQUIRED)
+endif ()
+
+esma_add_library(${this}
+  SRCS ${srcs}
+  DEPENDENCIES MAPL.shared PFLOGGER::pflogger ESMF::ESMF
```
tclune, line 15:
```suggestion
  DEPENDENCIES MAPL.shared ESMF::ESMF
```
No dependency on pflogger is needed here.

PR #4038, opened 2025-09-09
Multiple Rules in ExtData3G

bena-nasa:
Implement the multi rule capability in ExtData3G
Also turned off ExtData1G tests because they take a lot of time and no one should be using 1G by the time anyone sees this!

## Types of change(s)

## Checklist


gridcomps/ExtData3G/ExtDataGridComp.F90
```diff
@@ -144,6 +172,49 @@ subroutine run(gridcomp, importState, exportState, clock, rc)
...
+           num_rules => this%rules_per_export%at(i)
+           exit
+        end if
+     enddo
+     _ASSERT(found,"ExtData no item with basename '"//TRIM(base_name)//"' found")
+
+     item_index = -1
+     if (num_rules == 1) then
+        item_index = i_start
+     else if (num_rules > 1) then
```
tclune, line 204: Is this conditional needed? Looks like the else block might work for both.

bena-nasa: yes I need the conditional, don't want to loop if there's only one rule, nothing to do and the starting times for an item won't even be defined if there's only one rule

PR #4040, opened 2025-09-09
Allow lookups to return as value

darianboggs:
## Types of change(s)

## Checklist

## Description
ACG2 allows aliases for row values in the spec file. For example, a user could use 'xyz' for DIMS column instead of the value used in MAPL which is 'MAPL_DimsHorzVert'. However, ACG2 would not recognize 'MAPL_DimsHorzVert', previously. This PR extends ACG2 to recognize both the alias and the actual MAPL identifier. This holds for any column that allows aliases for the column values.

In addition, this PR improves the exception/error handling. More exceptions and errors produce more meaningful information about the state and short_name for any rows in the spec file that produce errors during processing.

Also, the PR adds two new command-line arguments: '--debug' and '--skip-checks'. '--debug' produces messages to standard output for any problems that occur when ACG2 runs. '--skip-checks' skips the checks on the compatibility of the values in a spec file row. The default behavior remains running the tests. '--skip-checks' allows a more permissive run for the sake of debugging ACG problems and generating include files based on automatically written spec files.

### DIFF
--- a/Apps/MAPL_GridCompSpecs_ACG.py
+++ b/Apps/MAPL_GridCompSpecs_ACG.py
@@ -437,8 +437,8 @@ def get_args():
     parser.add_argument("--" + LONGNAME_GLOB_PREFIX, dest=LONGNAME_GLOB_PREFIX,
                         action="store", nargs='?', default=None,
                         help="alternative prefix for long_name substitution")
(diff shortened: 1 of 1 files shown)


tclune (dismissed):
Did not look closely, but nothing jumped out.

I do note that the CI is failing though.

atrayano (approved):

PR #4056, opened 2025-09-15
A few error message improvements

mmanyin:
## Types of change(s)
- [ x] Trivial change (affects only documentation or cleanup)

## Checklist
- [ x] Tested this change with a run of GEOSgcm

## Description
Added a few error messages that proved helpful during a recent overhaul of CTM


tclune (commented):
This is fine.

I would be very interested in finding a better way to insert high-level error messages such as was done here.    The usual approach is to give a useful message at the lowest code level where the error occurred.  But often there is not sufficient information at that level to compose a message that has any meaning to the end user.   OTOH, at higher levels, we have this nifty `_RC` macro ...

PR #4083, opened 2025-09-22
Fixes #4059

### LINKED ISSUE #4059: Regrid brackets only when attribute indicates
The plan for this had always been to put something in FieldBundleSet (and Get) to specify when a field node in a bracket has changed.   

The tricky part is how to *clear* this entry.    In theory this can be done inside of RegridTransform, but this will fail if the same export from ExtData is regridded to distinct geoms.     Instead, I think ExtData will need to clear the setting on entry.

bena-nasa:
Fixes #4059 

This allows extdata to set a flag that the regridding can queue off of to decide whether the bracket needs regridding.
As usual confirmed it work via prints but couldn't think of a good test.

I didn't know how far to take this. I added a new FieldBundle method to detect if the attribute is present, but only for the new one, I didn't do it for every option for fieldGet/Set

Also this is only for fieldBundles so no way if you are just regridding a field you could do something similar.

This was really the minimum needed...
Thought I would put it out here before going more crazy

## Types of change(s)

## Checklist

### DIFF
--- a/generic3g/transforms/RegridTransform.F90
+++ b/generic3g/transforms/RegridTransform.F90
@@ -15,6 +15,10 @@ module mapl3g_RegridTransform
    private
 
    public :: RegridTransform
+!   public :: COUPLER_IMPORT_NAME
+!   public :: COUPLER_EXPORT_NAME
+! import[1]
+! export[1]
 
    type, extends(ExtensionTransform) :: ScalarRegridTransform
       type(ESMF_Geom) :: src_geom
@@ -34,6 +38,8 @@ module mapl3g_RegridTransform
       module procedure :: new_ScalarRegridTransform
    end interface RegridTransform
 
+!   character(len=*), parameter :: COUPLER_IMPORT_NAME = 'coupler_import'
+!   character(len=*), parameter :: COUPLER_EXPORT_NAME = 'coupler_export'
 contains
 
    function new_ScalarRegridTransform(src_geom, dst_geom, dst_param) result(transform)
@@ -73,8 +79,8 @@ contains
 
       regridder_manager => get_regridder_manager()
 
-      this%src_geom = get_geom(importState, 'import[1]')
-      this%dst_geom = get_geom(exportState, 'export[1]')
+      this%src_geom = get_geom(importState, COUPLER_IMPORT_NAME)
+      this%dst_geom = get_geom(exportState, COUPLER_EXPORT_NAME)
       spec = RegridderSpec(this%dst_param, this%src_geom, this%dst_geom)
       this%regrdr => regridder_manager%get_regridder(spec, _RC)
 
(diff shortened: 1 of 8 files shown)


field_bundle/FieldBundleInfo.F90
```diff
@@ -209,4 +226,28 @@ end function to_string
              
    end subroutine fieldbundle_set_internal
 
+   subroutine fieldbundle_is_present_internal(info, unusable, namespace, do_regrid_transform, rc)
```
tclune, line 229: I don't immediately see why this procedure is needed.

generic3g/transforms/RegridTransform.F90
```diff
@@ -142,13 +143,46 @@ subroutine update(this, importState, exportState, clock, rc)
...
+   function check_do_transform(fb_in, fb_out, rc) result(do_transform)
+      logical :: do_transform
+      type(ESMF_FieldBundle), intent(in) :: fb_in
+      type(ESMF_FieldBundle), intent(in) :: fb_out
+      integer, optional, intent(out) :: rc
+
+      logical :: do_trans_in, do_trans_out, present_in, present_out 
+      integer :: status
+
+      call MAPL_FieldBundleIsPresent(fb_in, do_regrid_transform=present_in, _RC)
```
tclune, line 165: OK - I see "why" now, but ...   Can't we simply arrange it so that there is always a value?    Just add a call to FieldBundleSet() in BracketClassAspect::create().

generic3g/transforms/RegridTransform.F90
```diff
@@ -142,13 +143,46 @@ subroutine update(this, importState, exportState, clock, rc)
...
+      if (present_in) then
+         call MAPL_FieldBundleGet(fb_in, do_regrid_transform=do_trans_in, _RC)
+      end if
+      call MAPL_FieldBundleIsPresent(fb_out, do_regrid_transform=present_out, _RC)
```
tclune, line 169: No - the output field should not be responsible for controlling do_regrid transform.   We probably need to discuss.

field_bundle/FieldBundleGet.F90
```diff
@@ -53,6 +54,7 @@ subroutine bundle_get(fieldBundle, unusable, &
       character(:), optional, allocatable, intent(out) :: standard_name
       character(:), optional, allocatable, intent(out) :: long_name
       type(StateItemAllocation), optional, intent(out) :: allocation_status
+      logical, optional, intent(out) :: do_regrid_transform
```
tclune, line 57: I'm thinking that we can probably find a better name here, as it does not quite reflect the intent.

We are not so much telling upstream to regrid (e.g., if the geoms are the same, it should not trigger a regrid), as saying that the bundle data has been updated.

Something like `bracket_data_changed`.  (Not wed to the details here.)

tclune (approved):

PR #4085, opened 2025-09-25
Initialize field to zero

pchakraborty:
## Types of change(s)

## Checklist

## Description

Instead of initializing the field in `FieldClassAspect::allocate`, we initialize the field right after its completed, in `MAPL_FieldEmptyComplete`

### DIFF
--- a/field/FieldCreate.F90
+++ b/field/FieldCreate.F90
@@ -73,7 +73,7 @@ contains
       _UNUSED_DUMMY(unusable)
    end function field_create
 
-   subroutine field_empty_complete( field, &
+   subroutine field_empty_complete(field, &
         typekind, unusable, &
         gridToFieldMap, ungridded_dims, &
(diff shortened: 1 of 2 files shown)


field/FieldCreate.F90
```diff
@@ -116,6 +116,9 @@ subroutine field_empty_complete( field, &
            ungriddedUBound=bounds%upper, &
            _RC)
 
+      ! Initialize field to zero
+      call ESMF_FieldFill(field, dataFillScheme="const", const1=0.d0, _RC)
```
tclune, line 120: As per the discussion yesterday, I think that setting the initial value to 0 will tend to hide bugs.    In the use case you mentioned, the 2nd field should not have been referenced, and indeed should not even exist.

If you need this as a short term workaround, then fine.  But please then come back to either delete or initialize to some absurd value.

PR #4144, opened 2025-10-22
Changes to fix the crashing when calling the reduce function where th…

aoloso:
…e MPI user defined type and operator function were returning MPI_TYPE_NULL and MPI_OP_NULL

## Types of change(s)
- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [X ] Ran the Unit Tests (`make tests`)

## Description
The call to main_prof%reduce() was crashing because the user defined type and operator that were subsequently passed to MPI_Reduce had values of MPI_TYPE_NULL and MPI_OP_NULL respectively. The issue was traced to call to prof%finalize before the reduce call. The call to finalize appeared to be freeing both the data type and the operator prematurely. The fix was to replace finalize with stop.

## Related Issue
4143


mathomp4:
I think the one failing job is our transient. I've asked Github to re-run

mathomp4 (approved):

PR #4300, opened 2026-01-09
ExtData2G backwards compatibility with non-CF vertical coord

lizziel:
## Types of change(s)

## Checklist
- [X] Tested this change with a run of GCHP

## Description

This update is needed for ExtData2G compatibility with GCHP due to use of input files that are not CF-compliant. These files have vertical dimension variable with name 'lev', units 'level', and no positive attribute.

Changes are:
1. Categorize dimensions with name `lev` as vertical coordinate in ExtData2G (update provided by Atanas) as an alternative criteria to CF definition of vertical coordinate.
2. Assign `positive` attribute in ExtData2G for the case of vertical coordinate with no `positive` attribute in file and without pressure units. Do this by looking at `standard_name`, `units,` and order of dim values. Stop with error if `standard_name` is not `level` or `levels`,  and units are not `level` or `1`. Assumes `positive:up` if level values are increasing, and `positive:down` if decreasing.

## Related Issue

None

### DIFF
--- a/vertical/VerticalCoordinate.F90
+++ b/vertical/VerticalCoordinate.F90
@@ -113,7 +113,7 @@ contains
             standard_name = coord_var%get_attribute_string("standard_name")
             ! metadata combinations that imply integer levels
             if ( any(standard_name == ["level ", "levels"])  .and. &
-                 any(tmp_units == ["1    ", "level"])) then
+                 any(temp_units == ["1    ", "level"])) then
                vertical_coord%positive = "up"
                if (vertical_coord%levels(1) >= vertical_coord%levels(2)) then
                   vertical_coord%positive = "down"


vertical/VerticalCoordinate.F90
```diff
@@ -103,6 +105,27 @@ function new_verticalCoordinate(metadata, var_name, rc) result(vertical_coord)
...
+            ! metadata combinations that imply integer levels
+            if ( ((trim(standard_name) == "level" )  .or.  &
+                  (trim(standard_name) == "levels")) .and. &
+                 ((trim(temp_units) == "1"     ) .or. &
+                  (trim(temp_units) == "level")) ) then
+               if (vertical_coord%levels(1) < vertical_coord%levels(2)) then
+                  vertical_coord%positive = "up"
+               else
+                  vertical_coord%positive = "down"
+               endif
```
tclune:
```suggestion
               vertical_coord%positive = "up"
               if (vertical_coord%levels(1) >= vertical_coord%levels(2)) then
                  vertical_coord%positive = "down"
               endif
```

vertical/VerticalCoordinate.F90
```diff
@@ -103,6 +105,27 @@ function new_verticalCoordinate(metadata, var_name, rc) result(vertical_coord)
...
+         ! now test if no positive attribute and does not have pressure units
+         ! for backwards compatibility with non-cf files
+         if ((.not. coord_var%is_attribute_present("positive")) .and. &
+              (.not. has_pressure_units)) then
+            standard_name = coord_var%get_attribute_string("standard_name")
+            ! metadata combinations that imply integer levels
+            if ( ((trim(standard_name) == "level" )  .or.  &
+                  (trim(standard_name) == "levels")) .and. &
+                 ((trim(temp_units) == "1"     ) .or. &
+                  (trim(temp_units) == "level")) ) then
```
tclune: Fortran string comparison ignores trailing spaces,  so these expressions can be made a bit more legible in the following manner:
```suggestion
            if ( any(standard_name == ["level ", "levels"])  .and. &
                 any(tmp_units == ["1    ", "level"])) then
```

lizziel:
No worries. I should have done a more thorough review and tested it before committing.

PR #4395, opened 2026-02-12
add the options to create locstream with local_id

weiyuan-jiang:
## Types of change(s)

## Checklist

## Description

Add an option local_id to create locstream from a simple Nx1 grid


base/MAPL_LocStreamMod.F90
```diff
@@ -2922,11 +2922,12 @@ end subroutine MAPL_GridCoordAdjust
 
 !A special subroutine for Nx1 Grid in river-routing grid comp
 !Some information in the locstream is not filled
-subroutine MAPL_LocstreamCreateSimple(Locstream, grid, rc)
+subroutine MAPL_LocstreamCreateSimple(Locstream, grid, local_id, rc)
```
tclune: Can you provide a use case where "local_id" would not be a contiguous list of integers?

I really don't think this option is needed.

PR #4398, opened 2026-02-13
Use ESMF_HConfigAdd, instead of ESMF_HConfigSet

pchakraborty:
## Types of change(s)

## Checklist

## Description
Use ESMF_HConfigAdd, instead of ESMF_HConfigSet, to add a non-duplicate key to the merged hconfig. ESMF_HConfigSet’s behavior seems to have changesd, because it used to work earlier. This needs to be confirmed.

Also, there was a duplicate version of hconfig_merge in add_child_by_spec.F90. That was removed in favor of the merge routine from HConfigUtilities.F90


pchakraborty:
> > > I do not see where the merge functionality went? The behavior is very precisely prescribed, even if the implementation got it wrong.
> > 
> > 
> > It’s using the `merge_hconfig` routine from `hconfig_utils/HConfigUtilities.F90`
> 
> AH. That does not show in the PR. Let me go look at that code to be sure it has what I want, and then I'll approve.



> > > I do not see where the merge functionality went? The behavior is very precisely prescribed, even if the implementation got it wrong.
> > 
> > 
> > It’s using the `merge_hconfig` routine from `hconfig_utils/HConfigUtilities.F90`
> 
> AH. That does not show in the PR. Let me go look at that code to be sure it has what I want, and then I'll approve.

Oh. It’s visible on my side though - in the `use` section.

PR #4476, opened 2026-03-03
New home for MAPL_GetHorzIJIndex - in geom, as MAPL_GridGetHorzIJIndex

pchakraborty:
## Types of change(s)

## Checklist

## Description

Ported `MAPL_GetHorzIJIndex` from `base` to `geom`. Renamed as `MAPL_GridGetHorzIJIndex`

> [!NOTE]  
> Goal is for each grid kind (lat-lon, cubed-sphere, etc.) to have its own implementation of `GridGetHorzIJIndex`. This is to be addressed in a separate PR

## Related Issue
 #4475


geom/tests/Test_GridGetHorzIJIndex.pf
```diff
@@ -0,0 +1,365 @@
...
+
+      ! Set input points (in radians)
+      lonR8(1) = 0.0d0
+      latR8(1) = 0.0d0
+      lonR8(2) = 1.5707963267948966d0  ! pi/2
+      latR8(2) = 0.7853981633974483d0  ! pi/4
+      lonR8(3) = 3.1415926535897932d0  ! pi
+      latR8(3) = -0.7853981633974483d0  ! -pi/4
+      lonR8(4) = -1.5707963267948966d0  ! -pi/2
+      latR8(4) = 1.5707963267948966d0  ! pi/2 (north pole)
```
tclune: Not worth fixing, but ... why not use `pi/2` and such directly rather than putting in the decimal representation?   Apologies if this was a pre-existing test that you only copied.

PR #4525, opened 2026-03-13
Feature/wdboggs/#4292 test verticalgrid aspect

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR implements tests for the `update_payload` and `update_from_payload` methods for the `VerticalGridAspect` type.

## Related Issue
#4292


generic3g/specs/VerticalGridAspect.F90
```diff
@@ -513,6 +513,9 @@ subroutine update_payload(this, field, bundle, state, rc)
       mirror_grid = MirrorVerticalGrid()
       vgrid = mirror_grid
       if(.not. this%is_mirror()) then
+#ifdef __GFORTRAN__
+         deallocate(vgrid)
+#endif
```
tclune: Ah.  Yes - this is a common problem with gFortran.  I've tried to be good about using this pattern.  

Note that you do not need the #ifdef.  It is ok to delete on all compilers.

field/FieldInfo.F90
```diff
@@ -150,7 +150,10 @@ subroutine field_info_set_internal(info, unusable, &
       if (present(quantity_type_metadata)) then
          quantity_info = quantity_type_metadata%make_info(_RC)
          call MAPL_InfoSet(info, namespace_ // KEY_QUANTITY_TYPE_METADATA, quantity_info, _RC)
+         _HERE
```
atrayano:
```suggestion
```

darianboggs:
Atanas fixed the deallocation problem. `TestVerticalGridAspect test_update_field_no_vgrid` is failing because the `ESMF_Field` (payload) initially has 3 dimensions, and the `VerticalGridAspect` has no `VerticalGrid` and therefore represents a purely horizontal grid. `FieldSet` and `FieldDelta` don't handle the resulting change from rank-3 to rank-2 in the physical memory.

PR #4549, opened 2026-03-20
Fix Monthly and other averages in History/Statistics

bena-nasa:
This fixes issues I found when attempting to extend History and the Statistics GridComp to do longer averages such as monthly or DJF etc. Partially done with Claude and opencode.

Just wanted to get this out, needs cleanup as I say below, but the core logic is right. It is just a questions if things should move and how this templating needed to specify the reference time should be specified. I have something that works but this can change as I say below.

The big thing was that the HistoryCollectionGridComps need to run if they would run "next".
Also, this notion of a time offset needs to be generalized to create the correct alarms.
You need to be able to specify a reference, month, day, and (time hours, minutes etc...)

I changed this so that in history and statistics you specify a grads style template with an underscore between the date part and time now. Basically when constructing the initial runtime for the history and statistics alarms it looks looks at the template and you are saying **I want the alarm to ring at such a time and day as an example, but it can be any month or year**. Any place you have a grads token it uses the current item from the clock time (year, month, day, hour, minute). Anytime that it spots a non-token, it uses that in the time instead of the item from the clock time.

For example if you want daily averages relative to 21z you specify

%y4%m2%d2_210000

If you want a monthly average from the start of the month

%y4%m00_000000

if you want a DJF

%y41200_0000

This is ugly, maybe we should have a tokenized ISO time instead...

%y4-%m2-%d2T%h2:%n2:00

or 

YYYY-MM-DDTHH:MM:SS

just to throw some things out there.

There are still things to do but this is getting to be a big PR already

To do in future PRs
- I had to disable the timestep compatibility check in `initialize_set_clock.F90` this needs to be generalized
- There is duplicate routines in `initialize_set_clock.F90` and `statisticsGridComp.F90` to do the detemplating of the template used to create the run alarms for history and statistics.
- I want to replace ESMF_Alarm with SimpleAlarm in StatisticsGridComp

## Types of change(s)

## Checklist


generic3g/OuterMetaComponent/initialize_set_clock.F90
```diff
@@ -36,8 +35,8 @@ module recursive subroutine initialize_set_clock(this, outer_clock, unusable, rc
 
       user_offset = this%user_offset
 
-      call check_compatibility(user_timestep, timeStep, compatible, offset=user_offset, _RC)
-      _ASSERT(compatible, 'The user timestep and offset are not compatible with the outer timestep.')
+      !call check_compatibility(user_timestep, timeStep, compatible, offset=user_offset, _RC)
+      !_ASSERT(compatible, 'The user timestep and offset are not compatible with the outer timestep.')
```
tclune: I understand the intent here, but if we keep this PR I almost want a warning to print that we should come back to this.

PR #4557, opened 2026-03-24
Change expr key in var_list in history

bena-nasa:
This changes the key that specifies the component/short_name in a collection var_list from `expr` to `source`
## Types of change(s)

## Checklist


Tests/MAPL3G_Component_Testing_Framework/test_cases/case06/history2.yaml
```diff
@@ -19,4 +19,4 @@ collections:
     geom: *geom1
     time_spec: *one_hour
     var_list:
-      E1: {expr: E_1}
+      E1: {source: E_1}
```
tclune: Not urgent, but if we have real cases where we only need the one value, maybe it is worth allowing a scalar.  E.g.,

```suggestion
      E1: dyn.E1
```

Issue #4594, opened 2026-04-01
ACG: support use_field_dictionary column in RC files


tclune:
## Background

PR #4577 introduced the Field Dictionary (FD) singleton and `make_VariableSpec` integration.
The follow-on redesign (branch `feature/#4440-field-dict-core-infrastructure`) makes FD
lookup **opt-in per variable** via `use_field_dictionary=.true.` in `make_VariableSpec`.

## Problem

Components that use the Auto-Code Generator (ACG) to emit `make_VariableSpec` calls from
RC files currently have no way to opt in to FD lookup on a per-variable basis.

## Requested feature

Add a `use_field_dictionary` boolean column to the ACG RC-file spec schema.
When set to `true` for a variable, the generated `make_VariableSpec` call should include
`use_field_dictionary=.true.`.

Default should be `false` (preserving current behaviour for all existing RC files).

## Acceptance criteria

- ACG reads an optional `use_field_dictionary` column (boolean, default `.false.`).
- Generated code passes `use_field_dictionary=.true.` only when the column is `true`.
- Existing RC files with no such column continue to compile and behave identically.
- Documentation for RC-file schema updated to describe the new column.

## Related

- Closes (partial): #4440
- See also: #4578 (strict-mode follow-on)

PR #4626, opened 2026-04-08
Feature/wdboggs/#4594 add update field dictionary acg3

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR implements the ACGv3 changes in the related issue. The documentation updates will be in a separate PR.

## Related Issue
#4594


Apps/MAPL_GridCompSpecs_ACGv3.py
```diff
@@ -25,6 +25,7 @@
 
 ARGS = 'args' 
 AS = 'as'
+BOOL = 'bool'
```
tclune, line 29: This is our first `bool` option?  That surprises me a bit.

Apps/MAPL_GridCompSpecs_ACGv3.py
```diff
@@ -183,7 +189,9 @@ def get_options(args):
         'prec': PRECISION,
         'vloc': VSTAGGER,
         'vlocation': VSTAGGER,
-        'add2export': ADD_TO_EXPORT
+        'add2export': ADD_TO_EXPORT,
+        'field dictionary': USE_FIELD_DICTIONARY,
+        'field_dictionary': USE_FIELD_DICTIONARY
```
tclune, line 208: Why is this entry duplicated?

tclune (changes requested):
There was one duplicated entry in a dictionary.  Once that is resolved, I will approve.

Apps/MAPL_GridCompSpecs_ACGv3.py
darianboggs: ### Previous:
`LOGICAL` transformed logical-like strings, like ".T.", "TRUE", ".F.", etc., into Fortran logicals.

### Now:
- `BOOL` transforms logical-like strings into python `bool` values.
- `LOGICAL` transforms python `bool` values into Fortran logicals.
- `STRLOGICAL` transforms logical-like strings into Fortran logicals.

Previously, ACGv3 transformed logical-like strings to Fortran logicals, and then it transformed those into python `bool` values for conditional logic.

Apps/MAPL_GridCompSpecs_ACGv3.py
darianboggs: Those are two different aliases for `use_field_dictionary`. We set a precedent with `long_name` for which there is an alias `long name`.

tclune (dismissed):

pchakraborty (approved):
LGTM

PR #4632, opened 2026-04-08
Feature/bmauer/fixes #4598

### LINKED ISSUE #4598: Port `MAPL_Bundle_Read` and `MAPL_Bundle_Write` routines to MAPL3
Background
`MAPL_Bundle_read` and `MAPL_Bundle_write` are two helper procedures that live in the GriddedIO directory that were meant to replace the old `MAPL_CFIORead/MAPL_CFIOWrite` calls. These were just simple routines that allow users to read and write `ESMF_Bundles` to and from a file in simple utility programs for example. Both of these leveraged this `GriddedIO` class that 
1. Provided mechanism to interact with the pfio server to make the appropriate calls to stage or retrieve data from the server
2. Regrid if needed.

`MAPL_Bundle_write` was very simple, it wrote a bundle was is, no transformation done.

`MAPL_Bundle_read` was a bit more complicated. If the bundle that was passed in to fill was empty it populated the empty bundle using the file. If the bundle was not empty it would check that the fields corresponded to variables in the file and read it but the griddedIO layer would by design regrid if the grids did not match.

Now in MAPL3 I have made a new GeomIO layer that handles the PFIO interactions but does not do regridding since for History and ExtData the generic layer handles that and it was not necessary. 

So I propose we redo `MAPL_Bundle_read` and `MAPL_Bundle_write` to use the `GeomIO` layer. Since the writing was not doing any regridding anyway, this is trivial. The read will need to explicitly do the regridding since GeomIO does not. Here is the pseudo code for what `MAPL_bundle_read` should do:

```
MAPL_Bundle_Read(bundle, file_template, time, vars_to_read, regrid_method,.,bells and whistles)

De-template file name with time 
Get collection metadata and determine ESMF horizontal grid, MAPL vertical grid, and times on file
Confirm time is on file
If (bundle is empty) then
     Populate_with_mapl3_compatible_fields from file (modulo vars_to_read)
Else
     Confirm each bundle variable is in file

Check if file_grid matches bundle_grid

If (file_grid == bundle_grid) 
    File_bundle = bundle
Else
    File_bundle = clone(output_bundle, file_grid)

! this used to be a single call to griddedIO, now make 2 calls
Use GeomIO layer to read file_bundle
If (file_grid /= bundle_grid) regrid_file_bundle_to_output_bundle (horizontal only for now)
```

The two items that might take a little work are
1. a method that given a file, populates an ESMF bundle with MAPL3 compatible fields using the information from the bundle (variables + horizontal grid + vertical grid)
2. a routine to regrid a bundle to another bundle (given that they have the same named fields but on different grids) using the new regridder manager.

bena-nasa:
Fixes #4598 

Making to see if the CI passes

This gets the Fieldbundle read and write moved to GeomIO in a form I need for my Regrid_Util.F90

there are few features the MAPL2 version supported that the new version won't for reasons such as no tripolar factory yet in Geom

I've isolated it to a few calls that Regrid_Util.F90 still needs from MAPL2 that should be obvious from the use statements and that will need to be ported. This used some of the AI generated code Tom came up with for the reading.

## Types of change(s)

## Checklist


tclune (dismissed):
Lots that I may want to improve (e.g.. fArgParse for command line arguments). But priority is to get something that works into the v3 directories that will survive.

mathomp4:
I can say that some of GNU errors I see are because GNU does not like multi-line macros. They have to be all on one-line.

This one:
```
/__w/MAPL/MAPL/src/Shared/@MAPL/GeomIO/FieldBundleRead.F90:293:43:

  293 |       file_mapl_geom => get_geom_manager()%get_mapl_geom(metadata, _RC)
      |                                           1
Error: The leftmost part-ref in a data-ref cannot be a function reference at (1)
```

is an @tclune special. Standards language!

tclune:
> I can say that some of GNU errors I see are because GNU does not like multi-line macros. They have to be all on one-line.
> 
> This one:
> 
> ```
> /__w/MAPL/MAPL/src/Shared/@MAPL/GeomIO/FieldBundleRead.F90:293:43:
> 
>   293 |       file_mapl_geom => get_geom_manager()%get_mapl_geom(metadata, _RC)
>       |                                           1
> Error: The leftmost part-ref in a data-ref cannot be a function reference at (1)
> ```
> 
> is an @tclune special. Standards language!

Clearly an AI mistake.  Fortran does not support chaining function calls like this.   Need an intermediate variable for the manager.

tclune (dismissed):

bena-nasa:
arg, NAG failure of the unit test, will fix

bena-nasa:
@tclune Fixed the NAG build/test issues. Surprise, surprise, needed targets in places

tclune (approved):

PR #4643, opened 2026-04-15
Fix SIGFPE (integer divide by zero) in MAPL_LoadBalanceMod

mathomp4:
## Summary
- Added safety check in `MAPL_BalanceWork4` and `MAPL_BalanceWork8` to handle `Idim == 0` preventing division by zero.
- This occurs when the max work difference evaluates to 0 in load balancing, causing `NumMax` (and thus `Idim`) to be 0 (e.g. single sunlit gridpoint among processes).
- Updated `CHANGELOG.md` with an entry for this fix.

### DIFF
--- a/shared/MAPL_LoadBalance.F90
+++ b/shared/MAPL_LoadBalance.F90
@@ -109,7 +109,11 @@ contains
     logical :: SEND, RECV
     integer, pointer :: NOP(:,:)
 
-    Jdim = size(A)/Idim
+    if (Idim == 0) then
+       Jdim = 1
+    else
+       Jdim = size(A)/Idim
+    end if
 
     if(present(Handle)) then
        ISTRAT = Handle
@@ -213,7 +217,11 @@ contains
     logical :: SEND, RECV
     integer, pointer :: NOP(:,:)
 
-    Jdim = size(A)/Idim
+    if (Idim == 0) then
+       Jdim = 1
+    else
+       Jdim = size(A)/Idim
+    end if
 
     if(present(Handle)) then
        ISTRAT = Handle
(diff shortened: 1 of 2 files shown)


mathomp4:
CC'ing @wmputman has he saw the issue this might fix

mathomp4:
Per @wmputman this seems to have ... not hurt. That is, when he added this change, he did not get an FPE in MAPL anymore. The error just sort of "moved".

So my guess is this was acting more like a trap. Something bad happened and the divide-by-zero here was the first place the model tripped.

shared/MAPL_LoadBalance.F90
```diff
@@ -109,7 +109,11 @@ subroutine MAPL_BalanceWork4(A, Idim, Direction, Handle, rc)
     logical :: SEND, RECV
     integer, pointer :: NOP(:,:)
 
-    Jdim = size(A)/Idim
+    if (Idim == 0) then
+       Jdim = 1
+    else
+       Jdim = size(A)/Idim
+    end if
```
tclune:
```suggestion
    Jdim = 1
    if (Idim /= 0) then
       Jdim = size(A)/Idim
    end if
```
Never-elser

shared/MAPL_LoadBalance.F90
```diff
@@ -213,7 +217,11 @@ subroutine MAPL_BalanceWork8(A, Idim, Direction, Handle, rc)
...
```
tclune:
```suggestion
    Jdim = 1
    if (Idim /= 0) then
       Jdim = size(A)/Idim
    end if
```

mathomp4:
Ahh. I forgot about doing that. I need to remind myself!

tclune (approved):

PR #4646, opened 2026-04-17
Fix OpenMP crash in MAPL profiler for NAG fortran

weiyuan-jiang:
This PR fixes OpenMP-related segmentation faults in MAPL profiler tests for both NAG Fortran and Intel ifort compilers using compiler-specific preprocessing directives. NAG Fortran 7.2 (macOS): Required parallel region context for !$omp master directives to work correctly. But Intel ifort would crash if the parallel directives are added 

## Types of change(s)

## Checklist


mathomp4:
@weiyuan-jiang ifort seems *very* unhappy with this for some reason. GNU and ifx seem to be okay.

weiyuan-jiang:
> @weiyuan-jiang ifort seems _very_ unhappy with this for some reason. GNU and ifx seem to be okay.

I don't have ifort on my laptop. Let me try it on discover

mathomp4:
> > @weiyuan-jiang ifort seems _very_ unhappy with this for some reason. GNU and ifx seem to be okay.
> 
> I don't have ifort on my laptop. Let me try it on discover

It's also on bucy if that's easier :)

weiyuan-jiang:
Why do we put the openmp directives inside the profilers? I think we should shift the responsorbility to users. @tclune @aoloso

tclune:
> Why do we put the openmp directives inside the profilers? I think we should shift the responsorbility to users. @tclune @aoloso

1. We have a mechanism whereby MAPL itself can instrument a hierarchy of components that are column based.
2. MAPL does its own calls to the profiler in the generic layer.   So there is no way for users to protect those calls.

tclune:
@weiyuan-jiang (or @mathomp4) Can you create a small standalone program that demonstrates the original (NAG) error?   I want to be sure we are fixing the right thing.

mathomp4:
NOTE: The latest OpenMP standards say one should use `masked` instead of `master`, so I'll refer to that below.

The OpenMP Standard says for the [**`masked`** construct](https://www.openmp.org/spec-html/5.2/openmpse61.html#x177-18300010.5):

> **Binding**
> 
> The binding thread set for a **`masked`** region is the current team. A **`masked`** region binds to the innermost enclosing parallel region.

That does seem to read that you can only use `omp masked` inside an `omp parallel` region

weiyuan-jiang:
@tclune @mathomp4 Please see issue #4673 . This PR fixed that issue

tclune:
After a bit more playing with NAG,  I had a thought that it might simply be that NAG assumes there will be at least one parallel region prior to the use of `!$omp master`.  And indeed that appears to be the case with a simple reproducer.  (Still a bit puzzled that the debug version works regardless.)   Also not sure if I should submit this reproducer to NAG.

```
! Standalone reproducer for NAG compiler bug:
!   !$omp master used outside an explicit parallel region crashes
!   with optimization flags but works with debug flags.
!
! Compile (debug - works):
!   nagfor -O0 -g -gline -C=all -openmp omp_master_reproducer.F90 -o reproducer_debug
!
! Compile (optimized - crashes):
!   nagfor -O2 -openmp omp_master_reproducer.F90 -o reproducer_opt
!
! Run:
!   ./reproducer_debug    ! succeeds
!   ./reproducer_opt      ! segfaults in __NAGf90_OpenMP_filter2

program omp_master_reproducer
   implicit none

   ! Dummy parallel region to force NAG's OpenMP runtime to initialize
   ! before the !$omp master is encountered outside a parallel region.
   !$omp parallel
   !$omp end parallel

   call initialize_something()
   print *, 'Done.'

contains

   subroutine initialize_something()
      integer :: x

      ! Called during serial initialization - no enclosing !$omp parallel.
      ! Per the OpenMP standard the implicit initial parallel region
      ! (single-thread team) provides the binding region, so this is
      ! conforming.  NAG's optimized runtime crashes here anyway.
      x = 0
      !$omp master
      x = 42
      !$omp end master

      print *, 'x =', x

   end subroutine initialize_something

end program omp_master_reproducer
```

tclune:
@weiyuan-jiang Please try the above workaround in MAPL.  (Look in mapl3g/MaplFramework.F90 and add it to the initialization of MAPL.)  If that fixes MAPL, let's use that approach.

tclune:
**Note: This comment is from Claude (Anthropic), posted via @tclune's account at his request.**

---

Further investigation has refined our understanding of why the workaround needs to be placed carefully.

**The root cause is per-DSO lazy initialization in NAG's optimized OpenMP runtime.**

When NAG links a shared library with `-openmp`, each DSO gets its own copy of the OpenMP runtime support code and its own associated state block. The `__NAGf90_OpenMP_filter2` routine — which implements the `!$omp master` check — reads from that per-DSO state block to determine which thread is the master. In the optimized runtime, that state block is not populated until the first `!$omp parallel` region executes *within that DSO*.

This is why placing the no-op `!$omp parallel` / `!$omp end parallel` in `MaplFramework::initialize()` (in `libMAPL.dylib`) was not sufficient — it initializes the runtime state for `libMAPL.dylib`, but `libMAPL.profiler.dylib` is a separately linked shared library with its own uninitialized state block. When `__NAGf90_OpenMP_filter2` in `libMAPL.profiler.dylib` subsequently ran, it read from *its own* uninitialized state block, found a null pointer at offset `0x4`, and crashed.

**The fix therefore needs to be inside `libMAPL.profiler.dylib` itself** — specifically at the top of the `initialize` subroutine in `profiler/MAPL_Profiler.F90`, before any call that reaches `!$omp master` in `BaseProfiler.F90`.

The debug runtime does not exhibit this because it either initializes the per-DSO state block eagerly at library load time, or its `__NAGf90_OpenMP_filter2` has a null-check guard before dereferencing. The optimized runtime skips both for performance.

This is analogous to calling `MPI_Comm_rank` before `MPI_Init` — it may accidentally work in a debug build due to zero-initialized memory, but crashes in an optimized build.

**Standalone reproducer** (confirmed: passes with `-O0`, segfaults with `-O2` on NAG 7.2 Build 7241):

```fortran
program omp_master_reproducer
   implicit none

   call initialize_something()
   print *, 'Done.'

contains

   subroutine initialize_something()
      integer :: x
      x = 0
      !$omp master
      x = 42
      !$omp end master
      print *, 'x =', x
   end subroutine initialize_something

end program omp_master_reproducer
```

Adding a no-op parallel region *before* the `!$omp master` call (in the same or calling DSO) fixes the crash:

```fortran
   !$omp parallel
   !$omp end parallel
   call initialize_something()
```

mathomp4:
Huh. So you could just sprinkle some:
```fortran
   !$omp parallel
   !$omp end parallel
```
in some places and then it might work "simply"?

weiyuan-jiang:
I feel we should put this no op parallel directive right inside at the beginning of initialize_something(). I will do more tests

weiyuan-jiang:
How about using this PR #4715 to replace the current one?

weiyuan-jiang:
This PR is replaced by the PR #4715

PR #4671, opened 2026-04-23
v3: Fixes for Spack, NAG, and Python 3.11

mathomp4:
## Types of change(s)

## Checklist

## Description

In testing with NAG on macOS with Spack, I encountered an issue that I think is due to spack-stack using Python 3.11. 

This PR "fixes" things, but my fear is I've now broken everything everywhere else! I'm going to make this PR draft for now and ask @darianboggs to take a look and make sure it looks good to him.


Apps/tests/acg3/acg3_unittests.py
```diff
@@ -88,7 +88,7 @@ def message(e, a):
             return general_msg(a, e)
         test_params = (
                 TestParams(value, equal_test(expected), partial(message, expected)) for value, expected in
-                ((None, None), ('XX', r"'XX'"), ('*XX', f"'{INTERLUDE}XX'"))
+                ((None, None), ('XX', r" 'XX' "), ('*XX', " ''//trim(comp_name)//'XX' "))
```
darianboggs, line 91: The replacement of INTERLUDE with "'//trim(comp_name)//'" is fine, and the use of a str instead of a f-string is fine. The additional spaces is odd. It may be that it works with the additional spaces, but those spaces should not be necessary.

Apps/CMakeLists.txt
```diff
@@ -3,11 +3,16 @@
...
+# Note: We use configure_file(... COPYONLY) instead of file(COPY ...) here
+# so that CMake registers these scripts as dependencies. If you edit the source 
+# scripts, the build system will automatically detect the change and re-copy 
+# them to the build/install directories without needing a manual re-run of CMake.
 
-file (COPY MAPL_GridCompSpecs_ACG.py DESTINATION ${esma_etc}/MAPL)
-file (COPY MAPL_GridCompSpecs_ACGv3.py DESTINATION ${esma_etc}/MAPL)
-file (COPY mapl_acg.pl DESTINATION ${esma_etc}/MAPL)
-file (COPY mapl_stub.pl DESTINATION ${esma_etc}/MAPL)
+configure_file (MAPL_GridCompSpecs_ACG.py ${esma_etc}/MAPL/MAPL_GridCompSpecs_ACG.py COPYONLY)
```
darianboggs, line 12: It's odd that this is necessary, but OK.

Apps/MAPL_GridCompSpecs_ACGv3.py
```diff
@@ -648,9 +648,9 @@ def make_else_block(name=None):
         STRING: lambda value: add_quotes(value),
         STRINGVECTOR: lambda value: construct_string_vector(value),
         ARRAY: lambda value: mk_array(value),
-        MANGLED: lambda name: f"'{rm_quotes(name).replace("*","'//trim(comp_name)//'")}'" if name else None,
+        MANGLED: lambda name: f''' '{rm_quotes(name).replace("*", "'//trim(comp_name)//'")}' ''' if name else None,
```
darianboggs, line 651: This seems to be the only change that is more than adding spaces which has no effect in this file. 

I suspect this is Claude bloat overkill since it adds unnecessary changes. Using triple quotes to surround a string allows you use both " and ' characters in a string, but usually it is unnecessary, as it is in this case. In the f-string, the portion in the braces is code that is executed, so the quote marks in the braces are independent of the quote marks outside the braces. Replacing the outer quote marks with triple apostrophes is unnecessary.
  
If this line is the one that caused problems, I suspect that it is a bug with Python 3.11 or Spack. NAG should use the same Python version as the other compilers. If the tests pass only with the change, then OK.

darianboggs (commented):
Notes applied. It seems like a bug in Python 3.11 or Spack and/or code bloat by Claude.

Apps/CMakeLists.txt
mathomp4: Yeah. I found that when I changed the ACG script, it didn't get re-installed.

mathomp4:
> Notes applied. It seems like a bug in Python 3.11 or Spack and/or code bloat by Claude.

I'm going to try setting "Ready for review" to let CI do its thing.

Apps/CMakeLists.txt
weiyuan-jiang: Do you think this is related to the error I reported this morning like MAPL.generic3g.scenarios .........***Failed    4.18 sec
           MAPL: Field dictionary not loaded: "geos_field_dictionary.yaml" not found. Dictionary defaults (units, long_name) will not be applied.
           MAPL: Field dictionary not loaded: "geos_field_dictionary.yaml" not found. Dictionary defaults (units, long_name) will not be applied.
           MAPL: Field dictionary not loaded: "geos_field_dictionary.yaml" not found. Dictionary defaults (units, long_name) will not be applied.
           MAPL: Field dictionary not loaded: "geos_field_dictionary.yaml" not found. Dictionary defaults (units, long_name) will not be applied

tclune:
@mathomp4 @darianboggs Can this PR be merged?

mathomp4:
> @mathomp4 @darianboggs Can this PR be merged?

If @darianboggs is okay with it, then I think so. I've pressed the Update Branch button.

Apps/MAPL_GridCompSpecs_ACGv3.py
darianboggs: I'm resolving this conversation since there is a workaround.

darianboggs (approved):

PR #4675, opened 2026-04-23
Fix vertical regrid 

weiyuan-jiang:
## Types of change(s)

## Checklist

## Description

When built with NAG, two tests fail. 1) The weight is wrong 2) nag needs continuous memory when passing an array to a subroutine


tclune:
I do not understand.  NAG has been working for me for some time with this layer.  What has changed?

tclune:
> When built with NAG, two tests fail. 1) The weight is wrong 2) nag needs continuous memory when passing an array to a subroutine

(1) In what way is the weight wrong?  Is there a compiler bug or a code bug?
(2) This simply not true, at least as stated.

weiyuan-jiang:
At least this is a bug. Make it to 0.5 or choose one of them as 1.[weight
](https://github.com/GEOS-ESM/MAPL/blob/2beb4587c48c66112bb12ea268ac20a164215b3b/generic3g/vertical/VerticalLinearMap.F90#L103)

weiyuan-jiang:
> I do not understand. NAG has been working for me for some time with this layer. What has changed?

Maybe I am using X86_64 based MacBook ? But wait, that fixed the test issue on bucy too.

generic3g/transforms/VerticalRegridTransform.F90
```diff
@@ -414,19 +415,21 @@ subroutine compute_interpolation_matrix_(method, v_in_coord, stagger_in, src_ali
...
-             associate(src => vv_in(horz, :, ungrd), dst => vv_out(horz, :, ungrd))
-               if (method == VERTICAL_REGRID_LINEAR) then
-                  call compute_linear_map(src, dst, matrix(horz), _RC)
-               else if (method == VERTICAL_REGRID_CONSERVATIVE) then
-                  call compute_conservative_map(src, dst, matrix(horz), _RC)
-               else
-                  _FAIL("Unknown vertical regridding method")
-               end if
-             end associate
+             src_levels = vv_in(horz, :, ungrd)
```
weiyuan-jiang, line 424: Ok, I take back this change. Maybe my build directory is too dirty and get an run time error. But this change may make it faster

tclune: I was not looking for optimization.  If you believe it is faster, please explain why.

generic3g/vertical/VerticalLinearMap.F90
```diff
@@ -53,11 +53,11 @@ subroutine compute_linear_map(src, dst, matrix, rc)
       do ndx = 1, size(dst)
          val = dst(ndx)
          call find_bracket_(val, src, pair)
-         call compute_weights_(val, pair%value_, weight)
-         if (pair(1) == pair(2)) then
+         if (pair(1)%index == pair(2)%index) then
             ! matrix(ndx, pair(1)%index) = weight(1)
-            call add_row(matrix, ndx, pair(1)%index, [weight(1)])
+            call add_row(matrix, ndx, pair(1)%index, [1.0_REAL32])
```
tclune: agreed.

generic3g/vertical/VerticalLinearMap.F90
```diff
@@ -76,13 +76,22 @@ subroutine find_bracket_(val, array, pair)
...
+         ndx2 = size(array)
+      else
+         ndx1 = minloc(abs(array - val), 1)
+         if (array(ndx1) < val) then
+            ndx1 = max(1, ndx1 - 1)
+         end if
+         ndx2 = ndx1
+         if (array(ndx1) /= val) then
+            ndx2 = min(size(array), ndx1 + 1)
+         end if
```
tclune: I cannot convince myself that this more complex logic is necessary or even correct.


Consider this bit: 
```f90
   ndx1 = minloc(abs(array - val), 1)
   if (array(ndx1) < val) then
      ndx1 = max(1, ndx1 - 1)
   end if
```
With array = [1., 2., 3.] and val = 1.2.  Then ndx starts as 1 and the max leaves it as 1.    But with val = 2.2, you initially get ndx = 2, and then the correction makes it 1, which is wrong.  But I suspect the original code is also wrong.

The original is also wrong.  How about this version:
```
      n = size(array)
      nearest = minloc(abs(array - val), 1)
      if (array(nearest) < val) then
         ndx1 = max(1,nearest - 1)
      else
         ndx2 = min(n,ndx1+1)
      endif
```

tclune: Nope.  Even that is not quite right.

```
      n = size(array)
      nearest = minloc(abs(array - val), 1)
      if (array(nearest) < val) then
         ndx1 = max(1,nearest - 1)
         ndx2 = min(n,nearest)
      else
         ndx1 = nearest
         ndx2 = min(n,nearest+1)
      endif

Comment (tclune) [reply generic3g/vertical/VerticalLinearMap.F90]: Of course an even more efficient solution would be a bisection search:  Something like:

```
idx1 = 1
idx2 = n
mid  = n/2
if (val < array(mid)) then
   idx2 = mid
else
   idx1 = mid
end 
<recurse>
```
Lots of details to work out and would need some proper unit testing to go with it.

Comment (weiyuan-jiang) [reply generic3g/vertical/VerticalLinearMap.F90]: I think this logic is correct

Comment (weiyuan-jiang) [reply generic3g/vertical/VerticalLinearMap.F90]: n = size(array)
      nearest = minloc(abs(array - val), 1)
      if (array(nearest) < val) then
         ndx1 = max(1,nearest - 1)
         ndx2 = min(n,nearest)
      else
         ndx1 = nearest
         if (array(nearest) == val) then
            ndx2 = nearest  ! Exact match
         else
            ndx2 = min(n,nearest+1)
         endif
      endif

Comment (tclune) [reply generic3g/vertical/VerticalLinearMap.F90]: > I think this logic is correct

Not sure which variant your comment is addressing.   At this point I think we need a solid set of unit tests and then we can use any version that passes the tests.

**Review by tclune — CHANGES_REQUESTED** [review]
I want a strong unit test for this before we proceed.   I can write if you don't feel comfortable. 

Basic sketch:
```f90
array = [1., 2., 3.]

call find_bracket(0.9, array, pairs)
@assert_that(pairs(1)%index, is(1))
@assert_that(pairs(2)%index, is(1))

call find_bracket(1.1, array, pairs)
@assert_that(pairs(1)%index, is(1))
@assert_that(pairs(2)%index, is(2))

call find_bracket(1.9, array, pairs)
@assert_that(pairs(1)%index, is(1))
@assert_that(pairs(2)%index, is(2))

call find_bracket(2.1, array, pairs)
@assert_that(pairs(1)%index, is(2))
@assert_that(pairs(2)%index, is(3))

call find_bracket(2.9, array, pairs)
@assert_that(pairs(1)%index, is(2))
@assert_that(pairs(2)%index, is(3))

call find_bracket(3.1, array, pairs)
@assert_that(pairs(1)%index, is(3))
@assert_that(pairs(2)%index, is(3))
```

generic3g/transforms/VerticalRegridTransform.F90
```diff
@@ -414,19 +415,21 @@ subroutine compute_interpolation_matrix_(method, v_in_coord, stagger_in, src_ali
         end if
 
        allocate(matrix(n_horz))
+       allocate(src_levels(shape_in(2)))
+       allocate(dst_levels(shape_out(2)))
 
        ! TODO: Convert to a `do concurrent` loop
        do horz = 1, n_horz
           do ungrd = 1, n_ungridded
```
weiyuan-jiang, line 423: Here is what Claude said:
The Problem
Original code (using associate):
associate(src => vv_in(horz, :, ungrd), dst => vv_out(horz, :, ungrd))
  call compute_linear_map(src, dst, matrix(horz), _RC)
end associate
The associate construct creates aliases to array sections vv_in(horz, :, ungrd) and vv_out(horz, :, ungrd). These are non-contiguous array slices in memory because they're extracted from a 3D array.
Issue with NAG compiler: 
The NAG Fortran compiler has strict requirements and different behavior regarding non-contiguous array sections. When passing these non-contiguous slices to subroutines, it can cause:
- Memory layout issues
- Incorrect array access patterns in find_bracket (which uses operations like minloc(abs(array - val), 1))
- Potential runtime instabilities or wrong results
The Solution
New code (creating contiguous copies):
src_levels = vv_in(horz, :, ungrd)    ! Creates contiguous 1D copy
dst_levels = vv_out(horz, :, ungrd)   ! Creates contiguous 1D copy
call compute_linear_map(src_levels, dst_levels, matrix(horz), _RC)
By explicitly copying the array sections into new 1D allocatable arrays (src_levels and dst_levels), you ensure:
1. Memory contiguity - The arrays are guaranteed to be stored contiguously in memory
2. Predictable behavior - All compilers (NAG, gfortran, Intel) handle these consistently
3. Safe array operations - Operations in find_bracket like minloc, array arithmetic, and indexing work reliably

tclune: It is correct that NAG is strict, but Claude is just guessing that there is a problem here.   If the called procedures need a contiguous copy, NAG will generally do this correctly.  Always possible that there is a bug in this particular case.   And explicitly doing the copy is unlikely to have a significant performance difference.  (In any event, we should not be worried much about performance for now.)

tclune (dismissed):

tclune (approved):

weiyuan-jiang:
When I re-used build directory , sometimes I saw those failed tests. I am not sure why. Does CI test use a brand new directory? @mathomp4

tclune:
> When I re-used build directory , sometimes I saw those failed tests. I am not sure why. Does CI test use a brand new directory? @mathomp4

Yeah.  CI for MAPL3 is in a bit of a sad state right now.   ifx tends to fail on case 24.  on-prem nag sometimes fails.  And on laptop with debug, MAPL.generic3g.scenarios fail.  For now, we just merge anyway.

PR #4701, opened 2026-04-25
Add global field fill default override (#4690)

### LINKED ISSUE #4690: Add ability in MAPL3 to overwrite default initialization values for ESMF fields
## Problem

`FieldFill` initializes all newly allocated ESMF fields to a sentinel value:
- R4/R8 fields: IEEE signaling NaN (sNaN)
- I4/I8 fields: `-huge(...)`

This is intentional — it catches code that reads from uninitialized fields. However, sNaN causes hardware floating-point traps when ifort is compiled with debug flags (e.g. `-fpe0`), making it impossible to run ifort debug builds when fields remain uninitialized before their first use.

A per-field override (`default_value`) already exists at the `FieldClassAspect`/`VariableSpec` level and is settable via YAML. But there is no way to change the global sentinel applied to **all** fields at initialization time.

## Solution

Add a global override mechanism for the default field fill value, configured at MAPL initialization time via `MaplFramework`.

### Scope

- Only R4 and R8 typekind fields are affected. Integer fields (I4/I8) are not subject to the ifort sNaN trap issue and are exceedingly rare, so they retain their existing `-huge` fill behavior.

### New module: `mapl3g_FieldFillDefault` (`field/FieldFillDefault.F90`)

Holds module-level allocatable state for the global fill overrides:

```fortran
real(REAL32), allocatable :: fill_default_r4  ! unallocated => use sNaN
real(REAL64), allocatable :: fill_default_r8  ! unallocated => use sNaN
```

Public procedures:
- `set_field_fill_default_r4(value)`
- `set_field_fill_default_r8(value)`
- `get_field_fill_default_r4(value) result(is_set)` — returns `.true.` and fills `value` if override is active
- `get_field_fill_default_r8(value) result(is_set)`
- `reset_field_fill_defaults()` — deallocates both (used for testing/cleanup)

### Modified: `field/FieldFill.F90`

Before using sNaN, consult the global override:
```fortran
if (.not. get_field_fill_default_r4(fill_r4)) then
   fill_r4 = ieee_value(fill_r4, ieee_signaling_nan)
end if
```

### Modified: `mapl3g/MaplFramework.F90`

New **public** type-bound procedure `initialize_field_fill_defaults` reads from the MAPL HConfig:

| YAML key | Effect |
|---|---|
| `field_default_fill_value:` | Sets both R4 and R8 to this value |
| `field_default_fill_value_r4:` | Sets R4 only |
| `field_default_fill_value_r8:` | Sets R8 only |

**Conflict rules (errors, not warnings):**
- If both `field_default_fill_value:` and `field_default_fill_value_r4:` (or `_r8:`) are present → error
- If both Fortran argument and YAML key specify a value for the same typekind → error

Called in `MaplFramework%initialize()` alongside the other `initialize_*` calls.

### Modified: `MAPL_Initialize` public wrapper

Two new optional Fortran arguments:
```fortran
real(REAL32), optional, intent(in) :: field_default_fill_value_r4
real(REAL64), optional, intent(in) :: field_default_fill_value_r8
```

These are set **before** `the_mapl_object%initialize` is called. If the YAML config also specifies a value for the same typekind, an error is raised.

### YAML example

```yaml
mapl:
  field_default_fill_value: 0.0   # sets both R4 and R8; use instead of sNaN
```

Or typekind-specific:
```yaml
mapl:
  field_default_fill_value_r4: 0.0
  field_default_fill_value_r8: 0.0
```

### New tests: `field/tests/Test_FieldFill.pf`

- Default R4/R8 fill is sNaN when no override is set
- Override R4 fill value is applied correctly
- Override R8 fill value is applied correctly
- `reset_field_fill_defaults` restores sNaN behavior
- Getter returns `.false.` when no override is set

## Files changed

| File | Change |
|---|---|
| `field/FieldFillDefault.F90` | **New** |
| `field/FieldFill.F90` | Use global override if set |
| `field/API.F90` | Export new public names |
| `field/CMakeLists.txt` | Add new source |
| `mapl3g/MaplFramework.F90` | Add `initialize_field_fill_defaults` (public); update `MAPL_Initialize` signature |
| `field/tests/Test_FieldFill.pf` | **New** |
| `field/tests/CMakeLists.txt` | Register new test |

tclune:
## Types of change(s)

## Checklist
- [x] Ran the Unit Tests (`ctest`)

## Description

Adds a global override mechanism for the default ESMF field fill value in MAPL3.

- New `FieldFillDefault.F90` singleton module with lazy initialization (sNaN by default for R4 and R8)
- `FieldFill.F90` reads fill values directly from the singleton — no conditionals
- `MaplFramework` gains a public `initialize_field_fill_defaults` procedure, callable at startup via YAML keys (`field_default_fill_value_r4:`, `field_default_fill_value_r8:`, `field_default_fill_value:`) or Fortran optional arguments
- Integer fields (I4/I8) retain existing `-huge(...)` fill behavior unchanged
- 8 new pFUnit tests in `Test_FieldFill.pf` covering defaults, overrides, reset, getters, and R4/R8 independence

## Related Issue

Fixes #4690


atrayano (dismissed):

field/API.F90
```diff
@@ -2,6 +2,10 @@ module mapl3g_Field_API
    use mapl3g_FieldGet, only: MAPL_FieldGet => FieldGet
    use mapl3g_FieldSet, only: MAPL_FieldSet => FieldSet
    use mapl3g_FieldFill, only: MAPL_FieldFill => FieldFill
+   use mapl3g_FieldFillDefault, only: &
+        MAPL_InitializeFieldFillDefaults => initialize_field_fill_defaults, &
+        MAPL_SetFieldFillDefaults => set_field_fill_defaults, &
+        MAPL_ResetFieldFillDefaults => reset_field_fill_defaults
```
pchakraborty: Maybe we don’t make the FieldFillDefault routines public at the moment. We can always revisit if the need arises.

pchakraborty (changes requested):
One comment about not making the FieldFielDefault routines public.

field/API.F90
tclune: Can you do ths with the +/-.  I can't do it on my own code.

tclune:
```suggestion
```

pchakraborty (approved):
LGTM

PR #4704, opened 2026-04-26
Add EASEGeomSpec and EASEGeomFactory to MAPL.geom

### LINKED ISSUE #4700: Port EASEGridFactory to MAPL3 EASEGeomSpec/EASEGeomFactory; remove MAPL.base dependency from MAPL.geom
## Summary

Two related cleanup tasks:

1. The legacy `MAPL_EASEGridFactory` (Equal-Area Scalable Earth grid, used by land/soil components) has not yet been ported to the MAPL3 geometry layer.
2. `MAPL.geom` has an unnecessary dependency on `MAPL.base` in `geom/CMakeLists.txt`. All modules actually used by `geom/` (`MAPL_Constants`, `MAPL_InternalConstants`, `MAPL_EarthConstants`, `MAPL_MinMaxMod`, `MAPL_RangeMod`, `mapl_KeywordEnforcer`, etc.) live in `MAPL.shared`, not `MAPL.base`. The dependency is a historical holdover and should be removed.

These are combined because removing `MAPL.base` from `MAPL.geom` is too small to stand alone as a PR.

### EASE factory
- Implement `EASEGeomSpec` extending `GeomSpec` in `geom/EASE/`
- Implement `EASEGeomFactory` extending `GeomFactory`; port `MAPL_EASEConversion.F90` logic as needed
- Register with `GeomManager`
- Add pfUnit tests
- Wire into `geom/CMakeLists.txt`

### MAPL.base dependency removal
- Remove `MAPL.base` from the `DEPENDENCIES` line in `geom/CMakeLists.txt`
- Verify build succeeds (all needed modules are already available via `MAPL.shared`)

## Reference

Legacy implementation preserved in git history at the last MAPL2 tag: `MAPL_EASEGridFactory.F90`, `MAPL_EASEConversion.F90`.

## Notes

This is a blocker for complete removal of the MAPL2 legacy grid factory layer from `MAPL.base`.

tclune:
## Summary

- Adds `EASEGeomSpec` and `EASEGeomFactory` to `MAPL.geom`, supporting EASEv1 and EASEv2 grids
- New standalone `mapl3g_EASEConversion` module (no `MAPL.base` dependency)
- `EASEGeomFactory` registered in `GeomManager`; pFUnit tests included

## Types of change(s)

## Checklist
- [x] Ran the Unit Tests (`make tests` or `ctest`)

## Related Issue

Closes #4700


geom/EASE/EASEDecomposition.F90
```diff
@@ -0,0 +1,148 @@
...
+   type :: EASEDecomposition
+      private
+      integer, allocatable :: lon_distribution(:)
+      integer, allocatable :: lat_distribution(:)
+   contains
+      procedure :: get_lon_distribution
+      procedure :: get_lat_distribution
+   end type EASEDecomposition
+
+   interface EASEDecomposition
```
weiyuan-jiang, line 27: We need a decomposition by reading ims and jms from a file

tclune: It should be handled by this procedure:  
```
module function make_EASEGeomSpec_from_metadata(file_metadata, rc) result(spec)
```

weiyuan-jiang: But how? file_metadata does not have the information

weiyuan-jiang (dismissed):

geom/EASE/EASEDecomposition.F90
tclune: For metadata it has to just use a default decomposition, there is really no other choice unless EASE data files have decomposition information in them.  (We don't do that for any other grids.)

In this case it is just ExtData and it needs some decomposition so that it can do regridding.   Maybe even a noop if extdata never uses EASE grids.

weiyuan-jiang (approved):

PR #4715, opened 2026-04-28
Fix NAG OpenMP per-DSO lazy initialization crash in profiler

weiyuan-jiang:
NAG's optimized OpenMP runtime uses per-DSO lazy initialization where each shared library gets its own OpenMP state block that is not populated until the first ! parallel region executes within that DSO.

When ! master directives in BaseProfiler.F90 ran before the state was initialized, __NAGf90_OpenMP_filter2 would dereference a null pointer at offset 0x4, causing segmentation faults.

This fix adds a no-op ! parallel region at the beginning of each profiler constructor (TimeProfiler, MemoryProfiler, DistributedProfiler) to ensure the OpenMP state is initialized before any ! master directives are encountered.

This approach is minimal and optimal because:
- Constructors are the true entry points for all profiler objects
- Once initialized, the OpenMP state persists for the DSO
- No redundant initialization overhead in frequently-called methods
- Works for both production code and unit tests

The debug runtime does not exhibit this crash because it either initializes eagerly or includes null-check guards.

## Types of change(s)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Workaround to satisfy NAG's lazy initialization of openmp region
 - Workaround to pass ifx 2025.1 tests in debug mode
 - Profile reporting has been relocated into the `./profile` directory.
 - Improved diagnostic message for profiler imbalances at end of run.
--- a/profiler/DistributedProfiler.F90
+++ b/profiler/DistributedProfiler.F90
@@ -38,6 +38,16 @@ contains
       integer, intent(in) :: comm
       integer, optional, intent(in) :: comm_world
 
+      ! NAG OpenMP runtime workaround: Initialize per-DSO OpenMP state
+      !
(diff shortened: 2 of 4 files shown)


tclune:
Technically, anyone that uses this layer without going through the profiler initialize() routine is not using the tool as "documented".  (Not that the documentation exists.)

Do you have scenarios in mind where MAPL_Profiler:initialize() is not called?

weiyuan-jiang:
> Technically, anyone that uses this layer without going through the profiler initialize() routine is not using the tool as "documented". (Not that the documentation exists.)
> 
> Do you have scenarios in mind where MAPL_Profiler:initialize() is not called?
The program here dont call initiliazaion

profiler/demo/

tclune:
> > Technically, anyone that uses this layer without going through the profiler initialize() routine is not using the tool as "documented". (Not that the documentation exists.)
> > Do you have scenarios in mind where MAPL_Profiler:initialize() is not called?
> > The program here dont call initiliazaion
> 
> profiler/demo/

OK Fair point.    

I'm just worried that there is a small overhead to doing this for each constructor.  Hopefully not measurable.

tclune (approved):

PR #4716, opened 2026-04-28
Install Configurable gridcomp

pchakraborty:
## Types of change(s)

## Checklist


tclune (approved):

weiyuan-jiang:
why do you need to install this? The PR #4717 can build the lib if just run "make tests"

tclune:
> why do you need to install this? The PR #4717 can build the lib if just run "make tests"

@pchakraborty is now using this component to support Held-Suarez configuration in a clean manner.   In MAPL2 there were some kludges to satisfy fv imports.   ConfigurableGridcomp gives a reasonably elegant mechanism to support Held-Suarez without putting kludges into the dyncore.   Not an ideal situation.   And we can imagine that over time ConfigurableGridComp becomes even more, er, configurable and may be useful for users in other contexts.    Only for expert users though.

PR #4719, opened 2026-04-29
Add ability to checkpoint History3G (really statisticsGridComp)

bena-nasa:
This was supposed to be a simple to to allow the StatisticsGridComp to checkpoint/restart but was not

It turned out to be just a ton of little changes that were needed to accomplish this including but not limited to:
- Adding a "bootstrap" option
- Moving "get_checkpoint_subdir" to a common location.
- Adding a procedure to get the ESMF_Geom of a state and return an error if all fields in state do not have same Geom
- Had to modify PFIO and field utilities to support reading/write I4 variables and getting the pointer for I4 ESMF_Fields

Changes to statsisticsGridComp
- It now uses the internal state to store the intermediate fields
- Uses the regular restart mechanism
- Still has custom write method for now
- Replaced ESMF_Alarm with SimpleAlarm due to another ESMF_Alarm malfunction

Limitations for now
- Requires that a StatisticsGridComp instantiation have the same Geom and Alarms ring times for all the stat items. In practice this is fine for now given that only History uses this and the way it is used this will always be true.

I did modify case42 to test the ability start/stop when doing time averaging in History

## Types of change(s)

## Checklist


generic3g/RestartHandler.F90
```diff
@@ -84,9 +85,10 @@ subroutine read(this, state, filename, rc)
       _RETURN_UNLESS(item_count>0)
 
       inquire(file=filename, exist=file_exists)
-      if (.not. file_exists) then
-         _FAIL("Restart file " // trim(filename) // " does not exist")
+      if ((.not. file_exists ) .and. bootstrap) then
+         _RETURN(_SUCCESS)
```
tclune: check my logic ...
```suggestion
_RETURN_IF(bootstrap .and. (.not. file_exists))
```

gridcomps/History3G/HistoryGridComp_private.F90
```diff
@@ -115,6 +115,8 @@ subroutine add_stats_gc(gridcomp, child_name, child_hconfig, rc)
       end if
 
       stats_list = ESMF_HConfigCreate(_RC)
+      stats_mapl_section = create_stats_mapl_section(_RC)
+      call ESMF_HConfigAdd(stats_hconfig, stats_mapl_section, addKeyString='mapl', _RC)
```
tclune, line 119: I may need to understand what is happening here.  We have a named parameter for the mapl section of a config file.

gridcomps/History3G/HistoryGridComp_private.F90
```diff
@@ -135,6 +137,16 @@ subroutine add_stats_gc(gridcomp, child_name, child_hconfig, rc)
 
    end subroutine add_stats_gc
 
+   function create_stats_mapl_section(rc) result(stats_mapl_section)
+      type(ESMF_HConfig) :: stats_mapl_section
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+
+      stats_mapl_section = ESMF_HConfigCreate(content='{misc: {checkpoint: {import: False, internal: False}, restart: {bootstrap: True, import: False, internal: True}}}', _RC)
```
tclune, line 146: This looks like test code.   You'll need to 'splain to me what is happening here.

gridcomps/StatisticsGridComp/StatisticsGridComp.F90
```diff
@@ -143,56 +149,22 @@ subroutine modify_advertise_item(iter, rc)
...
-              ungridded_dims=ungridded_dims, &
-              units=units, &
-              typekind=typekind, &
-              vgrid=vertical_grid, &
-              vert_staggerloc=vstagger, &
-              _RC)
-
-         call mapl_StateGet(exportState, itemName=name, field=f_out, _RC)
-
-         call mapl_FieldSet(f_out, &
```
tclune, line 184: I am very curious where this FieldSet went.

gridcomps/StatisticsGridComp/StatisticsGridComp.F90
```diff
@@ -357,100 +316,75 @@ subroutine run(gridcomp, importState, exportState, clock, rc)
...
+             stat => iter%of()
+             alarm = stat%get_alarm()
+             is_ringing = alarm%is_ringing(currTime, _RC)
+            if (first_item) then
+               first_ringing = is_ringing
+               first_item = .false.
+            else
+               _ASSERT(is_ringing .eqv. first_ringing, &
+                    'Inconsistent alarm state: not all statistics alarms have the same ringing status')
+            end if
```
tclune: Some compilers do not allow multiline cpp macros with `&`
```suggestion
            if (first_item) then
               first_ringing = is_ringing
               first_item = .false.
               cycle
            endif
            _FAIL(is_ringing .eqv. first_ringing, 'Inconsistent alarm state: not all statistics alarms have the same ringing status')
```

gridcomps/StatisticsGridComp/TimeAverage.F90
```diff
@@ -35,133 +35,201 @@ module mapl3g_TimeAverage
...
+           units=units, &
+           typekind=typekind, &
+           vgrid=vertical_grid, &
+           vert_staggerloc=vstagger, &
+           standard_name='foo', &
+           has_deferred_aspects=.false., &
+           _RC)
+
+      call esmf_StateGet(internal_state, 'sum_'//name, field=sum_f, _RC)
+      call mapl_FieldSet(sum_f, &
```
tclune, line 84: OK - all the FieldSet's  when here then right?  If so you can resolve this and the one before.

tclune: I am a bit concerned then that we need to do this for every stat subtype (avg, max, min, ...) But removing that duplication can wait for a rainy day.

state/StateGetGeom.F90
```diff
@@ -0,0 +1,98 @@
...
+      type(ESMF_Geom)  :: field_geom
+      type(ESMF_State) :: nested_state
+      type(MaplGeom), pointer :: mapl_geom
+
+      call ESMF_StateGet(state, itemCount=item_count, _RC)
+
+      ! An empty (nested) state is allowed — nothing to check here.
+      if (item_count == 0) then
+         _RETURN(_SUCCESS)
+      end if
```
tclune:
```suggestion
      _RETURN_UNLESS(item_count > 0)
```

state/StateGetGeom.F90
```diff
@@ -0,0 +1,98 @@
...
+      type(ESMF_State), intent(in)    :: state
+      type(ESMF_Geom),  intent(inout) :: geom
+      logical,          intent(inout) :: found
+      integer, optional, intent(out)  :: rc
+
+      integer :: status, item_count, i
+      character(ESMF_MAXSTR), allocatable :: item_names(:)
+      type(ESMF_StateItem_Flag), allocatable :: item_types(:)
+      type(ESMF_Field) :: field
+      type(ESMF_Geom)  :: field_geom
```
tclune:
```suggestion
      type(ESMF_Geom)  :: trial_geom
```

state/StateGetGeom.F90
```diff
@@ -0,0 +1,98 @@
...
+             if (status /= _SUCCESS) then
+                if (present(rc)) rc = status
+                return
+             end if
+
+         else
+            _ASSERT(.false., 'state items must be fields or nested states')
+         end if
+
+      end do
```
tclune, line 87:
```suggestion
      do i = 1, item_count

         if (item_types(i) == ESMF_STATEITEM_FIELD) then

            call ESMF_StateGet(state, itemName=item_names(i), field=field, _RC)
            call ESMF_FieldGet(field, geom=trial_geom, _RC)
            mapl_geom => get_mapl_geom(trial_geom, _RC)

            if (found) then
                _ASSERT(MAPL_SameGeom(trial_geom, geom), 'all fields must share the same geom')
            else
               geom  = trial_geom
               found = .true.
            end if

         else if (item_types(i) == ESMF_STATEITEM_STATE) then

            call ESMF_StateGet(state, itemName=item_names(i), nestedState=nested_state, _RC)
            call collect_geom(nested_state, geom, found, _RC)

         else
            _ASSERT(.false., 'state items must be fields or nested states')
         end if

      end do
```

utilities/OSUtilities.F90
```diff
@@ -0,0 +1,49 @@
+#include "MAPL.h"
```
tclune, line 1: Not sure this is the right dir for this, but it can stay here for now.

tclune (commented):
A few questions before I approve.

tclune (dismissed):
There are some needed changes, but developer has agreed to follow up with those rather than put this PR in jeopardy.

gridcomps/History3G/HistoryGridComp_private.F90
bena-nasa: we will fix in subsequent PR

gridcomps/History3G/HistoryGridComp_private.F90
bena-nasa: we will fix in subsequent PR

tclune (approved):

PR #4772, opened 2026-05-05
Export WRITE_PARALLEL from mapl_base3g umbrella module (#4771)

### LINKED ISSUE #4771: Export WRITE_PARALLEL from MAPL umbrella module
The `WRITE_PARALLEL` routine from `mapl3g_FileIO` (in `base3g/`) is not currently re-exported by the `MAPL` umbrella module. This means components that have migrated off `MAPL2` and use only `use MAPL` cannot access `WRITE_PARALLEL` without importing a specific internal `mapl3g_*` module directly, which violates the intended API boundary.

## Fix
Add `WRITE_PARALLEL` to `mapl_base3g` (base3g/API.F90), which is already re-exported by the `MAPL` umbrella.

tclune:
## Summary
- Adds `WRITE_PARALLEL` from `mapl3g_FileIO` to `mapl_base3g` (base3g/API.F90)
- Makes `WRITE_PARALLEL` accessible via `use MAPL` without requiring direct use of internal `mapl3g_*` modules

## Motivation
Components migrating from MAPL2 to MAPL3 need access to `WRITE_PARALLEL` through the `MAPL` umbrella module. Previously it was only available by importing `mapl3g_FileIO` directly, violating the intended API boundary.

## Depends on
Nothing — this can merge independently.

## Blocked by
None.

Closes #4771

### DIFF
--- a/base3g/API.F90
+++ b/base3g/API.F90
@@ -22,6 +22,7 @@ module mapl_base3g
           MAPL_SunGetDaylightDuration, MAPL_SunGetDaylightDurationMax, &
           MAPL_SunGetLocalSolarHourAngle, MAPL_SunOrbit
    use MAPL_TimeInterpolation, only: MAPL_Interp_Fac, MAPL_ClimInterpFac
+   use mapl3g_FileIO, only: WRITE_PARALLEL
    implicit none(type,external)
    private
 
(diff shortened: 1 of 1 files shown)


tclune:
Superseded by #4773 which already includes this change.

Issue #4855, opened 2026-05-11
Delete MAPL_Base module from Base_Base.F90


tclune:
MAPL_Base only re-exports MAPL_UNDEF from MAPL_Constants. Replace the use in MAPL_BaseMod with a direct use of MAPL_Constants and delete the MAPL_Base module. External callers in GEOSsurface_GridComp are currently inactive in CMake.\n\nPart of #4847.

tclune:
Resolved by #4863 which deletes Base_Base.F90 entirely.

Issue #4847, opened 2026-05-11
Eliminate Base_Base by migrating all callers to geom equivalents


tclune:
## Summary

`Base_Base.F90` and `Base_Base_implementation.F90` contain the last remaining legacy v2 interfaces in MAPL base. This umbrella issue tracks the full sequence of PRs needed to migrate all callers to v3 equivalents and delete these files.

## Remaining interfaces

| Legacy | v3 Replacement | Module |
|--------|---------------|--------|
| `MAPL_GRID_INTERIOR` / `MAPL_GridGetInterior` | `grid_get_interior` | `mapl3g_GridGet` |
| `MAPL_GridGetCorners` | `grid_get_corners` | `mapl3g_GridGet` |

(`MAPL_UNDEF` is already available directly from `mapl_Constants` — no migration needed.)

## Sequence of PRs (must merge in order)

1. **GEOS-ESM/MAPL#4846** — Migrate MAPL-internal callers and delete `Base_Base`
2. **GEOS-ESM/GEOSgcm_GridComp#1406** — Migrate GEOSgcm component callers
3. **GEOS-ESM/geos-chem#34** — Migrate `Chem_GridCompMod.F90`

## Definition of done

`Base_Base.F90` and `Base_Base_implementation.F90` are deleted from MAPL.

PR #4856, opened 2026-05-11
Delete MAPL_Base module from Base_Base.F90

tclune:
## Summary

- `MAPL_Base` only re-exported `MAPL_UNDEF` from `MAPL_Constants`
- Replaced `use MAPL_Base` in `MAPL_BaseMod` with `use MAPL_Constants, only: MAPL_UNDEF` directly
- External callers in `GEOSsurface_GridComp` (`CatchmentRst.F90`, `rmTinyCatchParaMod.F90`, `mod_process_hres_data.F90`) are currently inactive in CMake

Refs #4855, #4847


tclune:
Superseded by #4863 which deletes Base_Base.F90 entirely.

PR #4869, opened 2026-05-12
Add TimeAccumulate with unit tests.

darianboggs:
## Types of change(s)

## Checklist

## Description
Statistic computations were implemented through FrequencyAspect and the associated transforms previously. Now the computations are handled with StatisticsGridComp. FrequencyAspect and the associated transforms were removed in a previous PR. One statistic handled by FrequencyAspect, accumulation (summation) in time, was not implemented with StatisticGridComp. This PR adds a new type derived from AbstractStatisticsGridComp called TimeAccumulate to calculate the missing accumulation statistic. The PR includes unit tests of the new type. Since the procedures are bound to TimeAccumulate type, the unit tests mirror the tests individually. An integration test requires creating a TimeAccumulate object, and it is out of the scope of this PR.


gridcomps/StatisticsGridComp/tests/Test_TimeAccumulate.pf
```diff
@@ -0,0 +1,556 @@
...
+   ! Helper: fill an R4 field with a constant value
+   ! -----------------------------------------------------------------------
+   subroutine fill_r4(field, value, rc)
+      type(ESMF_Field), intent(inout) :: field
+      real(ESMF_KIND_R4), intent(in) :: value
+      integer, optional, intent(out) :: rc
+      integer :: status
+      call ESMF_FieldFill(field, dataFillScheme='const', const1=real(value, ESMF_KIND_R8), _RC)
+      _RETURN(_SUCCESS)
+   end subroutine fill_r4
```
tclune, line 64: Don't change, but wrapping a single procedure in a wrapper seems unnecessarily verbose here.

gridcomps/StatisticsGridComp/StatisticsGridComp.F90
```diff
@@ -111,6 +112,12 @@ subroutine advertise_item(gridcomp, iter, rc)
                has_deferred_aspects=.true., _RC)
           call MAPL_GridCompAddVarSpec(gridcomp, varspec, _RC)
           call advertise_time_max_internal_fields(gridcomp, name, _RC)
+       case ('accumulate')
+          period = mapl_HConfigAsTimeInterval(hconfig, keystring='period', _RC)
```
tclune: This is done for all cases.  Please refactor it out of the SELECT CASE.

gridcomps/StatisticsGridComp/TimeAccumulate.F90
```diff
@@ -0,0 +1,313 @@
...
+      call MAPL_GridCompGetInternalState(gridcomp, internal_state, _RC)
+      call mapl_FieldGet(this%f, short_name=name, _RC)
+      call esmf_StateGet(internal_state, 'sum_'//name, field=sum_f, _RC)
+
+      call MAPL_AssignFptr(this%f, f, _RC)
+      call MAPL_AssignFptr(sum_f, sum_f_ptr, _RC)
+
+      where (f /= MAPL_UNDEF)
+         sum_f_ptr = sum_f_ptr + f
+      end where
```
tclune: When F is MAPL_UNDEF, don't you also need to set sum_F_ptr to UNDEF?   Otherwise the accumulation is uneven?   For averaging one can change the denominator.  But here?

tclune (changes requested):
One change and one important question (inline comments)

gridcomps/StatisticsGridComp/tests/Test_TimeAccumulate.pf
darianboggs: Yes. I don't know why Claude does that.

gridcomps/StatisticsGridComp/StatisticsGridComp.F90
darianboggs: I will. It seems pointless. It isn't used in the procedure.

gridcomps/StatisticsGridComp/TimeAccumulate.F90
darianboggs: I will. I told Claude to convert accumulator to a AbstractTimeStatistic before I got rid of the accumulator, and I kept the file it generated. I think that Claude tried to mimic TimeAverage instead of analyzing the calculations in accumulator.

gridcomps/StatisticsGridComp/StatisticsGridComp.F90
darianboggs: Done. Removed completely from `advertise_item`.

gridcomps/StatisticsGridComp/TimeAccumulate.F90
darianboggs: Done. do concurrent would be more efficient, but I used nested where blocks to minimize the change. Could be changed to do concurrent if greater efficiency is needed.

tclune (dismissed):

PR #4886, opened 2026-05-13
This adds adds a new option to ExtDataDriver.x

bena-nasa:
This adds a new option to ExtDataDriver.x so that I can have it run like the model, where the clock is ticked BEFORE history if I want.

For various reasons in my version of the cap. I was ticking the clock after 3 components run (extdata, root, history). But sometimes you might want to run like the cap used by the model. Useful for truly reproducing how the model runs for testing.

## Types of change(s)

## Checklist

### DIFF
--- a/Tests/ExtDataDriverGridComp.F90
+++ b/Tests/ExtDataDriverGridComp.F90
@@ -36,6 +36,7 @@ module ExtData_DriverGridCompMod
      type(ESMF_VM) :: vm
      type(ESMF_Time), allocatable :: times(:)
      logical :: run_fbf = .false.
+     logical :: run_like_model = .false.
    contains
      procedure :: set_services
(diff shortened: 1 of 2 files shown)


Tests/ExtDataDriverGridComp.F90
```diff
@@ -36,6 +36,7 @@ module ExtData_DriverGridCompMod
      type(ESMF_VM) :: vm
      type(ESMF_Time), allocatable :: times(:)
      logical :: run_fbf = .false.
+     logical :: run_like_model = .false.
```
tclune, line 39: Hoping for a better name/key if this is needed in MAPL3.  (Assuming it is not needed there.)

tclune (approved):

PR #4901, opened 2026-05-15
avoid getting values from un-associated array when built by gfortran 

weiyuan-jiang:
…uilt by gfortran

## Types of change(s)

## Checklist

## Description
check before assign tilelons and tilelats in MAPL_Locstreamget


base/MAPL_LocStreamMod.F90
```diff
@@ -252,23 +252,29 @@ subroutine MAPL_LocStreamGet(LocStream, NT_LOCAL, nt_global, TILETYPE, TILEKIND,
...
     end if
 
     if (present(tilelats)) then
 #ifdef __GFORTRAN__
-       allocate(tmp_rptr(lbound(locstream%Ptr%Local_GeoLocation,1):ubound(locstream%Ptr%Local_GeoLocation,1)))
-       do i = lbound(locstream%Ptr%Local_GeoLocation,1), ubound(locstream%Ptr%Local_GeoLocation,1)
-         tmp_rptr(i) = locstream%Ptr%Local_GeoLocation(i)%y
-       enddo
-       tilelats => tmp_rptr
+       tilelats => null()
```
tclune, line 270: @mathomp4 Has gfortran really not fixed such a terrible compiler problem in all these years?  The `#ifdef ` simply should not be necessary.

@weiyuan-jiang I'm not sure what the issue is, but ALLOCATE always creates a target, so I'm not sure how your changes affect anything.   My guess is that a different bug is lurking here and the old workaround an d the new one here are only bandaids.

tclune (approved):

base/MAPL_LocStreamMod.F90
weiyuan-jiang: In some simple situation, locstream%Ptr%Local_GeoLocation is not allocated, which would lead to crash without this check

weiyuan-jiang: I agree the gfortran fix is terrible. It creates a memory leak.

PR #4910, opened 2026-05-15
Update MAPL from 2.59 to 2.69.0

lizziel:
This is a zero diff update to bring in MAPL 2.69. Prior to this version we used MAPL 2.59. For a list of changes per MAPL release see https://github.com/GEOS-ESM/MAPL/releases.

The reason for this update is to bring in fixes and error handling improvements for ExtData2G which is currently beta in GCHP.


lizziel:
Sorry, wrong target Github organization!

PR #4933, opened 2026-05-18
Feature/wdboggs/add variance statistic

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR adds variance to StatisticsGridComp.


gridcomps/StatisticsGridComp/tests/Sampler.F90
```diff
@@ -0,0 +1,281 @@
+#include "MAPL.h"
+module mapl3g_Sampler
```
tclune: This is the test layer, so the name may survive.  But we already have the term "Sampler" for the I/O capabilities that Yonggang implemented.     

Also, since this is just to support it does not need the usual `mapl*_` name prefix.

No action for now.  Mostly reminding myself of the issue.

gridcomps/statistics/TimeVariance.F90
```diff
@@ -0,0 +1,276 @@
...
+
+   public :: TimeVariance
+   public :: advertise_time_variance_internal_fields
+   public :: WELFORD, SHIFTED
+
+   enum, bind(c)
+      enumerator :: VARIANCE_ALGORITHM = 0
+      enumerator :: WELFORD
+      enumerator :: SHIFTED
+   end enum
```
tclune, line 25: Names should be more symmetric?

First is not public?

Will need another PR to align this with the other enum-like entities that MAPL exports.  (But wait for my existing reorg to complete.)

gridcomps/StatisticsGridComp/TimeVariance.F90
```diff
@@ -0,0 +1,276 @@
...
+      procedure :: get_alarm
+   end type TimeVariance
+
+   interface TimeVariance
+      module procedure new_TimeVariance
+   end interface TimeVariance
+
+contains
+
+   function new_TimeVariance(unusable, gridcomp, f, var_f, alarm, algorithm, biased, rc) result(stat)
```
tclune: Constructors are not permitted to have failure modes.   Will need to either to lazy initialization and just copy the args or switch to some sort of init() (or both).

gridcomps/StatisticsGridComp/WelfordCovarianceKernel.F90
```diff
@@ -0,0 +1,309 @@
...
+   subroutine advertise(this, gridcomp, name, rc)
+      class(WelfordCovarianceKernel), intent(inout) :: this
+      type(esmf_GridComp), intent(inout) :: gridcomp
+      character(*), intent(in) :: name
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+      type(VariableSpec) :: varspec
+
+      _UNUSED_DUMMY(this)
```
tclune: move to end of procedure

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Relaxed the comparison standard for grid_is_ok in case a grid is r4
 
 ### Added
+- variance to StatisticsGridComp
```
tclune:
```suggestion
- Extended StatisticsGridComp to support variance of a single field.
```

gridcomps/statistics/StatisticsGridComp.F90
```diff
@@ -206,6 +212,9 @@ function make_item(name, iter, clock, rc) result(stat)
           case ('accumulate')
              deallocate(stat) ! gfortran workaround
              stat = make_accumulate_stat(name, iter, alarm, _RC)
+          case ('variance')
+             deallocate(stat) ! gfortran workaround
+             stat = make_variance_stat(name, iter, alarm, _RC)
```
tclune, line 217: Note to self:  For diurnal variation there will need to be two distinct time intervals at play.

tclune (changes requested):
Some minor changes requested inline.

gridcomps/StatisticsGridComp/tests/Sampler.F90
darianboggs: Claude ignored it. For the sake of testing, I'll just use the Claude `call random_number(x)` instead of Sampler.F90 which generates a random gaussian distribution.

gridcomps/statistics/TimeVariance.F90
darianboggs: Changed to avoid the need for VARIANCE_ALGORITHM enumerator.

gridcomps/StatisticsGridComp/WelfordCovarianceKernel.F90
darianboggs: Done

gridcomps/StatisticsGridComp/TimeVariance.F90
darianboggs: Fixed with private module initialize subroutine that is called once by update

tclune (approved):

PR #4966, opened 2026-05-22
Replace  MAPL_GridCompSpecs_ACG.py with  MAPL_GridCompSpecs_ACGv3.py in MAPL3

darianboggs:
## Types of change(s)

## Checklist

## Description
For MAPL 3, I created a new script, MAPL_GridCompSpecs_ACGv3.py, to perform automatic code generation, while MAPL 2 continues to use MAPL_GridCompSpecs_ACGv.py for the same function. This prevented merges from MAPL 2 from overwriting the MAPL 3 automatic code generation. This required adding an argument in CMakeLists.txt to specify which ACG to use. Now that MAPL 3 is more mature, and we are not merging from MAPL 2, we no longer need to keep two versions of ACG in MAPL 3.

This PR:
1. removes the MAPL 2 version of ACG,
2. renames MAPL_GridCompSpecs_ACGv3.py to MAPL_GridCompSpecs_ACG.py so that the MAPL 3 automatic generation is performed for any existing calls to MAPL_GridCompSpecs_ACG.py
3. removes the argument in CMakeLists.txt for calling MAPL_GridCompSpecs_ACGv3.py.


darianboggs:
Added @tclune and @mathomp4 as reviewers for their feedback

cmake/mapl_acg.cmake
```diff
@@ -85,11 +85,9 @@ function (mapl_acg target specs_file)
     set (_generator_dir ${esma_etc}/MAPL)
   endif ()
 
-  if (ARGS_3g)
```
pchakraborty, line 88: Does this change make the arg `3g` redundant? The GCM builds have been using this extra arg to call the ACGv3, but the removal does not seem to throw any errors.

tclune: Yes - pretty much any string with "3g" is now needing to be fixed.

tclune (commented):
I suspect this cannot pass CI because the client subrepos explicitly mention 3g.   So you'll need one PR in MAPL that has both 3g and not-3g pointing to the same thing.  Then you can PR in the other repos to get rid of 3g.  ANd then finally in MAPL to get rid of 3g.

cmake/mapl_acg.cmake
tclune: > Yes - pretty much any string with "3g" is now needing to be fixed.

OTOH, it clearly did. :-)

mathomp4:
@darianboggs Can you add a changelog entry for this. This is significant enough to warrant a mention I think.

darianboggs:
CHANGELOG.md entry added to local repository branch.

darianboggs:
I'll push up my local branch, and that should fix the failing test. I designed the change so that existing MAPL2 CMakeLists.txt in GCM and elsewhere would not need to change the call. Though spec files will need to change between MAPL 2 and MAPL 3, the CMake should not need to change. So the `acg3` option is still there in mapl_acg.cmake, but it does nothing. Once that flag is removed from MAPL 3 components that have been converted, we will remove that option. In the meantime, it is ignored. ACG will still function the same in MAPL 2 as it has.

tclune (approved):

PR #5006, opened 2026-05-27
Bugfix/wdboggs/fix acg lookup dict

darianboggs:
## Types of change(s)

## Checklist
The ACG unit tests were run.

## Description
ACG allows aliases for row values in some columns like DIMS. We have advertised that users can use aliases OR the values the aliases point to. ACG aliases do not appear in MAPL, so users may be more familiar with the actual values.

This "bi-directional" lookup did not work. This PR fixes that and adds unit tests of this mechanism in ACG.


darianboggs:
Originally this PR would have merged the changes into `main`. That was incorrect.
To be safe, I am closing it and creating a new PR.

PR #5007, opened 2026-05-27
Bugfix/wdboggs/fix acg lookup dict

darianboggs:
## Types of change(s)

## Checklist
(Ran only the ACG unit tests)

## Description
ACG allows aliases for row values in certain columns, like DIMS. It should have allowed aliases or actual values, but it did not do that. This PR fixes that and adds unit tests to test that it works.


tclune (approved):
You do now need to get in the practice of opening a GH issue even for small things.  Claude makes this pretty easy.

PR #5027, opened 2026-05-29
Remove KE => KeywordEnforcer alias workaround

tclune:
## Types of change(s)

## Description

Removes the `KE => KeywordEnforcer` alias that was used as a workaround for an Intel compiler issue with ambiguous/multiple import paths to `KeywordEnforcer`. With the cleaner module layering now in place, there is a single canonical import path, making the alias unnecessary.

All `class(KE)` dummy argument declarations replaced with `class(KeywordEnforcer)` directly across `infrastructure/` and `superstructure/`. NAG build passes cleanly.

This is a trial PR to confirm the Intel compiler no longer needs the workaround.

## Checklist
- [x] Ran the Unit Tests (NAG local build passes)


mathomp4:
@tclune This is interesting. This PR has no files changed. Was it superseded by another PR at some point?

Issue #5031, opened 2026-06-01
Establish gridcomp-specific prefix convention for entities defined within gridcomps


tclune:
Entities defined within a gridcomp (e.g. types, functions, constants local to `extdata/`, `statistics/`, `history/`, etc.) should carry a gridcomp-specific prefix (e.g. `ExtData_`, `Statistics_`, `History_`) rather than being unprefixed or using the `MAPL_` prefix. This is analogous to the `MAPL_` prefix convention enforced for the MAPL library itself (see issue #5030). This would make the provenance of every public symbol immediately clear from its name.

Issue #5030, opened 2026-05-31
Phase 2: Enforce mapl_ prefix on all public API symbols


tclune:
## Parent issue
Part of #5026

## Prerequisite
Phase 1 (#5029) must be complete so the public API is well-defined before renaming begins.

## Goal
Every symbol exposed through an `Export.F90` umbrella must carry the `mapl_` (or `MAPL_`) prefix, making the namespace unambiguous for downstream consumers.

## Strategy
1. Audit the trimmed `Export.F90` lists from Phase 1 for symbols lacking the prefix.
2. Rename each non-conforming symbol and update all call sites across MAPL.
3. Open coordinated PRs in dependent repos (GEOSgcm components, GEOS, etc.) to update their call sites.
4. Sequence PRs so that CI passes in each repo: rename in MAPL first (with a compatibility alias if needed), then remove the alias once downstream repos are updated.

## Notes
- This phase **requires coordinated PRs across multiple repositories**.
- Compatibility aliases (or a transition branch strategy) may be needed to keep CI green during the rollout.

## Acceptance criteria
- All symbols in every `Export.F90` carry the `mapl_`/`MAPL_` prefix.
- CI passes in MAPL and all known downstream repos.

Issue #5026, opened 2026-05-29
Enforce use of Internal/Export umbrellas for all cross-directory USE statements


tclune:
## Background

A recent audit of the \`Internal.F90\` umbrella modules (#5021) found that **8 of 11 are effectively dead** — consumers across MAPL consistently bypass the umbrellas and \`use\` leaf modules directly. Only \`mapl_Enums_internal\` has been consistently adopted.

## Problem

The Internal/Export umbrella pattern provides no consolidation benefit if consumers ignore it. Cross-directory \`use mapl_<leaf>_mod\` statements create a web of fine-grained dependencies that are hard to audit and maintain.

## Goal

1. **Enforce umbrella adoption**: Migrate all cross-directory \`use mapl_<leaf>_mod\` statements to \`use mapl_<layer>_internal, only: ...\` (for MAPL-internal consumers) or \`use mapl_<layer>_export, only: ...\` (for external consumers).

2. **Trim overly-inclusive Internal umbrellas**: Once adoption is enforced, it becomes clear which symbols are genuinely needed cross-directory. Remove symbols that are not actually consumed externally.

3. **Special case — enums**: Enum *parameters/constants* (e.g. \`MAPL_CONSERVE_NONE\`) are intended for broad consumption and should stay in \`mapl_Enums_internal\`. Enum *types* (e.g. \`MAPL_ConservationType\`) may not be needed outside their defining layer and should be audited separately.

4. **Best candidate to start**: \`infrastructure/vertical/vertical_grid/\` — the most diverse set of external consumers (7+ directories, 6+ leaf modules) with a dead umbrella.

## Audit findings (from #5021)

| Internal module | Dead? | External consumers of leaf modules |
|---|---|---|
| \`mapl_Enums_internal\` | No — actively used | 9 directories |
| \`mapl_field_internal\` | Yes | 5 directories |
| \`mapl_field_bundle_internal\` | Yes | 3 directories |
| \`mapl_esmf_internal\` | Yes (only own Export.F90) | everywhere |
| \`mapl_geom_internal\` | Yes | 4 directories |
| \`mapl_regridder_mgr_internal\` | Yes | 5 directories |
| \`mapl_vertical_grid_internal\` | Yes | 7+ directories |
| \`mapl_mp_utils_internal\` | Mostly (1 symbol externally) | — |
| \`mapl_generic_internal\` | Yes | 4 directories |
| \`mapl_state_internal\` | Yes | 4 directories |
[...]

PR #5032, opened 2026-06-01
Lock down Export umbrella modules; rename unprefixed enums to MAPL_-prefixed (Phase 1, #5026/#5029)

### LINKED ISSUE #5029: Phase 1: Lock down Export umbrella modules to minimal public API
## Parent issue
Closes part of #5026

## Goal
Restrict each `Export.F90` umbrella module to expose only the symbols genuinely needed by external consumers. Start by making everything private (or removing re-exports) and then re-add only what is proven necessary by compiling MAPL gridcomps and downstream code (e.g. GEOS).

## Strategy
1. For each `Export.F90` in MAPL, reduce the `use` / `only` lists to the bare minimum.
2. Build MAPL and its gridcomps; restore any symbol that causes a compile failure with a comment explaining why it must be public.
3. Build downstream (GEOS) against the trimmed API and repeat.
4. The resulting failures open the necessary discussions about whether a symbol *should* be public.

## Notes
- No PRs to other repos are required during this phase; all changes are contained in MAPL.
- This phase is a prerequisite for Phase 2 (enforcing the `mapl_` prefix on all public symbols).

## Acceptance criteria
- Every `Export.F90` umbrella contains only symbols that are demonstrably consumed by code outside their defining layer.
- MAPL builds cleanly.

tclune:
## Summary

Implements Phase 1 of the Export umbrella lock-down (#5026/#5029): all 12 `Export.F90` umbrella modules now have `private` statements with explicit `public ::` declarations, enforcing a clean boundary between MAPL-internal and external-consumer APIs.

### Export umbrella lock-down
- Added `private` + explicit `public ::` to all 12 `Export.F90` files
- Deleted three now-redundant `API.F90` shim files (`field/`, `field_bundle/`, `vertical_grid/`)
- Removed symbols with no external consumers from export umbrellas:
  - `geom/Export.F90`: removed `XYGeomSpec`, `XYGeomFactory`, `CubedSphereGeomSpec`, `CubedSphereDecomposition` and related helpers
  - `vertical_grid/Export.F90`: removed `IntegerPair`
  - `generic/Export.F90`: removed `AbstractUserSetServices`, `DSOSetServices`

### Vertical-grid public API
- Added `BasicVerticalGrid`, `BasicVerticalGridSpec`, `BasicVerticalGridFactory`, and `VerticalGridManager`/`get_vertical_grid_manager` to `vertical_grid/Export.F90` as part of the canonical public API

### Enum renames (~35 files)
All unprefixed enum constants and types renamed to their `MAPL_`-prefixed equivalents:
- `FIELDBUNDLETYPE_*` → `MAPL_FIELDBUNDLETYPE_*`
- `STATEITEM_ALLOCATION_*` → `MAPL_STATEITEM_ALLOCATION_*`
- `GENERIC_COUPLER_*` → `MAPL_GENERIC_COUPLER_*`
- `VectorBasisKind` / `VECTOR_BASIS_KIND_*` → `MAPL_VectorBasisKind` / `MAPL_VECTOR_BASIS_KIND_*`
- `FieldBundleType_Flag` → `MAPL_FieldBundleType_Flag`

### Module structure cleanup
- All production files that had vtable workarounds (scoped `use` inside subroutines, direct leaf imports) restored to canonical module-level imports via `mapl_vertical_grid_export`
- 14 test files cleaned up to use the export umbrella rather than direct leaf imports; only `Test_BasicVerticalGrid.pf` and `Test_VerticalGridManager.pf` retain direct leaf imports (appropriate for layer-specific tests)

## Testing

Full clean build (model + tests) passes with GFortran.

## Related issues

- Closes #5029 (Phase 1)
- Part of #5026
- Phase 2 (symbol prefix enforcement) tracked in #5030
- Gridcomp prefix convention tracked in #5031


apps/Regrid_Util/Regrid_Util.F90
```diff
@@ -430,7 +426,7 @@ subroutine main()
                                startTime=tSeries(1), _RC )
 
    bundle=ESMF_FieldBundleCreate(name="cfio_bundle",_RC)
-   call MAPL_FieldBundleSet(bundle, fieldBundleType=FIELDBUNDLETYPE_BASIC, _RC)
+   call MAPL_FieldBundleSet(bundle, fieldBundleType=MAPL_FIELDBUNDLETYPE_BASIC, _RC)
```
pchakraborty, line 429: Why are we exposing FIELDBUNDLETYPE_BASIC?

enums/Export.F90
```diff
@@ -7,60 +7,44 @@ module mapl_Enums_export
...
    ! CouplerPhases
    public :: MAPL_GENERIC_COUPLER_INITIALIZE, MAPL_GENERIC_COUPLER_UPDATE
    public :: MAPL_GENERIC_COUPLER_INVALIDATE, MAPL_GENERIC_COUPLER_CLOCK_ADVANCE
 
+   ! GenericPhases
+
+   ! CouplerPhases
+
    ! StateItemAllocation
+   public :: MAPL_StateItemAllocation
```
pchakraborty, line 30: Do we expect `StateItemAllocation` used anywhere in GEOS?

enums/Export.F90
```diff
@@ -7,60 +7,44 @@ module mapl_Enums_export
...
    ! FieldBundleType_Flag
+   public :: MAPL_FieldBundleType_Flag
    public :: MAPL_FIELDBUNDLETYPE_BASIC, MAPL_FIELDBUNDLETYPE_VECTOR
    public :: MAPL_FIELDBUNDLETYPE_BRACKET, MAPL_FIELDBUNDLETYPE_VECTORBRACKET
    public :: MAPL_FIELDBUNDLETYPE_SERVICE, MAPL_FIELDBUNDLETYPE_SERVICE_AGGREGATE
    public :: MAPL_FIELDBUNDLETYPE_SERVICE_SEPARATE, MAPL_FIELDBUNDLETYPE_INVALID
 
+   ! VectorBasisKind
+   public :: MAPL_VectorBasisKind
+   public :: MAPL_VECTOR_BASIS_KIND_INVALID, MAPL_VECTOR_BASIS_KIND_GRID
```
pchakraborty, line 45: I can maybe see MAPL_VECTOR_BASIS_KIND_NS being used in GEOS, but do we need the others?

pchakraborty (dismissed):
I’d prefer an `API.90` to `Export.F90` - “export" is typically used in a different context.

Also, I think the `MAPL_` prefix should be reserved for routines/types that are used in client code, and not internally in MAPL. And it maybe more urgent to make that change that I previously thought.

infrastructure/esmf/Export.F90
darianboggs, line 1: It looks like there are several comment lines that are no longer relevant with deletions.

enums/Export.F90
atrayano: I cannot see we need to expose it to GEOS

darianboggs (commented):
Remove some unneeded comments.

I find 'export' confusing because it suggests data export.

I queried ChatGSFC about the naming for this type of situation, a "Facade" pattern, according to best practices in the Fortran community. Given possilities:
`    module foo_API`
OR
    module foo_export
it recommended:
`    module foo_api`

I queried other suggestions. Apparently, DOE, NCAR, and NASA have naming conventions that it suggests (self-referential logic/tautology noted). It gave suggestions including a "Highly Recommended" convention. The "Highly Recommended" convention resembled our past convention, not surprisingly. One suggestion included the suffix "_facade", which doesn't seem appropriate to our user base. I don't recommend the suggestions, but the analysis of the current conventions is germane, and I include it as what follows:

1. foo_export

Advantages:
It is action-oriented. It clearly tells the developer that this module is responsible for exporting entities to the outside world.

Disadvantages (Not Recommended):
Dealbreaker / Major Confusion: In the Fortran and C-interop world, "export" carries heavy compiler-specific baggage. On Windows/Intel Fortran, developers use directives like !DEC$ ATTRIBUTES DLLEXPORT to expose symbols to dynamic link libraries (DLLs). Naming a module foo_export strongly implies to a systems programmer that this module handles compiler-level symbol exporting or C-binding wrappers, not a high-level API Facade.

Domain Confusion: In scientific computing, "export" is often used to describe data output (e.g., exporting a field to a NetCDF file). A user might think use foo_export gives them routines to write data to disk.


2. foo_API (or foo_api)

Advantages:
Universally Understood: API (Application Programming Interface) is the exact computer science term for what you are building. It unambiguously signals "this is the designated boundary for external users."

Self-Documenting: If an external user sees use foo_internal vs use foo_api, it is immediately obvious which one they are supposed to be using.

Disadvantages:
Fortran is case-insensitive, so it will typically be written and compiled as foo_api. This isn't inherently bad, but "API" is a term more frequently seen in web development or C/C++ libraries than in traditional Fortran codebases.

There are no dealbreakers here. foo_api is a very safe, robust choice.

bena-nasa:
What was wrong with API.F90, why did this change Export.F90 for the public facing routines?

enums/Export.F90
tclune: `GRID` variant is needed in DYN for some internal vectors I think.  And I think  the CTM community may want this aswell.   the `INVALID` case is less crucial, but seems like we should allow it for comparison purposes.

enums/Export.F90
tclune: I fear we need it for components that have sophisticated modify_advertise phases.  E.g., Statistics.   But we can probably tighten it down - or even encapsulate within logical queries.  E.g.,  `if (mapl_IsConnected(item)`.

tclune:
> I queried ChatGSFC about the naming for this type of situation, a "Facade" pattern, according to best practices in the Fortran community. Given possilities:
> ` module foo_API`

I already assigned a separate GH issue for this to you.   Details can go there, but it sounds like Claude is already ok with @pchakraborty 's suggestion.

atrayano (approved):

infrastructure/esmf/Export.F90
tclune: Am going to merge.   @darianboggs you'll need to create a separate GH issue to fix these.   I don't want to second guess and don't want to rerun the CI for this.

PR #5055, opened 2026-06-03
v2: Test using gen2 circleci resources

mathomp4:
I want to test out the new gen2 resources at CircleCI:

https://circleci.com/docs/guides/execution-managed/using-docker/#docker-gen2-benefits

See if they are worth the cost.

Per that page, gen2 is 1.4x faster, but they are also 1.2x more costly...


darianboggs (approved):

mathomp4:
Yeah. These seem to be a bit twitchy compared to "gen1". I might hold off on this for now. I'm trying to get a complete set of runs to see what happens

PR #5078, opened 2026-06-05
MAPL: add FieldBundleFilter; refactor StateGet and RestartHandler

pchakraborty:
Add `FieldBundleFilter`, a new infrastructure utility that removes fields from a `FieldBundle` according to a caller-supplied predicate function.

Overload the `StateGet` generic interface with a `state_get_bundle` procedure that serialises an `ESMF_State` into a flat `FieldBundle`, expanding contained `FieldBundle`s into aliased fields named `<bundle_name>_<index>`.

Simplify `RestartHandler` by replacing the private `get_field_bundle_from_state_` method with a call to the new `MAPL_StateGet` overload, and replacing the two inline filter loops with `MAPL_FieldBundleFilter` and dedicated predicate functions (`predicate_skip_restart_`, `predicate_incomplete_`).

Add unit tests for `FieldBundleFilter` (`Test_FieldBundleFilter.pf`) and the `StateGet` bundle overload (`Test_StateGetBundle.pf`).


infrastructure/field_bundle/FieldBundleFilter.F90
```diff
@@ -0,0 +1,49 @@
...
+   end interface
+
+   abstract interface
+      function field_predicate(field, rc) result(remove)
+         use ESMF
+         type(ESMF_Field), intent(in) :: field
+         integer, optional, intent(out) :: rc
+         logical :: remove
+      end function field_predicate
+   end interface
```
tclune, line 25:
```suggestion
   abstract interface
      function I_field_predicate(field, rc) result(remove)
         use ESMF
         type(ESMF_Field), intent(in) :: field
         integer, optional, intent(out) :: rc
         logical :: remove
      end function I_field_predicate
   end interface
```
making consistent with other MAPL abstract interfaces.

tclune: Technically `I_predicate` seems clear enough.

infrastructure/field_bundle/FieldBundleFilter.F90
```diff
@@ -0,0 +1,49 @@
...
+         integer, optional, intent(out) :: rc
+         logical :: remove
+      end function I_field_predicate
+   end interface
+
+contains
+
+   subroutine bundle_filter(bundle, predicate, rc)
+      type(ESMF_FieldBundle), intent(inout) :: bundle
+      procedure(field_predicate) :: predicate
```
tclune:
```suggestion
      procedure(I_field_predicate) :: predicate
```

infrastructure/field_bundle/FieldBundleFilter.F90
```diff
@@ -0,0 +1,49 @@
...
+   use ESMF
+   use mapl_FieldBundleGetImpl_mod, only: FieldBundleGet
+   use mapl_ErrorHandling_mod, only: MAPL_Return, MAPL_Verify
+
+   implicit none(type, external)
+   private
+
+   public :: FieldBundleFilter
+
+   interface FieldBundleFilter
```
tclune, line 14: I was going to suggest that we might want to "publish" some common filters such as `FilterFieldName(name)` but Fortran functions are not of sufficient power for this.   Did you give some thought about using an abstract type instead that has a `filter` method?

tclune (approved):

infrastructure/field_bundle/FieldBundleFilter.F90
pchakraborty: No, not really. The filtering predicate seemed to be the obvious solution, so went with that.

PR #5083, opened 2026-06-05
v3: Clean up changelog

mathomp4:
I had Claude take a look and clean up our changelog.


tclune:
The question is whether we should apply now or wait for alpha.1.  If the latter, then we need to be careful about including any new commits.  Or repeat the AI prompts.

tclune (approved):

PR #5089, opened 2026-06-05
Suppress NAG module timestamp warnings for Spack-installed ESMF

mathomp4:
## Types of change(s)

## Checklist

## Summary

This PR adds a small CMake workaround for noisy NAG Fortran module timestamp warnings when building against Spack-installed ESMF libraries.

When ESMF is installed from a Spack binary cache, its installed `.mod` files may have normalized/epoch-like timestamps for reproducibility. NAG then emits many warnings of the form:

```text
File ... containing module ... needs recompilation
```

even though the dependency was intentionally installed this way and the build is otherwise valid.

The new CMake logic checks, at configure time, whether:

* the Fortran compiler is NAG, and
* ESMF `.mod` files appear to have ancient/epoch timestamps

If so, it adds NAG’s targeted `-nocheck_modtime` option. This suppresses the timestamp-based warnings without disabling other NAG diagnostics.

## Details

The workaround is intentionally narrow:

* It only applies for NAG.
* It only triggers automatically when old ESMF `.mod` timestamps are detected.
* It preserves normal NAG module timestamp checking for regular source-built dependency stacks.
* Debug details are available through CMake’s normal log-level mechanism, e.g. `--log-level=DEBUG`.

The behavior can still be controlled manually through the CMake cache policy:

```bash
-DMAPL_NAG_MODTIME_POLICY=AUTO     # default: detect old ESMF .mod files
-DMAPL_NAG_MODTIME_POLICY=CHECK    # keep NAG timestamp checking enabled
-DMAPL_NAG_MODTIME_POLICY=NOCHECK  # force -nocheck_modtime
```

## Motivation

This reduces warning noise in NAG builds that use Spack-built dependency stacks, while keeping NAG’s useful stale-module diagnostics enabled in normal cases.


tclune (approved):
That's a lot of code.  But at least it is isolated.

mathomp4:
> That's a lot of code. But at least it is isolated.

Yeah, Claude went a bit verbose. Once it's in, I can try it out with a spack set of libraries. It might just choke on that.

mathomp4:
Oh wait. this was for spack. duh

Issue #5133, opened 2026-06-10
Migrate Cap.F90 and GEOS.F90 to ./mapl (out of GridComps)


tclune:
Some thought is needed in advance to be sure this is a good idea.

My primary motivation is that Cap needs some things exposed that we don't think other gridded components would ever need.  E.g., to directly create a GriddedComponentDriver.     Other thing that may not need to be public as a consequence:  MAPL_GENERIC_INIT_PHASE_SEQUENCE  and server logic (as yet unimplemented).

Note: GEOS.F90 has to follow Cap.F90 in any event.

If this works, then Cap would use CapGridComp via DSO.   This raises the possibility of having variant implementations of CapGridComp in the future, though I don't immediately see any requirements that point in that direction.  But flexibility is flexibility.

PR #5138, opened 2026-06-10
Fixes #5073

### LINKED ISSUE #5073: Hide user_setservices in MAPL_GridComps
Yet another type that we don't want to have to expose to the world via `use MAPL`, update `MAPL_GridCompAddChild` so as to not have to do this.

bena-nasa:
Fixes #5073
This adds new interface for MAPL_GridCompAddChild so they can pass either a DSO or procedure, rather than a `user_setservices` 
note that issue #5133 will also address this as `Cap.F90` also uses `user_setservices`
The legacy interface that uses `user_serservices` in the interface will be removed once the gridcomps outside of MAPL have been updated.

## Types of change(s)

## Checklist


tclune (approved):
Looks like there is a merge issue in the ChangeLog.

PR #5140, opened 2026-06-10
Fix run-nag-build-discover nag-build failure in MAPL.esmf_utils.tests

aoloso:
### Description

This PR fixes the failing GitHub Actions job `run-nag-build-discover / nag-build` by addressing a runtime failure in `MAPL.esmf_utils.tests`.

From run `27289225662` (job `80604889462`), the failure occurs in the R4 info retrieval path (`InfoGetR4`) during `MAPL_InfoGetShared`.  
The fix updates `info_get_r4` in `InfoUtilities.F90` to retrieve the value through an R8 temporary and safely cast to R4, avoiding the NAG runtime abort while preserving intended behavior.

## :file_folder:  Modified files

- `infrastructure/esmf/InfoUtilities.F90`

### DIFF
--- a/superstructure/state/API.F90
+++ b/superstructure/state/API.F90
@@ -2,13 +2,14 @@
 module mapl_state_api
 
    use mapl_StateArithmeticParser_mod, only: MAPL_ParserVariablesInExpression => parser_variables_in_expression
-   use mapl_StateArithmeticParser_mod, only: mapl_StateEval
-   use mapl_StateAddMethod_mod, only: mapl_StateAddMethod
-   use mapl_StateDestroy_mod, only: MAPL_StateDestroy
+   use mapl_StateArithmeticParser_mod, only: mapl_StateEval => StateEval
+   use mapl_StateAddMethod_mod, only: mapl_StateAddMethod => StateAddMethod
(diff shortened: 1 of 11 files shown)


superstructure/state/API.F90
```diff
@@ -18,9 +19,10 @@ module mapl_state_api
    public :: mapl_StateAddMethod
    public :: MAPL_StateDestroy
    public :: MAPL_StateGet
+   public :: MAPL_StateSet
    public :: MAPL_StateGetGeom
    public :: MAPL_StateGetPointer
    public :: mapl_StateEval
 
-   public :: StateMask
+   public :: MAPL_StateMask
```
tclune, line 27: This looks like a change to the public interface.   I fear it is used in GOCART.

tclune (commented):
Is there a reason for the one name change?  Have you checked that it does not break GEOS?   (I'll look at CI in a moment.)

tclune (approved):

superstructure/state/API.F90
aoloso: I should build the full GCM then.

tclune: CI does that.

PR #5152, opened 2026-06-11
Move stats gc to generic checkpointing

### LINKED ISSUE #5056: Enable Statistics to use framework write_restart
For the moment, you will need to kludge something to set the geometry in the gridcomp.  (Does not need to be same as collection as it won't be used - just put something into the mapl section in the hconfig.)

See if you can then set the controls.

Substeps:   

1. Move CheckpointControls into  its own module.   Make components private and add accessors.  (tedious, but claude ...)
2. Add something like `mapl_GridcompSetCheckpointControls(gridcomp, import=..., _RC)`
    - args should be optional, I would think.

bena-nasa:
Fixes #5056 
Apparently someone (not me) added a line to set a geom already in stats gc, so the has_geom issue in write restart is moot for this case

## Types of change(s)

## Checklist


pchakraborty (approved):
LGTM

PR #5176, opened 2026-06-15
Move ESMF_Interfaces.F90 to superstructure/interfaces/ (new MAPL.ESMF_Interfaces)

pchakraborty:
## Summary

`ESMF_Interfaces.F90` was living in `infrastructure/esmf/` only because it could not be placed in `superstructure/generic/` — doing so would create a circular CMake dependency: `MAPL.generic` depends on `MAPL.state`, which in turn already depends on `MAPL.esmf`, and `MAPL.esmf` would have needed to depend on `MAPL.generic` to consume the interfaces. Breaking that cycle requires a dedicated library that sits below both `MAPL.state` and `MAPL.generic` in the dependency graph, with no upward dependencies of its own.

The new `MAPL.ESMF_Interfaces` library (`superstructure/interfaces/`) fills exactly that role: it depends only on `ESMF::ESMF`, so both `MAPL.state` and `MAPL.generic` can safely list it as a dependency without introducing a cycle.

## Changes

- **`infrastructure/esmf/CMakeLists.txt`**: remove `ESMF_Interfaces.F90` from `MAPL.esmf` sources
- **`superstructure/interfaces/`**: new directory with `ESMF_Interfaces.F90`, `API.F90`, and `CMakeLists.txt` defining the `MAPL.ESMF_Interfaces` library
- **`superstructure/CMakeLists.txt`**: add `add_subdirectory(interfaces)` and `MAPL.ESMF_Interfaces` dependency
- **`superstructure/state/CMakeLists.txt`**: add `MAPL.ESMF_Interfaces` dependency
- **`superstructure/generic/CMakeLists.txt`**: add `MAPL.ESMF_Interfaces` dependency
- **`superstructure/generic/API.F90`**: remove now-stale commented-out aliases for `UserCompGetInternalState` / `UserCompSetInternalState`
- **`superstructure/API.F90`**: re-export `mapl_esmf_interfaces_api`
- **`superstructure/generic/{InnerMetaComponent,MAPL_Generic,OpenMP_Support,OuterMetaComponent/*,transforms/CouplerMetaComponent}`**: no module-level changes needed; `mapl_ESMF_Interfaces_mod` is still the same module name

### DIFF
--- a/infrastructure/esmf/CMakeLists.txt
+++ b/infrastructure/esmf/CMakeLists.txt
@@ -6,7 +6,7 @@ esma_set_this (OVERRIDE MAPL.esmf)
 
 set (srcs
     # From esmf_utils/
-    ESMF_Interfaces.F90 # cannot move to MAPL.generic, causes circular dependency
+    # ESMF_Interfaces.F90 moved to superstructure/interfaces/
     ESMF_Utilities.F90
     TypeKind.F90
     InfoUtilities.F90
--- a/superstructure/interfaces/API.F90
+++ b/superstructure/interfaces/API.F90
@@ -1,0 +1,13 @@
+! Export umbrella for the MAPL.ESMF_Interfaces library.
+module mapl_esmf_interfaces_api
+
+   use mapl_esmf_interfaces_mod, only: MAPL_UserCompGetInternalState => UserCompGetInternalState
+   use mapl_esmf_interfaces_mod, only: MAPL_UserCompSetInternalState => UserCompSetInternalState
+
+   implicit none
+   private
+
+   public :: MAPL_UserCompGetInternalState
+   public :: MAPL_UserCompSetInternalState
+
+end module mapl_esmf_interfaces_api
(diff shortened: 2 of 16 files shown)


infrastructure/esmf/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ esma_set_this (OVERRIDE MAPL.esmf)
 
 set (srcs
     # From esmf_utils/
-    ESMF_Interfaces.F90 # cannot move to MAPL.generic, causes circular dependency
+    # ESMF_Interfaces.F90 moved to superstructure/interfaces/
```
tclune, line 9: Shouldn't this line just be deleted - do we really care where it once was?

superstructure/interfaces/API.F90
```diff
@@ -0,0 +1,13 @@
+! Export umbrella for the MAPL.ESMF_Interfaces library.
```
tclune, line 1: Do we really need an entire library for not-even-exceutable code?

tclune (commented):
I'd like to have just a brief discussion to see if there is a way to avoid the heavy hammer of an entire library for a file that does not even contain executable code.

tclune:
Given our updated ESMF understanding, and some new ESMF published interfaces, most of this file has become unnecessary.  Not sure what remains to be done.  @pchakraborty can you reassess?

Issue #5137, opened 2026-06-10
Fixup PythonBridge


tclune:
1. Needs to be exported from "MAPL" umbrella.
2. If it uses `ESMF_Info`, we need to protect the namespace.
3. We need to confirm external interface before we release MAPL.  Coordinate with @FlorianDeconinck

PR #5183, opened 2026-06-18
Add py PyBridge to MAPL

aoloso:
## Types of change(s)

## Checklist

## Description
Added PythonBridge to MAPL interface

## Related Issue
[5137](https://github.com/GEOS-ESM/MAPL/issues/5137)


python/API.F90
```diff
@@ -0,0 +1,5 @@
+module mapl_PythonBridge_api
+
+   use mapl_PythonBridge_mod
```
pchakraborty, line 3: This would expose all public routines in `mapl_PythonBridge_mod`. Is that what we want here?

tclune: Someone should reach out to @FlorianDeconinck to see which procedures should be public.   To be safe we can make *none* of them public until they are needed.    (We can always add public interfaces, but we cannot take them back ...)

aoloso: I had @FlorianDeconinck  take a look and he said the changes were good. I added API.F90 instead of exporting `mapl_PythonBridge_mod` directly because according to Florian, it is possible additional modules may be added later in the python directory that we may want to expose.

FlorianDeconinck: So the `MAPL_PythonBridge` was already organized with a public API, e.g. everything that is in [Python/PythonBridge.F90](https://github.com/GEOS-ESM/MAPL/blob/main/Python/PythonBridge.F90) is public.

The private helpers, as well as the functions called from Python upward, are burried in the `MAPL_PythonBridge` directory below.

We can easily forward declare into `API.F90` - or merge files.

Do we have anything else in the `Python/` subdirectory we want to move in the public API?

pchakraborty: I think renaming `PythonBridge.F90` as `API.F90`, with the module name `mapl_PythonBridge_api` is a good idea.

aoloso: As we just discussed in the Python meeting, the idea is for API.F90 to be content free. We only need to expose objects that should be seen externally. For now I'll expose only the objects that are public in `PythonBridge.F90`, not anything below.

pchakraborty (dismissed):

pchakraborty (approved):

PR #5185, opened 2026-06-18
Fix typo in allreducesum.H preprocessor guard

mathomp4:
## Types of change(s)

## Checklist

## Description

Fix a typo in `infrastructure/esmf/comms/allreducesum.H` line 6 where the preprocessor guard was inconsistent: the conditional checked for `NAMESTR_` but the `#undef` attempted to undefine `NAMESTr_` (with lowercase 'r').

This fix ensures the guard properly executes, preventing potential macro collision issues when the template is included multiple times with different `RANK_` and `VARTYPE_` combinations in `infrastructure/esmf/comms/Comms.F90`.

The fix changes:
```fortran
#ifdef NAMESTR_
#undef NAMESTr_    <- was this (typo)
#endif
```

To:
```fortran
#ifdef NAMESTR_
#undef NAMESTR_    <- now this (correct)
#endif
```

This is part of the C preprocessor template pattern where macros are repeatedly defined and cleaned up between multiple template header inclusions.

### DIFF
--- a/infrastructure/esmf/comms/allreducesum.H
+++ b/infrastructure/esmf/comms/allreducesum.H
@@ -3,7 +3,7 @@
 #endif
 
 #ifdef NAMESTR_
-#undef NAMESTr_
+#undef NAMESTR_
 #endif
 
 #define NAME_ comms_allreduce_sum_


mathomp4:
Note to @pchakraborty or @atrayano or whoever sees this. I was fiddling with an app and had it look at MAPL. It seemed to have found this. I think it's a bug? I'm not sure. But it seems simple enough.

pchakraborty (approved):
LGTM

Issue #5165, opened 2026-06-12
Remove pointer declaration and “GetPointer” for vectors


pchakraborty:

darianboggs:
OK.

darianboggs:
Closed by #5192

PR #5192, opened 2026-06-22
Feature/wdboggs/#5165 acg remove pointers vectors

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR adds a new behavior. Previously, ACG generated declaration and get pointer declaration files according to the command line options for all specs. Now, the ACG generates the files only for MAPL_STATEITEM_FIELD. Existing ACG spec files do not specify ITEMTYPE, but MAPL_STATEITEM_FIELD is the default ITEMTYPE. So ACG filters for ITEMTYPE is MAPL_STATEITEM_FIELD or unspecified. It uses a new general spec filter function to do this. The spec filter function decorates a simple function on a value and creates a function that filters a spec. The spec filter function has two optional arguments: key and default, which become parameters in the decorator. If the key parameter is present, the simple function acts on the value in the key value pair. Otherwise the simple function acts on the entire spec. When the key parameter is present, but it is not found in the spec, the value of the default parameter is returned.

## Related Issue
#5165

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - Add new `color_message` function
     - Add helper script for regression test work
 - For ACG, only declare pointer and get_pointer for MAPL_STATEITEM_FIELD
+- For ACG, add spec_filters to generalize testing specs
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


tclune (approved):
I think we need to declare pointers for vectors, but that can be a separate PR.

Requires a naming convention, of course.

darianboggs:
> I think we need to declare pointers for vectors, but that can be a separate PR.
> 
> Requires a naming convention, of course.

That makes sense. Let me know when you want to move forward with that. We can discuss with all relevant parties what the naming convention is.

PR #5193, opened 2026-06-23
fixes #3747

### LINKED ISSUE #3747: Issue creating R4 exports from R8
I have several R8 exports from one gridded component to another. I can output these arrays with MAPL 2.26 History by including the names in `HISTORY.rc`. This works because `MAPL_HistoryGridComp.F90` creates R4 exports from the R8. However, I am not able to do this with MAPL 2.55.

The failure occurs within a call to `MAPL_FieldCreate` using an R4 field that was just created from the R8 export.
https://github.com/GEOS-ESM/MAPL/blob/9b372646a4f8bfb4a29b53096eacbfcc6d3d2a1a/gridcomps/History/MAPL_HistoryGridComp.F90#L2081

That call invokes `MAPL_FieldCreateRename` in `base/Base/Base_Base_implementation.F90` which in turn calls `ESMF_FieldCreate`:
https://github.com/GEOS-ESM/MAPL/blob/9b372646a4f8bfb4a29b53096eacbfcc6d3d2a1a/base/Base/Base_Base_implementation.F90#L1234

This leads to an ESMF error:
```
20250523 161150.823 ERROR            PET00 ESMF_FieldCreate.F90:1747 ESMF_FieldCreateGBArray Value unrecognized or out of range  - - Array indexflag must be the same as the Grid indexflag
20250523 161150.823 ERROR            PET00 ESMF_FieldCreate.F90:32772 ESMF_FieldCreateFromField() Value unrecognized or out of range  - Internal subroutine call returned Error
```

There is a note in `MAPL_FieldCreateRename` warning that the data and the grid must be the same:
https://github.com/GEOS-ESM/MAPL/blob/9b372646a4f8bfb4a29b53096eacbfcc6d3d2a1a/base/Base/Base_Base_implementation.F90#L1213-L1216

I printed the `ESMF_Field` passed to the function in MAPL 2.26 and MAPL 2.55 and they are identical. I am using identical grid resolutions for both runs and all configuration files are the same. The exports are between FV3 and GEOS-Chem and neither of those gridded components changed.

Have you tested MAPL 2.55 with exports declared as R8? In theory they should be converted to REAL4 and successfully output from History. Let me know if you are able to do this.

bena-nasa:
This fixes a long standing bug that History could not write R8 exports downgraded to R4 in files as it used to do. I think I found the issue and fixed. 

I've also update by ExtDataDriver.x application so I can create R8 fields when populating the states to test this. That is why lots of RC files changed. I had to add new column to my dynamic state description.

The issue was that when the fields are downgraded to R4 they were done in a way that had a different index space that we usually create fields with which caused problems later on. The import item was the change in Base

## Types of change(s)

## Checklist

### DIFF
--- a/Tests/VarspecDescription.F90
+++ b/Tests/VarspecDescription.F90
@@ -1,8 +1,4 @@
-#define _SUCCESS      0
-#define _FAILURE     1
-#define _VERIFY(A)   if(  A/=0) then; if(present(rc)) rc=A; PRINT *, Iam, __LINE__; return; endif
-#define _ASSERT(A)   if(.not.A) then; if(present(rc)) rc=_FAILURE; PRINT *, Iam, __LINE__; return; endif
-#define _RETURN(A)   if(present(rc)) rc=A; return
+#include "MAPL_Generic.h"
 
 module VarspecDescriptionMod
    use MAPL
@@ -34,7 +30,6 @@ contains
       logical :: lcomp
       integer, optional, intent(out) :: rc
 
-      character(len=*), parameter :: Iam = 'new_VarspecDescriptionFromConfig'
       integer :: status
 
       type(StringVector) :: svec
@@ -50,7 +45,7 @@ contains
       enddo
 
       lcomp = (svec%size()==6 .or. svec%size()==7)
-      _ASSERT(lcomp)
+      _ASSERT(lcomp, 'Wrong number of columns in state descriptor')
       VarspecDescr%short_name = svec%at(1)
       
       ! Parse and validate precision (column 2)
@@ -99,7 +94,6 @@ contains
       integer, optional, intent(out) :: rc
 
       integer :: status
-      character(len=*), parameter :: Iam = "addNewSpec"
 
       if (specType == "IMPORT") then
          call MAPL_AddImportSpec(GC, &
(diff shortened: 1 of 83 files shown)


Tests/VarspecDescription.F90
```diff
@@ -48,29 +49,42 @@ function new_VarspecDescriptionFromConfig(cf,nwords,rc) result(VarspecDescr)
...
+      ! Parse and validate precision (column 2)
+      tmpstring = svec%at(2)
+      if (trim(tmpstring) == 'R4') then
+         VarspecDescr%precision = MAPL_R4
+      else if (trim(tmpstring) == 'R8') then
+         VarspecDescr%precision = MAPL_R8
+      else
+         PRINT *, Iam, ': Invalid precision "', trim(tmpstring), '"'
+         PRINT *, Iam, ': Must be either R4 or R8'
+         _ASSERT(.false.)
```
tclune:
```suggestion
        _FAIL('Invalid precision "'// trim(tmpstring) // '": Must be either R4 or R8')

```

Tests/VarspecDescription.F90
```diff
@@ -48,29 +49,42 @@ function new_VarspecDescriptionFromConfig(cf,nwords,rc) result(VarspecDescr)
          call svec%push_back(trim(tmpstring))
       enddo
 
-      lcomp = (svec%size()==5 .or. svec%size()==6)
+      lcomp = (svec%size()==6 .or. svec%size()==7)
```
tclune, line 47: Magic numbers ... but won't fix because it is MAPL2.

tclune (dismissed):

tclune (approved):

PR #5199, opened 2026-06-26
Add MAPL_StateMerge to state library

weiyuan-jiang:
## Types of change(s)

## Checklist


superstructure/state/StateMerge.F90
```diff
@@ -0,0 +1,142 @@
...
+   function merge_states(state_a, state_b, unusable, name, rc) result(merged)
+      type(ESMF_State), intent(inout) :: state_a
+      type(ESMF_State), intent(inout) :: state_b
+      class(KeywordEnforcer), optional, intent(in) :: unusable
+      character(len=*), optional, intent(in) :: name
+      integer, optional, intent(out) :: rc
+      type(ESMF_State) :: merged
+
+      integer :: status
+      character(len=ESMF_MAXSTR) :: merged_name
```
tclune: Our usual convention is just to add an underscore for the local variable that defaults to the optional input.
`merged_name` -> `name_`

superstructure/state/StateMerge.F90
```diff
@@ -0,0 +1,142 @@
...
+
+      merged_name = 'merged'
+      if (present(name)) merged_name = name
+
+      merged = ESMF_StateCreate(name=trim(merged_name), _RC)
```
tclune: What about StateIntent?    MAPL3 is using that a bit more heavily now, so you probably should get that right here.  But the logic is a bit messy.

1. If none have intent - no intent.
2. If only one has intent use that one.
3. If both have intent - check that they are the same.

Also consider adding an optional argument to specify StateIntent.

tclune (commented):
I put in some suggestions, but I don't want to obligate you to make those changes.  If you want to delay those fixes to a subsequent pass, I can go in and approve for now.

tclune (dismissed):

tclune (approved):

Issue #5195, opened 2026-06-24
Eliminate MAPL_UserCompSetInternalState


tclune:
(Also corresponding "Get" )

ESMF now provides proper interfaces for these, so there is no need for the MAPL wrapper.    User code uses our macros so no external client code needs to be updated.

1. Delete MAPL interface declarations in ESMF_Interfaces.F90
2. Update macros in ./include/MAPL_private_state.h to use ESMF procedures directly rather than MAPL interfaces.
3.

PR #5201, opened 2026-06-26
Mapl2esmf set get int state

aoloso:
## Types of change(s)

## Checklist

## Description
Replaced MAPL_UserComp[Set,Get]InternalState with ESMF_UserComp[Set,Get]InternalState

## Related Issue
[5195](https://github.com/GEOS-ESM/MAPL/issues/5195)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Replaced MAPL_UserComp[Set,Get]InternalState with ESMF_UserComp[Set,Get]InternalState
 - Added PythonBridge to MAPL interface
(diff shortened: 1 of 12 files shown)


tclune (changes requested):
1) you will need to first update external clients to use the ESMF interfaces, otherwise MAPL CI will fail.  *Only* fix the components that are currently ported:  GWD, GOCART, DYN.

2. OpenMP_SUpport is failing CI.  You should check to see if it can use the cpp macros that I introduced in MAPL3.  (MAPL_private_state.h)

tclune (commented):
AH - the ported gridcomps are already using my macros, so the only fix they need is to remove the "ONLY" clause that explicitly names the interfaces your PR eliminates. Still need to fix those first so that MAPL will pass CI.

Tedious, but we'll need to get good at this as more of GEOS is ported or we'll start tripping ourselves.

PR #5212, opened 2026-06-30
fixes #5210

### LINKED ISSUE #5120: ./pfio

### LINKED ISSUE #5210: R8 History Output in MAPL3
I was playing around and in my History/ExtData tester I changed an export that would be written by History to R8 in the tester component that produces the exports (to mimic what I have done in the analogous test case in MAPL2). However, I noticed that the output file still had R4 variable. I found this odd as I don't recall any sort of games like we do in MAPL2 where R8 is copied to an R4 for output in History 2.0 so was expecting R8). I then found that the import coming into History was R4 even though the export it was connected to was R8. Not sure what is going on but I can't see why it should be coming into History as an R4 import.

I suspect there is a problem with the "mirroring" the imports do, they are not mirroring the connect export's typekind.
I think this is what is going on.
Seems like `specs/VariableSpec.F90` always passes a typekind (which is defaulted to R4 in the derived type definition) to the typekind aspect, thus making mirroring of the typekind impossible, as the constructor for typekindaspect only is set to mirror if the typekind is not passed to the constructor.

bena-nasa:
This fixes #5120
Had to expose MAPL_TYPEKIND_MIRROR
modified tests to exercise R8
modified ComponentDriverGridcomp to support both R4 and R8 imports and exports
Cleaned up some places missing return code when chasing something

## Types of change(s)

## Checklist


gridcomps/componentDriverGridComp/componentDriverGridComp.F90
```diff
@@ -420,10 +422,21 @@ subroutine compare_states(state, reference_state, threshold, rc)
...
-            call mapl_assignFptr(reference_field, reference_ptr, _RC)
-            if (any(abs(ptr - reference_ptr) > threshold)) then
-               _FAIL("state differs from reference state greater than allowed threshold")
+            call ESMF_FieldGet(field, typekind=typekind, _RC)
+            if (typekind == ESMF_TYPEKIND_R4) then
+               call mapl_assignFptr(field, ptr, _RC)
+               call mapl_assignFptr(reference_field, reference_ptr, _RC)
+               if (any(abs(ptr - reference_ptr) > threshold)) then
+                  _FAIL("state differs from reference state greater than allowed threshold")
+               end if
```
tclune, line 431: Don't change now, but "condition + msg" is what `_ASSERT` is for.

infrastructure/esmf/FieldPointerUtilities.F90
```diff
@@ -678,12 +678,12 @@ subroutine copy_broadcast(x, y, rc)
       character(len=*), parameter :: UNSUPPORTED_TK = &
          'Unsupported typekind in FieldCOPY() for '
 
-      conformable = FieldsAreConformable(x, y)
+      conformable = FieldsAreConformable(x, y, _RC)
```
tclune: I think the `_RC` was intentionally omitted here. Please check with @darianboggs

infrastructure/esmf/FieldPointerUtilities.F90
```diff
@@ -802,7 +802,7 @@ subroutine copy(x, y, rc)
       logical :: y_is_double
       character(len=*), parameter :: UNSUPPORTED_TK = 'Unsupported typekind in FieldCOPY() for '
 
-      conformable = FieldsAreConformable(x, y)
+      conformable = FieldsAreConformable(x, y, _RC)
```
tclune: Same comment as above.

tclune (changes requested):
I think the new `_RC` are subtle bugs.  In each case there is logic further down that is exercised in the `.not. conformable` case.

bena-nasa:
> I think the new `_RC` are subtle bugs. In each case there is logic further down that is exercised in the `.not. conformable` case.

Ok, I will remove them, was just chasing something and saw things missing `_RC`

infrastructure/esmf/FieldPointerUtilities.F90
bena-nasa: I restored them

tclune (approved):

PR #5223, opened 2026-07-06
feat(configurable): support per-child run_phases in mapl.children config

pchakraborty:
## Summary

Each child in the `mapl.children` section of a Configurable gridcomp can now declare its own `run_phases` list:

```yaml
mapl:
  children:
    child_A:
      dso: libfoo.so
      config_file: child_A.yaml
      run_phases: [run]
    child_B:
      dso: libbar.so
      config_file: child_B.yaml
      run_phases: [Run1, Run2]
```

At run time, `ConfigurableGridComp` iterates over `mapl.children` and for each child calls `MAPL_GridCompRunChild` once per listed phase name. Children with no `run_phases` entry default to phase `"run"`, preserving existing behaviour. When no `mapl.children` section is present the original `MAPL_GridcompRunChildren` fallback is used.

## Changes

- Add `KEY_CHILDREN_SECTION` and `KEY_RUN_PHASES` parameters
- Introduce `run_children_phases_` helper that reads `run_phases` per child from hconfig
- Default run phase is `"run"` when `run_phases` is not specified


gridcomps/configurable/ConfigurableGridComp.F90
```diff
@@ -49,7 +51,7 @@ subroutine init(gridcomp, importState, exportState, clock, rc)
       ! ASSUME: mapl and states sections always exist
       mapl_cfg = ESMF_HConfigCreateAt(hconfig, keyString=MAPL_SECTION, _RC)
       states_cfg = ESMF_HConfigCreateAt(mapl_cfg, keyString=COMPONENT_STATES_SECTION, _RC)
-      has_export_section = ESMF_HConfigIsDefined(states_cfg, keyString=COMPONENT_EXPORT_STATE_SECTION, _RC)
+      has_export_section = ESMF_HConfigIsDefined(states_cfg, keyString=COMPONENT_EXPORT_STATE_SECTION, rc=status)
```
tclune: I don't understand why you dont' check the return status.   The value is not dependent on the result value - it should not throw an error simply because the keyString is not found.
```suggestion
      has_export_section = ESMF_HConfigIsDefined(states_cfg, keyString=COMPONENT_EXPORT_STATE_SECTION, _RC)
```

gridcomps/configurable/ConfigurableGridComp.F90
```diff
@@ -124,14 +126,60 @@ recursive subroutine run(gridcomp, importState, exportState, clock, rc)
...
+
+      integer :: status, i
+      type(ESMF_HConfig) :: mapl_cfg, children_cfg, child_cfg
+      logical :: has_mapl, has_children, has_run_phases
+      character(ESMF_MAXSTR) :: child_name
+      character(ESMF_MAXSTR), allocatable :: phase_names(:)
+      type(ESMF_HConfigIter) :: iter, b, e
+
+      has_mapl = ESMF_HConfigIsDefined(hconfig, keyString=MAPL_SECTION, _RC)
+      if (has_mapl) then
```
tclune: This is a significant level of nesting that I hope can be improved.  Looking ...

gridcomps/configurable/ConfigurableGridComp.F90
```diff
@@ -124,14 +126,60 @@ recursive subroutine run(gridcomp, importState, exportState, clock, rc)
...
+         mapl_cfg = ESMF_HConfigCreateAt(hconfig, keyString=MAPL_SECTION, _RC)
+         has_children = ESMF_HConfigIsDefined(mapl_cfg, keyString=KEY_CHILDREN_SECTION, _RC)
+         if (has_children) then
```
tclune: Possibly:
```suggestion
_RETURN_UNLESS(has_children)
```

gridcomps/configurable/ConfigurableGridComp.F90
```diff
@@ -124,14 +126,60 @@ recursive subroutine run(gridcomp, importState, exportState, clock, rc)
...
+               child_cfg = ESMF_HConfigCreateAtMapVal(iter, _RC)
+               has_run_phases = ESMF_HConfigIsDefined(child_cfg, keyString=KEY_RUN_PHASES, _RC)
+               if (has_run_phases) then
+                  phase_names = ESMF_HConfigAsStringSeq(child_cfg, keyString=KEY_RUN_PHASES, stringLen=ESMF_MAXSTR, _RC)
+                  do i = 1, size(phase_names)
+                     call MAPL_GridCompRunChild(gridcomp, child_name=trim(child_name), phase_name=trim(phase_names(i)), _RC)
+                  end do
+               else
+                  call MAPL_GridCompRunChild(gridcomp, child_name=trim(child_name), phase_name='run', _RC)
+               end if
```
tclune: A slight help:
```suggestion
               phase_names = ['run']
               if (has_run_phases) then
                  phase_names = ESMF_HConfigAsStringSeq(child_cfg, keyString=KEY_RUN_PHASES, stringLen=ESMF_MAXSTR, _RC)
               end if
               do i = 1, size(phase_names)
                  call MAPL_GridCompRunChild(gridcomp, child_name=trim(child_name), phase_name=trim(phase_names(i)), _RC)
               end do
```

gridcomps/configurable/ConfigurableGridComp.F90
tclune: But might leak the ConfigDestroy step.

gridcomps/configurable/ConfigurableGridComp.F90
tclune: Probably can do
```
_RETURN_UNLESS(has_mapl)
```
No extra config to destroy in this case.

tclune (changes requested):
I think the deep nesting that was added can me significantly ameliorated with some small logic negation.   Further improvement probably requires helper procs, but that is overkill for now.

My other concern is that this is having ConfigurableGridComp look directly inside  of he `mapl` section which is a little bit bad.     Other alternatives would be:

* Have separate dictionary outside the `mapl` section
* Create an accessor function so that a component can learn the phases of its children.
   - This is probably a useful thing, but is a slight violation of encapsulation.

All 3 paths have a bit of a smell.

gridcomps/configurable/ConfigurableGridComp.F90
pchakraborty: This is probably an ESMF bug - the function `ESMF_HConfigIsDefined` returns non zero `rc` if key does not exist.

gridcomps/configurable/ConfigurableGridComp.F90
pchakraborty: That will skip the fallback at the bottom

```Fortran
! Default: run the "run" phase of each child
call MAPL_GridcompRunChildren(gridcomp, phase_name='run', _RC)
```

tclune: But ... if there is no MAPL section ... there will be no children to run!  What am I missing?

pchakraborty: Indeed. I pushed the change.

tclune (dismissed):

tclune (approved):

PR #5238, opened 2026-07-07
fix(configurable): guard against missing states section in init

pchakraborty:
## Problem

`ESMF_HConfigCreateAt` returns a null/undefined HConfig (without signaling an error) when the requested key does not exist in the parent map. Passing that null object to `ESMF_HConfigIsDefined` with a `keyString` then fails internally with:

```
HConfig object MUST be map when key specified
```

This was triggered for any component whose YAML has a `mapl:` section but no `states:` subsection (e.g. `root.yaml` in the `gocart+seasalt+dust` regression test), causing spurious ESMF ERROR entries in every PET log file on every rank.

## Fix

Check for the `states:` key via `ESMF_HConfigIsDefined` before navigating into it, and return early (`_RETURN_UNLESS`) if it is absent.

Also promotes the `has_export_section` rc check from the silent `rc=status` form to `_RC`, since `states_cfg` is now guaranteed to be a valid map at that point.

## Testing

Verified against the `gocart+seasalt+dust` regression test — PET log files no longer contain ESMF HConfig ERROR entries.

### DIFF
--- a/gridcomps/configurable/ConfigurableGridComp.F90
+++ b/gridcomps/configurable/ConfigurableGridComp.F90
@@ -40,7 +40,7 @@ contains
 
       character(:), allocatable :: field_name
       type(ESMF_HConfig) :: hconfig, mapl_cfg, states_cfg, export_cfg, field_cfg
-      logical :: has_export_section, has_default_vert_profile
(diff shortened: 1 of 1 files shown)


tclune (approved):
Glad you found this.

PR #5243, opened 2026-07-09
Add named default pfio server constants

### LINKED ISSUE #5242: Config-driven pfio server selection for History and geom_io
##  Human role

- Human will review and test changes
- Agent can stage, but should not commit changes without approval

## Background

Currently, all pfio client lookups across MAPL are hardwired to the string literals `'o_client'` and `'i_client'`. This prevents components like `HistoryGridComp` from dynamically selecting which server they use, and makes the client/server naming convention confusing (server and client have different names for the same connection).

The original intent of `MaplFramework` was to encapsulate all framework singletons — but the current implementation leaks low-level details (hardwired strings, direct `DirectoryService` access) to components that should not know about them.

## Goals

1. Allow `HistoryGridComp` to select its pfio output server via config (`server:` key), enabling a dedicated history server without code changes.
2. Introduce named Fortran `parameter` constants for default servers so no code ever again hardwires a string literal.
3. Add `MAPL_ConnectToServer` as a clean public API — components never touch `DirectoryService` directly.
4. Unify server/client naming: server and client use the **same name** (1-to-1 only). Old `i_client`/`o_client` names eliminated.
5. Propagate server name from `HistoryGridComp` to its collection children via `make_child_hconfig` — collections look up the client by name but do not connect; parent History owns the connect/signal.
6. Make `GeomPFIO` server-name-aware via optional args, defaulting to the named parameter constants.

## Design decisions

- **No relocation of `MaplFramework`** (vertical grid registration stays; lowering deferred).
- **Default servers are always local** and are wired during `MAPL_CreateServers` (inside `initialize_local_servers`). They are available on model PETs for all subsequent operations.
- **Non-default (remote) servers**: `MAPL_ConnectToServer` must be called after `MAPL_RunServers` returns on model PETs. The correct call site is a component's `GENERIC::INIT_USER` phase.
- **`HistoryGridComp`**: reads `server:` from hconfig in `setServices`; stores in private state (MAPL `_SET_NAMED_PRIVATE_STATE` macro pattern); calls `MAPL_ConnectToServer` in its `init` phase; propagates server name into child collection hconfigs via `make_child_hconfig`.
- **`ExtData`**: deferred — low priority, async read is hard to exploit.
- **Standalone path** (Regrid_Util): writes via `FieldBundleWriter` are already broken today (no `MAPL_CreateServers` → no clients). Not made worse. Separate issue.
- **`RestartHandler`**: forward-compatible — uses default server name parameter constants now; per-component config via MAPL hconfig inheritance deferred to a later issue.
- **`MAPL_ConnectToServer` client type**: defaults to `FastClientThread` for output (matching current `o_client` behavior), `ClientThread` for input. May add optional `client_subclass` arg later.

## Lifecycle timing

```
MAPL_Initialize        → ESMF, profilers, udunits, field dictionary
MAPL_CreateServers     → model_comm, DirectoryService, local default servers + clients
                         (remote server GridComps created — collective, blocks server PETs)
MAPL_CapCreate         → all setServices called recursively (History reads server: key here)
MAPL_RunServers        → server PETs block; model PETs return immediately
MAPL_CapRun
  MAPL_DriverInitializePhases
    GENERIC::INIT_USER → History calls MAPL_ConnectToServer here (safe: after RunServers)
```

### Task 1 — Named server parameter constants (`pfio/ServerNames.F90`)

New module `mapl_DefaultServerNames_mod` in `MAPL.pfio` (no upward deps):

```fortran
module mapl_DefaultServerNames_mod
   implicit none
   character(*), parameter :: MAPL_DEFAULT_INPUT_SERVER  = 'mapl_default_input_server'
   character(*), parameter :: MAPL_DEFAULT_OUTPUT_SERVER = 'mapl_default_output_server'
end module mapl_DefaultServerNames_mod
```

- Add `ServerNames.F90` to `pfio/CMakeLists.txt` source list.
- Export constants from `pfio/API.F90` (and therefore from `MAPL` umbrella).

### Task 2 — Rename default server/client registrations in `MaplFramework`

In `mapl/MaplFramework.F90`:

- `use mapl_DefaultServerNames_mod` at top.
- `initialize_local_servers` (no-`servers:`-section path): replace `'i_server'`/`'i_client'` and `'o_server'`/`'o_client'` with the named parameter constants (same name for both server and client).
- `initialize_clients` (no-`pfio_clients:`-section path): likewise use parameter constants when registering clients and connecting to servers.

### Task 3 — Add `MAPL_ConnectToServer` to `MaplFramework`

New public subroutine in `mapl/MaplFramework.F90`:

```fortran
subroutine MAPL_ConnectToServer(server_name, unusable, client_name, rc)
   character(*), intent(in) :: server_name
   character(*), optional, intent(in) :: client_name  ! defaults to server_name
```

Internally: allocate a `FastClientThread` (or `ClientThread` for input), call `add_client(client_name_, client)`, then `the_mapl_object%directory_service%connect_to_server(server_name, p_client)`.

Add `interface MAPL_ConnectToServer` and add to `public ::` list.

### Task 4 — Update all hardwired `get_client('o_client'/'i_client')` call sites

Replace hardwired string literals with the named parameter constants. Files:

| File | Change |
|------|--------|
| `superstructure/generic/RestartHandler.F90` | `get_client('o_client')` → `get_client(MAPL_DEFAULT_OUTPUT_SERVER)`; same for i_client |
| `infrastructure/geom_io/GeomPFIO.F90` | All 4 get_client calls → use stored `this%output/input_server_name` (see Task 5) |
| `infrastructure/geom_io/GridPFIO.F90` | Same |
| `infrastructure/geom_io/FieldBundleRead.F90` | i_client → `MAPL_DEFAULT_INPUT_SERVER` |
| `infrastructure/geom_io/FieldBundleWrite.F90` | o_client → `MAPL_DEFAULT_OUTPUT_SERVER` |
| `gridcomps/extdata/ExtDataFileReader.F90` | i_client → `MAPL_DEFAULT_INPUT_SERVER` (no config-driven change yet) |
| `gridcomps/extdata/PrimaryExport.F90` | Same |

All files add `use mapl_ServerNames_mod` (or get constants via `use MAPL`).

### Task 5 — Server-name-aware `GeomPFIO`

Add optional stored server names to `GeomPFIO` abstract type:

```fortran
type, abstract :: GeomPFIO
   private
   ...
   character(:), allocatable :: output_server_name
   character(:), allocatable :: input_server_name
```

- `init_with_metadata` gains optional `output_server_name` arg; defaults to `MAPL_DEFAULT_OUTPUT_SERVER`. Stores name; uses `get_client(this%output_server_name)`.
- `init_with_filename` gains optional `input_server_name` arg; defaults to `MAPL_DEFAULT_INPUT_SERVER`.
- `update_time_on_server`, `stage_time_to_file` use `get_client(this%output_server_name)`.
- `GridPFIO.F90`, `FieldBundleRead.F90`, `FieldBundleWrite.F90` propagate the server name arg through to the underlying `GeomPFIO` calls.

No null-guard / standalone special case — the existing behavior (assert if client not found) is preserved.

### Task 6 — Config-driven server selection in `HistoryGridComp`

In `gridcomps/history/HistoryGridComp.F90`:

**Private state** (add type before `contains`):
```fortran
type :: HistoryPrivateState
   character(:), allocatable :: server_name
end type HistoryPrivateState
```
Use `_SET_NAMED_PRIVATE_STATE` / `_GET_NAMED_PRIVATE_STATE` macros (same pattern as `HistoryCollectionGridComp`).

**`setServices`**:
- Read `server:` from hconfig; default to `MAPL_DEFAULT_OUTPUT_SERVER`.
- Store in private state.

**`init`** (GENERIC::INIT_USER phase — runs after MAPL_RunServers):
- Retrieve private state.
- If `server_name /= MAPL_DEFAULT_OUTPUT_SERVER`: call `MAPL_ConnectToServer(server_name)`.
  (Default server connection is already established by `MaplFramework`.)

**`run`**:
- Retrieve private state; use `mapl_get_client(server_name)` instead of hardwired `'o_client'`.

### Task 7 — Propagate server name to collection children

In `gridcomps/history/HistoryGridComp_private.F90`, `make_child_hconfig`:

After the existing `ESMF_HConfigAdd(child_hconfig, content=collection_name, addKeyString='collection_name', ...)`, add:

```fortran
! Propagate server: key from History hconfig into each collection's hconfig.
! Collections use the server name to look up their client (but do not connect).
has_server = ESMF_HConfigIsDefined(hconfig, keystring='server', _RC)
if (has_server) then
   server_name = ESMF_HConfigAsString(hconfig, keystring='server', _RC)
   call ESMF_HConfigAdd(child_hconfig, content=server_name, addKeyString='server', _RC)
end if
```

The History-level `hconfig` (the parent) must be threaded into `make_child_hconfig` — it already receives it as the first arg. If the History hconfig has no `server:` key, no key is added and the collection defaults to `MAPL_DEFAULT_OUTPUT_SERVER` (Task 5 default).

### Task 8 — `HistoryCollectionGridComp` reads server name

In `gridcomps/history/HistoryCollectionGridComp.F90`, `init`:

```fortran
! Read server name propagated by make_child_hconfig; default to MAPL_DEFAULT_OUTPUT_SERVER
has_server = ESMF_HConfigIsDefined(hconfig, keystring='server', _RC)
server_name = MAPL_DEFAULT_OUTPUT_SERVER
if (has_server) server_name = ESMF_HConfigAsString(hconfig, keystring='server', _RC)

! Pass to writer initialization so it uses the right client
call collection_gridcomp%writer%init_with_metadata(metadata, geom, &
     output_server_name=server_name, _RC)
```

The collection does **not** call `MAPL_ConnectToServer` — only `HistoryGridComp` does. The collection only looks up the already-registered client by name.

### Existing config (no change needed — backward compatible)

No `servers:` section → default local servers created under `MAPL_DEFAULT_INPUT/OUTPUT_SERVER`. History reads no `server:` key → uses `MAPL_DEFAULT_OUTPUT_SERVER`. Identical behavior to today.

### New config with dedicated history server

```yaml
mapl:
  servers:
    mapl_default_input_server:
      local: true
    mapl_default_output_server:
      local: true
    pfio_history:
      num_nodes: 2
      sharedOb: libMAPL.history_server.so
      userRoutine: setservices_
```

History component hconfig:
```yaml
server: pfio_history
```

## Forward-compat notes

- **RestartHandler per-component config**: When MAPL hconfig inheritance is added, `RestartHandler` can read `mapl: { io_server: pfio_history }` from inherited config and call `get_client(io_server_name)`. No structural change needed now.
- **Shared server between History and RestartHandler**: Both use name-based lookup — if configured to the same server name, they share it automatically.
- **ExtData config-driven**: Same pattern as History when the time comes.
- **`MAPL_ConnectToServer` client type**: Currently defaults to `FastClientThread` for output. A future optional `client_subclass` arg or YAML key can refine this.

tclune:
## Types of change(s)

## Checklist
- [x] Ran the Unit Tests (`ctest`)

## Description

Introduces named constants `MAPL_DEFAULT_INPUT_SERVER` and `MAPL_DEFAULT_OUTPUT_SERVER`
in a new module `mapl_DefaultServerNames_mod` (exported via `pfio/API.F90`).
All hardcoded `'i_client'`/`'o_client'`/`'i_server'`/`'o_server'` string literals
are replaced with these constants throughout the codebase.

Also fixes a latent bug: `MAX_LEN_PORT_NAME` was 16, causing silent truncation of port
names longer than 16 characters, which caused `connect_to_server` to miss the local-port
fast path and deadlock in MPI remote discovery. Increased to 64.

## Related Issue

Fixes #5242

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Named default pfio server constants (#5242): new module `mapl_DefaultServerNames_mod`
+  exports `MAPL_DEFAULT_INPUT_SERVER` and `MAPL_DEFAULT_OUTPUT_SERVER`; all hardcoded
+  `'i_client'`/`'o_client'`/`'i_server'`/`'o_server'` string literals replaced with these
+  constants throughout `MaplFramework`, `RestartHandler`, `GeomPFIO`, `GridPFIO`,
+  `FieldBundleRead`, `FieldBundleWrite`, `HistoryGridComp`, `ExtDataFileReader`, and
+  `PrimaryExport`; fixed `MAX_LEN_PORT_NAME` (16 → 64) to support longer port names
+
 - Added `MAPL_StateMerge` to combine two `ESMF_State` objects into one without allocating new field memory
 - MAPL3 initialization lifecycle (#5231): new 6-call application lifecycle
   (`MAPL_Initialize`, `MAPL_CreateServers`, `MAPL_CapCreate`, `MAPL_RunServers`,
--- a/gridcomps/extdata/ExtDataFileReader.F90
+++ b/gridcomps/extdata/ExtDataFileReader.F90
@@ -1,10 +1,11 @@
 #include "MAPL.h"
 module mapl_ExtDataReader_mod
    use esmf
-   use gftl2_StringStringMap
-   use gftl2_StringIntegerMap
-   use MAPL
-   use pFlogger, only: logger
+    use gftl2_StringStringMap
+    use gftl2_StringIntegerMap
+    use MAPL
+    use mapl_DefaultServerNames_mod, only: MAPL_DEFAULT_INPUT_SERVER
+    use pFlogger, only: logger
    use, intrinsic :: iso_c_binding, only: c_ptr
    implicit none(type,external)
    private
@@ -95,7 +96,7 @@ module mapl_ExtDataReader_mod
          _RETURN(_SUCCESS)
       end if
 
-      i_client => mapl_get_client('i_client', _RC)
+      i_client => mapl_get_client(MAPL_DEFAULT_INPUT_SERVER, _RC)
 
       call MAPL_FieldBundleGet(this%accumulated_fields, fieldList=field_list, _RC)
       do i=1,size(field_list)
--- a/gridcomps/extdata/ExtDataFileReader.F90
+++ b/gridcomps/extdata/ExtDataFileReader.F90
@@ -95,7 +95,7 @@ module mapl_ExtDataReader_mod
          _RETURN(_SUCCESS)
       end if
 
-      i_client => mapl_get_client_thread('i_client', _RC)
+      i_client => mapl_get_client('i_client', _RC)
 
       call MAPL_FieldBundleGet(this%accumulated_fields, fieldList=field_list, _RC)
       do i=1,size(field_list)
--- a/gridcomps/extdata/PrimaryExport.F90
+++ b/gridcomps/extdata/PrimaryExport.F90
@@ -94,7 +94,7 @@ module mapl_PrimaryExport_mod
          call primary_export%bracket%set_node(NODE_LEFT, left_node)
          call primary_export%bracket%set_node(NODE_RIGHT, right_node)
          call primary_export%file_selector%get_file_template(file_template)
-         i_client => mapl_get_client_thread('i_client', _RC)
+         i_client => mapl_get_client('i_client', _RC)
          primary_export%client_collection_id = i_client%add_data_collection(file_template, _RC)
          call primary_export%bracket%set_parameters(time_interpolation=sample%time_interpolation)
          allocate(primary_export%start_and_end, source=time_range)
--- a/gridcomps/extdata/PrimaryExport.F90
+++ b/gridcomps/extdata/PrimaryExport.F90
@@ -13,10 +13,11 @@ module mapl_PrimaryExport_mod
    use gftl2_StringVector
    use mapl_ExtDataRule_mod
    use mapl_ExtDataCollection_mod
-   use mapl_ExtDataSample_mod
-   use VerticalCoordinateMod
-   use pflogger, only: logger
-   implicit none(type,external)
+    use mapl_ExtDataSample_mod
+    use VerticalCoordinateMod
+    use pflogger, only: logger
+    use mapl_DefaultServerNames_mod, only: MAPL_DEFAULT_INPUT_SERVER
+    implicit none(type,external)
    private
 
    public PrimaryExport
@@ -94,7 +95,7 @@ module mapl_PrimaryExport_mod
          call primary_export%bracket%set_node(NODE_LEFT, left_node)
          call primary_export%bracket%set_node(NODE_RIGHT, right_node)
          call primary_export%file_selector%get_file_template(file_template)
-         i_client => mapl_get_client('i_client', _RC)
+          i_client => mapl_get_client(MAPL_DEFAULT_INPUT_SERVER, _RC)
          primary_export%client_collection_id = i_client%add_data_collection(file_template, _RC)
          call primary_export%bracket%set_parameters(time_interpolation=sample%time_interpolation)
          allocate(primary_export%start_and_end, source=time_range)
--- a/gridcomps/history/HistoryCollectionGridComp.F90
+++ b/gridcomps/history/HistoryCollectionGridComp.F90
@@ -65,7 +65,9 @@ contains
       type(HistoryCollectionGridComp), pointer :: collection_gridcomp
       type(ESMF_HConfig) :: hconfig
       type(ESMF_Geom) :: geom
+      logical :: has_server
       character(len=ESMF_MAXSTR) :: name
+      character(len=:), allocatable :: server_name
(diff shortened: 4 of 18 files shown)


bena-nasa (dismissed):

bena-nasa:
@tclune the `Regrid_Util.x` application is broken due to something with the servers, dismissing my review. We need to fix this.

You can run this like so, I have files on bucy you can use
`mpirun -np 1 Regrid_Util.x -i /home/bmauer/geos_example_output/720x361_example.20000415_0000z.nc4 -o out.nc4 -ogrid PC360x181-DC`

Here is the error I'm seeing:
```
pe=00000 FAIL at line=00035    ClientManager.F90                        <Client 'mapl_default_input_server' not found in client manager>
pe=00000 FAIL at line=00144    GeomPFIO.F90                             <status=1>
pe=00000 FAIL at line=00368    FieldBundleRead.F90                      <status=1>
pe=00000 FAIL at line=00455    Regrid_Util.F90                          <status=1>
```

bena-nasa (changes requested):
`Regrid_Util.x` is broken because `MAPL_Initialize` does not add the local servers, so we need a way for the program to call `this%initialize_local_servers` in the `MaplFramework`

tclune:
Note - we could add a test that only runs if the input file is found in some standard place.  This would at least let CI check it even if users such as me don't look.

bena-nasa (approved):

Issue #5204, opened 2026-06-29
Create "apply" method


tclune:
The idea here is to enable service providers that would like to handle fields with ungridded dimensions in the natural fashion.  (Collections of 3D "slices".)

We want an interface that takes a FieldBundle and a subroutine and applies the function to each field in the bundle.

Needs several interfaces, extensions (see below)
* FieldBundleApplyUserRoutine
* FieldApplyUserRoutine  (iterating over ungridded of arbitrary rank is tricky)
* FieldGetPointerToSlice
  - overloaded for R4 and R8
  - default implementation for 3D, but might want/need 2D overloads.

```
subroutine  FieldBundleApplyUserRoutine(bundle, userRoutine, unusable, userrc, rc)
    <loop over fields in bundle>
   call FieldApplyProc(field, userRoutine, _USERRC)
end subroutine
```

```
subroutine FieldApplyUserRoutine(field, userRoutine, unusable, userrc, rc)
    integer, allocatable :: shape(:)
    call mapl_FieldGet(field, ungridded_shape=shape, _RC)

   idx = ???
   do
      call mapl_FieldGetPointerToSlice(field, slice=idx, _RC
      call userRoutine(field, rc=userrc)
      call increment(idx)
   end do
   
end subroutine
```

tclune:
The interface for the userRoutine should be
```
abstract interface
   subroutine I_userRoutine(field, unusable, rc)
       import esmf_Field
       import KeywordEnforcer
       type(ESMF_Field), intent(inout) :: field
       class(KeywordEnforcer), optional, intent(in) :: unusable
       integer, optional, intent(out) :: rc
   end subroutine
end interface
```

Might need a different name if it conflicts with other user procedures elsewhere.

tclune:
The provided interface is richer than it would first appear.  Consider the following user code:

```f90
subroutine outer(bundle, ...)
     
      u = ...
      v = ...

      call mapl_FieldBundleApplyUserRoutine(bundle, advect, _USERRC, _RC)

contains

      subroutine advect(field, unusable, rc)
          ...    ! has access to variables in outer scope; e.g., u, v
      end subroutine advect
end subroutine outer

PR #5260, opened 2026-07-16
Feature/aoloso/#5204 apply service providers

aoloso:
## Types of change(s)

## Checklist

## Description
Code changes to respond to #5204. See details in CHANGELOG. 

## Related Issue
#5204


infrastructure/field/FieldCondensedArray.F90
```diff
@@ -46,15 +50,99 @@ subroutine assign_fptr_condensed_array_r8(x, fptr, rc)
 
    end subroutine assign_fptr_condensed_array_r8
 
+   subroutine assign_fptr_condensed_array_slice3d_r4(x, fptr, rc)
```
tclune: We'll have to discuss this.  The entire point of "condensed" is that it is 3D no matter what the original field was:    Nh x Nv X No     We compress horizontal dimensions into a single dimension.  Vertical dimensions into a single dimension. (Size 1 if no vertical grid). And all non gridded dimensions as one dimension.  (Size 1 if no ungridded dims.)

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,481 @@
...
+
+      ! Slice pointers (ungridded dim(s) removed)
+      real(kind=ESMF_KIND_R4), pointer :: slice_r4(:,:)
+      real(kind=ESMF_KIND_R8), pointer :: slice_r8(:,:)
+      real(kind=ESMF_KIND_R4), pointer :: slice3d_r4(:,:,:)
+      real(kind=ESMF_KIND_R8), pointer :: slice3d_r8(:,:,:)
+
+      call ESMF_FieldGet(field, geom=geom, typekind=typekind, &
+           name=field_name, _RC)
+      call ESMF_FieldGet(field, ungriddedDimCount=ungrid_count, _RC)
```
tclune: ESMF considers our vertical dims to be ungridded,  so you need to be careful with this call.   

`mapl_FieldGet` has a way to return the ungridded dims in the MAPL sense, but does not have a `dimCount` variant.  That could be added here if it helps.

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,481 @@
...
+
+      if (ungrid_count == 0) then
+         n_slices = 1
+      else
+         ! Use condensed_slice_rank to resolve vertical-vs-second-ungridded
+         ! ambiguity before computing n_slices from native array dimensions.
+         slice_rank = condensed_slice_rank(field, _RC)
+
+         if (typekind == ESMF_TYPEKIND_R4) then
+            if (slice_rank == 3) then
```
tclune: This nested conditional for rank and kind should only happen once.  (I hope.)  Thinking ...

tclune: Yes.  n_slices is simply:
```
call mapl_FieldGet(field, ungriddedDims=ungriddedDims, _RC)
n_slices = ungriddedDims%get_num_ungridded()
```

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,481 @@
...
+      end if
+
+      do k = 1, n_slices
+         user_status = 0
+         ! Build a transient ESMF_Field pointing into slice k of the parent.
+         slice_field = FieldSliceToField(field, k, _RC)
+         call userRoutine(slice_field, rc=user_status)
+         call ESMF_FieldDestroy(slice_field, _RC)
+         if (user_status /= 0) exit
+      end do
```
tclune, line 154: For now this is ok.  I suspect we want to avoid creating/destroying fields in the usual case where there is just the 1 slice.

tclune (changes requested):
See inline comments.    I am hoping that the complexity you introduced in CondensedArray*.F90 can be simplified.   I'll try to give that some thought this morning.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: Ok. Also do we have case where we have two ungridded dims? If we always have one, the code can be simplified a bit.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: Ok, I am resolving this leaving `ESMF_FieldDestroy` in place.

infrastructure/esmf/FieldPointerUtilities.F90
```diff
@@ -36,6 +36,8 @@ module mapl_FieldPointerUtilities_mod
       module procedure assign_fptr_r8_rank2
       module procedure assign_fptr_r4_rank3
       module procedure assign_fptr_r8_rank3
+      !module procedure assign_fptr_r4_rank4
+      !module procedure assign_fptr_r8_rank4
```
tclune: Commented out.  Delete or uncomment?

infrastructure/field/tests/Test_FieldApplyUserRoutine.pf
```diff
@@ -0,0 +1,370 @@
+#include "MAPL_TestErr.h"
+#include "unused_dummy.H"
+
+module Test_FieldApplyUserRoutine
+   use mapl_FieldApplyUserRoutine_mod, only: MAPL_FieldApplyUserRoutine => FieldApplyUserRoutine
+   use mapl_FieldApplyUserRoutine_mod, only: MAPL_FieldGetPointerToSlice => FieldGetPointerToSlice
+   use mapl_FieldApplyUserRoutine_mod, only: MAPL_FieldCreateFieldSlice => FieldCreateFieldSlice
```
tclune: It is ok to use the rename here for the tests, but really not necessary.  The MAPL prefix is just to guard the namespace in "outside" contexts.

infrastructure/field/tests/Test_FieldApplyUserRoutine.pf
```diff
@@ -0,0 +1,370 @@
...
+      g_last_rank = dimcount
+
+      if (typekind == ESMF_TYPEKIND_R4) then
+         if (dimcount == 2) then
+            call ESMF_FieldGet(field, farrayPtr=data_r4, _RC)
+            data_r4 = real(g_call_count, kind=ESMF_KIND_R4)
+         else
+            call ESMF_FieldGet(field, farrayPtr=data3d_r4, _RC)
+            data3d_r4 = real(g_call_count, kind=ESMF_KIND_R4)
+         end if
```
tclune, line 58: Please use SELECT CASE with a default clause.   If/when othercases arise, it will be more natural to extend.

tclune: Never mind - forgot this was just test code.

infrastructure/field/FieldCondensedArray.F90
```diff
@@ -46,15 +47,65 @@ subroutine assign_fptr_condensed_array_r8(x, fptr, rc)
...
+   ! Return the rank of the "slice" produced by the condensed-array
+   ! machinery for this field:
+   !   1 — field has a single non-ungridded (grid) dimension, no vertical
+   !       (e.g. LocStream surface points)
+   !   2 — field has two non-ungridded dimensions: either 2-D horizontal
+   !       (Grid/Mesh) with no vertical, or 1-D horizontal (LocStream)
+   !       with vertical
+   !   3 — field has three non-ungridded dimensions: 2-D horizontal
+   !       (Grid/Mesh) with vertical
+   function condensed_slice_rank(f, rc) result(slice_rank)
```
tclune, line 59: Needs a better name.

tclune: Maybe `num_nonspatial_dims` if not `num_ungridded_dims`.

infrastructure/field/FieldCondensedArray.F90
```diff
@@ -46,15 +47,65 @@ subroutine assign_fptr_condensed_array_r8(x, fptr, rc)
...
+      n_spatial = count(gridToFieldMap /= 0)
+      if (has_vertical) n_spatial = n_spatial + 1
+
+      slice_rank = n_spatial
+
+      _RETURN(_SUCCESS)
+   end function condensed_slice_rank
+
    function get_fptr_shape(f, rc) result(fptr_shape)
       integer :: fptr_shape(ARRAY_RANK)
```
tclune: I don't immediately see what `ARRAY_RANK` is here.   I expected an allocatable array.

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,454 @@
...
+            call c_f_pointer(cptr, s_ptr_2d_r4, s_shape)
+            slice = ESMF_FieldCreate(geom, farrayPtr=s_ptr_2d_r4, &
+                 datacopyflag=ESMF_DATACOPY_REFERENCE, name=field_name, _RC)
+         case (3)
+            ! 2-D horizontal + vertical
+            call c_f_pointer(cptr, s_ptr_3d_r4, s_shape)
+            slice = ESMF_FieldCreate(geom, farrayPtr=s_ptr_3d_r4, &
+                 datacopyflag=ESMF_DATACOPY_REFERENCE, name=field_name, _RC)
+         case default
+            _FAIL('FieldCreateFieldSlice: unsupported slice rank.')
```
tclune, line 251: Remember to "fix" this comment when you rename the `condensed_slice_rank`.

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,454 @@
...
+      ! Obtain a C pointer to the start of the ith ungridded slice in the
+      ! condensed rank-3 view (gridded_product, vertical_or_1, ungridded_flat).
+      ! Then reinterpret that contiguous memory as an array of shape s_shape.
+      ! ------------------------------------------------------------------
+      if (typekind == ESMF_TYPEKIND_R4) then
+   
+         call assign_fptr_condensed_array(field, f_ptr_r4, _RC)
+         cptr = c_loc(f_ptr_r4(1, 1, ith))
+   
+         select case (s_rank)
```
tclune, line 234: I thought s_rank was the number of ungridded dims.  The SELECT CASE construct makes it look like the number of gridded dims.  ???

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,454 @@
...
+
+      nullify(ptr)
+      call mapl_FieldGet(field, ungridded_dims=ungridded_dims, _RC)
+      ungrid_num = ungridded_dims%get_num_ungridded()
+      _ASSERT(ungrid_num <= MAX_UNGRIDDED_DIMS, 'FieldGetPointerToSlice: ungrid_num > MAX_UNGRIDDED_DIMS is not supported.')
+      if (ungrid_num == 0) then
+         _ASSERT(slice_index == 1, 'slice_index must be 1 for a field with no ungridded dimension.')
+         call ESMF_FieldGet(field, farrayPtr=fptr2d, _RC)
+         ptr => fptr2d
+      else
```
tclune: I don't see how 4D and 5D arrays will be supported here.  Am I missing something?

tclune: ok - I see that you've split this across multiple procs ...

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,454 @@
...
+         call ESMF_FieldGet(field, farrayPtr=fptr3d, _RC)
+         ptr => fptr3d(:,:,slice_index)
+      end if
+      _RETURN(_SUCCESS)
+   end subroutine get_slice_r8
+
+   subroutine get_slice3d_r4(field, slice_index, ptr, rc)
```
tclune: Something still feels off here.   Not clear how to extend to higher ranks.  Maybe I missed something in my prototype from last week?

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,454 @@
...
+      call ESMF_FieldGet(field, status=field_status, _RC)
+      _RETURN_UNLESS(field_status == ESMF_FIELDSTATUS_COMPLETE)
+
+      call ESMF_FieldGet(field, typekind=typekind, _RC)
+      call mapl_FieldGet(field, ungridded_dims=ungridded_dims, _RC)
+      ungrid_num = ungridded_dims%get_num_ungridded()
+
+      _ASSERT(ungrid_num <= MAX_UNGRIDDED_DIMS, 'FieldApplyUserRoutine: ungrid_num > MAX_UNGRIDDED_DIMS is not supported.')
+
+      if (ungrid_num == 0) then
```
tclune: We need a "helper" procedure that returns the number of slices.   In the worst case, it at least removes the long conditional from the main procedure.  In the best case you can just use existing calls to directly get n_slices.   The idea was to use the "condensed" layout to get the num slices.   The condensed array layout is always {n_horz, n_vert, n_nonspatial}  so the size of the  3rd rank always gives you what you need.

tclune: Further, you don't want an R4 and an R8 version for this logic.

tclune (changes requested):
See inline comments.

infrastructure/field/FieldCondensedArray.F90
aoloso: I am confused. This procedure returns the number of 'gridded' dims i.e. the rank of the array sliced off of the original array with a given index k of the ungridded dim loop. It is not the number of ungridded dims. To make things clear, the procedure has the following comments"

`   ! Return the rank of the "slice" produced by the condensed-array
   ! machinery for this field:
   !   1 — field has a single non-ungridded (grid) dimension, no vertical
   !       (e.g. LocStream surface points)
   !   2 — field has two non-ungridded dimensions: either 2-D horizontal
   !       (Grid/Mesh) with no vertical, or 1-D horizontal (LocStream)
   !       with vertical
   !   3 — field has three non-ungridded dimensions: 2-D horizontal
   !       (Grid/Mesh) with vertical`

infrastructure/field/FieldCondensedArray.F90
aoloso: I didn't introduce this - it was already in the develop branch. `ARRAY_POINTER` is specified to be `3` in FieldCondensedArray_private.F90 (mapl_FieldCondensedArray_private_mod).

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: This is a no op, per earlier comment regarding `condensed_slice_rank`.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: `s_rank` is indeed the number pf gridded dims, per the earlier comments on `condensed_slice_rank`. The goal of the enclosing procedure `FieldCreateFieldSlice` is to create the appropriate field sliced off of the incoming `field` for the incoming `ith` ungridded index of the incoming `field`.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: I think you are right in the sense that the code is hardwired to support up to 4D with up to 2 ungridded dims. Need to rethink how to support an arbitrary rank? However are we limiting the number of ungridded dims to 2?

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: Ok, will think through this.

infrastructure/esmf/FieldPointerUtilities.F90
aoloso: Done

infrastructure/field/tests/Test_FieldApplyUserRoutine.pf
aoloso: Done

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: https://github.com/GEOS-ESM/MAPL/blob/47bca689154263256a01acb1955dbdd886404aa9/infrastructure/field/FieldApplyUserRoutine.F90#L394-L403

Will this work? Still R4/R8 -couldn't see how that can be avoided

infrastructure/field/FieldCondensedArray.F90
tclune: If that is the case, you don't need the function at all.  ESMF_FieldGet() can return the rank directly.

infrastructure/field/FieldCondensedArray.F90
tclune: Oh.  That is terrible.  we should fix while here.     Maybe `CONDENSED_RANK`

infrastructure/field/FieldApplyUserRoutine.F90
tclune: Arbitrary rank is probably not possible.  But I'm hoping just one new SELECT CASE for rank per KIND. (So 2 in total.)   Been a long time since I thought through the prototype, so this comment is pure instinct.

infrastructure/field/FieldApplyUserRoutine.F90
tclune: Yes.  But, note that the `n_slices = ...` can be moved outside the conditional now.   But also ...   assign_fptr_condensed_array() must already have logic to figure out n_slices for it to figure out what it already does.   You should extract that logic into a new procedure directly if possible, and then use it directly here.  But if that proves difficult (it might well require duplicating lots of code in there), then skip it.  This conditional here is the type we expect and just want to minimize.

infrastructure/field/FieldCondensedArray.F90
aoloso: `s_rank` is needed before the field `slice` is created so `ESMF_FieldGet` can not be used. 
`s_rank` is used in the case construct for the field `slice` creation to select the rank of the pointer to be created for `slice`

aoloso: Again up to a maximum of 3D 'gridded' dimensions is supported i.e. 

1D original field (like locstream) with 0 ungrid dim => 1D field slice
2D original field with 1 ungrid dim => num_ungrid 1D slices
2D original field with 0 ungrid dim => 2D field slice
3D original field with 0 ungrid dim => 3D field slice
3D original field with 1 ungird dim => num_ungrid 2D slices
3D original field with 2 ungrid dims => num_ungrid1*num_ungrid2 1D slices
....
ditto for 4D original field with an assumption that there will be at least one ungrid dim in that case thus.

Just like in the case of get_slice..., generalizing this to arbitrary rank and arbitrary number of ungrid dims might not be straight forward. I think at least one may need to exploit run-time unlimited polymorphism but I am not even sure about that.

infrastructure/field/FieldCondensedArray.F90
aoloso: Done

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: `assign_fptr_condensed_array()` doesn't already have the logic to figure out n_slices. The logic can be added to pass `n_slices` as an optional argument. However unless the dummy argument `fptr` is also made optional in which case `assign_fptr_condensed_array()` can be called just with `n_slices`, it doesn't reduce the complexity of the conditional. However making `fptr` optional will necessitate the callee `assign_fptr` with overloaded interface from within `assign_fptr_condensed_array()` that the corresponding dummy argument for them be made optional also. Looks like it is not worth the effort. Therefore I am leaving the code as is.

infrastructure/field/FieldCondensedArray.F90
aoloso: I think the logic for `get_field_layout` in infrastructure/field/FieldCondensedArray.F90 already appropriate uses ESMF_FieldGet where possible. It looks like there isn't a simpler, yet cleaner way to obtain `s_rank`.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: Just to be sure, did you mean trying to reduce the number of procedures in the overloaded interface FieldGetPointerToSlice such that there'll be 2 procedures get_slice_r4 and get_slice_r8 and then use select case in each procedure for rank?

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,419 @@
...
+      else
+         _FAIL('FieldCreateFieldSlice: unsupported typekind (expected R4 or R8).')
+      end if
+   
+      ! ------------------------------------------------------------------
+      ! Copy all ESMF attributes from the original field to the slice.
+      ! TODO: Remove ungridded-dimension-specific metadata so the slice
+      !       field does not misrepresent its structure.
+      ! ------------------------------------------------------------------
+      call ESMF_AttributeCopy(field, slice, _RC)
```
tclune: MAPL3 does not use Attributes.  Only ESMF_Info.    Please check to see if this routine does anything and if not, check back with me on how to adapt Info.

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,419 @@
...
+
+      nullify(ptr)
+      call mapl_FieldGet(field, ungridded_dims=ungridded_dims, _RC)
+      ungrid_num = ungridded_dims%get_num_ungridded()
+      _ASSERT(ungrid_num <= MAX_UNGRIDDED_DIMS, 'FieldGetPointerToSlice: ungrid_num > MAX_UNGRIDDED_DIMS is not supported.')
+      if (ungrid_num == 0) then
+         _ASSERT(slice_index == 1, 'slice_index must be 1 for a field with no ungridded dimension.')
+         call ESMF_FieldGet(field, farrayPtr=fptr2d, _RC)
+         ptr => fptr2d
+      else
```
tclune: I'm terribly confused.  Arrays could be 2d, 3d or ever 1d in this case.

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,419 @@
...
+         call ESMF_FieldGet(field, farrayPtr=fptr3d, _RC)
+         ptr => fptr3d(:,:,slice_index)
+      end if
+      _RETURN(_SUCCESS)
+   end subroutine get_slice_r8
+
+   subroutine get_slice3d_r4(field, slice_index, ptr, rc)
```
tclune: What is the relationship between get_slice_r4() and get_slice3d_r4()?

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,419 @@
...
+   ! ESMF_Field that wraps that slice; this field is passed to userRoutine
+   ! and then destroyed.  The number of slices is derived from the actual
+   ! last dimension(s) of the native array, disambiguated by
+   ! condensed_slice_rank to correctly handle the vertical-vs-second-
+   ! ungridded-dim ambiguity.
+   !
+   ! ungrid_num > MAX_UNGRIDDED_DIMS is caught by an assertion before
+   ! the slice loop.
+   ! --------------------------------------------------------------------------
+   subroutine FieldApplyUserRoutine(field, userRoutine, unusable, userrc, rc)
```
tclune: Please move this procedure to the top.  It should be the most visible.

tclune (commented):
Very close now.  I made a few minor suggestions in line.

But we are still left with a big issue with the actual slice mechanics.  I'll try to find my prototype and see why your version is more complex than I expected.

infrastructure/field/FieldApplyUserRoutine.F90
```diff
@@ -0,0 +1,419 @@
...
+   use mapl_FieldPointerUtilities_mod, only: FieldGetLocalElementCount
+   use mapl_UngriddedDims_mod, only: ungriddedDims
+   use mapl_FieldGet_mod, only: MAPL_FieldGet => FieldGet
+   use mapl_ErrorHandling_mod
+
+   implicit none(type, external)
+   private
+
+   public :: FieldCreateFieldSlice
+   public :: FieldGetPointerToSlice
```
tclune: Is this actually used?  It is the set of procedures that this contains that have been confusing me.

aoloso: They are used only in the tests for FieldApplyUserRoutine. However the tests themselves are irrelevant for procedures that are not used. The procedures and the tests have been deleted.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: Procedure not used and not relevant. Deleted.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: Oh the intent is to copy the attributes of the original field to the new field (slice) without the ungridded dim. Will look into ESMF_Info.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: No longer relevant. Encompassing procedure has been deleted.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: No longer relevant. Code deleted.

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: Done!

infrastructure/field/FieldApplyUserRoutine.F90
aoloso: ESMF_Info is now used to copy shared and private attributes from the original field to the 'slice' field.

tclune (approved):

PR #5282, opened 2026-07-30
v3: Improve Spack library support

mathomp4:
## Types of change(s)
- [x] Refactor (no functional changes, no API changes)

## Checklist
- [x] Ran the relevant CMake configuration and targeted tests

## Description

Improves MAPL3 configuration with external Spack libraries by delegating dependency-target setup and ESMF-version validation to ESMA_cmake.

Also fixes two build-tree issues exposed by this configuration:
- Use the copied ACG generator for standalone MAPL builds when `MAPL_BASE_DIR` is defined.
- Keep the generic `specs` source directory out of a binary directory named `specs`, avoiding the Spack Intel `ifort #10417` compiler-environment failure.

Adds the NCCS Intel MPI test environment setting required by the MAPL server-utilities test.

### Updates since opening:
- **Component Updates (`components.yaml`):**
  - Updated `ESMA_cmake` to `v4.46.0` (delegates dependency-target setup and ESMF validation; adds support for `ESMFConfig.cmake`)
  - Updated `ESMA_env` to `v5.26.0`
  - Updated `ecbuild` to `geos/v3.15.2`
- **CI / Testing Updates:**
  - Updated Baselibs to `v8.33.0` across CircleCI and GitHub Actions container images
  - Added `build_test_mapl_gnu16` CI job (`gfortran-16` / GCC 16.2.0)
  - Updated the downstream `geosgcm_build_tests` matrix to test `gfortran-16` (replacing `gfortran-14`)

## Related Issue
None.


mathomp4:
Note to @tclune this has the ugly `specs/` workaround. Not sure if that's how we want to handle it or not, but it does seem reproducible for now. Still, I'm not merging this until I can fully test.

I am...sort of surprised all the GEOSgcm runs worked. I was sort of expecting a failure. Hmm.

superstructure/generic/CMakeLists.txt
```diff
@@ -58,7 +58,10 @@ esma_add_library(${this}
   TYPE SHARED
   )
 
-add_subdirectory(specs)
+# Do not name this binary directory "specs". With the Spack Intel toolchain,
+# ifort's GCC probe fails from this directory when ./specs exists, surfacing as
+# the misleading ifort #10417 compiler-environment error.
+add_subdirectory(specs generic_specs)
```
tclune, line 75: I would love to understand this better.  It is such a weirdly specific problem.

As long as you can hold off on the merge, I'll contemplate other names.  Ideally we can come up with something that allows src and build dir to be same string.

mathomp4: Yeah. It's odd. GPT 5.6 Terra couldn't reproduce it in a simple case. But in MAPL, it definitely sees it. Still, I might lean on Terra to see if we could reproduce it somehow...

mathomp4: @tclune I've released the ESMA_cmake tag with the needed CMake updates. But this PR still has the `specs/` oddity. Any recent thoughts?

tclune:
So long as the change is only for the build dir, this is fine.

mathomp4:
> So long as the change is only for the build dir, this is fine.

Okay. Let me make the comment read a bit better for future devs...

tclune (approved):

PR #5292, opened 2026-08-04
Improve connection assertion messages

### LINKED ISSUE #5291: Improve SimpleConnection assertion messages
SimpleConnection activation reports unknown virtual connection points without identifying enough context. Include source or destination, component name, and virtual point full name so failures can be traced to configuration.

tclune:
## Types of change(s)

## Checklist
- [ ] Ran the Unit Tests (`make tests` or `ctest`)

## Description
Report source or destination, component name, and virtual point full name when `SimpleConnection` activation cannot find a virtual connection point. Keep `_ASSERT` invocations on one physical line for gfortran preprocessing compatibility.

Static checks: `git diff --check`; Fortran line-length check. No configured build tree was available.

## Related Issue

Fixes #5291

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Improved `SimpleConnection` assertion messages for unknown virtual connection points
 - Fixed bug in FieldBundleRead when file grid and output bundle grid are different grid classes
 - Buggy logic in server initialization (#5214)
 - Missing call to initialize error handling in MPI context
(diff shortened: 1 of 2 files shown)


tclune:
@pchakraborty @lizziel This is a quick patch to provide a better error message for the situation you encountered today.

My longer term plan would be to use logging debug messages.  Ordinary runs would produce no output in this layer, but if MAPL.connections level is set to DEBUG, then one would se a line emitted for each connection.   The last output would be the connection that failed.    At that point, I'll switch back to the simpler messages for the actual failure.

pchakraborty (approved):
LGTM

lizziel:
Thanks!

Issue #5281, opened 2026-07-29
Create Child GridComp based on Child ESMF_HConfig


darianboggs:
Currently in MAPL3, a Parent GridComp creates a Child GridComp based on information from the Parent ESMF_HConfig, including the shared object filename and SetServices procedure name. The Child ESMF_HConfig does not contain this information. We would like to push shared object and SetServices information down into the Child ESMF_HConfig. This will require changes to how the Child GridComp is created.

Currently we have a constructor `MAPL_GridcompCreate()` that (a) creates a raw ESMF gridcomp (ESMF_GridCompCreate), (b) attaches a private state (OuterMetaComponent), and (c) sets the user-level DSO information.   This constructor should be deleted, and calls to it should be replaced with simle ESMF_GRidCompCreate.   Then SetServices in GenericGridcomp should have new logic to attach the private state.  And finally, the setservices method in OuterMetaGridcomp should use the hconfig to find and set the user setservices.

darianboggs:
Tagging @tclune

darianboggs:
To avoid the dependence between the YAML files and Fortran code, the plan is to add the map key/value pairs to the Child HConfig YAML files before implementing the code changes. Once the code changes are implemented, the unused map key/value pairs will be eliminated. See sub-issues for more details.

PR #5296, opened 2026-08-04
Changes to GridComp yaml files for DSO refactor

darianboggs:
## Types of change(s)

## Checklist

## Description
This PR makes additions to the YAML files of child GridComps based on the parent GridComp YAML files. It does not modify the parent YAML files, and only adds a map for `setServices:` under `mapl:` in the child YAML file. Where a `mapl:` key does not exist, it adds a map with key `mapl:` in the child YAML.
## Related Issue
#5281

### DIFF
--- a/gridcomps/cap/tests/parent_child_captest/AGCM.yaml
+++ b/gridcomps/cap/tests/parent_child_captest/AGCM.yaml
@@ -25,3 +25,7 @@ mapl:
     vertical_grid:
       class: basic
       num_levels: 4
+
+  setServices:
+    sharedObj: libconfigurable_gridcomp
+    userRoutine: setservices_
--- a/gridcomps/cap/tests/write_restart/AGCM.yaml
+++ b/gridcomps/cap/tests/write_restart/AGCM.yaml
@@ -45,3 +45,7 @@ mapl:
     vertical_grid:
       class: basic
       num_levels: 4
+
+  setServices:
+    sharedObj: libconfigurable_gridcomp
+    userRoutine: setservices_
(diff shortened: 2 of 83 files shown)


tclune (commented):
Nothing is jumping out at me when I reviewed this.  But ...

- unchanged YAML files will not show up, so only tests can verify that the information is where it needs to be.
- As we discussed, the PR must be updated to include removal of the old (parent level) DSO information before we merge.    Very hard to test that the info was removed.   Though I suppose we could temporarily add a weird assertion in child processing that throws an error if the old DSO is found.  (We'd want to delete that before we merge though.

darianboggs:
> Nothing is jumping out at me when I reviewed this. But ...
> 
> * unchanged YAML files will not show up, so only tests can verify that the information is where it needs to be.
> * As we discussed, the PR must be updated to include removal of the old (parent level) DSO information before we merge.    Very hard to test that the info was removed.   Though I suppose we could temporarily add a weird assertion in child processing that throws an error if the old DSO is found.  (We'd want to delete that before we merge though.

Yes. I have a YAML file that maps the parent YAML files to the children they reference, which I told Claude to generate. The map is limited to those parent YAML files that have the child config YAML file value and the dso/sharedObj value. So I can use that to remove the extra information from the parent files. The python script uses `git ls-tree -d -r HEAD --name-only`, `git diff --name-only`, and the `CMakeLists.txt` files to filter on repository files, check for changes, and filter out repository directories that are no longer used in the build due to the umbrella refactoring.

tclune (approved):

PR #5306, opened 2026-08-08
Fix truncated length for 'SetServices' key in parse_child

pchakraborty:
## Summary
- `userProcedure_keys` in `parse_child.F90` was declared with `len=10`, one character short of `SetServices` (11 chars), silently truncating it to `SetService` and breaking the key match.
- Also drops trailing whitespace on an unrelated line.

## Test plan
- [ ] CI checks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pchakraborty:
Wrong base

PR #5334, opened 2026-08-17
Move child DSO services into configs

tclune:
## Types of change(s)
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Trivial change
- [x] Refactor

## Description
Moves DSO-backed child setServices metadata into child-owned configuration and supports external startup through ESMF_GridCompCreate followed by ESMF_GridCompSetServices. Migrates generic, CAP, configurable-gridcomp, and MAPL3G fixtures.

## Testing
- NAG full ctest: 69/73 passed.
- Generic, CAP, configurable-gridcomp, and all 27 MAPL3G tests passed.
- Four Regrid_Util tests remain unavailable because LOCAL_REGRESSION_DATA_DIR regression data is not configured.

## Related Issue
Fixes #5333

PR #5347, opened 2026-08-24
ACG: split DeclarePointer/GetPointer headers by state intent

### LINKED ISSUE #5342: ACG: split generated DeclarePointer/GetPointer headers by state intent (Import/Export/Internal)
`MAPL_GridCompSpecs_ACG.py` already splits the state-spec headers by intent - separate `Import___.h`, `Export___.h`, and `Internal___.h` files (`-i`/`-x`/`-p`).

The generated `DeclarePointer___.h` and `GetPointer___.h` headers (`-d`/`-g`) don't follow the same pattern - each is a single combined file covering all three state categories together.

Requesting ACG generate these split by intent too, e.g.:

- `DeclarePointer_Import___.h` /`GetPointer_Import___.h`
- `DeclarePointer_Export___.h` /`GetPointer_Export___.h`
- `DeclarePointer_Internal___.h` /`GetPointer_Internal___.h`

instead of one combined `DeclarePointer___.h`/`GetPointer___.h` per component. This would let a `Run`-like method `#include` only the intent(s) it actually needs (e.g. a helper subroutine that only touches `EXPORT` state), mirroring how the Import/Export/Internal spec headers are already consumed independently.

pchakraborty:
Contingent on #5334 

## Description

Closes #5342.

`MAPL_GridCompSpecs_ACG.py` already splits the state-spec headers by intent (separate `Import___.h`, `Export___.h`, `Internal___.h`). This PR extends the same pattern to the generated `DeclarePointer___.h` and `GetPointer___.h` headers, emitting additional per-intent companions:

- `DeclarePointer_Import___.h` / `GetPointer_Import___.h`
- `DeclarePointer_Export___.h` / `GetPointer_Export___.h`
- `DeclarePointer_Internal___.h` / `GetPointer_Internal___.h`

This lets a `Run`-like method `#include` only the intent(s) it actually needs.

> [!NOTE]  
> The existing combined `DeclarePointer___.h`/`GetPointer___.h` outputs are preserved unchanged so downstream code that includes them continues to work without modification.

## Changes

- `apps/MAPL_GridCompSpecs_ACG.py`: emit per-state pointer headers in addition to the combined ones; fix a `reduce()` crash when a state has zero gettable fields; pass state as a set to the field-state filter instead of a raw string.
- `cmake/mapl_acg.cmake`: register the new per-state files as additional generated outputs for `GET_POINTERS`/`DECLARE_POINTERS`.
- `apps/tests/acg3/acg3_unittests.py`: add unit tests for the new pointer filename helper.

## Testing

- Ran the `acg3` unit tests (`unittest discover`) — all passing.
- Generated headers from the `acg3` `compile_test.acg` fixture and verified combined files contain all states while split files contain only their respective state.


darianboggs (approved):
It looks good. It seems to follow the established patterns in ACG, which is good. I don't see the passing as set instead of raw string, but I don't think it is relevant to the change, so it doesn't matter if it is not there.

PR #5358, opened 2026-08-26
Bugfix/wdboggs/fix extdata dangling pointer

darianboggs:
## Types of change(s)

## Checklist

## Description
When iterating on a StringStringMap, the NAG compiler produced a dangling pointer error. This PR adds a target attribute to the ExtDataReader object that is passed into the problem proceduere. Also, it changes the iterator to use ftn_begin() and ftn_end() instead of begin() and end(). This is consistent with most iterators in MAPL.


tclune (approved):
The gFTL changes are irrelevant, but preferable style.

PR #5361, opened 2026-08-26
v3: Fix omission of setting allocation_status when creating ServiceClassAspect

lizziel:
## Types of change(s)

## Checklist
- [X] Tested with GCHP 

## Description

While testing a two-child case in GCHP using MAPL3 I ran into the following run issue:

```
pe=00001 FAIL at line=00125    InfoUtilities.F90                        <Key not found in info object: /MAPL/internal/allocation_status>
pe=00001 FAIL at line=00103    FieldBundleInfo.F90                      <status=1>
pe=00001 FAIL at line=00116    FieldBundleGet.F90                       <status=1>
pe=00001 FAIL at line=00707    StateItemSpec.F90                        <status=1>
pe=00001 FAIL at line=00177    StateItemSpec.F90                        <status=1>
```

Claude traced it to a problem in `ServiceClassAspect.F90` subroutine `create`, where a field bundle is created for type `MAPL_FIELDBUNDLETYPE_SERVICE`, but the `allocation_status=MAPL_STATEITEM_ALLOCATION_CREATED` is not set. This differs from all other `ClassAspect` types and is therefore deemed a bug, assuming they should all have this. I fixed it with the recommended fix, and that solved it.

The reason this was triggered was `AdvCore` declares bundle `TRADV` as a provider but I did not give it any subscribers. This is because I have not hooked up GEOS-Chem yet, which will be the subscriber. If the intention is for the model to not allow the case of declaring a provider without having a subscriber, then that would need to be a separate update in the error handling.

Note I am relying heavily on Claude (Sonnet 5) for this analysis. I am looking to @tclune for whether all of this makes sense. 

## Related Issue

none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed documentation workflows so manual runs publish only from trusted branches and v2 and MAPL3 documentation deployments preserve each other's output
 - Removed deployment and build-cache credentials from pull request jobs and restricted PR workflow tokens to read-only access
 - Dangling pointer in ExtDataFileReader due to a missing target attribute on ExtDataReader
+- Fixed omission of setting FieldBundle allocation status in create() for ServiceClassAspect
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


tclune (approved):
Yes - looks like a bizarre corner case that is unfortunately not covered by our tests.

mathomp4:
@lizziel Can you press the "Update branch" button on your side?

lizziel:
Done!

tclune (approved):

mathomp4:
@lizziel I just merged your other PR. Can you now re-update this one?

(Note: it turns out I can't press "Update branch" for you because your fork is an organizational fork. I learned something today!)

lizziel:
Done, again. :)

PR #5362, opened 2026-08-27
Feature/#5208 internal checkpoint

tclune:
## Types of change(s)

## Checklist


superstructure/generic/OuterMetaComponent/initialize_realize.F90
```diff
@@ -25,7 +25,9 @@ module recursive subroutine initialize_realize(this, importState, exportState, c
       call recurse(this, phase_idx=MAPL_GENERIC_INIT_REALIZE, _RC)
 
       user_states = this%user_gc_driver%get_states()
-      tmp_states = MultiState(importState=user_states%importState)
+      tmp_states = MultiState(importState=user_states%importState, &
+           exportState=user_states%exportState, &
+           internalState=user_states%internalState)
```
tclune: Claude wanted this change, but I don't think it is necessary.  I will play with it in a bit and likely undo this change.

tclune: I honestly can't remember why importState is special here - but it looked intentional.  OTOH the tests all pass either way.  Further, we will soon be refactoring how each phase updates states, so we can take a deeper dive at that point.

atrayano (dismissed):

atrayano (approved):

PR #5364, opened 2026-08-27
[Fix | MAPLPy] Move py -> fortran API into  declaration for .so fetch

FlorianDeconinck:
## Types of change(s)

## Checklist

## Description

Previously we didn't expose the public API of the python -> fortran functions. This was not necessary (it's called from Python, Fortran code shouldn't call into it) and the `.so`/`.dylib` would fetch the symbol as an unresolved .text symbol then resolve it via dynamic loading.

Starting with GCC 16 and/or GEOS v11.10.1+, the dynamic resolving of those text symbols fail and we get an `undefined symbols` issue.

We propose here to expose the function as a full `public` symbol, which will force the `.so` to resolve it while building (e.g. going from `t` symbols to `T`).

To mitigate the "fortran code should not use this", we make sure to not re-expose the functions to the higher module (e.g. `PythonBridge.F90`)


mathomp4 (dismissed):

mathomp4 (approved):

mathomp4:
@FlorianDeconinck This will be released in v2.72.0 soon

Issue #5355, opened 2026-08-25
Change "mapl" section in cap.yaml


tclune:
(I thought I had created an issue for this before, but could not find proof just now.)

Currently in cap.yaml we have a conventional `mapl` section under `cap`, and another top level `mapl` section for framework wide sections.   The reuse of the same mapping key is confusing.

1. Decide on a better name for the top level `mapl`.
2. Make the change.

Current thoughts for alternatives:

* framework
* mapl_framework
* mapl_global
* global

tclune:
Further discussion has broadened the scope of this issue.

1. We want to split cap.yaml into **3** files:   `mapl.yaml`, `cap_driver.yaml`, and `cap_gridcomp.yaml`
2. mapl.yaml
     - locate template in ./mapl/mapl.yaml
     - keys:  {esmf, resources, and driver}
3. cap_driver.yaml
     - location TBD;  leave in ./mapl/cap_driver.yaml for now
     - keys: {cap_restart, clock, cap_gridcomp}
     - Move `clock` subsection from old `cap.yaml:` into driver.


New templates:
```yaml
--- # mapl.yaml
app:
  config:  cap_driver.yaml

esmf:
  logKindFlag: ESMF_LOGKIND_MULTI_ON_ERROR
  defaultCalKind: ESMF_CALKIND_GREGORIAN
  logAppendFlag: false

pflogger: pflogger.yaml

app_petcount: 1

servers:
  mit:
    num_nodes: 4
    sharedObj: libmit
    userRoutine: init_comm_

  pfio:
    num_nodes: 9

  app:
    num_nodes: any

--- # cap_driver.yaml

restart: cap_restart.yaml
skip_restart_write: false

clock:
  start: 1891-03-01T00:00:00
  stop: 2999-03-02T21:00:00
  dt: PT900S
#   repeat_duration: P1Y
  segment_duration: P1H

--- # cap_gridcomp.yaml
checkpointing:
  enabled: true
  final: true
  path: checkpoints
  alarms:
    - {frequency: PT6H, refTime: '1891-03-01T00:00:00'}
    - {times: ['1891-03-01T00:00:00']}
    - {offsets: ['P6H']} # relative to segment start

root_name: &root GCM
extdata_name: &extdata EXTDATA
history_name: &history HIST

mapl:
  children:
    *root :
      config_file: GCM.yaml
    *extdata :
      config_file: extdata.yaml
    *history :
      config_file: history.yaml
...
```

[...]

PR #5371, opened 2026-08-30
Refactor/aoloso/split cap yaml #5355

aoloso:
## Types of change(s)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

## Description
Split cap.yaml

## Related Issue
5355


aoloso:
Is this on right track? If so, I can remove the old cap.yaml.

mapl/Cap.F90
```diff
@@ -281,6 +285,122 @@ function get_timestamp(clock, rc) result(path)
       _RETURN(_SUCCESS)
    end function get_timestamp
 
+   recursive subroutine merge_cap_hconfig(dst_hconfig, src_hconfig, rc)
```
tclune: This was unexpected.  I don't immediately see why something like this would be needed.  If you can explain - that would be great.

mapl/CapDriver.F90
```diff
@@ -314,8 +434,14 @@ function make_cap_options(hconfig, is_model_pet, rc) result(options)
...
       integer :: status
+      logical :: has_name
 
-      options%name = esmf_HConfigAsString(hconfig, keystring='name', _RC)
+      has_name = esmf_HConfigIsDefined(hconfig, keystring='name', _RC)
+      if (has_name) then
+         options%name = esmf_HConfigAsString(hconfig, keystring='name', _RC)
+      else
+         options%name = 'CAP'
+      end if
```
tclune, line 311: Preferred style.
```suggestion
      options%name = 'CAP'
      has_name = esmf_HConfigIsDefined(hconfig, keystring='name', _RC)
      if (has_name) then
         options%name = esmf_HConfigAsString(hconfig, keystring='name', _RC)
      end if
```

mapl/MaplServerUtilities.F90
```diff
@@ -74,12 +74,19 @@ integer function get_model_petCount(vm, hconfig, rc) result(model_petCount)
 
       integer :: status
       logical :: has_model_petcount
+      logical :: has_app_petcount
```
tclune, line 76: model_petcount should go away.  Replaced by app_petcount.

mapl/CapDriver.F90
```diff
@@ -3,6 +3,7 @@
 module mapl_Cap_mod
 
    use MAPL
+   use mapl_HConfigUtilities_mod, only: merge_hconfig
```
tclune, line 6: Please rename this file to CapDriver.F90 to reflect the new understanding.

mapl/cap_gridcomp.yaml
```diff
@@ -0,0 +1,22 @@
+checkpointing:
```
tclune: I think my instructions were wrong here.  Looking at the code logic, checkpointing is actually in the driver layer.  So the hconfig for checkpointing should go there.

mapl/Cap.F90
aoloso: This was meant to be an intermediate step where the original way of using cap.yaml is retained. I wanted to make sure my splitting approach is what you wanted before going further. Both `merge_cap_hconfig` and `resolve_cap_hconfig` will go away in the next pass.

mapl/CapDriver.F90
```diff
@@ -107,19 +122,32 @@ subroutine mapl_run_driver(hconfig, is_model_pet, unusable, servers, rc)
...
       type(esmf_Clock) :: clock
       type(CapOptions) :: options
+      type(esmf_HConfig) :: app_hconfig, cap_driver_hconfig, cap_gridcomp_hconfig
+      character(:), allocatable :: config_file, gridcomp_config_file
       integer :: status
 
-      options = make_cap_options(hconfig, is_model_pet, _RC)
-      clock = make_clock(hconfig, options%lgr, _RC)
-      driver = make_driver(clock, hconfig, options, _RC)
+      app_hconfig = ESMF_HConfigCreateAt(hconfig, keystring='app', _RC)
```
tclune: This seems to duplicate logit on line 97 above.

gridcomps/cap/cap_gridcomp.yaml
```diff
@@ -0,0 +1,15 @@
+root_name: &root GCM
```
tclune, line 1: This file goes in ./gridcomps/cap

mapl/CapDriver.F90
aoloso: `cap_driver.yaml` is parsed from disk twice for every run — once in mapl_cap_create, once in mapl_cap_run. There's no caching in between.

- MAPL_Get(hconfig=hconfig) is cheap — it returns the already-parsed top-level mapl.yaml hconfig cached in the MAPL singleton (parsed once in MAPL_Initialize), not a fresh file read.
- But cap_driver_hconfig/cap_gridcomp_hconfig are not cached anywhere in the singleton or in driver/CapOptions — each function independently re-derives them via ESMF_HConfigCreate(filename=...), which does re-read and re-parse the file from disk.
- This looks like each public entry point (MAPL_CapCreate, MAPL_CapRun) was written to be independently self-sufficient — deriving everything it needs from the MAPL singleton rather than depending on hidden state passed between calls — likely to keep the public API minimal (MAPL_CapRun(driver, rc) intentionally doesn't take an hconfig argument). But that convenience comes at the cost of redundant file I/O/parsing on every invocation.

However we can, based on suggestions from Claude analysis:
- either have mapl_cap_create stash cap_driver_hconfig somewhere mapl_cap_run can retrieve it — e.g., a private module-level saved variable in mapl_CapDriver_mod, since both calls happen in the same process/module scope.
- or extend CapOptions (or MAPL_GriddedComponentDriver) to carry the resolved cap_driver_hconfig, and have mapl_cap_run pull it from driver/options instead of re-resolving — cleaner, avoids new global state, but requires exposing it through the driver type.

Would you prefer one over the other, or shall we keep the original code with the duplicate?

gridcomps/cap/cap_gridcomp.yaml
aoloso: It looks like ./gridcomps/cap already has a file called CapGridComp.yaml that looks similar in structure to the current ./mapl/cap_gridcomp.yaml. Do we delete ./gridcomps/cap/CapGridComp.yaml and move ./mapl/cap_gridcomp.yaml to ./gridcomps/cap?

tclune: Exactly. Or rather edit the one that is already there.

apps/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ install (PROGRAMS
   split_restart.py
   mapl_acg.pl
   mapl_stub.pl
+  latlon_to_face.py
```
tclune, line 22: Is this new file related to this pull request?

mapl/CapDriver.F90
```diff
@@ -58,17 +59,15 @@ subroutine mapl_cap_create(driver, unusable, config, rc)
 
       integer :: status
       logical :: is_model_pet
-      type(ESMF_HConfig) :: hconfig, app_hconfig, cap_driver_hconfig, cap_gridcomp_hconfig
+      type(ESMF_HConfig) :: cap_driver_hconfig, cap_gridcomp_hconfig
```
tclune: The driver hconfig is the one being passed in under the new scheme.

tclune (changes requested):
It looks like you merged in other changes in develop to this PR.  Always better to rebase.  (Apologies if you did, and I'm misinterpreting the situation.)

Only substantive change - it still appears that cap driver is deriving its config file rather than using the driver hconfig that was supposed to be passed in.

apps/CMakeLists.txt
aoloso: No. I didn't add it - It is part of the code base for the `develop` branch.

mapl/CapDriver.F90
aoloso: Cleaned up the interface. Yes the driver hconfig was the one being passed.

tclune (approved):
Whew!

PR #5373, opened 2026-09-03
Add vector overloads to StateGetPointer

### LINKED ISSUE #5354: Extend StateGetPointer to support vector quantities
* 2 arguments
* enforce keywords for order
* Only need to support 3D
* Only support R4 for now

Notify @lizziel when it is available

pchakraborty:
## Description

Add `state_get_vector_ptr_r4_2d` and `state_get_vector_ptr_r4_3d` subroutines to `StateGetPointer`, allowing retrieval of paired (u, v) field pointers from a vector-type ESMF FieldBundle stored in a state. Returns `isPresent=.false.` without error when the bundle is declared but not connected (0 fields).

Adds `Test_StateGetPointerVector.pf` covering the new overloads.

Fixes #5354

@lizziel


lizziel:
@pchakraborty, thanks for pinging me on this!

superstructure/state/StateGetPointer.F90
```diff
@@ -84,4 +88,80 @@ module mapl_StateGetPointer_mod
...
+      _ASSERT(item_type == ESMF_STATEITEM_FIELDBUNDLE, 'expected vector (fieldBundle) for shortname: <'//itemName//'>')
+
+      call ESMF_StateGet(state, itemName, bundle, _RC)
+      call MAPL_FieldBundleGet(bundle, fieldBundleType=fieldBundleType, _RC)
+      _ASSERT(fieldBundleType == MAPL_FIELDBUNDLETYPE_VECTOR, 'expected vector fieldBundleType for shortname: <'//itemName//'>')
+
+      ! An optional vector export may be declared but not connected (0 fields)
+      call MAPL_FieldBundleGet(bundle, fieldCount=field_count, _RC)
+      if (field_count == 2) then
+         if (present(isPresent)) isPresent = .true.
```
tclune, line 120: Do we say "not present" for an unallocated scalar that is in the state?   I'd have thought this statement should go outside the conditional.

tclune (commented):
Please see inline question.

superstructure/state/StateGetPointer.F90
pchakraborty: Yes, anything other than `field_count == 2` would set `isPresent` to the default value of `.false.`. The default is set on line 108
```Fortran
if (present(isPresent)) isPresent = .false.
```

tclune (approved):

PR #5378, opened 2026-09-10
Fix get_geom kluge

bena-nasa:
## Types of change(s)

## Checklist


gridcomps/history/HistoryCollectionGridComp.F90
```diff
@@ -76,7 +76,7 @@ subroutine init(gridcomp, importState, exportState, clock, rc)
       _GET_NAMED_PRIVATE_STATE(gridcomp, HistoryCollectionGridComp, PRIVATE_STATE, collection_gridcomp)
       collection_gridcomp%output_bundle = create_output_bundle(hconfig, importState, _RC)
 
-      geom = detect_geom(collection_gridcomp%output_bundle, name, _RC)
+      geom = MAPL_FieldBundleGetGeom(collection_gridcomp%output_bundle, _RC)
```
tclune, line 79: There are two different senses of getting a geom from a bundle.   ESMF allows an ESMF bundle to have a geom independent of its contained fields.    We may want to rename this interface to something like GetUniqueGeom.   (Not a great name.)

tclune: Maybe GetCommonGeom

infrastructure/field_bundle/FieldBundleGetGeom.F90
```diff
@@ -0,0 +1,40 @@
...
+      integer, optional, intent(out) :: rc
+
+      integer :: status
+      integer :: i
+      type(ESMF_Field), allocatable :: fields(:)
+      type(ESMF_Geom) :: trial_geom
+
+      call FieldBundleGet(bundle, fieldList=fields, _RC)
+      do i = 1, size(fields)
+         call ESMF_FieldGet(fields(i), geom=trial_geom, _RC)
```
tclune, line 31: Due to the repeated init phases, we may eventually need a check first to see if the field is *COMPLETE* (or at least gridset) before checking its geom.    The failure mode would be different in that case - maybe just skip the phase and wait to come back.

superstructure/generic/tests/Test_IntegratedNormalization.pf
```diff
@@ -247,9 +247,6 @@ contains
       ! Verify mass conservation (conservative regridding should preserve mass for uniform mixing ratio)
       @assertTrue(mass_in > 0.0_ESMF_KIND_R8, 'Input mass should be positive')
       rel_error = abs(mass_out - mass_in) / mass_in
-      _HERE, mass_out
-      _HERE, mass_in
-      _HERE, rel_error, TOLERANCE_LOOSE
```
tclune, line 252: These must be ancient.  I did not touch this layer in months.

tclune (approved):

PR #5380, opened 2026-09-11
Fix Generic parent VM inheritance, Update to ESMA_cmake v4.48.0

### LINKED ISSUE #5379: Fix Generic components to inherit parent VM
## Description

`mapl_GenericSetServices()` created its direct user GridComp without a context flag, so ESMF defaulted it to `ESMF_CONTEXT_OWN_VM`. The Generic scenario test root also used the default context. Repeated scenario setup therefore accumulated separate VM communicator resources until MPICH exhausted context IDs.

## Proposed fix

- Create the direct Generic user GridComp with `ESMF_CONTEXT_PARENT_VM`.
- Create the Generic scenario test root with `ESMF_CONTEXT_PARENT_VM`.

## Validation

With NAG 7.2.43 and MPICH 5.0.1 on macOS, the original aggregate `MAPL.generic.scenarios` CTest entry passes all 230 cases after these changes.

This issue is limited to VM-context resource exhaustion. The separate `MPI_REAL` allreduce failure is an MPICH 5.0.1 bug fixed in MPICH 5.0.2rc1.

mathomp4:
## Types of change(s)
- [ ] Refactor (no functional changes, no API changes)

## Checklist
- [x] Ran the Unit Tests

## Description

Make GridComps created through the direct Generic SetServices path inherit their parent VM, matching MAPL’s standard Generic factory and child-component paths. The Generic scenario fixture also explicitly uses the parent VM for its root component. This prevents repeated scenario setup from allocating independent VM communicator contexts and exhausting MPICH context IDs.

The original aggregate `MAPL.generic.scenarios` CTest entry passes all 230 cases with NAG 7.2.43 and MPICH 5.0.1 on macOS.

This PR also fixes NetCDF quantization and Zstandard capability detection with Spack. `FindNetCDF` exposes the `NetCDF::NetCDF_C` imported target but does not set the legacy `NETCDF_INCLUDE_DIRS` variable, so the capability probes could not locate `netcdf_meta.h` and reported both features as unavailable. The probes now use the imported target’s include directories, which is also compatible with Baselibs. A clean Spack configure confirms both capability checks succeed. This is the MAPL3 counterpart to the MAPL v2 fix in #5388.

---

We also update to [ESMA_cmake v4.48.0](https://github.com/GEOS-ESM/ESMA_cmake/releases/tag/v4.48.0). This has fixes needed for macOS as well as updated NAG flags from investigations by @tclune.

## Related Issue

Fixes #5379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed Generic components created through direct SetServices to inherit their parent VM, preventing communicator-context exhaustion during repeated setup.
 - Fixed corrupted cubed-sphere coordinate endpoints in NAG-generated output
 - Fixed documentation workflows so manual runs publish only from trusted branches and v2 and MAPL3 documentation deployments preserve each other's output
 - Removed deployment and build-cache credentials from pull request jobs and restricted PR workflow tokens to read-only access
--- a/superstructure/generic/GenericGridComp.F90
+++ b/superstructure/generic/GenericGridComp.F90
@@ -74,7 +74,8 @@ contains
          call ESMF_GridCompGet(gridcomp, name=name, _RC)
          call ESMF_ConfigGet(config, hconfig=hconfig, _RC)
 
-         user_gridcomp = ESMF_GridCompCreate(name=trim(name), _RC)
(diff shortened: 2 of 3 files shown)


tclune (dismissed):

mathomp4:
@pchakraborty and I are trying to figure out the weird gfortran + debug issues.

tclune (approved):

PR #5392, opened 2026-09-17
Vector Statistics - Part 1

bena-nasa:
I have modified history and statistics grid comp to enable vectors to be handled for the mean, min, max, and accumulate cases.
This is part 1 because I have not done the variance case yet as they are more complicated.
Cluade has a plan but I need to understand this better.

This was already a big PR so I thought I should make it now.
I updated the scenario, statistics_real test to test all 8 possibilities for vector and scalar and updated configurable gridcomp as necessary

Part 2 will handle the variance 

## Types of change(s)

## Checklist


gridcomps/statistics/StatisticsGridComp.F90
```diff
@@ -72,29 +72,34 @@ subroutine advertise_item(gridcomp, iter, rc)
...
       action = esmf_HConfigAsString(hconfig, keystring='action', _RC)
       name = esmf_HConfigAsString(hconfig, keystring='name', _RC)
-
-      call MAPL_GridCompAddSpec(gridcomp, ESMF_STATEINTENT_IMPORT, name, typekind=MAPL_TYPEKIND_MIRROR,  _RC)
+      item_type_char = esmf_HConfigAsString(hconfig, keystring='itemtype', _RC)
+      if (item_type_char == 'field') then
+         item_type = MAPL_STATEITEM_FIELD
+      else if (item_type_char == 'vector') then
+         item_type = MAPL_STATEITEM_VECTOR
+      end if
```
tclune, line 90: Annoying, but you probably an ELSE clause with a fail message in case it is something else.   Possibly you could put a single assertion up in the StatisticsGridComp.F90 layer.

superstructure/generic/tests/scenarios/statistics_real/A.yaml
```diff
@@ -1,6 +1,20 @@
...
   TS: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
   PS: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
+  QV_MIN: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
+  SLP_ACCUM: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
+  UV:
+    - [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
+    - [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23]
+  UV_MIN:
+    - [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
+    - [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23]
```
tclune, line 11: This works fine.  If we ever really want the users to use it for actual GEOS work, we'll want to transpose things.  Time sequence of vector pairs as opposed to a pair of time sequences.  But definitely easier to read the way you have done it.  (You'd either get a long string with nested braces or a very long list with one pair per line.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed History and StatisticsGridComp so one can take average, min, max, and accumlation for vectors
```
tclune:
```suggestion
- Extended History and StatisticsGridComp so one can take average, min, max, and accumlation for vectors
```

tclune (dismissed):
Approving.  But please think about my comment regarding the ELSE clause.

gridcomps/statistics/StatisticsGridComp.F90
bena-nasa: @tclune made that change

tclune (approved):

PR #5396, opened 2026-09-18
Support for arbitrary order

darianboggs:
## Types of change(s)

## Checklist

## Description
The ACG Writer was not handling cases where an optional argument comes before the non-optional arguments in the subroutine. 

For example:
``` fortran
call MAPL_AddImportSpec(GC,
   LONG_NAME='Nancy Drew',
   SHORT_NAME='ND',
   ...
   _RC)
```
Now, it handles this case. It assumes that any arguments without a dummy argument name are positional, non-optional arguments. After processing the `dummy=value` arguments, optional or non-optional, it fills in the positional arguments that did not appear as `dummy=value` arguments.


atrayano (dismissed):

pchakraborty (approved):
LGTM

